PR #6, opened 2015-07-29
ThingShema Update

janko-isidorovic:
Added stuff to ThingShema:
- Created date (date),
- Enabled status (boolean),
- Claimable status (boolean)

Name is now required parameter.


server.js
```diff
@@ -32,8 +32,8 @@ var port = process.env.PORT || config.port;        // set our port
 
 // ROUTES FOR OUR API
 // =============================================================================
-app.use('/status', require('./app/routes/status'));
```
drasko, line 35: Why? Do you really want `api.mainflux.com/api/<something>` route? Why not just `api.mainflux.com/<something>`?

drasko:
Commit history is messed up. This can be corrected via `git cherry-pick` or `git squash`. This should be only one commit, not 5.

drasko:
In fact this should be 2 different commits - `.gitignore` change has nothing to do with `ThingSchema` and should have be done in a separate commit.

drasko:
Speaking about `api` prefix to the routes I suggest not using it, but using only `api.mainflux.com/things` for example. Question we have to resolve is how do we handle api version change?

drasko:
As per discussion, API versioning will probably be handled through subdomain name change (api.v2.mainflux.com), but most probably not needed as we will use JSON-API spec http://jsonapi.org/, which will guarantee that JSON objects do not change, and for resources - links should never change.

drasko:
This PR must be refused due two things:
- Commit history messed up
- `/api` route change should not exist (this commit is not needed)

Please fix these and re-post PR.

N.B. If you change `.gitignore` keep this in a spearate comit, as it is now.

PR #9, opened 2015-07-30
extra brake line removed

janko-isidorovic:
unnecessary brake line removed


drasko:
Extra break, not extra brake.

PR cancelled due bad commit subject.

PR #63, opened 2016-10-08
Fixed docker-compose and updated paho package, to resolve travis issue https://github.com/eclipse/paho.mqtt.golang/issues/62

nmarcetic:


docker-compose.yml
```diff
@@ -21,6 +21,9 @@
 ###
 mongo:
   image: mongo:latest
+  ports:
```
drasko, line 24: Why are you exposing these ports to local host? This is not needed.

Docker container `mainflux-server` should be capable to directly access container `mongo`, no need need for Mongo DB ports exposed to the host.

Problem is of course that your Mongo client on your dev host will not work - you have to log into container and check DB contents.

But anyway - for dev I would expect that everybody use local MongoDB on host, and Docker is for deployments...

drasko:
@mijicd any opinions on this?

@nmarcetic this was our previous `docker-compose.yml`: https://github.com/Mainflux/depricated-mainflux/blob/master/docker-compose.yml - no ports were exposed there and it worked nicely...

drasko (commented):
@nmarcetic also please provide PR description and please sign the commits.

Apache-2.0 guarantees patent-free submissions, and we must have commiters signature that it respects DCO: http://elinux.org/Developer_Certificate_Of_Origin

Follow the guidelines here: https://github.com/Mainflux/mainflux/blob/master/CONTRIBUTING.md#pull-requests

mijicd (changes requested):
If MongoDB is to be used exclusively by the mainflux-server, we can omit the ports exposure.

As the matter of fact, I would even prefer to edit the container networking over linking them together, but we can discuss this in the future.

At the moment I'd suggest the following:
- remove the ports section
- utilize the docker-compose inheritance to create the development and production composition

drasko:
@mijicd +1

PR #69, opened 2016-11-04
add revision number in mainflux banner, delete white spaces 

manuio:


drasko:
Not bad... however - this will demand `git` installed on deployment machine. Or in the `docker`, which is even worst (I do not install git in deployment dockers)

I do not like adding this dependency - too heavy.

@mijicd @nmarcetic opinion?

drasko:
For the sake of keeping deployment without git dependencies, I propose that we do not include this PR, but rather find the way how to include build revision into `cosnt.go` or similar file where `Version` is.

I think we should implement something like this: https://www.atatus.com/blog/golang-auto-build-versioning/

PR #95, opened 2017-09-24
Add normalizer

drasko:


glide.yaml
```diff
@@ -1,23 +1,25 @@
...
+- package: github.com/nats-io/go-nats
+  version: 1.3.0
+- package: github.com/prometheus/client_golang
+  version: 0.8.0
+- package: github.com/sirupsen/logrus
+  version: 1.0.3
+- package: golang.org/x/crypto
+  subpackages:
+  - bcrypt
+- package: go.uber.org/zap
```
mijicd: I'd rather that we replace logrus completely instead of adding another logger to the list of dependencies.

normalizer/messages.go
```diff
@@ -0,0 +1,61 @@
...
+package normalizer
+
+import (
+	"github.com/cisco/senml"
+	"github.com/mainflux/mainflux/writer"
+	"go.uber.org/zap"
+)
+
+// Message represents a message emitted by the mainflux adapters layer.
+type MainfluxMessage struct {
```
mijicd: Please avoid prefixing domain objects with `Mainflux`. It does not bring any value except code clutter.

cmd/normalizer/main.go
```diff
@@ -0,0 +1,107 @@
...
+
+		if err := json.Unmarshal(m.Data, &msg); err != nil {
+			logger.Error("Failed to unmarshal JSON",
+				zap.Error(err),
+			)
+			return
+		}
+
+		if msgs, err := normalizer.Normalize(logger, msg); err != nil {
+			logger.Error("Failed to normalize message",
```
mijicd: No need for a line break.

mijicd (changes requested):

normalizer/messages.go
mijicd: Also, note that this change will require a `writer` and `http` update as well.

drasko: Writer has been updated already, I'll update http.

Any propositions for message struct name? `AdapterMessage`?

mijicd (approved):

PR #99, opened 2017-09-24
Add /identity

drasko:

### DIFF
--- a/manager/api/endpoint.go
+++ b/manager/api/endpoint.go
@@ -167,3 +167,17 @@ func canAccessEndpoint(svc manager.Service) endpoint.Endpoint {
 		return accessRes{}, nil
 	}
 }
+
+func identityEndpoint(svc manager.Service) endpoint.Endpoint {
+	return func(_ context.Context, request interface{}) (interface{}, error) {
+		req := request.(identityReq)
+
+		id, err := svc.Identity(req.key)
+		if err != nil {
+			return nil, err
+		}
+
+		res := identityRes{id: id}
+		return res, nil
+	}
+}
--- a/manager/api/resources.go
+++ b/manager/api/resources.go
@@ -15,6 +15,29 @@ type apiResponse interface {
 	empty() bool
 }
 
+// Identity
+type identityReq struct {
+	key string
+}
+
+type identityRes struct {
+	id string
+}
+
+func (res identityRes) headers() map[string]string {
+	return map[string]string{
+		"X-Client-Id": res.id,
+	}
+}
+
+func (res identityRes) code() int {
+	return http.StatusOK
+}
+
+func (res identityRes) empty() bool {
+	return true
+}
+
 type clientReq struct {
 	key    string
 	id     string
--- a/manager/api/transport.go
+++ b/manager/api/transport.go
@@ -20,6 +20,13 @@ func MakeHandler(svc manager.Service) http.Handler {
 
 	r := bone.New()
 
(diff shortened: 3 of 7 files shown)


manager/api/resources.go
```diff
@@ -15,6 +15,27 @@ type apiResponse interface {
 	empty() bool
 }
 
+// Identity
+type identityReq struct {
+	Key string `json:"key"`
+}
+
+type identityRes struct {
+	ClientID string `json:"id"`
```
mijicd: Rename to `ID`.

manager/api/resources.go
```diff
@@ -15,6 +15,27 @@ type apiResponse interface {
...
+}
+
+func (res identityRes) headers() map[string]string {
+	return map[string]string{}
```
mijicd: We might consider returning it as a header to optimize the flow.

manager/api/transport.go
```diff
@@ -124,6 +131,15 @@ func MakeHandler(svc manager.Service) http.Handler {
 	return r
 }
 
+func decodeIdentity(_ context.Context, r *http.Request) (interface{}, error) {
```
mijicd, line 134: The same remark applies to requests: utilise headers to pass the token and identity.

mijicd (changes requested):

drasko:
Re-pushed. Please review.

manager/api/endpoint.go
```diff
@@ -167,3 +167,18 @@ func canAccessEndpoint(svc manager.Service) endpoint.Endpoint {
 		return accessRes{}, nil
 	}
 }
+
+func identityEndpoint(svc manager.Service) endpoint.Endpoint {
+	return func(_ context.Context, request interface{}) (interface{}, error) {
+		req := request.(identityReq)
+
+		res := identityRes{}
```
mijicd: Move after if, and "inline":

```
res := identityRes{
  id: id
}
```

manager/api/resources.go
```diff
@@ -15,6 +15,31 @@ type apiResponse interface {
...
+	key string
+}
+
+type identityRes struct {
+	id string
+}
+
+func (res identityRes) headers() map[string]string {
+	return map[string]string{
+		"Location": res.id,
```
mijicd: `Location` should be used to provide resource URL, absolute or relative (in our case only the relative ones were used). I would recommend using a custom header for this, for example `X-client-id`.

manager/swagger.yaml
```diff
@@ -56,6 +56,24 @@ paths:
...
+        Retrieves the entity identifier given its secret key.
+      tags:
+        - identity
+      parameters:
+        - $ref: "#/parameters/Authorization"
+      responses:
+        200:
+          description: ID retrieved
+          headers:
+            Location:
```
mijicd: Update according to the change recommended above.

mijicd (changes requested):

manager/mocks/security.go
```diff
@@ -28,6 +28,10 @@ func (idp *identityProviderMock) TemporaryKey(id string) (string, error) {
 		return "", manager.ErrUnauthorizedAccess
 	}
 
+	if id == "mySecretJwtToken" {
```
mijicd: Unnecessary.

mijicd (changes requested):

mijicd (approved):

PR #111, opened 2017-09-30
Improves API request validation

### LINKED ISSUE #106: DELETE client returns 500 internal server error if ID is wrong
When trying to DELETE a client with wrong or empty ID the manager returns "500 internal server error". Should be "403 Forbidden".

To reproduce:
`curl -s -S -i --noproxy localhost -X DELETE -H "Content-Type: application/senml+json" -H "Authorization: <user_token>" localhost:8180/clients/`
`curl -s -S -i --noproxy localhost -X DELETE -H "Content-Type: application/senml+json" -H "Authorization: <user_token>" localhost:8180/clients/<inexistent_ID>`

mijicd:
This pull request fixes #106 . On top of that, it provides following validations:
- user's email is validated
- authorization keys must be non-empty
- resource identifiers must be valid UUIDs


drasko:
:clap:

PR #128, opened 2017-12-06
Add HTTP auth check

drasko:

### DIFF
--- a/http/api/transport.go
+++ b/http/api/transport.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"os"
 
 	kithttp "github.com/go-kit/kit/transport/http"
 	"github.com/go-zoo/bone"
@@ -76,14 +77,32 @@ func decodeRequest(_ context.Context, r *http.Request) (interface{}, error) {
 	return msg, nil
 }
 
-// TODO: contact an auth provider
 func authorize(r *http.Request) (string, error) {
 	var apiKey string
+	var id string
+
 	if apiKey = r.Header.Get("Authorization"); apiKey == "" {
 		return "", errUnauthorizedAccess
 	}
 
-	return apiKey, nil
+	identityURL := os.Getenv("HTTP_ADAPTER_MANAGER_URL") + "/identity"
+	req, err := http.NewRequest("POST", managerURL, nil)
+	if err != nil {
+		return "", err
+	}
+
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("Authorization", apiKey)
+
+	c := &http.Client{}
+	resp, err := c.Do(req)
+	if err != nil {
+		return "", err
+	}
+
+	id = resp.Header.Get("x-client-id")
+
+	return id, nil
 }
 
 func checkContentType(r *http.Request) (string, error) {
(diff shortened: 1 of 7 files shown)


http/api/transport.go
```diff
@@ -76,14 +77,32 @@ func decodeRequest(_ context.Context, r *http.Request) (interface{}, error) {
...
 func authorize(r *http.Request) (string, error) {
 	var apiKey string
+	var id string
+
 	if apiKey = r.Header.Get("Authorization"); apiKey == "" {
 		return "", errUnauthorizedAccess
 	}
 
-	return apiKey, nil
+	identityURL := os.Getenv("HTTP_ADAPTER_MANAGER_URL") + "/identity"
```
mijicd: This can be set at top-level, as a constant, instead of setting it per each request.

http/api/transport.go
```diff
@@ -76,14 +77,32 @@ func decodeRequest(_ context.Context, r *http.Request) (interface{}, error) {
...
+		return "", err
+	}
+
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("Authorization", apiKey)
+
+	c := &http.Client{}
+	resp, err := c.Do(req)
+	if err != nil {
+		return "", err
```
mijicd: Make sure this error is properly handled. The safest way is to utilize domain errors.

mijicd (changes requested):
Make sure the build is fixed.

http/api/transport.go
drasko: What do you mean by this? I thought that upper layer above `transport.go` handles this error, so I just propagate it.

mijicd: Safely ignore this one, it was deprecated by #138 .

http/api/transport.go
```diff
@@ -76,14 +79,29 @@ func decodeRequest(_ context.Context, r *http.Request) (interface{}, error) {
...
 func authorize(r *http.Request) (string, error) {
 	var apiKey string
+	var id string
+
 	if apiKey = r.Header.Get("Authorization"); apiKey == "" {
 		return "", errUnauthorizedAccess
 	}
 
-	return apiKey, nil
+	authURL := os.Getenv("HTTP_ADAPTER_MANAGER_URL")
```
mijicd: It's better to perform this load & fallback at main, and just pass the configured value to this module.

http/api/transport.go
```diff
@@ -76,14 +79,29 @@ func decodeRequest(_ context.Context, r *http.Request) (interface{}, error) {
 	return msg, nil
 }
 
-// TODO: contact an auth provider
 func authorize(r *http.Request) (string, error) {
 	var apiKey string
+	var id string
```
mijicd: Not needed, it is created and initialized below.

mijicd (changes requested):

http/api/transport.go
drasko: Sure, but how to pass to this module :)

mijicd: Store the client reference to the service itself. Modify [this signature](https://github.com/mainflux/mainflux/blob/master/cmd/http/main.go#L47) and use it as an entrypoint for injecting the client. Then, store the client reference when initializing the transport layer through `MakeHandler` method and use it to authorize requests.

PR #144, opened 2018-01-14
fix external_link dependency for http adaptor service

kelvinji2009:
Signed-off-by: kelvinji <kelvinji2009@gmail.com>

`http adaptor` service will call the API `can_access` in `manager` service by `ManagerClient`, so we need to link the `manager` container to `http adaptor` container.


drasko:
LGTM

drasko:
Merged! Thanks a lot @kelvinji2009

Issue #135, opened 2017-12-25
Provide stable local environment


mijicd:
Due to multiple reported issues with the docker-compose approach using bash script to spawn containers, it is a top-priority to provide a stable environment for local development / demonstrations. Such system can be distributed either in Vagrant box(es) or as Minikube deployment.

chombium:
Hi @mijicd,

I also experienced the same problem with the docker-compose approach and I started testing something which will fix it. I will create a pull request tonight.

Running Mainflux on Minikube will also be nice and I was also thinking to start experimenting with Minkube when I finish with the fix for the docker-compose approach.

Best Regards,
Jovan

chombium:
Hi,

I also started reading and testing how we can improve the stability of the development environment.
The problem with docker-compose is that we need to ensure proper start up order of the services regarding their dependencies.

Possible solutions:

* Docker
The usual [docker-compose way](https://docs.docker.com/compose/startup-order/) is to define dependencies between the containers and add commands to which will check if (wait until) the dependencies are ready. That is usually done with the [depends_on](https://docs.docker.com/compose/compose-file/#depends_on) usually by adding a script which will check if some tcp port of the other container is opened or even calling some service in the container in order to make sure that the needed service is running. The other way of doing this is by using [healthchecks](https://docs.docker.com/compose/compose-file/#healthcheck).
The problem is that the depends_on is deprecated and will [not be ported](https://github.com/docker/compose/issues/4305) to docker-compose v3. The healthchecks are also ignored if someone uses "docker deploy" instead "docker-compose".
It seems that some people have used depends_on with docker-compose v3 [1](https://codefresh.io/docker-tutorial/deploy-docker-compose-v3-swarm-mode-cluster/), [2](https://codefresh.io/docker-tutorial/deploy-docker-compose-v3-swarm-mode-cluster/). 
I'll check this out.

* Kubernetes 
If we plan to use minikube for our devenv, Kubernetes has two probes for checking container health, the [Liveness and Readiness Probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/). The first one shows if the container is alive and the second shows if the container is ready to accept traffic. We can either define a check command or provide http or tcp endpoints with which the health of the container will be checked.

In both cases we need to define commands or api endpoints with which we can check the health status of the containers.

[...]

PR #151, opened 2018-02-03
change timeout for waiting for cassandra to 100 seconds;

achildrenmile:
changed the docker parameter to use param -i instead of -it to work with current docker version on windows


drasko:
@achildrenmile PR is unsigned and DCO is failing for that reason. Please refer to https://github.com/mainflux/mainflux/blob/master/.github/CONTRIBUTING.md#pull-requests, point 4.

.vs/ProjectSettings.json
```diff
@@ -0,0 +1,3 @@
+{
```
drasko, line 1: Please remove all editor-related files.

.vs/VSWorkspaceState.json
```diff
@@ -0,0 +1,7 @@
+{
```
drasko, line 1: Visual Studio files should be removed.

bin/mainflux-docker.sh
```diff
@@ -81,13 +81,13 @@ _start() {
   printf "\nWaiting for Cassandra to start. This takes time, please be patient...\n"
 
   # Wait until Cassandra is ready to accept cqlsh commands
-  # or timeout after 15 sec
+  # or timeout after 100 sec
```
drasko, line 84: Timeout of 100 sec looks too much to me. Would 30 sec do for your machine? Please find a minimal TO.

bin/mainflux-docker.sh
```diff
@@ -81,13 +81,13 @@ _start() {
...
   c_on=0
   retries=0
-  while [[ $retries -lt 15 ]]
+  while [[ $retries -lt 100 ]]
   do
     # Check if C* port 9042 is open
-    if $(docker exec -it mainflux-cassandra cqlsh -e "exit" > /dev/null 2>&1)
+    if $(docker exec -i mainflux-cassandra cqlsh -e "exit" > /dev/null 2>&1)
```
drasko, line 90: How does not having TTY affects Linux execution? If there is no difference, then this change is OK with me. Although I am not sure that we assure Windows support by design - Mainflux is developed and deployed on Unix machines, there is no guarantee that it will work on Windows.

drasko (changes requested):

drasko:
Please remove binary blobs sent in this PR.

chombium:
There is a better way of doing this... Instead of waiting some time  to let Cassandra start, we should wait until the Cassandra ports are opened and check the cluster state. We need something like this: https://stackoverflow.com/a/48035212 or

```bash
docker exec -i -t cassandra1 sh -c 'nodetool status'
```

I've already started working on this and I've started adjusting the Dockerfiles and the docker-compose yml configs to ensure the proper start up order of all containers as part of a pull request for #135, but I will not finish the pull request until the middle of the next week.

Best Regards,
Jovan

achildrenmile:
Hi all, pull request was done by accident. Will check everything in my private fork first. @chombium: thanks ;-)

drasko:
@chombium yes - I've tried to find a `cshql` command that makes the most sense. I do not see why `nodetool status` is better than pure `exit` (https://github.com/mainflux/mainflux/blob/master/bin/mainflux-docker.sh#L90). From my understanding, if system can log-in `cshql` (i.e. `cshql` interpreter is active), then it it is up and running.

I am really not sure where does docker-compose instability comes from, it is not only C* problem apparently, but I think it is the way of stoping/cleaning containers also...

PR #176, opened 2018-03-12
Add ARM cross-compile

drasko:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,18 +3,22 @@ BUILD_DIR=build
 all: manager http writer coap
 .PHONY: all manager http writer coap
 
+define compile_service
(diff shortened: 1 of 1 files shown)


mijicd (commented):
Move to bin as well?

drasko:
It is usual (even from FSF and C-era projects) that Makefiles sits in the root.

Some examples:
- Docker: https://github.com/moby/moby
- Tendermint: https://github.com/tendermint/tendermint
- Prometheus: https://github.com/prometheus/prometheus
- Traefik: https://github.com/containous/traefik

etc, etc...

mijicd (approved):

PR #177, opened 2018-03-12
Add mainflux launcher script

drasko:

### DIFF
--- a/bin/mainflux-launch.sh
+++ b/bin/mainflux-launch.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+###
+# Launches all Mainflux Go binaries when they are installed globally.
+# Also launches NATS broker instance, expecting that
+# `gnatsd` is installed globally.
+#
+# Expects that Cassandra is already installed and running.
+#
+# Does not launch NodeJS MQTT service - this one must be launched by hand for now.
+###
+
+# Kill all mainflux-* stuff
+function cleanup {
+	pkill mainflux
+}
+
+gnatsd &
+# Wait a bit for NATS to be on
+sleep 0.1
+mainflux-http &
+mainflux-manager &
+mainflux-writer &
+mainflux-coap &
+
+trap cleanup EXIT
+
+while : ; do sleep 1 ; done


bin/mainflux-launch.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+###
+# Launches all Mainflux Go binaries when they are installed globally.
+# Also launches NATS broker instance, expecting that
+# `gnatsd` is installed globally.
+#
+# Expects that Cassandra is already installed and running.
```
mijicd, line 8: No more Cassandra.

bin/mainflux-launch.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+###
+# Launches all Mainflux Go binaries when they are installed globally.
```
mijicd, line 4: This does not make sense. If we are providing launcher, we should automate the installation, or at least provide flags to indicate whether we want something to be installed or not.

bin/mainflux-launch.sh
```diff
@@ -0,0 +1,28 @@
...
+# Wait a bit for NATS to be on
+sleep 0.1
+mainflux-http &
+mainflux-manager &
+mainflux-writer &
+mainflux-coap &
+
+trap cleanup EXIT
+
+while : ; do sleep 1 ; done
```
mijicd, line 28: Why do we need this?

bin/mainflux-launch.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# Kill all mainflux-* stuff
+function cleanup {
+	pkill mainflux
+}
+
+gnatsd &
+# Wait a bit for NATS to be on
+sleep 0.1
+mainflux-http &
```
mijicd, line 21: The services should be built as well.

mijicd (changes requested):
Drip the mainflux prefix from script name.

bin/mainflux-launch.sh
drasko: Installation is automated via `Makefile`: https://github.com/mainflux/mainflux/pull/176

`make install` target will be added to install all the binaries.

bin/mainflux-launch.sh
drasko: This is the testing utility, not used for deployment. If we do not wait and catch CTRL+C signal, then kill all the `mainflux-` processes, then we will have to add a commaned to stop them, or chase them one by one and kill them manually.

I find it very convinient to stay in foreground mode with the script, and then kill all the processes with CTRL+C - this is very useful for fast starting whole system (all services) during development process.

bin/mainflux-launch.sh
drasko: This is developer helper script.

Maybe we should move it from the `bin` to `utils` or `helpers` or `scripts` or something?

bin/mainflux-launch.sh
mijicd: No it is not. If, as a user, I am enforced to install & configure everything, I don't see the point of providing an "automation" of easiest task of all - running 4 go binaries. Moreover, none of these will be configured properly (it will use the default configuration which might not be relevant).

bin/mainflux-launch.sh
mijicd: The main problem I have with this script is that it is useless. You define it as developer helper script. If we start from that premise, we're faced with the fact that:
- no infrastructure has been configured (NATS, PostgreSQL, nginx)
- no configuration can be provided to the services

While we can say "it is up to the developers to setup their infrastructure", the default configuration used by the services (since no other can be provided) is quite opinionated (e.g. it expects PostgreSQL to be available at `localhost:5432` which might not always be the case.

PR #178, opened 2018-03-12
Add Makefile and ARM support

drasko:


Makefile
```diff
@@ -0,0 +1,26 @@
+BUILD_DIR=build
```
mijicd, line 1: IMHO, having a "temporary" `bin` directory besides the `$GOBIN` is kinda redundant.

mijicd (changes requested):

Makefile
drasko: Cross-compiling for ARM on build host, for example.

mijicd (approved):

PR #186, opened 2018-03-15
MF-168 - Refactor CoAP adapter

dborovcanin:
1) Use NATS publish-subscribe to support CoAP observe messages
2) Use logger in sync with other services
3) Add metrics


coap/adapter.go
```diff
@@ -1,61 +1,257 @@
...
+func (ca *AdapterService) SendMessage(l *net.UDPConn, a *net.UDPAddr, m *coap.Message) *coap.Message {
+	ca.logger.Log("info", fmt.Sprintf("Got message in sendMessage: path=%q: %#v from %v", m.Path(), m, a))
+	var res *coap.Message
+	if len(m.Payload) == 0 && m.IsConfirmable() {
+		res.Code = coap.BadRequest
+		return res
+	}
+
+	// Uri-Query option ID is 15 (0xf).
+	token, err := ca.token(m.Option(0xf))
```
drasko: What's this token here? Is it CoAP token or some auth bearer token?

Because CoAP `token` is just a request ID: https://tools.ietf.org/html/rfc7252#section-5.3.1

coap/adapter.go
```diff
@@ -1,61 +1,257 @@
...
+	if err != nil {
+		res.Code = coap.Unauthorized
+		return res
+	}
+
+	cid := prefix + mux.Var(m, channel)
+
+	publisher, err := ca.mc.CanAccess(cid, token)
```
drasko: Looks like CoAP token is treated like JWT

drasko (commented):
Clarification of CoAP `token` vs authorization token is needed. They represent two different concepts.

Actually - it looks like CoAP depend on DTLS for auth: https://tools.ietf.org/html/rfc7252#section-9

While we were capable to "pass" JWT to MQTT via `password` field, I am not sure if CoAP has an appropriate header in which we can put JWT. Is this solved with "Option" feature: https://tools.ietf.org/html/rfc7252#section-3.1 - I see that you are using `Uri-Query` described here: https://tools.ietf.org/html/rfc7252#section-5.10

coap/adapter.go
dborovcanin: Its JWT token. JWT token is sent as url-query param, which is embeded in CoAP message options, with reserved code 15. Sorry for messy code, it's WIP version.

coap/adapter.go
dusanb94: Agree, I will do some refactoring to make it clear which token is used here.

dborovcanin: Since we use JWT-based auth, I found passing JWT as url query-param is pretty good workaround. Since token is wrapped as message option, it's not completely exposed. Also, obtaining token from message is straightforward and message payload stays clean. Possible problem is CoAP limitation of url query param length. 

Note: This implementation of CoAP adapter lacks self-notify checking. It will be implemented ASAP.

drasko: Yes, `Uri-Query` option field looks like the good candidate for JWT passing. Let's start with this and we'll see later.

Also, maybe we should note in the comment the max size of JWT allowed (alongside with some clarifications - that we use `Uri-Query`, etc).

dborovcanin: Agree, I will add some clarifications as I refactor code.

coap/adapter.go
```diff
@@ -1,61 +1,257 @@
...
 }
 
-type CoAPAdapter struct {
+// Subscriber represents token-NATS subscription map. Token is represented as string in
+// order to simplify access to subscription as well as to simplify unsubscribe feature.
+type Subscriber map[string]*Observer
+
+// AdapterService struct represents CoAP adapter service.
+type AdapterService struct {
 	obsMap map[string][]Observer
```
mijicd: Rename to `observers`. The map suffix is unnecessary since it is clear from the type that it is a map.

coap/adapter.go
```diff
@@ -1,61 +1,257 @@
...
+		res.Code = coap.Changed
+	}
+	return res
 }
 
-// BridgeHandler functions is a handler for messages recieved via NATS
-func (ca *CoAPAdapter) BridgeHandler(nm *broker.Msg) {
-	log.Printf("Received a message: %s\n", string(nm.Data))
+func (ca *AdapterService) notify(nm *broker.Msg) {
+	// TODO Authentication has to be added in order to prevent self-notifying.
```
mijicd: It's not the authentication, but the origin check (if I am the publisher, don't return me my message).

coap/adapter.go
```diff
@@ -1,61 +1,257 @@
...
+	m, err := ca.convertMsg(nm)
+	if err != nil {
+		return
+	}
+	ca.logger.Log(m.Publisher, m.Protocol, m.Channel, m.Payload)
+	ca.Transmit(m.Channel, m.Payload, m.Publisher)
+}
 
-	// And write it into the database
+func (ca *AdapterService) token(opt interface{}) (string, error) {
```
mijicd: Is it possible to use the concrete type instead of `interface{}`?

coap/adapter.go
```diff
@@ -1,61 +1,257 @@
...
+	if opt == nil {
+		return "", errUnauthorizedAccess
+	}
+	val := opt.(string)
```
mijicd: See the comment above. This typecase surely needs both the tests and the error check.

coap/adapter.go
```diff
@@ -1,61 +1,257 @@
...
+		ca.deregisterSub(m, cid)
+	}
+
+	if m.IsConfirmable() {
+		res.Code = coap.Valid
+	}
+	return res
+}
+
+func (ca *AdapterService) registerSub(o *Observer, cid string) error {
```
mijicd: Rename to `subscribe`.

coap/adapter.go
```diff
@@ -1,61 +1,257 @@
...
+	}
+	o.sub = res
+	token := base64.StdEncoding.EncodeToString(o.message.Token)
+	s[token] = o
+	ca.subs[cid] = s
+	ca.logger.Log("info", "Subscribed to: "+ca.subs[cid][token].sub.Subject)
+	return nil
+}
+
+func (ca *AdapterService) deregisterSub(m *coap.Message, cid string) error {
```
mijicd: Rename to `unsubscribe`.

coap/adapter.go
```diff
@@ -1,61 +1,257 @@
...
+	// Uri-Query option ID is 15 (0xf).
+	token, err := ca.token(m.Option(0xf))
+	if err != nil {
+		res.Code = coap.Unauthorized
+		return res
+	}
+
+	cid := prefix + mux.Var(m, channel)
+
+	_, err = ca.mc.CanAccess(cid, token)
```
mijicd: You can inline this: 
```golang
if _, err := ca.mc.CanAccess(cid, token); err != nil {
        res.Code = coap.Unauthorized
        return res
}
```

mijicd (changes requested):
This is a first, hawk-eye review. I will make another, more thorough review as you progress with the implementation. It would be good if @drasko could provide his opinion as well.

coap/adapter.go
dborovcanin: Since [go-coap library](https://github.com/dustin/go-coap/blob/master/message.go#L366) returns an interface, eventually it's possible to do some preprocessing/checking/parsing, but it looks like there is no other easy way to get what I want from options.

mijicd: Then it might be a good idea to extract the proper value before invoking this function, or if you want to keep the interface as is, don't forget to do the proper type checks.

coap/adapter.go
dborovcanin: @mijicd @drasko What do you suggest to do with notificataion messages? In this implementation, it's left to NATS, but, either way, there has to be some kind of inline structure to keep connections to be able to notify clients later. Previously, same thing was done with no NATS included at all. To keep it this way or to revert to drasko's implementation?

cmd/coap/main.go
```diff
@@ -43,7 +81,8 @@ func main() {
 
 	go func() {
 		coapAddr := fmt.Sprintf(":%d", cfg.Port)
-		errs <- ca.Serve(coapAddr)
+		logger.Log("info", fmt.Sprintf("Start CoAP server at %s", coapAddr))
```
drasko: Align the messages to other adapters, ex. https://github.com/mainflux/mainflux/blob/master/cmd/http/main.go#L78

drasko (changes requested):

coap/api/transport.go
```diff
@@ -0,0 +1,29 @@
+package api
+
+import (
+	"net"
+
+	mux "github.com/dereulenspiegel/coap-mux"
+	gocoap "github.com/dustin/go-coap"
```
drasko: `coap`, not `gocoap`, was used before - please stay consistent (change one or another)

drasko (changes requested):

drasko:
Additional concerns:

1. I see vendor updated - is this rebased to changes on the master I introduced recently ( I updated all vendoring)
2. If I remember well I had to traverse map of Observers and in quite inefficient way... This was OK for the PoC, but if needed we can maybe consider a better - DB-like (some indexing or in-RAM DB, dunno) to fetch the needed observer

coap/api/transport.go
dborovcanin: One `coap` is ours, one is library package. For that reason, I can't use `coap` name. To stay consistent, I will use `gocoap` name in other packages too.

drasko (changes requested):
Please align README.md to other adapters

coap/adapter.go
```diff
@@ -1,61 +1,129 @@
...
+	Observe(l *net.UDPConn, a *net.UDPAddr, m *gocoap.Message) *gocoap.Message
+	Recieve(l *net.UDPConn, a *net.UDPAddr, m *gocoap.Message) *gocoap.Message
 }
 
-type CoAPAdapter struct {
-	obsMap map[string][]Observer
-	logger *zap.Logger
+const (
+	token    string = "token"
+	prefix   string = "src.coap."
```
drasko: Please align to latest changes - NATS topics are now equivalent to Mainflux channels: https://github.com/mainflux/mainflux/pull/207#pullrequestreview-106327364

drasko (changes requested):

coap/adapter.go
```diff
@@ -1,61 +1,145 @@
...
+	"github.com/go-kit/kit/log"
 	"github.com/mainflux/mainflux"
 	broker "github.com/nats-io/go-nats"
-	"go.uber.org/zap"
 )
 
-const protocol string = "coap"
+// PubSub interface is used to access to NATS pub-sub model.
+type PubSub interface {
+	Publish(msg mainflux.RawMessage) error
```
mijicd: No need for naming parameters. Moreover, this will have to be updated once #188 is merged.

coap/adapter.go
```diff
@@ -1,61 +1,145 @@
...
+const (
+	key       string = "key"
+	channel   string = "id"
+	protocol  string = "coap"
+	maxPktLen int    = 1500
+)
+
+var (
+	errBadRequest = errors.New("bad request")
+	logger        log.Logger
```
mijicd: Why do you need logger variable? Let it be bound to the structure.

coap/adapter.go
```diff
@@ -1,61 +1,145 @@
...
 	return ca
 }
 
-// Serve function starts CoAP server
-func (ca *CoAPAdapter) Serve(addr string) error {
-	ca.logger.Info("Starting CoAP server", zap.String("address", addr))
-	return coap.ListenAndServe("udp", addr, ca.COAPServer())
-}
+// Recieve method processes message and pushes it to NATS.
+func (ca *AdapterService) Recieve(auth AuthProvider) MsgHandler {
```
mijicd: Fix typo, it should be `Receive` not `Recieve`.

coap/adapter.go
```diff
@@ -1,61 +1,145 @@
...
+
+		if value, ok := m.Option(gocoap.Observe).(uint32); ok && value == 0 {
+			subject := fmt.Sprintf("channel.%s", cid)
+			if _, err := ca.pubSub.Subscribe(subject, obsHandle(l, a, m, 120)); err != nil {
+				ca.logger.Log("error", fmt.Sprintf("Error occured during subscription to NATS %s", err))
+				res.Code = gocoap.InternalServerError
+				return res
+			}
+			res.AddOption(gocoap.Observe, 0)
+		} else {
```
mijicd: Either handle the condition you mentioned in the comment, or drop the entire branch. No "empty branches" should be present.

coap/adapter.go
```diff
@@ -1,61 +1,145 @@
...
+		if m.IsConfirmable() {
+			res.Code = gocoap.Changed
+		}
+		return res
 	}
+}
+
+// Observe method deals with CoAP observe messages.
+func (ca *AdapterService) Observe(auth AuthProvider) MsgHandler {
+	return func(l *net.UDPConn, a *net.UDPAddr, m *gocoap.Message) *gocoap.Message {
```
mijicd: While go enforces short names, it does not enforce meaningless names. Don't be cheap on characters :)

coap/adapter.go
```diff
@@ -1,61 +1,145 @@
...
 
-const protocol string = "coap"
+// PubSub interface is used to access to NATS pub-sub model.
+type PubSub interface {
+	Publish(msg mainflux.RawMessage) error
+	Subscribe(subject string, cb broker.MsgHandler) (*broker.Subscription, error)
+}
+
+// MsgHandler handles messages CoAP server recieved.
+type MsgHandler func(l *net.UDPConn, a *net.UDPAddr, m *gocoap.Message) *gocoap.Message
```
mijicd: No need for naming the parameters here.

coap/mock/nats/nats.go
```diff
@@ -0,0 +1,81 @@
+package nats
```
mijicd: Any reason for creating subpackage in `mocks`?

coap/mock/manager/client.go
```diff
@@ -0,0 +1,27 @@
+package mock
```
mijicd: Stay consistent, rename this package to `mocks`.

coap/mock/manager/client.go
```diff
@@ -0,0 +1,27 @@
...
+func (c Client) CanAccess(channel string, token string) (string, error) {
+	if c.Email == "" {
+		return "", errUnauthorized
+	}
+	return c.Email, nil
+}
+
+// New method creates new mocked manager service.
+func New(email string) Client {
+	return Client{
```
mijicd: Please change to `return Client{email}`.

coap/mock/nats/nats.go
```diff
@@ -0,0 +1,81 @@
...
+	"sync"
+	"testing"
+
+	"github.com/mainflux/mainflux"
+
+	"github.com/gogo/protobuf/proto"
+	broker "github.com/nats-io/go-nats"
+)
+
+const (
```
mijicd: Please change to `const prefix = "channel."`.

coap/mock/nats/nats.go
```diff
@@ -0,0 +1,81 @@
...
+	prefix = "channel."
+)
+
+var log *testing.T
```
mijicd: Why do you need this?

coap/utils.go
```diff
@@ -0,0 +1,120 @@
...
+	broker "github.com/nats-io/go-nats"
+)
+
+// AuthProvider represents ManagerClient which provides auth features.
+type AuthProvider interface {
+	CanAccess(string, string) (string, error)
+}
+
+// ObsHandle handles observe messages and keeps connection to client in order to send notifications.
+func obsHandle(l *net.UDPConn, a *net.UDPAddr, m *gocoap.Message, offset time.Duration) broker.MsgHandler {
```
mijicd: See the remark regarding naming.

coap/utils.go
```diff
@@ -0,0 +1,120 @@
...
+	"time"
+
+	gocoap "github.com/dustin/go-coap"
+	"github.com/gogo/protobuf/proto"
+	"github.com/mainflux/mainflux"
+	broker "github.com/nats-io/go-nats"
+)
+
+// AuthProvider represents ManagerClient which provides auth features.
+type AuthProvider interface {
```
mijicd: Please note that some of the functions below can be bound to the `AuthProvider` instead of passing the provider to them.

mijicd (changes requested):
@drasko please take a look at my comments, and add remarks regarding the CoAP details.

Gopkg.lock
```diff
@@ -392,6 +367,10 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
+<<<<<<< HEAD
```
anovakovic01: This should be resolved.

coap/api/transport.go
```diff
@@ -0,0 +1,201 @@
...
+
+		if err != nil {
+			res.Code = gocoap.Unauthorized
+			return res
+		}
+
+		if value, ok := msg.Option(gocoap.Observe).(uint32); ok && value == 1 {
+			id := fmt.Sprintf("%s-%x", publisher, msg.Token)
+			err := svc.Unsubscribe(id)
+			if err != nil {
```
anovakovic01: Can be part of `if` statement.

coap/adapter.go
```diff
@@ -1,61 +1,85 @@
...
-	log.Println("Calling obsTransmit()")
-	log.Println(m.Publisher, m.Protocol, m.Channel, m.Payload)
-	ca.obsTransmit(m)
+func (svc *adapterService) Unsubscribe(id string) error {
+	obs, ok := svc.Subs[id]
+	if !ok {
+		return nil
+	}
+	err := obs.Sub.Unsubscribe()
+	if err != nil {
```
anovakovic01: This can be part of `if` statement.

coap/api/transport.go
```diff
@@ -0,0 +1,201 @@
...
+			Type: gocoap.Acknowledgement,
+			Code: gocoap.NotFound,
+		}
+	}
+	return nil
+}
+
+// MakeHandler function return new CoAP server with GET, POST and NOT_FOUND handlers.
+func MakeHandler(svc coap.Service) gocoap.Handler {
+	// auth = mgr
```
anovakovic01: Remove commented code.

coap/api/transport.go
```diff
@@ -0,0 +1,201 @@
...
+			go handleSub(svc, id, conn, addr, msg, ch)
+			res.AddOption(gocoap.Observe, 0)
+		}
+		return res
+	}
+}
+
+func notify(svc coap.Service, id string, conn *net.UDPConn, addr *net.UDPAddr, msg *gocoap.Message, counter *uint32, count *[]byte) error {
+	err := sendMessage(conn, addr, msg, counter, count)
+	if err != nil {
```
anovakovic01: Can be part of `if` statement.

coap/api/transport.go
```diff
@@ -0,0 +1,201 @@
...
+	msg.SetOption(gocoap.Observe, (*count)[:3])
+	var err error
+	for i := 0; i < 3; i++ {
+		err = gocoap.Transmit(conn, addr, *msg)
+		if err != nil {
+			time.Sleep(5 * time.Millisecond)
+			continue
+		}
+		return nil
+	}
```
anovakovic01: Can you use circuit breaker here? Is this necessary?

coap/api/transport.go
```diff
@@ -0,0 +1,201 @@
...
+					return
+				}
+				res.Type = gocoap.NonConfirmable
+				res.Payload = rawMsg.Payload
+				if err := notify(svc, id, conn, addr, res, &counter, &count); err != nil {
+					return
+				}
+			}
+		}
+	}()
```
anovakovic01: Why do you need anonymous function here? Maybe you are missing `go` in front?

coap/nats/service.go
```diff
@@ -0,0 +1,21 @@
...
+
+// Service specifies NATS service API.
+type Service interface {
+	mainflux.MessagePublisher
+	// Subscribes to channel with specified id.
+	Subscribe(string, chan mainflux.RawMessage) (Subscription, error)
+}
+
+// Subscription interface represents subscription to messaging system.
+type Subscription interface {
```
anovakovic01: You can replace this with function type. Also, can you handle unsubscribe via channels?

coap/server.go
```diff
@@ -1,30 +1,107 @@
...
+	}
+	var respMsg *gocoap.Message
+	switch msg.Type {
+	case gocoap.Reset:
+		respMsg = &msg
+		cid := mux.Var(&msg, "id")
+		publisher, err := Authorize(&msg, respMsg, cid)
+		if err != nil {
+			respMsg.Code = gocoap.Unauthorized
+		} else {
```
anovakovic01: Avoid `else` by using `break`.

cmd/coap/main.go
```diff
@@ -6,44 +6,59 @@ import (
...
 
-	pub := nats.NewMessagePublisher(nc)
-	ca := adapter.NewCoAPAdapter(logger, pub)
-
-	nc.Subscribe("src.http", ca.BridgeHandler)
-	nc.Subscribe("src.mqtt", ca.BridgeHandler)
+	pub := nats.New(nc)
 
+	mgr := manager.NewClient(cfg.ManagerURL)
+	svc := coap.New(pub)
```
anovakovic01: Add logging and metrics.

coap/adapter.go
```diff
@@ -1,61 +1,85 @@
...
+	sub, err := svc.pubsub.Subscribe(chanID, ch)
+	if err != nil {
+		return ErrFailedSubscription
+	}
+	// Remove if entry already exists.
+	if _, ok := svc.Subs[clientID]; ok == true {
+		svc.Unsubscribe(clientID)
+	}
+	svc.mu.Lock()
+	svc.Subs[clientID] = nats.Observer{
```
anovakovic01: Maybe you should encapsulate set and delete methods for this map that will be thread safe (use lock).

coap/adapter.go
```diff
@@ -1,61 +1,85 @@
...
+	obs, ok := svc.Subs[id]
+	if !ok {
+		return nil
+	}
+	err := obs.Sub.Unsubscribe()
+	if err != nil {
+		return err
+	}
+	svc.mu.Lock()
+	delete(svc.Subs, id)
```
anovakovic01: Same as for setting value.

coap/server.go
```diff
@@ -1,30 +1,107 @@
-package adapter
+package coap
```
anovakovic01: Maybe this file should be part of `api` package?

coap/service.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// Service specifies coap service API.
+type Service interface {
+	mainflux.MessagePublisher
+	// Subscribes to channel with specified id and adds subscription to
+	// service map of subscriptions under given ID.
+	Subscribe(string, string, chan mainflux.RawMessage) error
+	// Unsubscribe method is used to stop observing resource.
+	Unsubscribe(string) error
```
anovakovic01: Can you avoid `Unsubscribe` method by using channel to unsubscribe?

anovakovic01 (changes requested):

coap/adapter.go
drasko: And use `err` here, not `ok`, in order to stay consistent.

coap/api/transport.go
```diff
@@ -0,0 +1,206 @@
...
+	err = binary.Write(buff, binary.LittleEndian, now)
+	if err != nil {
+		return err
+	}
+	msg.MessageID = *counter
+	msg.SetOption(gocoap.Observe, buff.Bytes()[:3])
+	for i := 0; i < 3; i++ {
+		err = gocoap.Transmit(conn, addr, *msg)
+		if err != nil {
+			time.Sleep(5 * time.Millisecond)
```
drasko: Where does this 5ms comes from? Spec?

coap/api/transport.go
```diff
@@ -0,0 +1,206 @@
...
+	*counter++
+	now := time.Now().UnixNano() / int64(time.Millisecond)
+	buff := new(bytes.Buffer)
+	err = binary.Write(buff, binary.LittleEndian, now)
+	if err != nil {
+		return err
+	}
+	msg.MessageID = *counter
+	msg.SetOption(gocoap.Observe, buff.Bytes()[:3])
+	for i := 0; i < 3; i++ {
```
drasko: Where does this 3 times comes from? Spec?

coap/server.go
```diff
@@ -1,30 +1,114 @@
...
+		respMsg = rh.ServeCOAP(conn, addr, &msg)
+	}
+	if respMsg != nil {
+		gocoap.Transmit(conn, addr, *respMsg)
+	}
+}
 
-	r.NotFoundHandler = coap.FuncHandler(notFoundHandler)
+// ListenAndServe binds to the given address and serve requests forever.
+func ListenAndServe(svc Service, mgr manager.ManagerClient, addr string, rh gocoap.Handler) error {
```
drasko: You are oveloading Dustin's Go server. This is OK for now - probably this way will be easier to bump up his version.

However, on the long-run, I think we'll need to rewrite this server because it is pretty unmaintained and has MIT license.

drasko (changes requested):

coap/api/transport.go
dusanb94: No, as I know, protocol specification is not explicit when it comes to this number, as well as timeout between confirmable message and corresponding acknowledgement or reject message.

dborovcanin (commented):
@drasko As I know, protocol spec is not explicit when it comes to a number of repeats before server gives up on notifying the client, nor about timeout between server's confirmable notification and corresponding acknowledgement or reject message from the client. There is no specific reason I took this approach, it looks just fine to me at the moment.

coap/api/transport.go
```diff
@@ -0,0 +1,203 @@
...
+			Publisher: publisher,
+			Protocol:  protocol,
+			Payload:   msg.Payload,
+		}
+
+		if err := svc.Publish(rawMsg); err != nil {
+			if msg.IsConfirmable() {
+				res.Code = gocoap.InternalServerError
+			}
+			return res
```
anovakovic01: This return statement is overhead.

coap/api/transport.go
```diff
@@ -0,0 +1,203 @@
...
+			go handleSub(svc, id, conn, addr, msg, ch)
+			res.AddOption(gocoap.Observe, 0)
+		}
+		return res
+	}
+}
+
+func notify(svc coap.Service, id string, conn *net.UDPConn, addr *net.UDPAddr, msg *gocoap.Message) error {
+	if err := sendMessage(conn, addr, msg); err != nil {
+		return svc.Unsubscribe(id)
```
anovakovic01: Maybe return `sendMessage` error instead of returning `Unsubscribe` error.

coap/api/transport.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	return nil
+}
+
+func sendMessage(conn *net.UDPConn, addr *net.UDPAddr, msg *gocoap.Message) error {
+	var err error
+	now := time.Now().UnixNano() / timestamp
+	buff := new(bytes.Buffer)
+	err = binary.Write(buff, binary.LittleEndian, now)
+	if err != nil {
```
anovakovic01: Can be part of `if` statement.

coap/server.go
```diff
@@ -1,30 +1,129 @@
...
+		Payload:   []byte{},
+	}
+	switch msg.Type {
+	case gocoap.Reset:
+		cid := mux.Var(&msg, "id")
+		res.Type = gocoap.Acknowledgement
+		publisher, err := Authorize(&msg, res, cid)
+		if err != nil {
+			res.Code = gocoap.Unauthorized
+		} else {
```
anovakovic01: Avoid `else` by using `break`.

coap/server.go
```diff
@@ -1,30 +1,129 @@
...
+			id := fmt.Sprintf("%s-%x", publisher, msg.Token)
+			svc.RemoveTimeout(id)
+			err := svc.Unsubscribe(id)
+			if err != nil {
```
anovakovic01: Can be part of `if` statement.

anovakovic01 (changes requested):

coap/service.go
dborovcanin: Yes, but I didn't want to enforce that interface where it is not necessary. I already got `Subscription` from NATS client and I've got a map for CoAP purposes, so I didn't want to enforce that kind of unsubscription handling and to make one more goroutine if it isn't really necessary. Instead, I use what I've already got.

coap/server.go
dborovcanin: Using `break` is probably faster, but since this is typical `if-else` case, I'd prefer to keep it this way.

coap/api/transport.go
dborovcanin: I call it as goroutine elsewhere in the code. I've used an anonymous function in order to avoid labelled loop; I used return instead of `break`, since `break` would only break corresponding `select`.

coap/api/server.go
```diff
@@ -0,0 +1,126 @@
...
+		Payload:   []byte{},
+	}
+	switch msg.Type {
+	case gocoap.Reset:
+		cid := mux.Var(&msg, "id")
+		res.Type = gocoap.Acknowledgement
+		publisher, err := authorize(&msg, res, cid)
+		if err != nil {
+			res.Code = gocoap.Unauthorized
+		} else {
```
anovakovic01: From [Effective Go](https://golang.org/doc/effective_go.html#if): `In the Go libraries, you'll find that when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted.` We should follow this guide and be consistent.

anovakovic01 (changes requested):

coap/api/transport.go
anovakovic01: That is ok, but it seems to me that this should be separate (named) function in that case. IMHO it will improve code readability.

anovakovic01 (dismissed):

Gopkg.lock
```diff
@@ -324,6 +324,7 @@
   version = "v1.1.1"
 
 [[projects]]
+<<<<<<< HEAD
```
mijicd: ... :)

coap/adapter.go
```diff
@@ -1,61 +1,116 @@
...
-	"github.com/dustin/go-coap"
 	"github.com/mainflux/mainflux"
+	"github.com/mainflux/mainflux/coap/nats"
 	broker "github.com/nats-io/go-nats"
-	"go.uber.org/zap"
 )
 
-const protocol string = "coap"
+const (
+	key     string = "key"
```
mijicd: It seems there are no usages for these constants.

coap/adapter.go
```diff
@@ -1,61 +1,116 @@
...
+		}
+	}
+	return nil
 }
 
-// Serve function starts CoAP server
-func (ca *CoAPAdapter) Serve(addr string) error {
-	ca.logger.Info("Starting CoAP server", zap.String("address", addr))
-	return coap.ListenAndServe("udp", addr, ca.COAPServer())
+func (svc *adapterService) Subscribe(chanID, clientID string, ch chan mainflux.RawMessage) error {
```
mijicd: Technically, you're not unsubscribing but replacing the connection, as far as I understand (with resource cleanup). Since that's the case, I would consider making it explicit through function naming (and their implementation), and rename the `Unsubscribe` to `CreateOrReplaceConnection`.

coap/api/server.go
```diff
@@ -0,0 +1,126 @@
...
+
+	mux "github.com/dereulenspiegel/coap-mux"
+	"github.com/mainflux/mainflux/coap"
+	manager "github.com/mainflux/mainflux/manager/client"
+
+	gocoap "github.com/dustin/go-coap"
+)
+
+func authKey(opt interface{}) (string, error) {
+	if opt == nil {
```
mijicd: You can remove this check. It is also performed as part of the type assertion below.

coap/api/server.go
```diff
@@ -0,0 +1,126 @@
...
+		return "", errBadRequest
+	}
+	val, ok := opt.(string)
+	if !ok {
+		return "", errBadRequest
+	}
+	arr := strings.Split(val, "=")
+	if len(arr) != 2 || strings.ToLower(arr[0]) != "key" {
```
mijicd: I suppose this was the place to use that unused constant defined above.

coap/api/server.go
```diff
@@ -0,0 +1,126 @@
...
+	res := &gocoap.Message{
+		Type:      gocoap.NonConfirmable,
+		Code:      gocoap.Content,
+		MessageID: msg.MessageID,
+		Token:     msg.Token,
+		Payload:   []byte{},
+	}
+	switch msg.Type {
+	case gocoap.Reset:
+		cid := mux.Var(&msg, "id")
```
mijicd: And I assume this is the place where the second constant should have been used.

coap/api/server.go
mijicd: I agree with this. Please omit `else`s wherever possible.

coap/api/transport.go
```diff
@@ -0,0 +1,191 @@
...
+	gocoap "github.com/dustin/go-coap"
+	"github.com/mainflux/mainflux"
+)
+
+var (
+	errBadRequest = errors.New("bad request")
+	errBadOption  = errors.New("bad option")
+	auth          manager.ManagerClient
+	maxPktLen     = 1500
+	network       = "udp"
```
mijicd: Shouldn't these be defined as constants?

coap/api/transport.go
```diff
@@ -0,0 +1,191 @@
...
+func sendMessage(conn *net.UDPConn, addr *net.UDPAddr, msg *gocoap.Message) error {
+	var err error
+	now := time.Now().UnixNano() / timestamp
+	buff := new(bytes.Buffer)
+	if err = binary.Write(buff, binary.LittleEndian, now); err != nil {
+		return err
+	}
+	msg.SetOption(gocoap.Observe, buff.Bytes()[:3])
+	timeout := time.Duration(5)
+	// Try to transmit 3 times; each time duplicate timeout between attempts.
```
mijicd: Is this specified in RFC? If so, then proceed. Otherwise, consider utilizing [exponential backoff](https://github.com/cenkalti/backoff).

coap/nats/pubsub.go
```diff
@@ -0,0 +1,64 @@
...
+
+const (
+	maxFailedReqs   = 3
+	maxFailureRatio = 0.6
+	prefix          = "channel"
+)
+
+type natsPublisher struct {
+	nc *broker.Conn
+	cb *gobreaker.CircuitBreaker
```
mijicd: While it is perfectly suitable place for a circuit breaker, I would advise that we drop it for now, for the sake of making the code slightly simpler.

Please note that this might be changed in the future.

coap/nats/pubsub.go
```diff
@@ -0,0 +1,64 @@
...
+		return err
+	}
+	_, err = pubsub.cb.Execute(func() (interface{}, error) {
+		return nil, pubsub.nc.Publish(fmt.Sprintf("%s.%s", prefix, msg.Channel), data)
+	})
+	return err
+}
+
+func (pubsub *natsPublisher) Subscribe(chanID string, ch chan mainflux.RawMessage) (Subscription, error) {
+	var sub *broker.Subscription
```
mijicd: No need for declaring it in advance.

coap/nats/service.go
```diff
@@ -0,0 +1,22 @@
+package nats
+
+import "github.com/mainflux/mainflux"
+
+// Service specifies NATS service API.
+type Service interface {
+	mainflux.MessagePublisher
+	// Subscribes to channel with specified id.
```
mijicd: Follow the godoc recommendations.

coap/nats/service.go
mijicd: Worth considering.

coap/service.go
mijicd: I have already suggested renaming this function. However, I am not sure whether it should be exported or not. Let's try handling this case via channels (message passing), and renaming it if needed. If that works, we can safely "unexport" it.

mijicd (changes requested):
Do the proper rebase (resolve all conflicts). Prune the dependencies. Align with logging and naming used in other services.

coap/nats/broker.go
```diff
@@ -0,0 +1,65 @@
...
+		var rawMsg mainflux.RawMessage
+		if err := proto.Unmarshal(msg.Data, &rawMsg); err != nil {
+			return
+		}
+		channel.Messages <- rawMsg
+	})
+
+	go func() {
+		<-channel.Closed
+		for i := 0; i < 3; i++ {
```
drasko: I still do not understand where this number 3 comes from.

drasko: Btw. did you want to use `maxFailedReqs` const here?

drasko (dismissed):

drasko:
@dusanb94 besides missing documentation, I do not see a single test in this implementation. Are you planning to add some?

coap/nats/broker.go
mijicd: If it is specified by standard, which I assume, please define it as a constant, instead of a magic value.

mijicd (approved):

mijicd:
@dusanb94 the contents of PR are good enough to be merged. Please make sure that remarks regarding documentation and tests given by @drasko are submitted as *NOISSUE* PRs as soon as possible. 

Also, make sure that coap is integrated into our Kubernetes setup.

Issue #179, opened 2018-03-13
docker-compose fails


drasko:
`docker-compose us` is running out-of-order dockers.

Without a backoff and re-try with timeour in services themselves, we can forget about using `docker-compose` and we must fallback to previous starting script.

```
drasko@Marx:~/go/src/github.com/mainflux/mainflux/docker$ docker-compose up
Creating mainflux-manager ... 
Creating mainflux-nats ... 
Creating mainflux-postgres ... 
Creating mainflux-normalizer ... 
Creating mainflux-manager
Creating mainflux-nginx ... 
Creating mainflux-nats
Creating mainflux-normalizer
Creating mainflux-nginx
Creating mainflux-nats ... done
Creating mainflux-http ... 
Creating mainflux-mqtt ... 
Creating mainflux-mqtt
Creating mainflux-mqtt ... done
Attaching to mainflux-manager, mainflux-nats, mainflux-normalizer, mainflux-postgres, mainflux-nginx, mainflux-http, mainflux-mqtt
mainflux-manager | {"error":"dial tcp: lookup postgres on 10.20.0.1:53: no such host","ts":"2018-03-13T02:17:55.877664432Z"}
mainflux-nats   | [1] 2018/03/13 02:17:56.008535 [INF] Starting nats-server version 1.0.2
mainflux-nats   | [1] 2018/03/13 02:17:56.008625 [INF] Starting http monitor on 0.0.0.0:8222
mainflux-normalizer | {"error":"Failed to connect: nats: no servers available for connection","ts":"2018-03-13T02:17:56.16486139Z"}
mainflux-nats   | [1] 2018/03/13 02:17:56.008662 [INF] Listening for client connections on 0.0.0.0:4222
mainflux-nats   | [1] 2018/03/13 02:17:56.008666 [INF] Server is ready
mainflux-postgres | The files belonging to this database system will be owned by user "postgres".
mainflux-postgres | This user must also own the server process.
mainflux-postgres | 
mainflux-nats   | [1] 2018/03/13 02:17:56.134986 [INF] Listening for route connections on 0.0.0.0:6222
mainflux-nginx  | 2018/03/13 02:17:56 [emerg] 1#1: host not found in upstream "mainflux-manager:8180" in /etc/nginx/nginx.conf:58
mainflux-postgres | The database cluster will be initialized with locale "en_US.utf8".
mainflux-http   | {"error":"nats: no servers available for connection","ts":"2018-03-13T02:17:56.689929955Z"}
[...]

PR #187, opened 2018-03-16
Fixed docker composition

nmarcetic:
Related to #179 
Signed-off-by: nmarcetic <n.marcetic86@gmail.com>


mijicd (approved):
Accepted as a step toward the Swarm solution.

PR #188, opened 2018-03-16
MF-171 - Extract websocket adapter as separate service

### LINKED ISSUE #171: Extract websocket adapter to separate service
At the moment, Mainflux relies on [MQTT adapter](https://github.com/mainflux/mqtt-adapter) to provide websocket interface toward the system. This approach should be abandoned by custom golang websocket adapter implementation.

anovakovic01:
This pull request resolves #171.

**Summary:**

- added WebSocket adapter that enables publishing and receiving of messages
- added logging and metrics to adapter
- added authorization on WebSocket handshake
- added test for WebSocket adapter publish method
- added API tests
- added WebSocket library in dependecies
- added Web Socket section in docs
- exposed version and metrics endpoints
- version raised to `0.2.0`


cmd/ws/main.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+const (
+	defPort       = "8180"
+	defNatsURL    = broker.DefaultURL
+	defManagerURL = "http://localhost:8180"
+	envPort       = "MF_WS_ADAPTER_PORT"
+	envNatsURL    = "MF_NATS_URL"
+	envManagerURL = "MF_MANAGER_URL"
+	topic         = "src.*"
```
drasko: How do we prevent loopback from messages on the topic `src.ws`, i.e. messages WS client is publishing?

mijicd: It actually makes sense, since 2 websocket clients should be able to “hear” eachother. However, it must be carefully handled to avoid making “self” dos.

drasko: With of-the-shelf MQTT this is a problem - MQTT loopbacks, but does not keep any "publisher" information, so you can not even drop the message you sent yourself. Here I suppose it will be easier.

anovakovic01: Currently, I am not sending message to subscriber who is publisher of that message at the same time. I determine if message should be sent by checking publisher id contained within the message.
https://github.com/anovakovic01/mainflux/blob/6a5605ebffb40229a7a164f708e2a36693daf742/ws/adapter.go#L59

drasko: @anovakovic01 yes, I see... OK for now, however we will have to think how we can improve design in the future to avoid these loopbacks, as on every `send` currently we have one additional `receive + ignore`.

anovakovic01: @drasko I'll try to think of better solution. Of course, I'm open for suggestions.

mijicd: @anovakovic01 @drasko see my second comment, it’s one way built around local state keeping instead of subscribe and filter.

drasko: I do not have one right now :).

But as I said - situation is even worst with MQTT, as we can change `on_publish` just to limited extent. Plus it is state-full protocol, and these brokers have their way of clustering often.

So, I think we have to find some elegant solution that we will apply to all adapters (all 4 of them face the same issue).

anovakovic01: @mijicd It's definitely better. By doing that we would avoid receive message from NATS that Drasko mentioned, so only ignore part stays.

ws/adapter.go
```diff
@@ -0,0 +1,83 @@
...
+	// HandleMessage that is received from message broker.
+	HandleMessage(*broker.Msg)
+
+	// AddConnection adds new client ws connection for given client and channel.
+	AddConnection(string, string, *websocket.Conn)
+}
+
+type adapterService struct {
+	pub    mainflux.MessagePublisher
+	conns  map[string]map[string]*websocket.Conn
```
mijicd: It is way simpler to use structs as keys. Check out https://blog.golang.org/go-maps-in-action, section *Key types*.

ws/adapter.go
```diff
@@ -0,0 +1,83 @@
...
+// Service contains publish and subscribe methods necessary for
+// message transfer.
+type Service interface {
+	mainflux.MessagePublisher
+
+	// HandleMessage that is received from message broker.
+	HandleMessage(*broker.Msg)
+
+	// AddConnection adds new client ws connection for given client and channel.
+	AddConnection(string, string, *websocket.Conn)
```
mijicd: Use the structure you've defined below (`idPair`), promote it, and pass it as a parameter here, instead of passing identifiers separately.

ws/api/transport.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/go-zoo/bone"
+	"github.com/gorilla/websocket"
+	"github.com/mainflux/mainflux"
+	manager "github.com/mainflux/mainflux/manager/client"
+	"github.com/mainflux/mainflux/ws"
+)
+
+const protocol = "ws"
+
+var errUnauthorizedAccess = errors.New("missing or invalid credentials provided")
```
mijicd: Add it to the group below.

ws/adapter.go
```diff
@@ -0,0 +1,83 @@
...
+		return
+	}
+	var rawMsg mainflux.RawMessage
+	if err := json.Unmarshal(msg.Data, &rawMsg); err != nil {
+		as.logger.Log("error", fmt.Sprintf("Unmarshalling failed: %s", err))
+		return
+	}
+
+	cid := rawMsg.Channel
+	for pid, conn := range as.conns[cid] {
```
mijicd: Publish can be wrapped in go routine.

ws/adapter.go
```diff
@@ -0,0 +1,83 @@
...
+		conns:  make(map[string]map[string]*websocket.Conn),
+		logger: logger,
+	}
+}
+
+func (as *adapterService) Publish(msg mainflux.RawMessage) error {
+	return as.pub.Publish(msg)
+}
+
+func (as *adapterService) HandleMessage(msg *broker.Msg) {
```
mijicd: Based on the discussion we had above, I would advise that you make this method private and run it as part of the `Publish`, i.e. no need to publish to NATS, and then consume it back (for now).

mijicd (changes requested):

ws/adapter.go
```diff
@@ -0,0 +1,84 @@
...
+
+var _ Service = (*adapterService)(nil)
+
+// Service contains publish and subscribe methods necessary for
+// message transfer.
+type Service interface {
+	mainflux.MessagePublisher
+
+	// BroadcastMessage broadcasts raw message to channel.
+	BroadcastMessage(mainflux.RawMessage)
```
mijicd: Rename to `Broadcast`.

ws/adapter.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// AddConnection adds new client ws connection for given client and channel.
+	AddConnection(IDPair, *websocket.Conn)
```
mijicd: Rename `IDPair` to `Subscription`.

ws/adapter.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (as *adapterService) Publish(msg mainflux.RawMessage) error {
+	as.BroadcastMessage(msg)
+	return as.pub.Publish(msg)
+}
+
+func (as *adapterService) BroadcastMessage(msg mainflux.RawMessage) {
+	chanID := msg.Channel
+	for pubID, conn := range as.conns[chanID] {
+		pid, sock := pubID, conn
```
mijicd: No need for this. If you are worried about enclosing scope of the variable, it is better to pass these as parameters to the goroutine.

ws/api/transport.go
```diff
@@ -0,0 +1,91 @@
...
+func MakeHandler(svc ws.Service, mc manager.ManagerClient) http.Handler {
+	auth = mc
+
+	mux := bone.New()
+	mux.GetFunc("/channels/:id/messages", handshake(svc))
+
+	return mux
+}
+
+func handshake(svc ws.Service) func(http.ResponseWriter, *http.Request) {
```
mijicd: It feels like this function is doing too much. As far as I understand the code, it handles both the connection establishing and listening to its "events".

ws/mocks/messages.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"errors"
+
+	"github.com/mainflux/mainflux"
+)
+
+var _ mainflux.MessagePublisher = (*publisher)(nil)
+
+var (
```
mijicd: Either group it with the type check above, or drop the group completely. No need for having a one-member variable group.

ws/swagger.yml
```diff
@@ -0,0 +1,38 @@
+swagger: "2.0"
```
mijicd: Swagger doc is a nice touch, though it might not be as suitable for documenting websocket APIs. Let's keep it for now.

ws/adapter_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/mainflux/mainflux/ws"
+	"github.com/mainflux/mainflux/ws/mocks"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/mainflux/mainflux"
+)
+
+var (
```
mijicd: Please drop the variable grouping. Also, I am not sure about this error, since it is duplicated in both mocks and ws_test. Perhaps we could promote it to the domain.

mijicd (changes requested):

cmd/ws/main.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+const (
+	defPort       = "8180"
+	defNatsURL    = broker.DefaultURL
+	defManagerURL = "http://localhost:8180"
+	envPort       = "MF_WS_ADAPTER_PORT"
+	envNatsURL    = "MF_NATS_URL"
+	envManagerURL = "MF_MANAGER_URL"
+	topic         = "src.ws"
```
mijicd: As far as I can see this constant is not used at all, so either use it or drop it. More general notes, since it is obvious that some naming standard needs to be introduced:

- Rename to `subject` just to avoid ambiguities. Also rename it in the HTTP adapter since that's the "source of ambiguity".
- Should this be the subject, or the prefix? Aren't we naming the subjects as `src.<protocol>.<channel>`?

ws/README.md
```diff
@@ -0,0 +1,54 @@
...
+a compose file template that can be used to deploy the service container locally:
+
+```yaml
+version: "2"
+services:
+  adapter:
+    image: mainflux/ws:[version]
+    container_name: [instance name]
+    ports:
+      - [host machine port]:8180
```
mijicd: Please update to: `- [host machine port]:[configured port]`.

ws/README.md
```diff
@@ -0,0 +1,54 @@
...
+To start the service outside of the container, execute the following shell script:
+
+```bash
+# download the latest version of the service
+go get github.com/mainflux/mainflux
+
+cd $GOPATH/src/github.com/mainflux/mainflux/cmd/ws
+
+# compile the app; make sure to set the proper GOOS value
+CGO_ENABLED=0 GOOS=[platform identifier] go build -ldflags "-s" -a -installsuffix cgo -o app
```
mijicd: With `Makefile` almost completed you can safely remove this reference from the docs, and replace it with `make <service_name>`.

ws/adapter.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+type adapterService struct {
+	pub    mainflux.MessagePublisher
+	logger log.Logger
+}
+
+// New instantiates the domain service implementation.
+func New(pub mainflux.MessagePublisher, logger log.Logger) Service {
+	return &adapterService{
```
mijicd: Please change to `&adapterService{pub, logger}`.

ws/api/transport.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/go-zoo/bone"
+	"github.com/gogo/protobuf/proto"
+	"github.com/gorilla/websocket"
+	"github.com/mainflux/mainflux"
+	manager "github.com/mainflux/mainflux/manager/client"
+	"github.com/mainflux/mainflux/ws"
+	broker "github.com/nats-io/go-nats"
+)
+
+const topic string = "src.*"
```
mijicd: See the naming remark given above.

ws/api/transport.go
```diff
@@ -0,0 +1,109 @@
...
+		if err != nil {
+			logger.Log("error", fmt.Sprintf("Failed to upgrade connection to websocket: %s", err))
+			return
+		}
+		socket := ws.NewSocket(conn)
+
+		// Subscribe to all NATS subjects.
+		topic := fmt.Sprintf("%s.%s", topic, sub.ChanID)
+		brokerSub, err := nc.Subscribe(topic, func(msg *broker.Msg) {
+			if msg == nil {
```
mijicd: This does not belong here. I would expect callback to be the part of the service.

ws/adapter.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// ErrFailedMessagePublish indicates that message publishing failed.
+	ErrFailedMessagePublish = errors.New("failed to publish message")
+	// ErrFailedMessageBroadcast indicates that message broadcast failed.
+	ErrFailedMessageBroadcast = errors.New("failed to broadcast message")
+)
+
+// Service contains publish and subscribe methods necessary for
+// message transfer.
+type Service interface {
```
mijicd: It feels like we should move `Broadcast` to the `MessagePublisher` interface. Also, it makes sense to create `MessageSubscriber` interface as well.

ws/adapter.go
```diff
@@ -0,0 +1,89 @@
...
+			Publisher: sub.PubID,
+			Protocol:  protocol,
+			Payload:   payload,
+		}
+		as.Publish(msg)
+	}
+}
+
+// Subscription contains publisher and channel id.
+type Subscription struct {
```
mijicd: Actually, this structure might be even more usefully than I originally thought. It makes sense to reuse it from other 2-way services, hence, it should be moved to `mainflux` package.

mijicd (changes requested):
Getting there.

ws/nats/publisher.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"fmt"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/mainflux/mainflux"
+	broker "github.com/nats-io/go-nats"
+)
+
+const topic string = "src.ws"
```
drasko: Needs to be aligned with new "per channel" publish.

drasko (dismissed):

ws/api/transport.go
mijicd: @anovakovic01 this remark still holds. Please take care of it! Also, consider using `http.HandlerFunc` as return type.

ws/api/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+		mm.latency.With("method", "broadcast").Observe(time.Since(begin).Seconds())
+	}(time.Now())
+
+	return mm.svc.Broadcast(msg, sendMsg)
+}
+
+func (mm *metricsMiddleware) Subscribe(channel string, onMessage func(mainflux.RawMessage)) (mainflux.Subscription, error) {
+	defer func(begin time.Time) {
+		mm.counter.With("method", "subscribe").Add(1)
+		mm.latency.With("method", "broadcast").Observe(time.Since(begin).Seconds())
```
mijicd: Why `broadcast`?

pubsub.go
```diff
@@ -0,0 +1,33 @@
...
+// MessageBroker specifies a message broadcasting API.
+type MessageBroker interface {
+	MessagePubSub
+	// Broadcasts raw message to channel.
+	Broadcast(RawMessage, func(RawMessage) error) error
+}
+
+// Subscription specifies subscription interface.
+type Subscription interface {
+	Unsubscribe() error
```
mijicd: I suggested that you share a structure. Subscription interface in this form is not that useful.

mijicd (changes requested):

pubsub.go
```diff
@@ -0,0 +1,33 @@
...
+
+// MessageSubscriber specifies a message subscription API.
+type MessageSubscriber interface {
+	// Subscirbes to channel. A non-nil error is returned to indicate
+	// operation failure.
+	Subscribe(string, func(RawMessage)) (Subscription, error)
+}
+
+// MessagePubSub specifies a message publishing and subscription API.
+type MessagePubSub interface {
```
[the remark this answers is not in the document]
anovakovic01: NATS implementation doesn't need broadcast method, that is why this interface exists. I will delete it if you think that this is not strong enough reason to leave it here.

pubsub.go
```diff
@@ -1,5 +1,11 @@
 package mainflux
 
+// WriteMessage is used to write message to channel.
+type WriteMessage func(RawMessage) error
```
mijicd: Strange type alias.

ws/adapter.go
```diff
@@ -48,27 +40,22 @@ func (as *adapterService) Publish(msg mainflux.RawMessage) error {
...
-func (as *adapterService) Subscribe(channel string, onMessage func(mainflux.RawMessage)) (mainflux.Subscription, error) {
-	sub, err := as.pubsub.Subscribe(channel, onMessage)
+func (as *adapterService) Subscribe(sub mainflux.Subscription, write mainflux.WriteMessage, read mainflux.ReadMessage) (func(), error) {
+	unsubscribe, err := as.pubsub.Subscribe(sub, write, nil)
 	if err != nil {
 		as.logger.Log("error", fmt.Sprintf("Failed to subscribe to a channel: %s", err))
 		return nil, ErrFailedSubscription
 	}
-	return sub, nil
+	go as.listen(sub, read, func() {
```
mijicd: Do you really need "wrapper" around `unsubscribe`?

ws/adapter.go
```diff
@@ -48,27 +40,22 @@ func (as *adapterService) Publish(msg mainflux.RawMessage) error {
...
+		unsubscribe()
+	})
+	return nil, nil
 }
 
-func (as *adapterService) Listen(socket Socket, sub Subscription, onClose func()) {
+func (as *adapterService) listen(sub mainflux.Subscription, read mainflux.ReadMessage, onClose func()) {
```
mijicd: This method signature went from bad to really bad.

mijicd (changes requested):
Recent changes obfuscated the code layout. Since the PR has been marked as work-in-progress for some time, I recommend that we pair on this and resolve.

mijicd (commented):
@anovakovic01 sorry for the late response. Here's the list of changes I recommend:

- Remove `MessageSubscriber` and `MessagePubSub` interfaces from `mainflux` package together with related artifacts. 

- **IMPORTANT:** use channels to communicate events between the transport and business layers! Channels should be the part of the signature above.

- Thus being said, handle all protocol-specific concerns in `transport.go` (including channel instantiation).

- Remove circuit breaker from NATS. It is useful, don't get me wrong, but let's introduce it gradually.

- Remove swagger docs, and document the way to use the adapter in MF docs.

- Logging should be handled in logging middleware. 

- Review errors. It does not seem all of the defined errors are needed.

- Report only metrics related to message publishing: error rate and request latency.

docs/getting-started.md
```diff
@@ -252,4 +252,15 @@ To publish message over channel, client should send following request:
 curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X POST -H "Content-Type: application/senml+json" -H "Authorization: <client_token>" https://localhost/channels/<channel_id>/messages -d -d '[{"bn":"some-base-name:","bt":1.276020076001e+09, "bu":"A","bver":5, "n":"voltage","u":"V","v":120.1}, {"n":"current","t":-5,"v":1.2}, {"n":"current","t":-4,"v":1.3}]'
 ```
 
-Note that you should always send array of messages in senML format.
\ No newline at end of file
+Note that you should always send array of messages in senML format.
+
+### Web Socket
```
mijicd: Please change to `### WebSocket`.

ws/adapter.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// Service specifies web socket service API.
+type Service interface {
+	mainflux.MessagePublisher
+	// Subscribes to channel with specified id.
+	Subscribe(string, Channel) error
+}
+
+type adapterService struct {
```
mijicd: Please split this file in a same way it's done for manager service.

ws/api/transport.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// Extract ID from /channels/:id/messages.
+	chanID := bone.GetValue(r, "id")
+	if !govalidator.IsUUID(chanID) {
+		return subscription{}, errNotFound
+	}
+
+	pubID, err := auth.CanAccess(chanID, authKey)
+	if err != nil {
+		return subscription{}, errUnauthorizedAccess
```
mijicd: Manager client exposes `ErrUnauthorizedAccess`. Please reuse it instead of defining custom one.

ws/api/transport_test.go
```diff
@@ -0,0 +1,105 @@
...
+	return ws.New(pubsub)
+}
+
+func newHTTPServer(svc ws.Service, mc manager.ManagerClient) *httptest.Server {
+	mux := api.MakeHandler(svc, mc, log.New(os.Stdout))
+	return httptest.NewServer(mux)
+}
+
+func newManagerServer() *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
mijicd: Please split this and name function.

ws/api/transport.go
```diff
@@ -0,0 +1,141 @@
...
+		channel := ws.Channel{make(chan mainflux.RawMessage), make(chan bool)}
+		if err = svc.Subscribe(sub.chanID, channel); err != nil {
+			logger.Warn(fmt.Sprintf("Failed to subscribe to NATS subject: %s", err))
+			w.WriteHeader(http.StatusExpectationFailed)
+			return
+		}
+		go broadcast(conn, channel.Messages)
+
+		// Start listening for messages from NATS.
+		go listen(svc, sub, conn, channel)
```
mijicd: Connection and channel should be part of the subscription structure. That will simplify `listen` signature. We can rethink whether these should be grouped into some socket/subscription structure.

ws/nats/pubsub.go
```diff
@@ -0,0 +1,71 @@
...
+	data, err := proto.Marshal(&msg)
+	if err != nil {
+		return err
+	}
+
+	return pubsub.nc.Publish(fmt.Sprintf("%s.%s", prefix, msg.Channel), data)
+}
+
+func (pubsub *natsPubSub) Subscribe(chanID string, channel ws.Channel) error {
+	var sub *broker.Subscription
```
mijicd, line 51: Do you really need it?

mijicd (changes requested):

ws/nats/pubsub.go
anovakovic01: I need it because I use it in subscription callback to unsubscribe.

mijicd: Agreed.

mijicd (approved):

PR #192, opened 2018-03-19
Use Protobuf to represent messages

### LINKED ISSUE #163: Reduce NATS event serialization footprint
Instead of using JSON as serialization format of all events published to NATS topics, binary serialization formats should be evaluated. Evaluation criteria should be performance, simplicity (i.e. learning curve) and library support.

Once we decide on a format, all publishers should use it instead of JSON.

mijicd:
Resolves #163, relates to #191.


mijicd (commented):
Created this PR on behalf of @drasko. Looks good to me, it should be merged once CI passes.

PR #205, opened 2018-03-21
Simplify Makefile

drasko:


mijicd (approved):
This is the step in the right direction. Eventually, I would like this to converge to principal build tool, i.e. to have `make release` and `make latest` as we discussed.

PR #211, opened 2018-03-30
MF-170 - Add manager API unit tests

### LINKED ISSUE #170: Raise the test coverage
We have tests only for the mainflux/manager and mainflux/manager/api. I think we should write more tests in order to be able to change things without worrying that we'll break something.

I will start writing some tests. and I want to start with the http and writer services. I will use the same tools/packages for testing which are used in the [manager_test.go](https://github.com/mainflux/mainflux/blob/master/manager/manager_test.go).

What do you think about this? Are there any parts that it does not make sense to write tests for?

Best Regards,
Jovan

anovakovic01:
This pull request resolves #170.

**Summary:**
- Add unit test for every manager API endpoint
- Update swagger documentation
- Fix connection request validation
- Fix EOF error handling
- Add invalid content type error handling
- Update README with codecov badge

### DIFF
--- a/manager/api/transport_test.go
+++ b/manager/api/transport_test.go
@@ -1,0 +1,718 @@
+package api_test
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/mainflux/mainflux/manager"
+	"github.com/mainflux/mainflux/manager/api"
+	"github.com/mainflux/mainflux/manager/mocks"
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	contentType  = "application/json; charset=utf-8"
+	invalidEmail = "userexample.com"
+	wrongID      = "123e4567-e89b-12d3-a456-000000000042"
+)
+
+var (
+	user    = manager.User{"user@example.com", "password"}
+	client  = manager.Client{Type: "app", Name: "test_app", Payload: "test_payload"}
+	channel = manager.Channel{Name: "test"}
+)
+
+type testRequest struct {
+	method      string
+	url         string
+	contentType string
+	token       string
+	body        io.Reader
+}
+
+func newService() manager.Service {
+	users := mocks.NewUserRepository()
+	clients := mocks.NewClientRepository()
+	channels := mocks.NewChannelRepository(clients)
+	hasher := mocks.NewHasher()
+	idp := mocks.NewIdentityProvider()
+
+	return manager.New(users, clients, channels, hasher, idp)
+}
+
+func newServer(svc manager.Service) *httptest.Server {
+	mux := api.MakeHandler(svc)
+	return httptest.NewServer(mux)
+}
+
+func makeRequest(client *http.Client, testReq testRequest) (*http.Response, error) {
+	req, err := http.NewRequest(testReq.method, testReq.url, testReq.body)
+	if err != nil {
+		return nil, err
+	}
+	if testReq.token != "" {
+		req.Header.Set("Authorization", testReq.token)
+	}
+	if testReq.contentType != "" {
+		req.Header.Set("Content-Type", testReq.contentType)
+	}
+	return client.Do(req)
+}
+
+func TestRegister(t *testing.T) {
+	svc := newService()
+	ts := newServer(svc)
+	defer ts.Close()
+	client := ts.Client()
+	url := ts.URL
+
+	body, _ := json.Marshal(user)
+	data := string(body)
+	invalidBody, _ := json.Marshal(manager.User{Email: invalidEmail, Password: "password"})
+	invalidData := string(invalidBody)
+
+	cases := []struct {
+		desc   string
+		req    string
+		status int
+	}{
+		{"register new user", data, http.StatusCreated},
+		{"register existing user", data, http.StatusConflict},
+		{"register user with invalid email address", invalidData, http.StatusBadRequest},
+		{"register user with invalid request format", "{", http.StatusBadRequest},
+		{"register user with empty JSON request", "{}", http.StatusBadRequest},
+		{"register user with empty request", "", http.StatusBadRequest},
+	}
+
+	for _, tc := range cases {
+		res, err := makeRequest(client, testRequest{
+			method:      http.MethodPost,
+			url:         fmt.Sprintf("%s/users", url),
+			contentType: contentType,
+			body:        strings.NewReader(tc.req),
+		})
+		assert.Nil(t, err, fmt.Sprintf("%s: unexpected error %s", tc.desc, err))
+		assert.Equal(t, tc.status, res.StatusCode, fmt.Sprintf("%s: expected status code %d got %d", tc.desc, tc.status, res.StatusCode))
+	}
+}
+
+func TestLogin(t *testing.T) {
+	svc := newService()
+	ts := newServer(svc)
+	defer ts.Close()
+	client := ts.Client()
+	url := ts.URL
+
+	token, _ := json.Marshal(map[string]string{"token": user.Email})
+	tokenData := string(token)
+	credentials, _ := json.Marshal(user)
+	data := string(credentials)
+	invalidEmailCredentials, _ := json.Marshal(manager.User{Email: invalidEmail, Password: "password"})
+	invalidEmailData := string(invalidEmailCredentials)
+	invalidCredentials, _ := json.Marshal(manager.User{"user@example.com", "invalid_password"})
+	invalidData := string(invalidCredentials)
+	nonexistentCredentials, _ := json.Marshal(manager.User{"non-existentuser@example.com", "pass"})
+	nonexistentData := string(nonexistentCredentials)
+	svc.Register(user)
+
+	cases := []struct {
+		desc   string
+		req    string
+		status int
+		res    string
+	}{
+		{"login with valid credentials", data, http.StatusCreated, tokenData},
+		{"login with invalid credentials", invalidData, http.StatusForbidden, ""},
+		{"login with invalid email address", invalidEmailData, http.StatusBadRequest, ""},
+		{"login non-existent user", nonexistentData, http.StatusForbidden, ""},
+		{"login with invalid request format", "{", http.StatusBadRequest, ""},
+		{"login with empty JSON request", "{}", http.StatusBadRequest, ""},
+		{"login with empty request", "", http.StatusBadRequest, ""},
+	}
+
+	for _, tc := range cases {
+		res, err := makeRequest(client, testRequest{
+			method:      http.MethodPost,
+			url:         fmt.Sprintf("%s/tokens", url),
+			contentType: contentType,
+			body:        strings.NewReader(tc.req),
+		})
+		assert.Nil(t, err, fmt.Sprintf("%s: unexpected error %s", tc.desc, err))
+		body, err := ioutil.ReadAll(res.Body)
+		assert.Nil(t, err, fmt.Sprintf("%s: unexpected error %s", tc.desc, err))
+		token := strings.Trim(string(body), "\n")
+
+		assert.Equal(t, tc.status, res.StatusCode, fmt.Sprintf("%s: expected status code %d got %d", tc.desc, tc.status, res.StatusCode))
+		assert.Equal(t, tc.res, token, fmt.Sprintf("%s: expected body %s got %s", tc.desc, tc.res, token))
+	}
+}
+
+func TestAddClient(t *testing.T) {
+	svc := newService()
+	ts := newServer(svc)
+	defer ts.Close()
+	cli := ts.Client()
+	url := ts.URL
+
+	client, _ := json.Marshal(client)
+	data := string(client)
+	invalidClient, _ := json.Marshal(manager.Client{
+		Type:    "foo",
+		Name:    "invalid_client",
+		Payload: "some_payload",
+	})
+	invalidData := string(invalidClient)
+	svc.Register(user)
+
+	cases := []struct {
+		desc   string
+		req    string
+		auth   string
+		status int
+	}{
+		{"add valid client", data, user.Email, http.StatusCreated},
+		{"add client with invalid data", invalidData, user.Email, http.StatusBadRequest},
+		{"add client with invalid auth token", data, "invalid_token", http.StatusForbidden},
+		{"add client with invalid request format", "}", user.Email, http.StatusBadRequest},
+		{"add client with empty JSON request", "{}", user.Email, http.StatusBadRequest},
+		{"add client with empty request", "", user.Email, http.StatusBadRequest},
+	}
+
+	for _, tc := range cases {
+		res, err := makeRequest(cli, testRequest{
+			method:      http.MethodPost,
+			url:         fmt.Sprintf("%s/clients", url),
+			contentType: contentType,
+			token:       tc.auth,
+			body:        strings.NewReader(tc.req),
+		})
+		assert.Nil(t, err, fmt.Sprintf("%s: unexpected error %s", tc.desc, err))
+		assert.Equal(t, tc.status, res.StatusCode, fmt.Sprintf("%s: expected status code %d got %d", tc.desc, tc.status, res.StatusCode))
+	}
+}
+
+func TestUpdateClient(t *testing.T) {
+	svc := newService()
+	ts := newServer(svc)
+	defer ts.Close()
+	cli := ts.Client()
+	url := ts.URL
+
+	update, _ := json.Marshal(client)
+	data := string(update)
+	invalidUpdate, _ := json.Marshal(manager.Client{
+		Type:    "foo",
+		Name:    client.Name,
+		Payload: client.Payload,
+	})
+	invalidData := string(invalidUpdate)
+	svc.Register(user)
+	id, _ := svc.AddClient(user.Email, client)
+
+	cases := []struct {
+		desc   string
+		req    string
+		id     string
+		auth   string
+		status int
+	}{
+		{"update existing client", data, id, user.Email, http.StatusOK},
+		{"update non-existent client", data, wrongID, user.Email, http.StatusNotFound},
+		{"update client with invalid id", data, "1", user.Email, http.StatusNotFound},
+		{"update client with invalid data", invalidData, id, user.Email, http.StatusBadRequest},
+		{"update client with invalid user token", data, id, invalidEmail, http.StatusForbidden},
+		{"update client with invalid data format", "{", id, user.Email, http.StatusBadRequest},
+		{"update client with empty JSON request", "{}", id, user.Email, http.StatusBadRequest},
+		{"update client with empty request", "", id, user.Email, http.StatusBadRequest},
+	}
+
+	for _, tc := range cases {
+		res, err := makeRequest(cli, testRequest{
+			method:      http.MethodPut,
+			url:         fmt.Sprintf("%s/clients/%s", url, tc.id),
+			contentType: contentType,
+			token:       tc.auth,
+			body:        strings.NewReader(tc.req),
+		})
+		assert.Nil(t, err, fmt.Sprintf("%s: unexpected error %s", tc.desc, err))
+		assert.Equal(t, tc.status, res.StatusCode, fmt.Sprintf("%s: expected status code %d got %d", tc.desc, tc.status, res.StatusCode))
+	}
+}
+
+func TestViewClient(t *testing.T) {
+	svc := newService()
+	ts := newServer(svc)
+	defer ts.Close()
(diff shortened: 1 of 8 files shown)


drasko (dismissed):
LGTM

chombium:
@anovakovic01 this looks cool, thanks for your work.
One question: I don't see the pact file in the commit. What are your thoughts on storing the pact files?
We need them available somewhere in order to verify the expected behavior of the "service provider".

I guess they are stored in the pact directory which you've added to the .gitignore.
I'll checkout your branch,  play a bit and give my comments.

Best Regards,
Jovan

anovakovic01:
@chombium Thanks for feedback. 

I didn't commit pact files because they are generated by running tests in `consumer_test`, so I think that you have everything you need to run pact tests. Also, if I am not mistaken, pact files are usually shared over pact broker which we won't need in our case (at least for now). We are not using pact in standard way because we don't have consumer service. Instead I wrote test use cases in `consumer_test` and generated pact files from them. If you think there are any reasons we should commit pact files, let me know.

Best Regards,
Aleksandar

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+	svc := newService()
+	mux := api.MakeHandler(svc)
+	ts := httptest.NewServer(mux)
+	defer ts.Close()
+	client := ts.Client()
+
+	test(svc, ts.URL, client)
+}
+
+func sendRequest(client *http.Client, method string, url string, contentType string, token string, body io.Reader) (*http.Response, error) {
```
mijicd: Please rename to `makeRequest`, and wrap all of the parameters in a structure called `testRequest`.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+
+		cases := []struct {
+			desc   string
+			req    string
+			status int
+		}{
+			{"register new user", string(validBody), http.StatusCreated},
+			{"try to register existing user", string(validBody), http.StatusConflict},
+			{"try to register user with invalid email address", string(invalidBody), http.StatusBadRequest},
+			{"try to register with invalid request format", "{", http.StatusBadRequest},
```
mijicd: Empty request would be viable test case as well.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+		validBody, _ := json.Marshal(validUser)
+		invalidBody, _ := json.Marshal(invalidUser)
+
+		cases := []struct {
+			desc   string
+			req    string
+			status int
+		}{
+			{"register new user", string(validBody), http.StatusCreated},
+			{"try to register existing user", string(validBody), http.StatusConflict},
```
mijicd: Please drop "try to". This applies for all specs.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+	users := mocks.NewUserRepository()
+	clients := mocks.NewClientRepository()
+	channels := mocks.NewChannelRepository()
+	hasher := mocks.NewHasher()
+	idp := mocks.NewIdentityProvider()
+
+	return manager.New(users, clients, channels, hasher, idp)
+}
+
+func runService(test func(manager.Service, string, *http.Client)) {
```
mijicd: Instead of passing a test callback, rename this method to `newServer` and return its instance in the same way it is done at `newService`. You can obtain all of the data in the test itself.

Just to be clear, this is not a bad way to remove duplication, but I'd prefer that we are as explicit as possible in our test cases.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+		nonexistentCredentials, _ := json.Marshal(nonexistentUser)
+		svc.Register(validUser)
+
+		cases := []struct {
+			desc   string
+			req    string
+			status int
+			res    string
+		}{
+			{"login with valid credentials", string(validCredentials), http.StatusCreated, string(token)},
```
mijicd: These `string(...)` are hard to read. Don't "torture" the table, do the cast during test data preparation. This applies to each of the further occurrences.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+	"github.com/mainflux/mainflux/manager/api"
+	"github.com/mainflux/mainflux/manager/mocks"
+	"github.com/stretchr/testify/assert"
+)
+
+const contentType = "application/json; charset=utf-8"
+
+var (
+	validUser       = manager.User{"user@example.com", "password"}
+	invalidPassUser = manager.User{"user@example.com", "invalid_password"}
```
mijicd: Since this is used only once, you can drop it from here and create an instance locally, based on the `validUser`.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+	invalidUser     = manager.User{"userexample.com", "password"}
```
mijicd: This "fellow" exists for the sole purpose of providing a false e-mail value. Define a constant that contains "bad" value instead (this can be reused for other purposes as well, see other tests as an example).

The only exception to the aforementioned statement is user registration. For that purpose you can create local instance based on the `validUser` but using the constant I mentioned as an e-mail.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+	nonexistentUser = manager.User{"non-existentuser@example.com", "pass"}
+	validClient     = manager.Client{Type: "app", Name: "test_app", Payload: "test_payload"}
+	invalidClient   = manager.Client{Type: "foo", Name: "invalid_client", Payload: "some_payload"}
```
mijicd: Same remark as `invalidUser` and `invalidPassUser`.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+	"github.com/mainflux/mainflux/manager"
+	"github.com/mainflux/mainflux/manager/api"
+	"github.com/mainflux/mainflux/manager/mocks"
+	"github.com/stretchr/testify/assert"
+)
+
+const contentType = "application/json; charset=utf-8"
+
+var (
+	validUser       = manager.User{"user@example.com", "password"}
```
mijicd: Please drop "valid" prefix from all valid entities. This will improve readability together with other changes.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+func sendRequest(client *http.Client, method string, url string, contentType string, token string, body io.Reader) (*http.Response, error) {
+	req, err := http.NewRequest(method, url, body)
+	if err != nil {
+		return nil, err
+	}
+	if token != "" {
+		req.Header.Set("Authorization", token)
+	}
+	if contentType != "" {
+		req.Header.Set("Content-Type", contentType)
```
mijicd: Add test cases that cover the missing content type. Since this might be the edge case, we should expect `415` error.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+			res, err := sendRequest(client, http.MethodPost, fmt.Sprintf("%s/clients", url), contentType, tc.auth, strings.NewReader(tc.req))
+			assert.Nil(t, err, fmt.Sprintf("%s: unexpected error %s", tc.desc, err))
+			assert.Equal(t, tc.status, res.StatusCode, fmt.Sprintf("%s: expected status code %d got %d", tc.desc, tc.status, res.StatusCode))
+		}
+	})
+}
+
+func TestUpdateClient(t *testing.T) {
+	runService(func(svc manager.Service, url string, client *http.Client) {
+		validUpdate, _ := json.Marshal(manager.Client{
```
mijicd: Why do you create new "valid" instance. Reuse!

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+			Type:    "device",
+			Name:    "test_device",
+			Payload: "device_payload",
+		})
+		invalidUpdate, _ := json.Marshal(manager.Client{
```
mijicd: Create invalid based on the existing valid. This will make the "error" more explicit.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+			Type:    "foo",
+			Name:    "test_device",
+			Payload: "device_payload",
+		})
+		svc.Register(validUser)
+		clientID, _ := svc.AddClient(validUser.Email, validClient)
```
mijicd: Please rename to `id`, it is sufficient.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+			assert.Nil(t, err, fmt.Sprintf("%s: unexpected error %s", tc.desc, err))
+			assert.Equal(t, tc.status, res.StatusCode, fmt.Sprintf("%s: expected status code %d got %d", tc.desc, tc.status, res.StatusCode))
+		}
+	})
+}
+
+func TestViewClient(t *testing.T) {
+	runService(func(svc manager.Service, url string, client *http.Client) {
+		svc.Register(validUser)
+		clientID, _ := svc.AddClient(validUser.Email, validClient)
```
mijicd: Please rename to `id`.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+			Name:    "test_device",
+			Payload: "device_payload",
+		})
+		svc.Register(validUser)
+		clientID, _ := svc.AddClient(validUser.Email, validClient)
+
+		cases := []struct {
+			desc     string
+			req      string
+			clientID string
```
mijicd: Please rename to `ID`.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+			auth     string
+			status   int
+		}{
+			{"update existing client", string(validUpdate), clientID, validUser.Email, http.StatusOK},
+			{"try to update non-existent client", string(validUpdate), "123e4567-e89b-12d3-a456-000000000042", validUser.Email, http.StatusNotFound},
```
mijicd: Wrong uuid can be made constant instead of repeating it.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+			clients = append(clients, client)
+		}
+
+		cases := []struct {
+			desc   string
+			auth   string
+			status int
+			res    []manager.Client
+		}{
+			{"fetch list of clients", validUser.Email, http.StatusOK, clients},
```
mijicd: Maybe add another case where we retrieve clients for user without any clients (you should create that user as well).

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+			assert.Equal(t, tc.status, res.StatusCode, fmt.Sprintf("%s: expected status code %d got %d", tc.desc, tc.status, res.StatusCode))
+			assert.ElementsMatch(t, tc.res, data["clients"], fmt.Sprintf("%s: expected body %s got %s", tc.desc, tc.res, data["clients"]))
+		}
+	})
+}
+
+func TestRemoveClient(t *testing.T) {
+	runService(func(svc manager.Service, url string, client *http.Client) {
+		svc.Register(validUser)
+		clientID, _ := svc.AddClient(validUser.Email, validClient)
```
mijicd: Rename to `id`.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+
+		cases := []struct {
+			desc   string
+			req    string
+			auth   string
+			status int
+		}{
+			{"create new channel", string(validData), validUser.Email, http.StatusCreated},
+			{"try to create new channel with invalid token", string(validData), invalidUser.Email, http.StatusForbidden},
+			{"try to create new channel with invalid data format", "{", validUser.Email, http.StatusBadRequest},
```
mijicd: Please add the empty JSON test.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+		cases := []struct {
+			desc   string
+			req    string
+			auth   string
+			status int
+		}{
+			{"add valid client", string(validClient), validUser.Email, http.StatusCreated},
+			{"try to add client with invalid data", string(invalidClient), validUser.Email, http.StatusBadRequest},
+			{"try to add client with invalid auth token", string(validClient), "invalid_token", http.StatusForbidden},
+			{"try to add client with invalid request format", "}", validUser.Email, http.StatusBadRequest},
```
mijicd: Please add the empty JSON test.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+			clientID string
+			auth     string
+			status   int
+		}{
+			{"update existing client", string(validUpdate), clientID, validUser.Email, http.StatusOK},
+			{"try to update non-existent client", string(validUpdate), "123e4567-e89b-12d3-a456-000000000042", validUser.Email, http.StatusNotFound},
+			{"try to update client with invalid id", string(validUpdate), "1", validUser.Email, http.StatusNotFound},
+			{"try to update client with invalid data", string(invalidUpdate), clientID, validUser.Email, http.StatusBadRequest},
+			{"try to update client with invalid user token", string(validUpdate), clientID, invalidUser.Email, http.StatusForbidden},
+			{"try to update client with invalid data format", "{", clientID, validUser.Email, http.StatusBadRequest},
```
mijicd: Please add the empty JSON test.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+		cases := []struct {
+			desc     string
+			clientID string
+			auth     string
+			status   int
+		}{
+			{"delete existing client", clientID, validUser.Email, http.StatusNoContent},
+			{"try to delete non-existent client", "123e4567-e89b-12d3-a456-000000000042", validUser.Email, http.StatusNoContent},
+			{"try to delete client with invalid id", "1", validUser.Email, http.StatusNoContent},
+			{"try to delte client with invalid token", clientID, invalidUser.Email, http.StatusForbidden},
```
mijicd: Fix typo.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+	})
+}
+
+func TestUpdateChannel(t *testing.T) {
+	runService(func(svc manager.Service, url string, client *http.Client) {
+		update, _ := json.Marshal(map[string]string{
+			"name": "updated_channel",
+		})
+		svc.Register(validUser)
+		chanID, _ := svc.CreateChannel(validUser.Email, validChannel)
```
mijicd: Remarks given for client-related tests regarding asking to rename to `id` apply here as well.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+			req    string
+			chanID string
+			auth   string
+			status int
+		}{
+			{"update existing channel", string(update), chanID, validUser.Email, http.StatusOK},
+			{"try to update non-existing channel", string(update), "123e4567-e89b-12d3-a456-000000000042", validUser.Email, http.StatusNotFound},
+			{"try to update channel with invalid token", string(update), chanID, invalidUser.Email, http.StatusForbidden},
+			{"try to update channel with invalid id", string(update), "1", validUser.Email, http.StatusNotFound},
+			{"try to update channel with invalid data format", "}", chanID, validUser.Email, http.StatusBadRequest},
```
mijicd: Please add the empty JSON test.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+		cases := []struct {
+			desc     string
+			chanID   string
+			clientID string
+			auth     string
+			status   int
+		}{
+			{"connect existing client to existing channel", chanID, clientID, validUser.Email, http.StatusOK},
+			{"try to connect existing client to non-existent channel", "123e4567-e89b-12d3-a456-000000000042", clientID, validUser.Email, http.StatusNotFound},
+			{"try to connect client with invalid id to channel", chanID, "1", validUser.Email, http.StatusNotFound},
```
mijicd: Add tests that will try to connect client from user A to the channel from user B and vice versa.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+		svc.Connect(validUser.Email, chanID, clientID)
+
+		cases := []struct {
+			desc     string
+			chanID   string
+			clientID string
+			auth     string
+			status   int
+		}{
+			{"disconnect connected client from channel", chanID, clientID, validUser.Email, http.StatusNoContent},
```
mijicd: Add tests that try to disconnect channel from user A from client from user B and vice versa. This is an upgrade to the invalid token test :)

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+			assert.Equal(t, tc.clientID, clientID, fmt.Sprintf("%s: expected %s got %s", tc.desc, tc.clientID, clientID))
+		}
+	})
+}
+
+func TestCanAccess(t *testing.T) {
+	runService(func(svc manager.Service, url string, client *http.Client) {
+		svc.Register(validUser)
+		clientID, _ := svc.AddClient(validUser.Email, validClient)
+		unconnectedClientID, _ := svc.AddClient(validUser.Email, validClient)
```
mijicd: Please rename to `notConnectedClientID`.

manager/api/transport_test.go
```diff
@@ -0,0 +1,535 @@
...
+		svc.Connect(validUser.Email, chanID, clientID)
+
+		cases := []struct {
+			desc      string
+			chanID    string
+			clientKey string
+			status    int
+			clientID  string
+		}{
+			{"validate token over existing channel id", chanID, clientID, http.StatusOK, clientID},
```
mijicd: Change descriptions to use the following "template":  `check access to [...] given [...]`.

manager/mocks/channels.go
```diff
@@ -29,7 +28,7 @@ func (crm *channelRepositoryMock) Save(channel manager.Channel) (string, error)
 	defer crm.mu.Unlock()
 
 	crm.counter += 1
-	channel.ID = strconv.Itoa(crm.counter)
+	channel.ID = fmt.Sprintf("123e4567-e89b-12d3-a456-%012d", crm.counter)
```
mijicd, line 33: Theoretically, this will start giving invalid UUIDs after 9 entities. 

Instead, let's be more explicit about the way we want to use this implementation. Predefine an array of n valid uuids, and cycle through it using the counter modulo n as an index. Yes, there will be duplicates, but this should be a stub anyhow, and later on we can improve its logic if it is necessary.

manager/mocks/clients.go
```diff
@@ -29,7 +28,7 @@ func (crm *clientRepositoryMock) Id() string {
 	defer crm.mu.Unlock()
 
 	crm.counter += 1
-	return strconv.Itoa(crm.counter)
+	return fmt.Sprintf("123e4567-e89b-12d3-a456-%012d", crm.counter)
```
mijicd, line 31: See the remark given for channels mock.

mijicd (changes requested):

manager/api/transport_test.go
anovakovic01: If content type is missing, handlers ignore that fact and work normally, so there is no need for this test case.

manager/mocks/channels.go
mijicd: Please ignore this comment. I misunderstood your implementation.

manager/api/transport_test.go
```diff
@@ -0,0 +1,718 @@
...
+
+	return manager.New(users, clients, channels, hasher, idp)
+}
+
+func newServer(svc manager.Service) *httptest.Server {
+	mux := api.MakeHandler(svc)
+	return httptest.NewServer(mux)
+}
+
+func makeRequest(client *http.Client, testReq testRequest) (*http.Response, error) {
```
mijicd: Even better approach is to bind this method, and the `http.Client` to the `testRequest`. Its signature would be `testRequest.make()`.

manager/api/transport_test.go
```diff
@@ -0,0 +1,718 @@
...
+	}
+	return client.Do(req)
+}
+
+func TestRegister(t *testing.T) {
+	svc := newService()
+	ts := newServer(svc)
+	defer ts.Close()
+	client := ts.Client()
+	url := ts.URL
```
mijicd: Drop this variable.

manager/api/transport_test.go
```diff
@@ -0,0 +1,718 @@
...
+
+	body, _ := json.Marshal(user)
```
mijicd: Please create helper `toJson(data interface{}) string`, and use it instead of creating "artificial" temporary names.

mijicd (changes requested):
Please correct these minor changes along with the explicitly given remark:
- add `Content-Type` check in code and documentation
- try to integrate with codecov.io

mijicd (approved):

chombium:
cool :) well done @anovakovic01 @mijicd

PR #212, opened 2018-04-03
MF-174 - Add logger interface and go-kit logger implementation

### LINKED ISSUE #174: Wrap the usage of external logger in our own implementation
Instead of directly using a logger provided by `go-kit`, implement Mainflux logger that depends on it (technically it can depend on any other 3rd party logger). 

**Acceptance criteria:**
- all logs are JSON formatted and emitted to stdout
- all logs must contain an UTC timestamp field named `ts` as ISO-8601 string
- logger should expose 3 methods: `Info`, `Warn`, `Error` (each of these will set `level` field)
- ensure that all services are using only our implementation

anovakovic01:
This pull request resolves #174.

**Summary:**
- Add wrapper logger interface and wrap go-kit logger with it
- Add tests for info, warn and error methods


drasko (dismissed):
LGTM

log/kitlog/logger.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// New returns wrapped go kit logger.
+func New(out io.Writer) log.Logger {
+	l := logkit.NewJSONLogger(logkit.NewSyncWriter(out))
+	l = logkit.With(l, "ts", logkit.DefaultTimestampUTC)
+	return &logger{log.All, l}
+}
+
+func (l *logger) SetLevel(lvl log.Level) {
```
mijicd: Please remove this.

log/kitlog/logger.go
```diff
@@ -0,0 +1,44 @@
...
+	l.lvl = lvl
+}
+
+func (l logger) Info(obj interface{}) {
```
mijicd: Logging methods should accept only the string parameter.

log/kitlog/logger.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (l logger) Warn(obj interface{}) {
+	l.Log(log.Warn, obj)
+}
+
+func (l logger) Error(obj interface{}) {
+	l.Log(log.Error, obj)
+}
+
+func (l logger) Log(lvl log.Level, obj interface{}) {
```
mijicd: Please remove this.

log/level.go
```diff
@@ -0,0 +1,30 @@
+package log
+
+const (
+	// All level is used when logging everything.
+	All Level = iota
```
mijicd: This won't exist.

log/level.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+var levels = map[Level]string{
+	Error: "error",
+	Warn:  "warn",
+	Info:  "info",
+}
+
+// AddLevel adds custom level's name to levels map.
+func AddLevel(lvl Level, name string) {
```
mijicd: Please remove this.

log/logger.go
```diff
@@ -0,0 +1,15 @@
+package log
+
+// Logger specifies logging API.
+type Logger interface {
```
mijicd: See my remarks given above. Also, rename the package to `logger`.

mijicd (changes requested):
Besides the explicitly requested changes, make sure that all logging usages are updated to use the new logger.

mijicd (changes requested):
In case that I missed something, use `warn` for all erroneous cases that does not "break" the service, e.g. failed normalization can happen for non-SenML messages. Use `error` whenever there is a infrastructure/service failure, e.g. database or NATS connection failure.

logger/logger_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"testing"
+
+	log "github.com/mainflux/mainflux/logger"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInfo(t *testing.T) {
+	cases := map[string]struct {
+		input string
```
mijicd: The test is missing the expected output. I would recommend to restructure it in such a way that is easy to acquire the contents and verify it, i.e. provide an implementation of `io.Writer` for test purposes.

mijicd (changes requested):
Getting there.

mijicd (approved):

PR #213, opened 2018-04-03
MF-139 - Expose metrics from all services

dborovcanin:
Add service interface in order to expose method to be monitored.


cmd/normalizer/main.go
```diff
@@ -8,14 +8,17 @@ import (
...
 	"github.com/mainflux/mainflux/normalizer"
+	"github.com/mainflux/mainflux/normalizer/api"
 	nats "github.com/nats-io/go-nats"
+	stdprometheus "github.com/prometheus/client_golang/prometheus"
 )
 
 const (
 	defNatsURL string = nats.DefaultURL
-	defPort    string = "8180"
+	defPort    string = "8900"
```
drasko: Why? I thought that @mijicd fixed all services to same port, demanding env variable to define the desired port.

dborovcanin: It was easiest for me to configure service that way, since I run all services locally, I don't use Docker for development. I accidentally committed that change.

normalizer/normalizer.go
```diff
@@ -16,31 +16,38 @@ const (
 	output  string = "normalized"
 )
 
+// Service represents Normalizer service.
+type Service interface {
```
mijicd: This is unnecessary, especially having in mind that it exposes a callback. I would prefer that you remove it and keep it as it is for now.

normalizer/normalizer.go
```diff
@@ -16,31 +16,38 @@ const (
...
+// New returns new Service.
+func New(nc *nats.Conn, logger log.Logger) Service {
+	return eventFlow{nc, logger}
+}
 
-		if err := proto.Unmarshal(m.Data, &msg); err != nil {
-			ef.logger.Log("error", fmt.Sprintf("Unmarshalling failed: %s", err))
-			return
-		}
+func (ef eventFlow) HandleMessage(m *nats.Msg) {
```
mijicd: Are there any benefits of replacing an anonymous callback method with exported one?

mijicd (changes requested):
At the moment, no metrics-related changes were applied.

normalizer/normalizer.go
dborovcanin: Since all important work happens inside callback function, it did make sense to export function in order to be able to monitor calls. This way, with interface and callback function exported, it is in sync with other services and we have the benefit of being able to monitor service behavior.

mijicd: You can monitor the service without exporting its internals.

drasko (commented):
LGTM

mijicd (approved):
As per our agreement, no additional interfaces were created. Once we introduce them, `normalizer` codebase will be updated. For now, this looks good to me.

PR #217, opened 2018-04-08
MF-165 - Add MQTT adapter

drasko:


drasko:
Problem with this PR is that not only adapter, but big part of MQTT server has to be implemented.

Some of the functionality/arch can be derived from https://github.com/surgemq/surgemq/tree/c88a02e0d607d081fc28d36513dc01bbf83b1b37/service and Go's HTTP server: https://golang.org/src/net/http/server.go

mijicd:
Closed due to recent requirements. We will work on improvements of Javascript solution we had before.

PR #223, opened 2018-04-11
NOISSUE - Add channel ID validation

anovakovic01:
Add channel id validation and remove unnecessary error. Update
swagger doc accordingly.

### DIFF
--- a/http/api/transport.go
+++ b/http/api/transport.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/asaskevich/govalidator"
 	kithttp "github.com/go-kit/kit/transport/http"
 	"github.com/go-zoo/bone"
 	"github.com/mainflux/mainflux"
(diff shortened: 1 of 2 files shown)


mijicd (approved):

drasko (commented):
To me "Not Found" would say - I searched ID in the DB and it is not present.

As I see here the case is that path is malformed, so I would maybe rather return 400 Bad Request, or something like this.

anovakovic01:
@drasko I followed manager's API logic.

chombium:
I agree with @drasko. Maybe we should improve this. If the id is invalid we should return bad request and if the id is valid but not found than not found.

On the other hand, if we don't want to introduce complex validation, invalid ids won't be found, so not found as a response code is also ok.

mijicd:
@chombium it is somewhat irrelevant. Neither response would be *wrong*, but 404 is more accurate one - a resource with non UUID identifier does not exist.

PR #227, opened 2018-04-17
MF-200 - Enable pagination of result sets

dborovcanin:


manager/api/logging.go
```diff
@@ -86,7 +86,7 @@ func (lm *loggingMiddleware) ViewClient(key string, id string) (client manager.C
 	return lm.svc.ViewClient(key, id)
 }
 
-func (lm *loggingMiddleware) ListClients(key string) (clients []manager.Client, err error) {
+func (lm *loggingMiddleware) ListClients(key string, offset int, limit int) (clients []manager.Client, err error) {
```
mijicd: Please change signature to `ListClients(key string, offset, limit int)`.

manager/api/logging.go
```diff
@@ -151,7 +151,7 @@ func (lm *loggingMiddleware) ViewChannel(key string, id string) (channel manager
 	return lm.svc.ViewChannel(key, id)
 }
 
-func (lm *loggingMiddleware) ListChannels(key string) (channels []manager.Channel, err error) {
+func (lm *loggingMiddleware) ListChannels(key string, offset int, limit int) (channels []manager.Channel, err error) {
```
mijicd: Same here.

manager/api/metrics.go
```diff
@@ -70,13 +70,13 @@ func (ms *metricsMiddleware) ViewClient(key string, id string) (manager.Client,
 	return ms.svc.ViewClient(key, id)
 }
 
-func (ms *metricsMiddleware) ListClients(key string) ([]manager.Client, error) {
+func (ms *metricsMiddleware) ListClients(key string, offset int, limit int) ([]manager.Client, error) {
```
mijicd: And here.

manager/api/metrics.go
```diff
@@ -115,13 +115,13 @@ func (ms *metricsMiddleware) ViewChannel(key string, id string) (manager.Channel
 	return ms.svc.ViewChannel(key, id)
 }
 
-func (ms *metricsMiddleware) ListChannels(key string) ([]manager.Channel, error) {
+func (ms *metricsMiddleware) ListChannels(key string, offset int, limit int) ([]manager.Channel, error) {
```
mijicd: And here.

manager/api/requests.go
```diff
@@ -110,16 +110,16 @@ func (req viewResourceReq) validate() error {
...
+	limit  int
 }
 
 func (req listResourcesReq) validate() error {
 	if req.key == "" {
 		return manager.ErrUnauthorizedAccess
 	}
 
-	if req.size > 0 && req.offset >= 0 {
+	if req.offset > 0 && req.limit >= 0 {
```
mijicd: While i agree with renaming `size` to `limit`, the condition was correct. You can start from the beginning (`offset >= 0`), and you must select at least 1 element (`limit > 0`).

manager/api/transport.go
```diff
@@ -247,10 +248,19 @@ func decodeView(_ context.Context, r *http.Request) (interface{}, error) {
 }
 
 func decodeList(_ context.Context, r *http.Request) (interface{}, error) {
+	offset, err := strconv.Atoi(r.FormValue("offset"))
```
mijicd: Please don't use form values. Instead, use the following approach:

```golang
q := r.URL.Query()
q.Get("offset")
q.Get("limit")
```

Check out the relevant godocs:
- https://golang.org/pkg/net/url/#URL.Query
- https://golang.org/pkg/net/url/#Values.Get

manager/channels.go
```diff
@@ -24,8 +24,8 @@ type ChannelRepository interface {
 	// by the specified user.
 	One(string, string) (Channel, error)
 
-	// All retrieves the channels owned by the specified user.
-	All(string) []Channel
+	// Bulk retrieves the bulk of channels owned by the specified user.
+	Bulk(string, int, int) []Channel
```
mijicd: Don't rename.

manager/clients.go
```diff
@@ -44,8 +44,8 @@ type ClientRepository interface {
 	// by the specified user.
 	One(string, string) (Client, error)
 
-	// All retrieves the clients owned by the specified user.
-	All(string) []Client
+	// Bulk retrieves the bulk of clients owned by the specified user.
+	Bulk(string, int, int) []Client
```
mijicd: Same here.

manager/manager.go
```diff
@@ -89,7 +89,7 @@ func (ms *managerService) ViewClient(key, id string) (Client, error) {
 	return ms.clients.One(sub, id)
 }
 
-func (ms *managerService) ListClients(key string) ([]Client, error) {
+func (ms *managerService) ListClients(key string, offset int, limit int) ([]Client, error) {
```
mijicd: See the remark regarding the method signatures above.

manager/manager.go
```diff
@@ -156,7 +156,7 @@ func (ms *managerService) ViewChannel(key, id string) (Channel, error) {
 	return ms.channels.One(sub, id)
 }
 
-func (ms *managerService) ListChannels(key string) ([]Channel, error) {
+func (ms *managerService) ListChannels(key string, offset int, limit int) ([]Channel, error) {
```
mijicd: Same here.

manager/manager_test.go
```diff
@@ -145,7 +145,7 @@ func TestListClients(t *testing.T) {
 	}
 
 	for desc, tc := range cases {
-		_, err := svc.ListClients(tc.key)
+		_, err := svc.ListClients(tc.key, 1, 2)
```
mijicd: Don't hardcode, use parameters and check various combinations.

manager/manager_test.go
```diff
@@ -254,7 +254,7 @@ func TestListChannels(t *testing.T) {
 	}
 
 	for desc, tc := range cases {
-		_, err := svc.ListChannels(tc.key)
+		_, err := svc.ListChannels(tc.key, 1, 2)
```
mijicd: Same here.

manager/mocks/channels.go
```diff
@@ -59,7 +59,7 @@ func (crm *channelRepositoryMock) One(owner, id string) (manager.Channel, error)
 	return manager.Channel{}, manager.ErrNotFound
 }
 
-func (crm *channelRepositoryMock) All(owner string) []manager.Channel {
+func (crm *channelRepositoryMock) Bulk(owner string, offset int, limit int) []manager.Channel {
```
mijicd: See the remarks given about method signature and renaming.

manager/mocks/clients.go
```diff
@@ -63,7 +63,7 @@ func (crm *clientRepositoryMock) One(owner, id string) (manager.Client, error) {
 	return manager.Client{}, manager.ErrNotFound
 }
 
-func (crm *clientRepositoryMock) All(owner string) []manager.Client {
+func (crm *clientRepositoryMock) Bulk(owner string, offset int, limit int) []manager.Client {
```
mijicd: Same here.

manager/postgres/channels.go
```diff
@@ -56,11 +56,10 @@ func (cr channelRepository) One(owner, id string) (manager.Channel, error) {
 	return channel, nil
 }
 
-func (cr channelRepository) All(owner string) []manager.Channel {
+func (cr channelRepository) Bulk(owner string, offset int, limit int) []manager.Channel {
```
mijicd: Same here.

manager/postgres/channels_test.go
```diff
@@ -113,7 +113,7 @@ func TestMultiChannelRetrieval(t *testing.T) {
 	}
 
 	for desc, tc := range cases {
-		n := len(chanRepo.All(tc.owner))
+		n := len(chanRepo.Bulk(tc.owner, 0, 10))
```
mijicd: See the remark about test parameters given above.

mijicd (changes requested):
Besides the requested remarks, please rename all `transport_test.go` to `endpoint_test.go` in a separate commit.

manager/channels.go
dborovcanin: @mijicd What do you think of changing offset and limit to uint32 or uint64, according to theirs nature?
Also, what do you think of an idea of wrapping pagination result in pagination object. Something like:
```
type Page struct {
	offset uint32
	limit  uint32
	total  uint32
	data   interface{}
}
```

mijicd: Type choice is correct, though I prefer to use less specific variant. Not sure how would this change affect the signatures (repository & manager). In generally, it's not a bad idea, though some clarification is needed.

dborovcanin: I'm not sure yet, but it should not be problem to Manager or Repository. Limit and offset are already passed as params and data interface is, basically, list of... something. So, repository method would return result type of Page. At the moment, I see only `total` questionable since I will call GORM's `Count` method inside repository method call. Not sure how that would affect performance. Response validation should be straightforward, I don't see any problems right now.

mijicd: I'm talking about request validation, we don't do any response validations. Regarding the second point, I don't recommend to avoid `Page` but return a tuple (array + count) as return value. Executing count will affect performance, but it can be handled, and these endpoints are not time critical anyhow.

manager/api/endpoint_test.go
```diff
@@ -332,7 +332,7 @@ func TestListClients(t *testing.T) {
 		req := testRequest{
 			client: cli,
 			method: http.MethodGet,
-			url:    fmt.Sprintf("%s/clients", ts.URL),
+			url:    fmt.Sprintf("%s/clients?offset=1&limit=2", ts.URL),
```
mijicd: I'd rather add more cases, covering both usage of default and custom values.

manager/api/requests.go
```diff
@@ -110,16 +110,16 @@ func (req viewResourceReq) validate() error {
...
+	limit  int
 }
 
 func (req listResourcesReq) validate() error {
 	if req.key == "" {
 		return manager.ErrUnauthorizedAccess
 	}
 
-	if req.size > 0 && req.offset >= 0 {
+	if req.offset >= 0 && req.limit > 0 {
```
mijicd: Reading the docs pointed me to another issue: `limit` cannot be infinite. Please set the upper bound and ensure the API documentation is synchronized.

P.S.: don't forget the tests.

manager/api/transport.go
```diff
@@ -14,6 +15,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+const MAX_LIMIT_SIZE = 100
```
mijicd: See the point given above. Also, **never use java conventions for naming go constants**.

manager/clients.go
```diff
@@ -44,8 +44,8 @@ type ClientRepository interface {
 	// by the specified user.
 	One(string, string) (Client, error)
 
-	// All retrieves the clients owned by the specified user.
-	All(string) []Client
+	// All retrieves the bulk of clients owned by the specified user.
```
mijicd: ...subset of...

manager/manager_test.go
```diff
@@ -244,17 +251,24 @@ func TestListChannels(t *testing.T) {
...
 	cases := map[string]struct {
-		key string
-		err error
+		key  string
+		size int
+		err  error
 	}{
-		"list channels":                        {key, nil},
-		"list channels with wrong credentials": {wrong, manager.ErrUnauthorizedAccess},
+		"list channels":                        {key, 5, nil},
```
mijicd: Please change to `n`.

manager/postgres/channels_test.go
```diff
@@ -113,7 +113,7 @@ func TestMultiChannelRetrieval(t *testing.T) {
 	}
 
 	for desc, tc := range cases {
-		n := len(chanRepo.All(tc.owner))
+		n := len(chanRepo.All(tc.owner, 0, 10))
```
mijicd: Parameterize the test case instead of hard coding values.

manager/postgres/clients_test.go
```diff
@@ -130,7 +130,7 @@ func TestMultiClientRetrieval(t *testing.T) {
 	}
 
 	for desc, tc := range cases {
-		n := len(clientRepo.All(tc.owner))
+		n := len(clientRepo.All(tc.owner, 0, 10))
```
mijicd: Parameterize the test instead of hard coding values.

manager/service.go
```diff
@@ -61,9 +61,9 @@ type Service interface {
 	// ID, that belongs to the user identified by the provided key.
 	ViewChannel(string, string) (Channel, error)
 
-	// ListChannels retrieves data about all clients that belongs to the user
-	// identified by the provided key.
-	ListChannels(string) ([]Channel, error)
+	// ListChannels retrieves data about bulk of channels specified by offset and limit
```
mijicd: - `bulk of` => `subset of`
- drop `specified by offset and limit`

mijicd (changes requested):
In general it looks quite good. Once the requested changes are made it'll be merged.

manager/api/transport.go
```diff
@@ -247,10 +248,32 @@ func decodeView(_ context.Context, r *http.Request) (interface{}, error) {
 }
 
 func decodeList(_ context.Context, r *http.Request) (interface{}, error) {
+
+	if len(r.URL.RawQuery) == 0 {
```
mijicd: This code cannot handle the case where either `offset` or `limit` are provided.

manager/manager_test.go
```diff
@@ -136,16 +136,25 @@ func TestListClients(t *testing.T) {
...
-		err error
+		key    string
+		offset int
+		limit  int
+		size   int
+		err    error
 	}{
-		"list clients":                        {key, nil},
-		"list clients with wrong credentials": {wrong, manager.ErrUnauthorizedAccess},
+		"list clients":                        {key, 0, 5, 5, nil},
```
mijicd, line 150: Missing offset/limit tests.

manager/manager_test.go
```diff
@@ -244,17 +253,26 @@ func TestListChannels(t *testing.T) {
...
-		err error
+		key    string
+		offset int
+		limit  int
+		size   int
+		err    error
 	}{
-		"list channels":                        {key, nil},
-		"list channels with wrong credentials": {wrong, manager.ErrUnauthorizedAccess},
+		"list channels":                        {key, 0, 5, 5, nil},
```
mijicd: Same here.

manager/postgres/channels_test.go
```diff
@@ -105,16 +105,19 @@ func TestMultiChannelRetrieval(t *testing.T) {
...
-		owner string
-		len   int
+		owner  string
+		offset int
+		limit  int
+		size   int
 	}{
-		"existing owner":     {email, n},
-		"non-existing owner": {wrong, 0},
+		"existing owner":     {email, 0, n, n},
```
mijicd, line 113: Missing invalid limit check and subset smaller than n.

manager/postgres/clients_test.go
```diff
@@ -122,16 +122,19 @@ func TestMultiClientRetrieval(t *testing.T) {
...
-		owner string
-		len   int
+		owner  string
+		offset int
+		limit  int
+		size   int
 	}{
-		"existing owner":     {email, n},
-		"non-existing owner": {wrong, 0},
+		"existing owner":     {email, 0, n, n},
```
mijicd, line 130: Same here.

manager/api/requests.go
```diff
@@ -110,16 +112,20 @@ func (req viewResourceReq) validate() error {
...
 }
 
-func (req listResourcesReq) validate() error {
+func (req *listResourcesReq) validate() error {
 	if req.key == "" {
 		return manager.ErrUnauthorizedAccess
 	}
 
-	if req.size > 0 && req.offset >= 0 {
+	if req.limit > maxLimitSize {
```
mijicd: Reject the request if this is the case. Once again, **don't forget to update the API documentation**.

mijicd (changes requested):

mijicd (approved):
There are minor inconsistencies between the docs and the implementations. These will be fixed in the follow up.

PR #229, opened 2018-04-18
modify the "up" position of docker-compose cmd && modify the curl cmd of "Sending messages  HTTP"

oldstorm:
1 modify the "up" position of docker-compose, otherwise it can't run well
2 modify the curl cmd of "Sending messages  HTTP", the cmd is one more "-d"


drasko (changes requested):
PRs need to be signed-off

Please always commit with `git commit -s`

PR #232, opened 2018-04-18
NOISSUE - Fix paging

mijicd:


docs/getting-started.md
```diff
@@ -124,13 +124,16 @@ Content-Length: 1105
...
 
 ```
-curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -H "Authorization: <user_auth_token>" https://localhost/channels?offset=0&limit=5
+curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -H "Authorization: <user_auth_token>" https://localhost/clients?offset=0&limit=5
 ```
-If you don't provide these params, default values will be used insted. Default value for
-offset is 0, and for limit is 10. Max value for limit is 100, and every value above that will be replaced by limit max value. Providing invalid values for these two params will be considered malformed request, so bad request will be received as response.
+
+If you don't provide them, default values will be used instead: 0 for `offset`,
+and 10. Note that `limit` cannot be set to values greater than 100. Providing
```
anovakovic01: Maybe to add for `limit` just to be clear?

docs/getting-started.md
```diff
@@ -187,13 +190,16 @@ Content-Length: 139
...
+group of channels. In that case, your request should look like:
 
 ```
 curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -H "Authorization: <user_auth_token>" https://localhost/channels?offset=0&limit=5
 ```
-If you don't provide these params, default values will be used insted. Default value for
-offset is 0, and for limit is 10. Max value for limit is 100, and every value above that will be replaced by limit max value. Providing invalid values for these two params will be considered malformed request, so bad request will be received as response.
+
+If you don't provide them, default values will be used instead: 0 for `offset`,
+and 10. Note that `limit` cannot be set to values greater than 100. Providing
```
anovakovic01: Maybe to add for `limit` just to be clear?

anovakovic01 (changes requested):

docs/getting-started.md
drasko: Agreed with @anovakovic01

docs/getting-started.md
drasko: +1

drasko (changes requested):

anovakovic01 (approved):
LGTM!

drasko (approved):

PR #234, opened 2018-04-20
NOISSUE - Define manager errors in the Service interface

manuio:


drasko (commented):
LGTM

drasko:
@manuIO please sign-off

mijicd:
This cannot be accepted, regardless of whether is signed or not. The errors you want to put in the service interface are transport related, not business related, hence they belong exactly where they were previously.

PR #241, opened 2018-04-24
NOISSUE - Add service name to version response

dborovcanin:

### DIFF
--- a/version.go
+++ b/version.go
@@ -9,12 +9,13 @@ const version string = "0.2.2"
 
 type response struct {
 	Version string
+	Service string
 }
 
 // Version exposes an HTTP handler for retrieving service version.
-func Version() http.HandlerFunc {
+func Version(service string) http.HandlerFunc {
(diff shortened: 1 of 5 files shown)


mijicd (approved):

version.go
```diff
@@ -9,12 +9,13 @@ const version string = "0.2.2"
...
 	Version string
+	Service string
 }
 
 // Version exposes an HTTP handler for retrieving service version.
-func Version() http.HandlerFunc {
+func Version(service string) http.HandlerFunc {
 	return http.HandlerFunc(func(rw http.ResponseWriter, _ *http.Request) {
-		res := response{Version: version}
+		res := response{Version: version, Service: service}
```
drasko, line 18: I would expect `{Service, Version}`, rather then `{Version, Service}`

drasko (commented):
Maybe we can also get some info from the undelaying instance (IP addr, OS, hostname, some other things that can be important for debugging and/or monitoring)?

PR #248, opened 2018-04-30
MF-164 - Extract users service

anovakovic01:
**Summary:**
- Add users service
- Add HTTP API for users service
- Add gRPC API for users service
- Add logging
- Add metrics
- Add README and swagger files
- Add tests for service implementation
- Add tests for API layer


mijicd (commented):
For now it looks good, I will take some more time to look it more thoroughly.

One remark though, you should (re)move manager service code that's related to the users service.

drasko (commented):
Looks good, no remarks from my side.

I note however that Identity is moved to `gRPC`-only, no HTTP counterpart. This will break all other services (adapters), so they must be changed simultaneously with this PR.

users/README.md
```diff
@@ -0,0 +1,78 @@
...
+make users
+
+# copy binary to bin
+make install
+
+# set the environment variables and run the service
+MF_USERS_DB_HOST=[Database host address] MF_USERS_DB_PORT=[Database host port] MF_USERS_DB_USER=[Database user] MF_USERS_DB_PASS=[Database password] MF_USERS_DB=[Name of the database used by the service] MF_USERS_HTTP_PORT=[Service HTTP port] MF_USERS_GRPC_PORT=[Service gRPC port] MF_USERS_SECRET=[String used for signing tokens] $GOBIN/mainflux-users
+```
+
+## Usage
```
mijicd, line 73: Eventually, we should deprecate these "single service API docs", and replace them with one, unified HTTP API doc.

cmd/users/main.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// Start gRPC server
+	go func() {
+		p := fmt.Sprintf(":%s", cfg.GRPCPort)
+		listener, err := net.Listen("tcp", p)
+		if err != nil {
+			logger.Error(fmt.Sprintf("Failed to listen on port %s: %s", cfg.GRPCPort, err))
+		}
+		server := grpc.NewServer()
+		service := grpcapi.NewService(svc)
```
mijicd: Don't name it, it has only one usage. It is a bit confusing to have `server`, `service` and `svc`.

cmd/users/main.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// Start HTTP server
+	go func() {
+		p := fmt.Sprintf(":%s", cfg.HTTPPort)
+		logger.Info(fmt.Sprintf("Users HTTP service started, exposed port %s", cfg.HTTPPort))
+		errs <- http.ListenAndServe(p, httpapi.MakeHandler(svc, logger))
+	}()
+
+	// Start gRPC server
+	go func() {
```
mijicd: Since this is quite big now, i propose that you extract these snippets to non-exported functions and use `main()` just to invoke them.

mijicd (changes requested):
Looks good. Fix these minor remarks and we can merge it. Side note: let's drop the "usage" section from individual services, and let's unify all of the individual API docs into the top-level API specification. There's not much sense into fiddling with individual service.

mijicd (changes requested):

manager/postgres/channels_test.go
```diff
@@ -12,18 +12,13 @@ import (
...
 	c1 := manager.Channel{Owner: email}
-	c2 := manager.Channel{Owner: wrong}
 
 	cases := map[string]struct {
 		channel manager.Channel
 		hasErr  bool
 	}{
-		"new channel, existing user":     {c1, false},
-		"new channel, non-existing user": {c2, true},
+		"create new channel": {c1, false},
```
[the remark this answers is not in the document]
anovakovic01: Currently, there are no relationships between users and channels in database. That is why this test case lost it's purpose.

api.go
```diff
@@ -0,0 +1,8 @@
+package mainflux
+
+// APIRes contains http response specifig methods.
+type APIRes interface {
```
mijicd: Godoc missing on methods. Rename to `Response`.

mijicd (changes requested):

mijicd (approved):

PR #257, opened 2018-05-08
NOISSUE - Update CoAP docs

dborovcanin:

### DIFF
--- a/cmd/coap/main.go
+++ b/cmd/coap/main.go
@@ -56,13 +56,13 @@ func main() {
 	svc = api.MetricsMiddleware(
 		svc,
 		kitprometheus.NewCounterFrom(stdprometheus.CounterOpts{
-			Namespace: "ws_adapter",
+			Namespace: "coap_adapter",
 			Subsystem: "api",
 			Name:      "request_count",
 			Help:      "Number of requests received.",
 		}, []string{"method"}),
 		kitprometheus.NewSummaryFrom(stdprometheus.SummaryOpts{
-			Namespace: "ws_adapter",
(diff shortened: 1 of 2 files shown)


docs/getting-started.md
```diff
@@ -282,3 +282,24 @@ If you are not able to send custom headers in your handshake request, send it as
...
+
+```
+coap get coap://localhost/channels/<channel_id>/messages?key=<client_token> -o
+```
+
+where `-o` in indicates that value of `Observe` option is 0.
+To unsubscribe from channel, client should send same message with value of `Observe` option set to 1
+or ignore confirmable message.
+
+Useful tools for sending CoAP messages can be found [here](https://github.com/mcollina/coap-cli) and [here](https://github.com/mkovatsc/Copper).
```
mijicd: Put this at the beginning and name the links according to the tool names, instead of as "here".

mijicd (changes requested):

cmd/coap/main.go
```diff
@@ -85,14 +85,5 @@ func main() {
...
-}
-
-func env(key, fallback string) string {
-	value := os.Getenv(key)
-	if value == "" {
-		return fallback
-	}
-
-	return value
+	logger.Info(fmt.Sprintf("CoAP adapter terminated: %s", c.Error()))
```
anovakovic01: Use `Error` instead of `Info`.

anovakovic01 (changes requested):

anovakovic01 (approved):

mijicd (approved):

PR #272, opened 2018-05-11
MF-216 - Integrate Bashflux into monorepo

manuio:


bashflux/bashflux.go
```diff
@@ -0,0 +1,80 @@
+package main
```
mijicd: Please move to `cmd`, and name according to the other services.

bashflux/bashflux.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"log"
+
+	"github.com/mainflux/mainflux/bashflux/cmd"
+	"github.com/spf13/cobra"
+)
+
+type Config struct {
```
mijicd: There is no need for exporting this. Please remove it, and use as an anonymous structure inside `main()`.

bashflux/bashflux.go
```diff
@@ -0,0 +1,80 @@
...
+	// Root Commands
+	rootCmd.AddCommand(cmd.CmdVersion)
+	rootCmd.AddCommand(cmd.CmdClients)
+	rootCmd.AddCommand(cmd.CmdChannels)
+	rootCmd.AddCommand(cmd.CmdMessages)
+	rootCmd.AddCommand(cmd.CmdSession)
+	rootCmd.AddCommand(cmd.CmdUsers)
+
+	// Clients
+	cmd.CmdClients.AddCommand(cmd.CmdCreateClient)
```
mijicd: Please do these initializations as part of the `cmd` package. Moreover, split the `cmd` package into `clients`, `users`, `channels` (for example), and provide a factory method into each of these.

bashflux/cmd/channels.go
```diff
@@ -0,0 +1,179 @@
...
+	"strings"
+
+	"github.com/mainflux/mainflux/clients"
+	"github.com/spf13/cobra"
+)
+
+var chanEndPoint = "channels"
+
+// Channels
+var CmdChannels = &cobra.Command{
```
mijicd: Based on the remark given above, you can do the following (draft code):

```golang
package clients

var commands :=[]cobra.Command{
    cobra.Command{ /* specify create */},
    cobra.Command{/* specify retrieve */},
    // etc.
}

// New does what godoc says...
func New() *cobra.Command {
    // package root
    cmd := cobra.Command { ... }

    for _, c := range commands {
        cmd.AddCommand(c)
    }

    return &cmd
}
```

The same "pattern" should be applied for each API component.

bashflux/cmd/channels.go
```diff
@@ -0,0 +1,179 @@
...
+				DeleteChannel(args[0], args[1])
+			}
+		} else {
+			LogUsage(cmdCobra.Short)
+		}
+	},
+}
+
+// CreateChannel - creates new channel and generates UUID
+func CreateChannel(msg string, token string) {
```
mijicd: Make a package-private API client (until we provide one from the services), and bind these methods to it. Later on use them as you're doing it right now.

mijicd (changes requested):
Let's make the first iteration. Once you complete it, we'll dig into a little bit more details regarding the HTTP clients.

PR #280, opened 2018-05-15
MF-268 - Replace JWT with plain string tokens in clients service

### LINKED ISSUE #268: Replace JWT with plain string tokens in things service
The current implementation falsely relies on extracting the thing identifier from JWT without checking whether the token exists in the database. Due to this, the usage of UUID as primary key is enforced. 

**Acceptance criteria:**
- replace JWT with plain string access key (e.g. UUID)
- enforce new key's uniqueness in the database schema
- check the token's existence via DB query

anovakovic01:
This pull request resolves #268.

**Summary:**
- Replace JWT identity provider with postgres identity provider
- Update existing tests
- Add tests for postgres identity provider


clients/hasher.go
```diff
@@ -0,0 +1,12 @@
+package clients
+
+// Hasher specifies an API for generating hashes of an arbitrary textual
+// content.
+type Hasher interface {
```
mijicd, line 5: Why do you need this?

clients/postgres/idp.go
```diff
@@ -0,0 +1,37 @@
...
+	"database/sql"
+
+	"github.com/mainflux/mainflux/clients"
+	"github.com/mainflux/mainflux/logger"
+	uuid "github.com/satori/go.uuid"
+)
+
+var _ clients.IdentityProvider = (*identityProvider)(nil)
+
+type identityProvider struct {
```
mijicd: This is over-engineering. `Identity` check can be performed as part of the queries in postgres repository, while Key generation can be kept inside the repository (as before).

mijicd (changes requested):

mijicd (approved):

PR #285, opened 2018-05-16
MF-165 - Add gRPC authorization calls to MQTT adapter

anovakovic01:
**Summary:**
- Remove unnecessary files
- Update environment variables
- Add gRPC authorization calls


mqtt/mqtt.js
```diff
@@ -87,125 +85,81 @@ nats.subscribe('channel.*', function (msg) {
...
+            var msg = {
+                Publisher: client.id,
+                Channel: channel,
+                Protocol: 'mqtt',
+                ContentType: packet.topic.split('/')[3],
+                Payload: packet.payload
+            };
+            var rawMsg = message.RawMessage.encode(msg);
+
+            console.log("msg:", msg);
```
nmarcetic: Remove log

mqtt/mqtt.js
```diff
@@ -87,125 +85,81 @@ nats.subscribe('channel.*', function (msg) {
...
+            console.log("packet:", util.inspect(packet, false, null));
```
nmarcetic: Remove log

mqtt/mqtt.js
```diff
@@ -87,125 +85,81 @@ nats.subscribe('channel.*', function (msg) {
...
 
             // Pub on NATS
-            nats.publish('channel.' + channel, msg);
+            nats.publish('channel.' + channel, rawMsg);
         } else {
             console.log('Publish not authorized');
```
nmarcetic: Remove log

mqtt/mqtt.js
```diff
@@ -87,125 +85,81 @@ nats.subscribe('channel.*', function (msg) {
...
+    
+    thingsClient.canAccess({
+        token: client.password,
+        chanID: channel
+    }, function (err, res) {
+        if (!err) {
             console.log('Subscribe authorized OK');
+            callback(null, packet);
         } else {
             console.log('Subscribe not authorized');
```
nmarcetic: Remove log

mqtt/mqtt.js
```diff
@@ -87,125 +85,81 @@ nats.subscribe('channel.*', function (msg) {
...
-        if (res && (res.statusCode === 200)) {
+    // Topics are in the form `channels/<channel_id>/messages/senml-json`
+    var channel = packet.topic.split('/')[1];
+    
+    thingsClient.canAccess({
+        token: client.password,
+        chanID: channel
+    }, function (err, res) {
+        if (!err) {
             console.log('Subscribe authorized OK');
```
nmarcetic: Remove

mqtt/mqtt.js
```diff
@@ -87,125 +85,81 @@ nats.subscribe('channel.*', function (msg) {
...
+    client.password = password.toString() || "";
+    callback(null, true);
 };
+
 /**
  * Handlers
  */
 aedes.on('clientDisconnect', function (client) {
-    var c = client;
     console.log('client disconnect', client.id);
```
nmarcetic: Remove logs

nmarcetic (commented):
@anovakovic01 Can we use `bunyan` logger ? Its already in our package.json as dependency please use it for service logging. All this console.logs should be replaced with logger, keep in mind that logging every message that coming from NATS is not good idea.

mqtt/mqtt.js
```diff
@@ -40,44 +44,38 @@ logging({
  */
 function startWs() {
     var server = http.createServer();
-    websocket.createServer({
-        server: server
-    }, aedes.handle);
+    websocket.createServer({server: server}, aedes.handle);
```
mijicd, line 39: You have extracted this to separate service, remember? :)

mqtt/mqtt.js
```diff
@@ -40,44 +44,38 @@ logging({
...
 
     var packet = {
         cmd: 'publish',
         qos: 2,
-        topic: 'mainflux/channels/' + m.Channel + '/messages/' + m.ContentType,
+        topic: 'channels/' + m.Channel + '/messages/' + m.ContentType,
         payload: m.Payload,
         retain: false
     };
+    console.log("Received message from NATS");
```
mijicd: If possible use proper logger, and consider carefully what to log.

mqtt/mqtt.js
```diff
@@ -14,16 +14,20 @@ var net = require('net');
...
 var nats = require('nats').connect(config.nats_url);
 
 var protobuf = require('protocol-buffers');
-const fs = require('fs');
+var grpc = require('grpc');
+var fs = require('fs');
 
 // pass a proto file as a buffer/string or pass a parsed protobuf-schema object
-var message = protobuf(fs.readFileSync('message.proto'));
+var message = protobuf(fs.readFileSync('../message.proto'));
```
mijicd: Please group `var`s:

```javascript
var message = ...,
    protobuf = ...
    // applies to all vars
```

mqtt/mqtt.js
```diff
@@ -87,125 +85,81 @@ nats.subscribe('channel.*', function (msg) {
...
-    };
-
-    request(options, function (err, res) {
-        var error = null;
-        var msg = {};
-        if (res && (res.statusCode === 200)) {
+    // Topics are in the form `channels/<channel_id>/messages/senml-json`
+    var channel = packet.topic.split('/')[1];
+
+    thingsClient.CanAccess({
```
mijicd: You can safely omit `Client` suffix.

mijicd (changes requested):

mqtt/mqtt.js
anovakovic01: I already have `things` here: `var things = grpc.load("../internal.proto").mainflux;`.

mqtt/mqtt.js
```diff
@@ -87,125 +57,77 @@ nats.subscribe('channel.*', function (msg) {
...
-            console.log(util.inspect(packet, false, null))
-
             // Pub on NATS
-            nats.publish('channel.' + channel, msg);
+            nats.publish('channel.' + channel, rawMsg);
         } else {
-            console.log('Publish not authorized');
-            error = 4; // Bad username or password
+            logger.warn("unauthorized publish: %s", err.message);
+            callback(4); // Bad username or password
```
drasko: Moving callback here (in case of error), you are not calling the callback in case there is no error.

I am not sure, but I think that it is expected to call `callback(0)` in that case - please check the doc.

mqtt/mqtt.js
```diff
@@ -87,125 +57,77 @@ nats.subscribe('channel.*', function (msg) {
...
+        chanID: channel
+    }, function (err, res) {
+        if (!err) {
+            logger.info('authorized subscribe');
+            callback(null, packet);
         } else {
-            console.log('Subscribe not authorized');
-            error = 4; // Bad username or password
+            logger.warn('unauthorizerd subscribe: %s', err);
+            callback(4, packet); // Bad username or password
```
drasko: Same here for callback.

mqtt/mqtt.js
```diff
@@ -87,125 +57,77 @@ nats.subscribe('channel.*', function (msg) {
...
-            c.password = password;
-            success = true;
-        } else {
-            error = new Error('Auth error');
-            error.returnCode = 4; // Bad username or password
-            success = false;
-        }
-        // Respond with auth error and success
-        callback(error, success);
-    });
```
drasko, line 193: Without authentication check you are letting any client connect, i.e. it will send `CONNECT` MQTT req, and servier will respond with `CONNACK` and TCP connection will be established, even though client did not do any SUB or PUB yet.

drasko (changes requested):

mqtt/mqtt.js
anovakovic01: I am not calling `callback` here on purpose. Message will be sent two times over MQTT if I call `callback` here. `callback` call is not necessary in this case.

mqtt/mqtt.js
anovakovic01: Here, `callback` is called if there is no error. Check `if` part.

mqtt/mqtt.js
drasko: So if I understand well - you are avoiding message to be published over MQTT by not calling this callback - you are just publishing to NATS and not to the other MQTT clients connected to MQTT broker.

Then message comes back over NATS and is published to other clients.

If this is correct - how does this not calling callback affects this NodeJS server (because obviously this is not the way that author imagined the use). @nmarcetic can there be some unexpected issues when not calling this callback (unreleased resources, memory leak, something other)?

docker/docker-compose.yml
```diff
@@ -144,3 +144,18 @@ services:
...
+
+  mqtt-adapter:
+    image: mainflux/mqtt:latest
+    container_name: mainflux-mqtt
+    depends_on:
+      - things
+    restart: on-failure
+    environment:
+      MF_MQTT_ADAPTER_PORT: 1883
+      MF_WS_PORT: 8880
```
drasko: If this is MQTT-over-WS I would propose `MF_MQTT_WS_PORT`

drasko (changes requested):

mqtt/mqtt.js
drasko: @nmarcetic @anovakovic01 any opinions on this?

docs/getting-started.md
```diff
@@ -305,3 +305,20 @@ coap get coap://localhost/channels/<channel_id>/messages?key=<client_token> -o
 
 where `-o` in indicates that value of `Observe` option is 0.
 To unsubscribe from channel, client should send same message with value of `Observe` option set to 1, ignore confirmable message received from service (by default, server will attempt to send confirmable message multiple times, as described in [RFC 7252](https://tools.ietf.org/html/rfc7252#page-28) or send `Reset` message with appropriate token.
+
+### MQTT
+
+To send and receive messages over MQTT you could use [Mosquitto](https://mosquitto.org),
```
drasko: Mosquitto tools, not just "Mosquitto" - because Mosquitto is MQTT broker, and Mosquitto tools are CLI tools `mosquitto_pub` and `mosquitto_sub`.

You can keep the same link, just add "tools"

drasko (changes requested):

mqtt/mqtt.js
anovakovic01: You are right, this callback should be called, in order to send `PUBACK` to publisher.

mqtt/package.json
```diff
@@ -17,28 +17,19 @@
     "aedes-logging": "^1.0.1",
```
drasko, line 17: Bump also the `aedes` version to latest one, `0.34.0`

drasko (dismissed):

mijicd (changes requested):

mqtt/mqtt.js
```diff
@@ -1,211 +1,139 @@
...
+aedes.authorizePublish = function (client, packet, publish) {
+    // Topics are in the form `channels/<channel_id>/messages`
+    var channel = packet.topic.split('/')[1];
+
+    things.CanAccess({
+        token: client.password,
+        chanID: channel
+    }, onAuthorize);
+
+    function onAuthorize(err, res) {
```
mijicd: Move this to the beginning of the function block, and use the variable syntax:

```javascript
var channel = packet.topic.split('/')[1],
    accessReq = {
        token: client.password,
        chanID: channel
    },
   onAuthorize = function (err, res) { ... };

things.CanAccess(accessReq, onAuthorize);
```

mqtt/mqtt.js
```diff
@@ -1,211 +1,139 @@
...
+        if (!err) {
+            logger.info('authorized publish');
+            
+            var rawMsg = message.RawMessage.encode({
```
mijicd: Since variables in javascript are function-scoped, not block-scoped, it makes sense to declare (or define) all locals at the beginning of the function (in this case only declare).

mqtt/mqtt.js
```diff
@@ -1,211 +1,139 @@
...
-        }
-    };
 
-    request(options, function (err, res) {
-        var error = null;
-        if (res && (res.statusCode === 200)) {
-            console.log('Subscribe authorized OK');
+aedes.authorizeSubscribe = function (client, packet, subscribe) {
+    // Topics are in the form `channels/<channel_id>/messages`
+    var channel = packet.topic.split('/')[1];
```
mijicd: See the remarks given above.

mqtt/mqtt.js
```diff
@@ -1,211 +1,139 @@
...
+    }, onAuthorize);
+
+    function onAuthorize(err, res) {
+        if (!err) {
+            logger.info('authorized subscribe');
+            subscribe(null, packet);
         } else {
-            console.log('Subscribe not authorized');
-            error = 4; // Bad username or password
+            logger.warn('unauthorizerd subscribe: %s', err);
```
mijicd: Please fix this typo.

mqtt/mqtt.js
```diff
@@ -1,211 +1,139 @@
...
-        var error = null;
-        var success = null;
-        if (res && (res.statusCode === 200)) {
-            // Set MQTT client.id to correspond to Mainflux device UUID
-            c.id = res.headers['x-client-id'];
-            // Store password for future references
-            c.password = password;
-            success = true;
+aedes.authenticate = function (client, username, password, acknowledge) {
+    var pass = password || "",
```
mijicd: I was thinking more about `var pass = (password || "").toString();`. Thus being said, it might be even better idea to apply the same refactoring here as well:

```javascript
// feel free to find more suitable variable names
var token = (password || "").toString(),
    identReq = {value: token},
    onIdentify = function (err, res) { ... };

things.identify(identReq, onIdentify);
```

mijicd (changes requested):

mijicd (approved):

PR #286, opened 2018-05-17
NOISSUE - Fixes the broken docker-compose

autolyticus:
Relative paths are completely supported in compose files


nmarcetic (dismissed):
@reisub0 Please sign-off your commit 
https://github.com/apps/dco

autolyticus:
Replaced PWD variables with simple dots which prevented ```docker-compose up``` from running successfully.

dborovcanin (dismissed):
You should remove `+ Signed of ` from commit message since it's obvious that commit is signed now.

mijicd (approved):

PR #283, opened 2018-05-16
MF-236 - Add support for storing messages in InfluxDB

dborovcanin:
This pull request resolves #236 

Also, this pull request implements service consuming messages from NATS and storing them in InfluxDB.

PR #290, opened 2018-05-18
NOISSUE - Fix potential race in gRPC tests

dborovcanin:

### DIFF
--- a/things/api/grpc/endpoint_test.go
+++ b/things/api/grpc/endpoint_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


users/api/grpc/endpoint_test.go
```diff
@@ -17,7 +17,7 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-const port = 8080
+const port = 8081
```
anovakovic01, line 20: Why have you changed port?

things/api/grpc/endpoint_test.go
```diff
@@ -109,3 +98,18 @@ func TestIdentify(t *testing.T) {
 		assert.Equal(t, tc.code, e.Code(), fmt.Sprintf("%s: expected %s got %s", desc, tc.code, e.Code()))
 	}
 }
+
+func newService(tokens map[string]string) things.Service {
```
anovakovic01: Please move these private functions at the beginning in order to be consistent (every other test file has private functions at the beginning).

anovakovic01 (dismissed):
Please check if `users` service needs this kind of update and add it here.

users/api/grpc/endpoint_test.go
dborovcanin: I've changed port because other test uses the same port, so running test concurrently could cause usage of the already taken port.

things/api/grpc/endpoint_test.go
dborovcanin: I didn't want anything to be before `TestMain` so, in order to be consistent with #283, I've placed unexported methods at the end of the file.

mijicd (approved):

PR #291, opened 2018-05-20
MF-269 - Replace UUID with BIGSERIAL as primary key

### LINKED ISSUE #269: Replace UUID PKs with auto-incremented values
`things` service relies on having UUID as primary key for their entities due to the legacy described in #268. Instead of generating identifiers, we should let the database do its "job", and replace UUID primary keys with `SERIAL`.

**Acceptance criteria:**
- using `SERIAL` as PK
- updated mocks
- updated tests

mijicd:
This pull request resolves #269.

**Summary:**
- dropped CoAP (incompatible API, poor library support, no tests)
- replaced UUID with `BIGSERIAL` (`uint64`) as things and channels primary key
- propagated these changes through protobuf schemas


things/api/http/requests_test.go
```diff
@@ -58,16 +58,16 @@ func TestAddThingReqValidation(t *testing.T) {
...
 	cases := map[string]struct {
 		thing things.Thing
-		id    string
+		id    uint64
 		key   string
 		err   error
 	}{
 		"valid thing update request": {thing, id, key, nil},
-		"non-uuid thing ID":          {thing, wrong, key, things.ErrNotFound},
+		"invalid thing ID":           {thing, 0, key, things.ErrNotFound},
```
drasko: Why `wrong` constant is not used here, and it is in many other places in tests? It would increase code self-documentation.

drasko (dismissed):

docs/getting-started.md
```diff
@@ -281,47 +281,3 @@ to send `Authorization` header with client authorization token.
...
-
-To subscribe to channel, client should send following confirmable message:
-
-```
-coap get coap://localhost/channels/<channel_id>/messages?key=<client_token> -o
-```
-
-where `-o` in indicates that value of `Observe` option is 0.
-To unsubscribe from channel, client should send same message with value of `Observe` option set to 1, ignore confirmable message received from service (by default, server will attempt to send confirmable message multiple times, as described in [RFC 7252](https://tools.ietf.org/html/rfc7252#page-28) or send `Reset` message with appropriate token.
-
```
dborovcanin, line 308: Should MQTT docs be removed?

dborovcanin (dismissed):
LGTM!

http/api/endpoint_test.go
```diff
@@ -60,31 +57,30 @@ func (tr testRequest) make() (*http.Response, error) {
...
 		contentType string
 		auth        string
 		status      int
 	}{
-		"publish message":                                  {id, msg, "application/senml+json", token, http.StatusAccepted},
-		"publish message with no authorization token":      {id, msg, "application/senml+json", "", http.StatusForbidden},
-		"publish message with invalid authorization token": {id, msg, "application/senml+json", invalidToken, http.StatusForbidden},
+		"publish message":                                  {id, msg, contentType, token, http.StatusAccepted},
+		"publish message with no authorization token":      {id, msg, contentType, "", http.StatusForbidden},
+		"publish message with invalid authorization token": {id, msg, contentType, invalidToken, http.StatusForbidden},
```
anovakovic01: Maybe add test case `publish message to non-existent channel`?

anovakovic01 (changes requested):

anovakovic01 (approved):
LGTM!

PR #292, opened 2018-05-21
NOISSUE - Remove redundant error check

dborovcanin:


writers/influxdb/influxdb.go
```diff
@@ -41,10 +41,7 @@ func (repo *influxRepo) Save(msg mainflux.Message) error {
 	}
 
 	bp.AddPoint(pt)
-	if err := repo.client.Write(bp); err != nil {
-		return err
```
anovakovic01, line 45: Maybe add public error?

anovakovic01 (dismissed):

drasko (commented):
Looks OK on first look, I think we can accept

writers/influxdb/influxdb.go
dborovcanin: We could. @mijicd Suggestions? Currently, I see no benefits of doing that, but I don't mind exporting an error, as well.

mijicd: There's no point of adding that.

mijicd (approved):

PR #293, opened 2018-05-21
NOISSUE - Add MQTT to kubernetes nginx

anovakovic01:
**Summary:**
- Add soft links for nginx configuration and certificates
- Add dashflux to nginx configuration for docker-compose
- Update documentation


k8s/nginx/nginx.conf
```diff
@@ -0,0 +1 @@
+/Users/novke/go/src/github.com/mainflux/mainflux/docker/nginx.conf
```
drasko: Hardcoded path?

docker/nginx.conf
```diff
@@ -214,6 +214,22 @@ http {
                 return 200;
             }
         }
+
+        location / {
```
drasko, line 218: Is this thing adding MQTT or Dashflux?

k8s/nginx/nginx.conf
nmarcetic: ?

k8s/nginx/ssl
```diff
@@ -0,0 +1 @@
+/Users/novke/go/src/github.com/mainflux/mainflux/docker/ssl
```
drasko: Hardcoded path?

nmarcetic: ?

drasko (changes requested):

nmarcetic (changes requested):
What are dose files with hard-coded local paths  ?

docker/nginx.conf
anovakovic01: This is adding Dashflux, because now, compose and k8s share same configuration, and k8s already had Dashflux.

nmarcetic (approved):

docker/nginx.conf
drasko: I see, but PR name is confusing then.

drasko (approved):

docker/nginx.conf
anovakovic01: Point of this PR is to add MQTT to nginx for kubernetes. This change was needed as I had to share nginx configuration.

PR #309, opened 2018-05-25
NOISSUE - Fix channel closing bug in WebSocket adapter

anovakovic01:
Fix channel closing bug by adding `closed` flag and mutex.

### DIFF
--- a/ws/adapter.go
+++ b/ws/adapter.go
@@ -4,6 +4,7 @@ package ws
 
 import (
 	"errors"
+	"sync"
 
 	"github.com/mainflux/mainflux"
 	broker "github.com/nats-io/go-nats"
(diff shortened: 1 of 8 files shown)


ws/adapter.go
```diff
@@ -25,17 +26,44 @@ type Service interface {
...
+		Closed:   make(chan bool),
+		closed:   false,
+		mutex:    sync.Mutex{},
+	}
+}
+
+// Send method send message over Messages channel.
+func (channel *Channel) Send(msg mainflux.RawMessage) {
+	channel.mutex.Lock()
+	defer channel.mutex.Unlock()
```
dborovcanin, line 53: Do we really need `defer`? Explicit Unlock is faster.

ws/adapter.go
```diff
@@ -25,17 +26,44 @@ type Service interface {
...
+	if !channel.closed {
+		channel.Messages <- msg
+	}
 }
 
 // Close channel and stop message transfer.
-func (channel Channel) Close() {
+func (channel *Channel) Close() {
+	channel.mutex.Lock()
+	defer channel.mutex.Unlock()
```
dborovcanin, line 63: Same here, but just to be consistent (performance isn't a problem here since this method is called only once), since `Send` method will be used much more.

dborovcanin (changes requested):

ws/adapter.go
anovakovic01: I am aware that it is faster, but it is more readable this way. I would keep this for now, and we can optimize it later if that kind of optimization is needed.

ws/adapter.go
anovakovic01: Same here.

dborovcanin (approved):

PR #324, opened 2018-06-11
NOISSUE - Raise test coverage

dborovcanin:
Raise test coverage.


drasko (dismissed):
LGTM

anovakovic01 (changes requested):

http/mocks/things.go
```diff
@@ -6,10 +6,15 @@ import (
...
 	"github.com/mainflux/mainflux/things"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 var _ mainflux.ThingsServiceClient = (*thingsClient)(nil)
 
+// ErrToken is used to simulate internal server error.
+const ErrToken = "unavailable"
```
anovakovic01: Rename this to something else, this is convention for `error` type.

ws/mocks/things.go
```diff
@@ -7,10 +7,15 @@ import (
...
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 var _ mainflux.ThingsServiceClient = (*thingsClient)(nil)
 
+// ErrToken is used to simulate internal server error.
+const ErrToken = "unavailable"
```
anovakovic01: Same here.

anovakovic01 (changes requested):
Almost there.

http/mocks/things.go
dborovcanin: Irrelevant. If you find the more appropriate name, feel free to rename.

anovakovic01: It is not irrelevant, we should be consistent, even in mock. This name is confusing. Call it `UnavailableToken`, or something like that.

nmarcetic: I agree with @anovakovic01 I

anovakovic01 (approved):
LGTM!

PR #330, opened 2018-06-15
NOISSUE - Expose SenML subject from Normalizer service

dborovcanin:

### DIFF
--- a/normalizer/normalizer.go
+++ b/normalizer/normalizer.go
@@ -13,8 +13,6 @@ import (
 )
 
 const (
-	// OutputSenML represents subject SenML messages will be published to.
-	OutputSenML   = "out.senml"
 	queue         = "normalizers"
 	input         = "channel.*"
 	outputUnknown = "out.unknown"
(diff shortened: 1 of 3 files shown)


anovakovic01 (commented):
Why do we need this?

drasko:
@anovakovic01 Dusan is trying to avoid duplicating strings and keeping them only in one place.

This is no necessary, because we have string only on 2 places for now, but I do not think that it is bad.

anovakovic01 (commented):
@drasko Thanks for explanation. I don't think that it is bad, I just wanted to know purpose of this change. @dusanb94 Maybe we should move this constant to project root? That way, we avoid writer being directly dependent on normalizer (import of normalizer is dependency).

drasko:
@anovakovic01 @dusanb94 yes - moving the libs/code that is shared between several components to the root dir is a good idea.

Note specially that in some scenarios `writer`s will be used (probably after some modifications) without `normalizer` - where `SenML` is not used, but some other custom message format.

drasko:
Additionally, separate file in the project root that will hold all topics (NATS subjects) used in the system with proper comments would be a great source of documentation.

drasko (changes requested):
PLease move topic information to the file in the project root.

anovakovic01 (approved):
LGTM!

drasko (approved):
LGTM

PR #331, opened 2018-06-25
MF-313 - Implement basic Cassandra reader

### LINKED ISSUE #313: Implement basic Cassandra reader
A basic reader that only reads messages and has a pagination should be implemented to facilitate experimenting with and demoing Mainflux system.

Reader exposes only one HTTP REST API endpoint to read messages from DB.

Something similar existed before for MongoDB: https://github.com/MainfluxLabs/mainflux-mongodb-reader

anovakovic01:
This pull request resolves #313.

**Summary:**
- Add common reader interface and shared HTTP API
- Add Cassandra reader implementation
- Add `docker-compose.yml` for Cassandra reader
- Add tests for HTTP API
- Add tests for Cassandra reader
- Add docker network for core services


readers/api/transport.go
```diff
@@ -0,0 +1,137 @@
...
+	auth = tc
+
+	opts := []kithttp.ServerOption{
+		kithttp.ServerErrorEncoder(encodeError),
+	}
+
+	mux := bone.New()
+	mux.Get("/channels/:chanID/messages", kithttp.NewServer(
+		listMessagesEndpoint(svc),
+		decodeList,
```
dborovcanin, line 43: Rename to `decodeRequest` or `decodeReq`.

readers/api/transport.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		return err
+	}
+
+	return nil
+}
+
+func getQuery(req *http.Request, name string, fallback uint64) uint64 {
+	vals := bone.GetQuery(req, name)
+	if len(vals) == 0 {
```
dborovcanin: Multiple limit and offset query params (`...&limit=3&offset=4&limit=32`) should cause an error.

readers/messages.go
```diff
@@ -0,0 +1,17 @@
...
+	"errors"
+
+	"github.com/mainflux/mainflux"
+)
+
+// ErrNotFound indicates that requested entity doesn't exist.
+var ErrNotFound = errors.New("entity not found")
+
+// MessageRepository specifies message reader API.
+type MessageRepository interface {
```
dborovcanin, line 13: We can rename `MessageReposistory` (reader as well as writer) to `Reader` and `Writer`.

dborovcanin (changes requested):
README.md and API documentation is missing, but those could be a part of another noissue pull request.

anovakovic01:
@dusanb94 I put WIP label for a reason. This PR is not ready for review. Don't waste your time with this until it is ready.

nmarcetic:
LGTM! 
I have two small remarks. You changed the docker network to `docker_mainflux-base-net` but didn't changed the network in `docker/addons` so I suppose that addons like influx, mongo etc... will not work with core composition. The second one is README file, I don't see any README also I think `offset` and `limit` should be mentioned in README also their default values (if not sent as query param what will happen). 

Nice job @anovakovic01  ;)

nmarcetic (approved):

readers/api/transport.go
```diff
@@ -0,0 +1,150 @@
...
+		limit:  limit,
+	}
+
+	return req, nil
+}
+
+func encodeResponse(_ context.Context, w http.ResponseWriter, response interface{}) error {
+	w.Header().Set("Content-Type", contentType)
+
+	if ar, ok := response.(mainflux.Response); ok {
```
dborovcanin, line 86: Rename `ar`.

dborovcanin (changes requested):

readers/api/transport.go
anovakovic01: This variable is called like this everywhere, so I don't see a reason for that. If you want to change it, send separate PR where you will update it in other services too.

dborovcanin (approved):

PR #347, opened 2018-08-13
Add network to influxdb. Fix #345

### LINKED ISSUE #345: No connection to InfluxDB
The influx docker-compose.yaml needs the same network change (‘docker_mainflux-base-net’) in order to be able to connect to it again.

ks156:
Signed-off-by: Paul RATHGEB <paul.rathgeb@skynet.be>

### DIFF
--- a/dashflux/src/app/components/things/thing-dialog/thing-dialog.component.html
+++ b/dashflux/src/app/components/things/thing-dialog/thing-dialog.component.html
@@ -23,7 +23,7 @@
       </mat-error>
     </mat-form-field>
     <mat-form-field>
-      <textarea formControlName="payload" matInput placeholder="Payload" matTextareaAutosize matAutosizeMinRows="2"
+      <textarea formControlName="metadata" matInput placeholder="Metadata" matTextareaAutosize matAutosizeMinRows="2"
                 matAutosizeMaxRows="5"></textarea>
     </mat-form-field>
   </div>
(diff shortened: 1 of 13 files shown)


drasko:
LGTM, but maybe generalizing the `networks` for all the service in this influxdb docker-compose add-on as @RubenN  suggested would be better solution. I guess for Grafana it is not necessary, because only Influx "links" to the services from other docker-compose file, but in more generic case there will be more services that will need this connection.

@dusanb94 @anovakovic01 @nmarcetic please review.

dborovcanin:
@drasko I share your opinion. I think networking should be handled the same way as in the other services, which is the approach @RubenN suggested.

drasko:
@ks156 can you please refine your PR to use approach that @RubenN suggested.

ks156:
@drasko Synced with master, and added the approach of @RubenN

drasko (approved):
LGTM

docker/addons/influxdb/docker-compose.yml
```diff
@@ -8,6 +8,10 @@
 
 version: "3"
 
+networks:
```
drasko: Actually - should be `external` - like in other `writers`: https://github.com/mainflux/mainflux/blob/master/docker/addons/mongodb-writer/docker-compose.yml#L12

drasko (changes requested):

docker/addons/influxdb/docker-compose.yml
ks156: Changes applied.

drasko: Actually, `docker_` prefix is missing. I think it is important for Kubernetes.

drasko (changes requested):

ks156:
#361 replaces this PR

PR #349, opened 2018-08-15
NOISSUE - Rename bashflux to cli and fix cert path

drasko:


cmd/cli/main.go
```diff
@@ -3,7 +3,7 @@ package main
 import (
 	"log"
 
-	bf "github.com/mainflux/mainflux/bashflux"
+	bf "github.com/mainflux/mainflux/cli"
```
dborovcanin: Please omit package name if there is no special reason to use `bf` instead of `cli`.

cmd/cli/main.go
```diff
@@ -19,7 +19,7 @@ func main() {
 
 	// Root
 	var rootCmd = &cobra.Command{
-		Use: "bashflux",
+		Use: "mainflux-cli",
```
dborovcanin, line 22: Maybe we can pick some other name for this? It will be easier to remember and use if we take shorter and simpler name. Simply `mainflux` (since it's obvious that it is a cli tool)?

dborovcanin (changes requested):

cmd/cli/main.go
drasko: Good spot, will change this.

cmd/cli/main.go
drasko: Actually, Makefile currently adds `mainflux-` prefix, and it's easy to add `cli` to current buils system like that. This `mainflux-` prefix is added because binaries are installed in `$GOBIN`. Let's keep it like this for now, and we'll see later for the refinement.

dborovcanin (approved):

anovakovic01 (approved):
LGTM!

PR #351, opened 2018-08-16
NOISSUE - Add Development Guide doc

drasko:

### DIFF
--- a/docs/dev-guide.md
+++ b/docs/dev-guide.md
@@ -1,0 +1,145 @@
+## Getting Mainflux
+
+Mainflux can be fetched from official [Mainflux GitHub repository](https://github.com/Mainflux/mainflux):
+
+```
+mkdir -p $GOPATH/src/github.com/mainflux
+cd $GOPATH/src/github.com/mainflux
+git clone https://github.com/mainflux/mainflux.git
+cd mainflux
+```
+
+## Building
+
+### Prerequisites
+Make sure that you have [Protocol Buffers](https://developers.google.com/protocol-buffers/) compiler (`protoc`) installed.
+
+[Go Protobuf](https://github.com/golang/protobuf) installation instructions are [here](https://github.com/golang/protobuf#installation).
+Go Protobuf uses C bindings, so you will need to install [C++ protobuf](https://github.com/google/protobuf) as a prerequisite.
+
+### Build All Services
+
+Use `GNU Make` tool to build all mainflux services:
+
+```
+make
+```
+
+Build artefacts will be put in the `build` directory.
+
+> N.B. All Mainflux services are built as a statically linked binaries. This way they can be portable (transfered to any platform just by placing them there and running them) as they contain all needed libraries and do not relay on system shared libs. This helps creating [FROM scratch](https://hub.docker.com/_/scratch/) dockers.
+
+### Build Individual Microservice
+Individual microservices can be built with command:
+
+```
+make <microservice_name>
+```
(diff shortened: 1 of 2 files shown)


chombium (changes requested):
well done @drasko. You have written the missing manual which will save us a ton of discussions in the mailing list and the Gitter channel.

I would add few warnings and one optional step as well.

As an optional step I would describe few changes how to add a volume to the ```things-db``` and ```users-db``` container, so that the data doesn't lost when this container is rebuilt and
executing again the db initialization of the [users](https://github.com/mainflux/mainflux/blob/master/users/postgres/init.go) and [thing](https://github.com/mainflux/mainflux/blob/master/things/postgres/init.go) services, so that the db state is the latest.

The thing with the migrations is that it would probably need a bit of refactoring so that they can be executed whenever that's needed. We should probably also replace the "CREATE TABLE" with "CREATE TABLE IF NOT EXISTS" (available from pgsql version 9.1). Do you think that this needs a ticket?

As a warning I would add that when the executing ```make dockers``` or ```make docker_things-db``` or ```make docker_users-db``` the data in the pgsql container will be lost.

docs/dev-guide.md
```diff
@@ -0,0 +1,161 @@
...
+mkdir -p $GOPATH/src/github.com/mainflux
+cd $GOPATH/src/github.com/mainflux
+git clone https://github.com/mainflux/mainflux.git
+cd mainflux
+```
+
+## Building
+
+### Prerequisites
+Make sure that you have [Protocol Buffers](https://developers.google.com/protocol-buffers/) compiler (`protoc`) installed.
```
dborovcanin, line 15: I think that `protoc` version matters and that we currently use `3.5.1`. Maybe it's a good idea to mention that here.

docs/dev-guide.md
```diff
@@ -0,0 +1,161 @@
...
+
+```
+make docker_http
+```
+
+> N.B. Mainflux creates `FROM scratch` docker containers which as compact and small in size.
+
+### MQTT Microservice
+MQTT Microservice in Mainflux is special, as it is currently the only microservice written in NodeJS. It is not compiled,
+but node modules need to be download in order to start the service:
```
anovakovic01: `downloaded`

docs/dev-guide.md
```diff
@@ -0,0 +1,161 @@
...
+npm install
+```
+
+After that MQTT Adapter can be started from top directory (as it needs to find `*.proto` files) with:
+```
+node mqtt/mqtt.js
+```
+
+### Protobuf
+Aforemntioned  `make` (which is an alias for `make all` target) is calling `protoc` command prior to compiling individual microservices.
```
anovakovic01: `Aforementioned`

docs/dev-guide.md
```diff
@@ -0,0 +1,161 @@
...
+
+To do this by hand, execute:
+
+```
+protoc --go_out=plugins=grpc:. *.proto
```
anovakovic01, line 89: Maybe add that it can be done also by calling `make proto`.

anovakovic01 (changes requested):

docs/dev-guide.md
drasko: I would leave this out of the generic doc, as it is just the implementation detail.

anovakovic01 (approved):
LGTM!

chombium (approved):

PR #355, opened 2018-08-16
NOISSUE - Use mainflux prefix for NginX certs

drasko:
`dhparam.pem` should be prefixed with `mainflux-`,
as are other certs in prder to avoid collisions
when copying them to a local system.


anovakovic01 (changes requested):
This will break `k8s` setup, so change and test `k8s` setup also, or lets do this some other time.

dborovcanin (commented):
Please be careful here. Does this affect docker-compose and k8s setup (since nginx.conf is shared)? Should we rename `/docker/ssl/dhparam.pem` accordingly?

drasko:
Indeed, this is touchy. I tried to setup NginX on my laptop and make it pick up local conf with `-c`, then started copying over files and wanted to facilitate this copying of certs by prefixing them with `mainflux-`.

I have corrected all occurrences of `dhparam`, but I would not like to break something :).

drasko:
Dropping this one for now, due unnecessary potential danger (and extreme low priority).

PR #383, opened 2018-09-04
MF-382 - Improve performance by adding Redis cache for message auth

### LINKED ISSUE #382: Improve performance by adding Redis cache for message auth
In Mainflux every publish/subscribe on the channel must be authenticated. For this to happen, adapter service sends `gRPC` request to `things` service, which does a lookup in the `PostgreSQL` DB.

Reading from unclustered Postgres is slow, so it is needed to add a Redis cache in the middle to improve the performance (throughput).

anovakovic01:
This pull request resolves #382.


drasko (dismissed):
LGTM

dborovcanin (dismissed):

cmd/things/main.go
```diff
@@ -14,64 +14,78 @@ import (
...
-	envGRPCPort = "MF_THINGS_GRPC_PORT"
-	envUsersURL = "MF_USERS_URL"
+	defDBHost    = "localhost"
+	defDBPort    = "5432"
+	defDBUser    = "mainflux"
+	defDBPass    = "mainflux"
+	defDBName    = "things"
+	defCacheURL  = "localhost:6379"
+	defCachePass = ""
+	defCacheDB   = "0"
```
ghost, line 44: I suggest adding ability to set cache expiration default if it's not already in

things/channels.go
```diff
@@ -50,3 +50,12 @@ type ChannelRepository interface {
...
 	HasThing(uint64, string) (uint64, error)
 }
+
+// ChannelCache contains channel-thing connection caching interface.
+type ChannelCache interface {
+	// Saves channel thing connection.
+	Save(uint64, uint64, string) error
+
+	// Connected checks if thing is connected to channel.
+	Connected(uint64, string) (uint64, error)
```
ghost: Maybe CheckIfConnected or something similar but shorter would fit better? You used an adjective instead of a verb as a method name

cmd/things/main.go
anovakovic01: We should leave that for some other PR, when we see if we need to cache some other stuff, and also, maybe this shouldn't be configurable. For now, lets leave it like this.

drasko (approved):
LGTM

drasko:
Boom! Merged!

PR #392, opened 2018-09-13
NOISSUE - Dev dockers

dborovcanin:
**Summary:**
This pull request provides building Docker images locally `for development purposes`. This is done in order to make building images much faster.
Building images for deployment is set not to use cache.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 BUILD_DIR = build
-SERVICES = users things http normalizer ws influxdb mongodb-writer mongodb-reader cassandra-writer cassandra-reader cli
+SERVICES = users things http normalizer ws influxdb-writer influxdb-reader mongodb-writer mongodb-reader cassandra-writer cassandra-reader cli
 DOCKERS = $(addprefix docker_,$(SERVICES))
 DOCKERS_DEV = $(addprefix docker_dev_,$(SERVICES))
 CGO_ENABLED ?= 0
@@ -28,10 +28,13 @@ clean:
 install:
 	cp ${BUILD_DIR}/* $(GOBIN)
 
+test:
+	GOCACHE=off go test -v -race -tags test $(shell go list ./... | grep -v 'vendor\|cmd')
+
 proto:
 	protoc --go_out=plugins=grpc:. *.proto
 
-$(SERVICES): proto
+$(SERVICES):
 	$(call compile_service,$(@))
 
 $(DOCKERS):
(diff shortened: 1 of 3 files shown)


drasko (changes requested):

Makefile
```diff
@@ -23,13 +28,10 @@ clean:
...
 	cp ${BUILD_DIR}/* $(GOBIN)
 
-test:
-	GOCACHE=off go test -v -race -tags test $(shell go list ./... | grep -v 'vendor\|cmd')
-
 proto:
 	protoc --go_out=plugins=grpc:. *.proto
 
-$(SERVICES):
+$(SERVICES): proto
```
drasko: We do not need anymore to make proto

drasko (changes requested):

docs/dev-guide.md
```diff
@@ -67,6 +67,22 @@ make docker_http
...
+make dockers_dev
+```
+
+or individually with
+
+```bash
+make docker_dev_<microservice_name>
+```
+
+Commands `make dockers` and `make dockers_dev` are similar. The main difference is that building images in the development mode is done on the local machine, rather than an intermediate image, which makes building images much faster. Before running this command, corresponding binary needs to be built in order to make changes visible. This can be done using `make` or `make <service_name>` command. Commands `make dockers_dev` and `make_docker_dev <service_name>` should be used only for development to speed up the process of image building. **For deployment images, commands from section above should be used.**
```
drasko: > Commands `make dockers_dev` and `make_docker_dev <service_name>`

Is there really a target `make_docker_dev <service_name>`? Or should it be `make docker_dev_<service_name>`?

drasko (changes requested):

ghost (dismissed):
LGTM

anovakovic01 (dismissed):
LGTM!

drasko (dismissed):
LGTM

docs/dev-guide.md
```diff
@@ -66,6 +66,22 @@ make docker_http
...
+make dockers_dev
+```
+
+or individually with
+
+```bash
+make docker_dev_<microservice_name>
+```
+
+Commands `make dockers` and `make dockers_dev` are similar. The main difference is that building images in the development mode is done on the local machine, rather than an intermediate image, which makes building images much faster. Before running this command, corresponding binary needs to be built in order to make changes visible. This can be done using `make` or `make <service_name>` command. Commands `make dockers_dev` and `make_docker_dev_<service_name>` should be used only for development to speed up the process of image building. **For deployment images, commands from section above should be used.**
```
anovakovic01: Shouldn't it be `make docker_dev_<service_name>`?

anovakovic01 (changes requested):

docs/dev-guide.md
drasko: yes - it should.

anovakovic01 (approved):
LGTM!

drasko (approved):
LGTM

Issue #379, opened 2018-08-30
Add debug log level to Mainflux logger


ghost:
Currently, the Mainflux logger is missing the debug log level. There are many cases were this level can be used. Also, add `SetLevel` method to Mainflux logger. Every service should have this logging level parameter set as environment variable `MF_<service_name>_LOG_LEVEL`.

mfinley3:
Along with adding the debug level, I was wondering if it would be beneficial to also be able to select the log level? That way someone can select if they want mainflux to write out all of the infos or not. 

Specifically something that would look like this: 
```
| Variable             | Description                              | Default      |
|----------------------|------------------------------------------|--------------|
| MF_USERS_LOG_LEVEL   | Log level for Users (info, warn, error)  | info         | 
```
For my team and I, everything that goes into stdout gets pulled into the ELK stack for us. So it would be beneficial for us to reduce the number of success messages printed out in each component.

Let me know your thoughts on this. If it is something that sounds useful I can get a PR open asap, work has been done just have some hoops to jump through.

anovakovic01:
This was the plan, just not written in issue description. I will update description.

drasko:
@mfinley3 looks like we are aligned, will you be sending PR?

mfinley3:
Yep! I will get a PR opened here ASAP. I am working on our teams path to contributing back now!

I made a few small cleanup changes to the internal PR. Going to have someone look those over and should have an open PR here tomorrow morning!

anovakovic01:
Resolved by #402.

PR #402, opened 2018-09-21
MF-379 - Log Level Option

mfinley3:
### What does this do?
Adds a log level option for each component of mainflux so users can set what logs they want.

#379 

Nothing should break/modify default functionality.

Unit tests for the log level have been added

Yes, readmes have been updated with the new log level env option.

### Notes
If any bad input is given or level is left unselected, level defaults to info. 

Regarding the size of this PR, let me know if you would me to break this up in some way since there were two vendor files added

Signed-off-by: mfinley3 <Michael.Finley@target.com>


chombium:
@mfinley3 is there any specific reason why you've set 'info' as the default log level?
Usually 'error' is set as default level so that  in production the logs are not flooded with messages which would show that the system is working properly and you keep the config files smaller. Using something other then 'error' in production would mean that you are examining something in the system.

I would suggest that you to change the default level to 'error'.

@mainflux/contributors any thoughts on this?

drasko:
@mfinley3 I agree with @chombium - `error` should be the default level.

cmd/cassandra-reader/main.go
```diff
@@ -30,18 +30,21 @@ import (
 const (
 	sep = ","
 
+	defLogLevel  = "info"
```
drasko: Please change to `error`

cmd/cassandra-writer/main.go
```diff
@@ -30,18 +30,21 @@ const (
 	sep   = ","
 
 	defNatsURL  = nats.DefaultURL
+	defLogLevel = "info"
```
drasko: Should be `error`.

cmd/http/main.go
```diff
@@ -28,27 +28,31 @@ import (
 
 const (
 	defPort      string = "8180"
+	defLogLevel  string = "info"
```
drasko: Same here.

cmd/influxdb-reader/main.go
```diff
@@ -21,6 +21,7 @@ import (
 
 const (
 	defThingsURL = "localhost:8181"
+	defLogLevel  = "info"
```
drasko: Same here.

cmd/influxdb-writer/main.go
```diff
@@ -28,6 +28,7 @@ const (
 	queue = "influxdb-writer"
 
 	defNatsURL   = nats.DefaultURL
+	defLogLevel  = "info"
```
drasko: And here.

cmd/mongodb-reader/main.go
```diff
@@ -29,12 +29,14 @@ import (
 
 const (
 	defThingsURL = "localhost:8181"
+	defLogLevel  = "info"
```
drasko: And here.

cmd/things/main.go
```diff
@@ -34,6 +34,7 @@ import (
 )
 
 const (
+	defLogLevel  = "info"
```
drasko: And here.

cmd/users/main.go
```diff
@@ -31,6 +31,7 @@ import (
 )
 
 const (
+	defLogLevel = "info"
```
drasko: And here

cmd/ws/main.go
```diff
@@ -28,27 +28,31 @@ import (
 
 const (
 	defPort      = "8180"
+	defLogLevel  = "info"
```
drasko: Same here.

ws/README.md
```diff
@@ -8,11 +8,12 @@ The service is configured using the environment variables presented in the
...
 default values.
 
-| Variable             | Description         | Default               |
-|----------------------|---------------------|-----------------------|
-| MF_WS_ADAPTER_PORT   | Service WS port     | 8180                  |
-| MF_NATS_URL          | NATS instance URL   | nats://localhost:4222 |
-| MF_THINGS_URL        | Things service URL  | localhost:8181        |
+| Variable                  | Description                   | Default               |
+|---------------------------|-------------------------------|-----------------------|
+| MF_WS_ADAPTER_LOG_LEVEL   | Log level for the WS Adapter  | info                  |
```
drasko: Default log level is error.

writers/mongodb/messages_test.go
```diff
@@ -26,7 +26,7 @@ import (
 var (
 	port       string
 	addr       string
-	testLog    = log.New(os.Stdout)
+	testLog    = log.New(os.Stdout, "info")
```
drasko: Strign consts should be defined at the top of the file.

users/README.md
```diff
@@ -18,6 +18,7 @@ default values.
 
 | Variable             | Description                              | Default      |
 |----------------------|------------------------------------------|--------------|
+| MF_USERS_LOG_LEVEL   | Log level for Users (info, warn, error)  | info         |
```
drasko: Default -> error

cmd/normalizer/main.go
```diff
@@ -26,24 +26,28 @@ import (
 )
 
 const (
-	defNatsURL string = broker.DefaultURL
-	defPort    string = "8180"
-	envNatsURL string = "MF_NATS_URL"
-	envPort    string = "MF_NORMALIZER_PORT"
+	defNatsURL  string = broker.DefaultURL
+	defLogLevel string = "info"
```
drasko: default to error

http/README.md
```diff
@@ -8,11 +8,12 @@ The service is configured using the environment variables presented in the
...
 default values.
 
-| Variable             | Description         | Default               |
-|----------------------|---------------------|-----------------------|
-| MF_HTTP_ADAPTER_PORT | Service HTTP port   | 8180                  |
-| MF_NATS_URL          | NATS instance URL   | nats://localhost:4222 |
-| MF_THINGS_URL        | Things service URL  | localhost:8181        |
+| Variable                    | Description                    | Default               |
+|-----------------------------|--------------------------------|-----------------------|
+| MF_HTTP_ADAPTER_LOG_LEVEL   | Log level for the HTTP Adapter | info                  |
```
drasko: Default -> err

drasko (changes requested):
Very good code here @mfinley3 - just fix default logging level and we are there as long as I am concerned.

writers/mongodb/messages_test.go
mfinley3: Good catch here, I have made this change to all files similar to this as well!
Ex: ws/api/transport_test.go

mfinley3:
Default level updated to error :)
Let me know if there is anything else!

drasko (dismissed):
LGTM

manuio (dismissed):
LGTM!

chombium:
well done @mfinley3 
LGTM!

logger/logger.go
```diff
@@ -23,27 +27,44 @@ type Logger interface {
 	Error(string)
 }
 
+var logLevels = map[string]level.Option{
+	"debug": level.AllowDebug(),
+	"info":  level.AllowInfo(),
+	"warn":  level.AllowWarn(),
+	"error": level.AllowError(),
+}
```
anovakovic01: I would suggest adding enumeration instead of plain strings for log levels. Also, add method for casting string to enumeration values so that it can be used from `main.go`.

drasko:
Thanks for review @chombium. Let's hold the merge until @anovakovic01 approve, he mentioned quickly that he would prefer enum instead of string map for log level. I'll let him comment.

logger/logger.go
```diff
@@ -23,27 +27,44 @@ type Logger interface {
...
 }
 
 // New returns wrapped go kit logger.
-func New(out io.Writer) Logger {
+func New(out io.Writer, logLevel string) Logger {
 	l := log.NewJSONLogger(log.NewSyncWriter(out))
+	if allowedLevel, ok := logLevels[strings.ToLower(logLevel)]; ok {
+		l = level.NewFilter(l, allowedLevel)
+	} else {
+		l = level.NewFilter(l, level.AllowError())
```
anovakovic01: Please return error if invalid log level is provided.

logger/logger.go
```diff
@@ -23,27 +27,44 @@ type Logger interface {
...
+
 var _ Logger = (*logger)(nil)
 
 type logger struct {
 	kitLogger log.Logger
 }
 
 // New returns wrapped go kit logger.
-func New(out io.Writer) Logger {
+func New(out io.Writer, logLevel string) Logger {
```
anovakovic01: Use this enumeration instead of string.

anovakovic01 (changes requested):

drasko:
Exaple for enumeration for log level and string representation: https://github.com/uber-go/zap/blob/f4243df2af592a2e0ccdccb77e1ad44f78d735e5/zapcore/level.go#L34

logger/logger.go
mfinley3: I'm struggling a bit with this one. 
I removed the enumeration in [mainflux/level.go](https://github.com/mainflux/mainflux/pull/402/files#diff-a8b26aac2e521bc9f1a99413d2713efaL1) in favor of [vendor/level.go](https://github.com/mainflux/mainflux/blob/b07c6ae26499758753de2b223fab747610dceef9/vendor/github.com/go-kit/kit/log/level/level.go#L178-L205).
It seemed go-kit/log had already provided this functionaily. 

You're able to do `level.InfoValue().String()` similarly to previously, `log.Info.String()` in order to use from `main.go`. 

Admittedly I do like the look of  `log.Info.String()` more, but this removes what would be duplicate code.

Regarding the map I have created, it is being used as a way to convert what would be the env-var logLevel (type string) into a function. Adding a level filter requires a function be passed in [vendor/level.go](https://github.com/mainflux/mainflux/blob/b07c6ae26499758753de2b223fab747610dceef9/vendor/github.com/go-kit/kit/log/level/level.go#L75). So this is how I was mapping that user input string into the function the vendor package needs. Ex: `l = level.NewFilter(l, level.AllowError())`

I can change it to:
```
var logLevels = map[string]level.Option{
	level.DebugValue().String(): level.AllowDebug(),
	level.InfoValue().String():  level.AllowInfo(),
	level.WarnValue().String():  level.AllowWarn(),
	level.ErrorValue().String(): level.AllowError(),
}
```
If that would be prefered.

@anovakovic01 / @drasko  Please let me know your thoughts and please let me know if I misunderstood what was being suggested here!

logger/logger.go
mfinley3: Please take a look at my above comment. I have New accept a string here because that will be coming directly from env-vars like [here](https://github.com/mainflux/mainflux/blob/b07c6ae26499758753de2b223fab747610dceef9/cmd/http/main.go#L51-L55)

drasko: You can use function similar to [this one](https://github.com/uber-go/zap/blob/f4243df2af592a2e0ccdccb77e1ad44f78d735e5/zapcore/level.go#L116) to unmarshal text into a level struct. But what @anovakovic01 refered to here is that he wants `New()` to take `logLevel` enum as a param, not a string.

logger/logger.go
drasko: Idea is to create our levels, independent of go-kit logger, i.e. create an abstraction/wrapper there, so that we can replace go-kit logger in the future - for example with [zap](https://github.com/uber-go/zap).

mfinley3: Ah ok that makes sense, I will work towards that approach and remove the dependency on go-kits levels

anovakovic01: Thanks @mfinley3.

dborovcanin (changes requested):

cmd/things/main.go
```diff
@@ -101,33 +104,35 @@ func main() {
...
-		CacheDB:   db,
+		CacheDB:   mainflux.Env(envCacheDB, defCacheDB),
 		HTTPPort:  mainflux.Env(envHTTPPort, defHTTPPort),
 		GRPCPort:  mainflux.Env(envGRPCPort, defGRPCPort),
 		UsersURL:  mainflux.Env(envUsersURL, defUsersURL),
 	}
 }
 
-func connectToCache(cacheURL, cachePass string, cacheDB int) *redis.Client {
+func connectToCache(cacheURL, cachePass string, cacheDB string, logger log.Logger) *redis.Client {
```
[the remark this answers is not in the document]
mfinley3: Good catch @dusanb94, I've updated it to use the cacheDB parameter in the lines below!

logger/logger.go
mfinley3: @anovakovic01 I've made some changes and it now uses the level wrapper and steers clear of go-kit/level. Please let me know your thoughts on the changes!

logger/logger.go
mfinley3: I've started returning an error and no longer default it to error if there is bad input.

Please let me know your thoughts on how I chose to handle it for now.
https://github.com/mainflux/mainflux/blob/2ffb2bd3ba384c7e9451fdb30862d9cbb0d0511f/cmd/normalizer/main.go#L95-L100
To be honest I'm not the biggest fan of the above; it got a little funky since there is no logger created yet. 

So I think a few options are have each component create a logger and then update it with the appropriate logging level, default if it is bad input, or the above here. 
Thoughts?

mfinley3:
I just realized there are some conflicts, I will work on getting those fixed up in the morning.

logger/logger_test.go
```diff
@@ -98,3 +118,68 @@ func TestError(t *testing.T) {
...
 }
+
+func TestLogLevel(t *testing.T) {
+	debugCases := map[string]struct {
+		input  string
+		output logMsg
+	}{
+		"debug log ordinary string": {"input_string", logMsg{"", ""}},
+		"debug log empty string":    {"", logMsg{"", ""}},
+	}
```
anovakovic01: You should move these into `TestDebug`. Just create new logger instance in for loop and set log level from test case. Same for other table maps in this function.

anovakovic01 (changes requested):
Almost there. Good job @mfinley3

cmd/cassandra-reader/main.go
```diff
@@ -77,7 +81,15 @@ func main() {
 }
 
 func loadConfig() config {
+	var logLevel log.Level
+	err := logLevel.UnmarshalText(mainflux.Env(envLogLevel, defLogLevel))
+	if err != nil {
+		fmt.Printf(`{"level":"error","message":"%s","ts":"%s"}`, err, time.RFC3339Nano)
+		os.Exit(1)
```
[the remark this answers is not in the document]
mfinley3: Sounds good!

mfinley3:
I've rebased and addressed the conflicts. 
The conflicts were primarily around: [cmd/influxdb-writer/main.go](https://github.com/mainflux/mainflux/pull/402/files#diff-10dba4d266dfb116ff41d69e3950fab7)

Please let me know if there is anything else!

cmd/cassandra-reader/main.go
```diff
@@ -77,15 +82,22 @@ func main() {
 }
 
 func loadConfig() config {
+	var logLevel logger.Level
+	err := logLevel.UnmarshalText(mainflux.Env(envLogLevel, defLogLevel))
```
drasko: Instead of unmarshalling the text in each of the `main`s maybe it would be finally better to just pass the string and unmarshall stuff in `logger.New()` - that can do the validation (by trying to unmarshal text) and return error if it fails.

@anovakovic01 @dusanb94 - opinions?

@mfinley3 sorry about this comment, I know we told you to use lvels struct, but for me it was a bit hard to see without implementation - this patch is pretty big.

cmd/influxdb-writer/main.go
```diff
@@ -80,8 +85,13 @@ func main() {
 	}
 	defer client.Close()
 
-	timeout := time.Duration(cfg.BatchTimeout) * time.Second
-	repo, err := influxdb.New(client, cfg.DBName, cfg.BatchSize, timeout)
+	batchTimeout, batchSize, err := unmarshalInfluxDBSettings(cfg.BatchTimeout, cfg.BatchSize)
```
drasko: No need for this function `unmarshalInfluxDBSettings()` IMHO. I think it hurts readability.

I would rather go with:
```
        batchTimeout, err = strconv.Atoi(cfgBatchTimeout)
	if err != nil {
		// log error
                // os.Exit(1)
	}
 	batchSize, err = strconv.Atoi(cfgBatchSize)
	if err != nil {
		// log error
                // os.Exit(1)
	}
```
directly here.

@anovakovic01 @dusanb94 opinions?

drasko (commented):
Just these two minor things from my side, but this is now neatpicking...

http/README.md
```diff
@@ -8,11 +8,12 @@ The service is configured using the environment variables presented in the
...
-|----------------------|---------------------|-----------------------|
-| MF_HTTP_ADAPTER_PORT | Service HTTP port   | 8180                  |
-| MF_NATS_URL          | NATS instance URL   | nats://localhost:4222 |
-| MF_THINGS_URL        | Things service URL  | localhost:8181        |
+| Variable                    | Description                    | Default               |
+|-----------------------------|--------------------------------|-----------------------|
+| MF_HTTP_ADAPTER_LOG_LEVEL   | Log level for the HTTP Adapter | error                 |
+| MF_HTTP_ADAPTER_PORT        | Service HTTP port              | 8180                  |
+| MF_NATS_URL                 | NATS instance URL              | nats://localhost:4222 |
+| MF_THINGS_URL               | Things service URL             | localhost:8181        |
```
dborovcanin, line 16: I think that there is a code block explaining how to run a service just under these tables. Should look something like:
```
MF_THINGS_URL=[Things service URL] MF_NATS_URL=[NATS instance URL] MF_HTTP_ADAPTER_PORT=[Service HTTP port] $GOBIN/mainflux-http
```
Please, add this env variable there, as well. Please do this in the other README.md files, as well.

dborovcanin (commented):
I agree with @drasko about `unmarshalInfluxDBSettings()` and `logger.New()`. I think what he proposed is cleaner if nothing else. Also, I've found **not handling zero-values for timeout and batch size in InfluxDB writer really dangerous**. It's my bad, so I'll send a separate NOISSUE pull request for that. Great work @mfinley3!

http/README.md
mfinley3: ~Ah yep, I missed this, I will get that cleaned up in the next commit!~

Should be added to all of the readmes now!

cmd/influxdb-writer/main.go
dborovcanin: @drasko I agree. @mfinley3 However, if you want to keep this method, I suggest you rename to something like `loadInfluxDBConfig`. I also suggest not to return an error but to simply log and kill in case of wrong env variable values.

mfinley3: I agree as well, I think the function hurts readability. I will try somethings and get this looking better

@dusanb94 I will type up both solutions and see which one looks better. I think I am leaning towards simply removing the function right now though

cmd/cassandra-reader/main.go
mfinley3: I liked string initially because of the cleanliness but I do see why passing in an enum for levels makes sense. So, I'll let @anovakovic01 chime in before implementing anything. I am fine with  going either way on this.

anovakovic01: @drasko purpose of enumeration is following: if you want to use logger with without environment variable, and set log level explicitly, it's better to use enum constant, than string (because compiler will give you error). In current usage this is not the case, so if you want, you can pass string for now.

anovakovic01 (changes requested):

cmd/influxdb-writer/main.go
mfinley3: I decided that it flowed better without calling a separate function, so went with keeping these in main().

cmd/cassandra-reader/main.go
mfinley3: Cool, converted it back to accepting a string!

logger/logger_test.go
```diff
@@ -39,6 +39,43 @@ type logMsg struct {
...
+	debugNotAllowedCases := map[string]struct {
+		input  string
+		output logMsg
+	}{
+		"debug log ordinary string": {"input_string", logMsg{"", ""}},
+		"debug log empty string":    {"", logMsg{"", ""}},
+	}
+
+	writer := mockWriter{}
+	logger := log.New(&writer, log.Debug)
```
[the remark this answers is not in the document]
mfinley3: Should now all be in one for loop!

logger/level_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}{
+		"select log level debug": {"debug", Debug},
+		"select log level DEBUG": {"DEBUG", Debug},
+		"select log level info":  {"info", Info},
+		"select log level INFO":  {"INFO", Info},
+		"select log level warn":  {"warn", Warn},
+		"select log level WARN":  {"WARN", Warn},
+		"select log level Error": {"Error", Error},
+		"select log level ERROR": {"ERROR", Error},
+	}
```
anovakovic01, line 26: Merge these test cases into one map and use one `for` loop.

anovakovic01 (changes requested):

logger/level_test.go
mfinley3: All cleaned up - and thanks again for the tip on adding err to the cases!

anovakovic01 (dismissed):
@mfinley3 Thanks for your contribution! LGTM!

drasko (dismissed):
We are there - LGTM.

Some great work here @mfinley3!

ws/README.md
```diff
@@ -8,11 +8,12 @@ The service is configured using the environment variables presented in the
...
-|----------------------|---------------------|-----------------------|
-| MF_WS_ADAPTER_PORT   | Service WS port     | 8180                  |
-| MF_NATS_URL          | NATS instance URL   | nats://localhost:4222 |
-| MF_THINGS_URL        | Things service URL  | localhost:8181        |
+| Variable                  | Description                   | Default               |
+|---------------------------|-------------------------------|-----------------------|
+| MF_WS_ADAPTER_LOG_LEVEL   | Log level for the WS Adapter  | error                 |
+| MF_WS_ADAPTER_PORT        | Service WS port               | 8180                  |
+| MF_NATS_URL               | NATS instance URL             | nats://localhost:4222 |
+| MF_THINGS_URL             | Things service URL            | localhost:8181        |
```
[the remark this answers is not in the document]
mfinley3: Ah shoot, you're right. I did miss it, sorry about that!

mfinley3:
@drasko @dusanb94 @anovakovic01 
I completely forgot the mqtt-adapter. I have it working with log levels as well though. 
Thankfully the change is only a few lines since it's js and the aedeas-logging package supports it already (with an update to the latest version). 
So hopefully not too much to review!
Please take look at the updated mqtt files: https://github.com/mainflux/mainflux/pull/402/commits/89a9b80ae00fc0669dec01619376fa218cedc3c7
Sorry about this guys.

dborovcanin (approved):

anovakovic01 (approved):
LGTM!

drasko:
Congrats @mfinley3 - you are now officially Mainflux contributor :)

mfinley3:
@drasko And first thing I do is break the tests haha!
Looks like the fix for that is in now though.
Happy to be working with you guys a bit closer though, our team will hopefully be giving back more now. I've documented some helpful things I did during this PR for our team while working on Mainflux PRs as well.

PR #406, opened 2018-09-26
NOISSUE - Fixing level_test.go

mfinley3:
@anovakovic01 found this.

Map order is not guaranteed and was causing issues with the test due to the pointers levels uses. So sometimes actual could be 2 or 4 for example while expected was always one.

We can fix this by moving the level def inside the for loop

Signed-off-by: Michael Finley <Michael.Finley@target.com>


anovakovic01 (approved):
LGTM!

drasko (approved):
LGTM

PR #408, opened 2018-09-29
Update mqtt to mqtt communication to support full set of protocol features

anovakovic01:
`MQTT2MQTT` communication now doesn't go over NATS, but over MQTT broker directly. `MQTT2ANYTHING_ELSE` still goes over NATS.

### DIFF
--- a/mqtt/mqtt.js
+++ b/mqtt/mqtt.js
@@ -63,6 +63,8 @@ function startMqtt() {
 
 nats.subscribe('channel.*', function (msg) {
     var m = message.RawMessage.decode(Buffer.from(msg)),
+        packet;
+    if (m && m.Protocol !== 'mqtt') {
         packet = {
             cmd: 'publish',
             qos: 2,
@@ -71,7 +73,8 @@ nats.subscribe('channel.*', function (msg) {
             retain: false
         };
 
-    aedes.publish(packet);
+        aedes.publish(packet);
+    }
 });
 
 aedes.authorizePublish = function (client, packet, publish) {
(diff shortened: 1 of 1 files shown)


mqtt/mqtt.js
```diff
@@ -63,6 +63,8 @@ function startMqtt() {
 
 nats.subscribe('channel.*', function (msg) {
     var m = message.RawMessage.decode(Buffer.from(msg)),
+        packet;
+    if (m && m.Protocol !== 'mqtt') {
```
drasko, line 67: This is good, however in the early days of Mainflux we had something like `<protocol>.<channel_nb>` for the NATS topic. This way adapter could subscribe only to wanted protocols. That would avoid processing of each message published over MQTT and then dropping it - with 200k mps that would be a lot of processing.

@anovakovic01 what is your opinion on this?

drasko (dismissed):
LGTM

mqtt/mqtt.js
anovakovic01: In that case you will have to know every other protocol that is used in the system (because you cannot say: subscribe to `channel.*` except `channel.mqtt`, as far as I know).

drasko: Yes, this is true. But Mainflux supports only 4 protocols so it would not be difficult to subscribe to all 3 of them other than MQTT - if this can seriously improve performance.

Never the less. let's merge this patch as it is today, and we'll see later for the improvements.

drasko:
@anovakovic01 please update the branch.

drasko (approved):
LGTM

dborovcanin (approved):

PR #420, opened 2018-10-18
NOISSUE - Refactor SDK and things service

anovakovic01:
### What does this do?
This PR improves SDK usability and separates HTTP requests and responses from model in `things` service.

CLI was updated to use new SDK version.

Not for SDK, that will be done in separate PR. For things service existing tests were updated. 

Documentation for SDK was updated.


sdk/go/channels.go
```diff
@@ -62,82 +69,123 @@ func (sdk *MfxSDK) Channels(token string) ([]things.Channel, error) {
...
+		}
 	}
 
 	return nil
 }
 
-// DeleteChannel - removes channel
-func (sdk *MfxSDK) DeleteChannel(id, token string) error {
-	url := fmt.Sprintf("%s/%s/%s", sdk.url, channelsEndpoint, id)
+func (sdk mfSDK) DeleteChannel(id, token string) error {
```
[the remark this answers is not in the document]
anovakovic01: IMHO it's clear from method name `DeleteChannel`.

sdk/go/sdk.go
```diff
@@ -8,69 +8,194 @@
...
+	Things(token string, offset, limit uint64) ([]Thing, error)
+
+	// Thing returns thing object by id.
+	Thing(id, token string) (Thing, error)
+
+	// UpdateThing updates existing thing.
+	UpdateThing(thing Thing, token string) error
+
+	// DeleteThing removes existing thing.
 	DeleteThing(id, token string) error
```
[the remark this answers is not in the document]
anovakovic01: I think it's clear from method name.

sdk/go/things.go
```diff
@@ -64,125 +73,175 @@ func (sdk *MfxSDK) Things(token string) ([]things.Thing, error) {
...
+		}
 	}
 
 	return nil
 }
 
-// DeleteThing - removes thing
-func (sdk *MfxSDK) DeleteThing(id, token string) error {
-	url := fmt.Sprintf("%s/%s/%s", sdk.url, thingsEndpoint, id)
+func (sdk mfSDK) DeleteThing(id, token string) error {
```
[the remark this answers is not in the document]
anovakovic01: It's clear from method name.

sdk/go/channels.go
```diff
@@ -62,82 +69,123 @@ func (sdk *MfxSDK) Channels(token string) ([]things.Channel, error) {
...
 	}
+
 	return c, nil
 }
 
-// UpdateChannel - update a channel
-func (sdk *MfxSDK) UpdateChannel(id, data, token string) error {
-	url := fmt.Sprintf("%s/%s/%s", sdk.url, channelsEndpoint, id)
-	req, err := http.NewRequest(http.MethodPut, url, strings.NewReader(data))
+func (sdk mfSDK) UpdateChannel(channel Channel, token string) error {
```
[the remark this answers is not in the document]
anovakovic01: `chan` is keyword in go, and `c` is ok for method body, but not sure if we'll have any benefit from shorter name (line has normal length).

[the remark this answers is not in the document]
manuio: As you want, is just to do it as int he rest of the code

cli/channels.go
```diff
@@ -38,36 +49,48 @@ var cmdChannels = []cobra.Command{
 				logUsage(cmd.Short)
 				return
 			}
+
 			if args[0] == "all" {
-				l, err := sdk.Channels(args[1])
+				l, err := sdk.Channels(args[1], uint64(Offset), uint64(Limit))
```
drasko, line 54: Would it make the sense set these `Offset` and `Limit` as optional and pass them as variadic parameters or via some other means? I would like to keep SDK interface as simple as possible, not to have to explain to beginners what Offset and Limit are.

cli/sdk.go
```diff
@@ -7,13 +7,12 @@
...
-)
+import mfxsdk "github.com/mainflux/mainflux/sdk/go"
 
 // Keep SDK handle in global var
-var sdk *mfxsdk.MfxSDK
+var sdk mfxsdk.SDK
 
-func SetSDK(s *mfxsdk.MfxSDK) {
+// SetSDK sets mainflux SDK instance.
+func SetSDK(s mfxsdk.SDK) {
```
drasko, line 16: Why not pass pointer?

cli/things.go
```diff
@@ -40,21 +51,24 @@ var cmdThings = []cobra.Command{
...
 			}
+
 			t, err := sdk.Thing(args[0], args[1])
 			if err != nil {
 				logError(err)
 				return
 			}
-			dump(t)
+
+			flush(t)
```
drasko, line 71: `flush` != `dump` - flush flushes the stdio buffer while `dump` just dumps/prints the stuff on screen. You can change `dump`, but make it something else than `flush`.

drasko (changes requested):

cli/channels.go
anovakovic01: No, these are required parameters (they are optional only in HTTP request), any user should be aware of these. Also, we should avoid package level variables (because this is easier to debug and also safer for concurrent use).

cli/sdk.go
anovakovic01: Because there is no need for pointer.

drasko: Can you please elaborate?

Passing a pointer is passing 4-8 bytes. Passing a whole data-structure is much more.

cli/things.go
anovakovic01: I know it's different, but `dump` sounds... how should I say this... wrong... Also, this CLI should be refactored anyway, so maybe this naming can be fixed in that PR?

anovakovic01 (commented):
@drasko This is premature optimization which is anti-pattern. I avoid using pointers if there's no need for them. They are not thread safe.

dborovcanin (approved):

dborovcanin (approved):

drasko (approved):
LGTM

PR #422, opened 2018-10-19
NOISSUE - Add info default value to log levels

darkodraskovic:

### What does this do?

Adds default "info" log level environment variables.

No.

No need for docs.

### DIFF
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -30,32 +30,32 @@ gnatsd &
 ###
 # Users
 ###
-$BUILD_DIR/mainflux-users &
+MF_USERS_LOG_LEVEL=info $BUILD_DIR/mainflux-users &
 
 ###
 # Things
 ###
-MF_THINGS_HTTP_PORT=8182 MF_THINGS_GRPC_PORT=8183 $BUILD_DIR/mainflux-things &
+MF_THINGS_LOG_LEVEL=info MF_THINGS_HTTP_PORT=8182 MF_THINGS_GRPC_PORT=8183 $BUILD_DIR/mainflux-things &
 
 ###
 # HTTP
 ###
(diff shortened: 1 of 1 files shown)


chombium (changes requested):
Could you please elaborate why you've set info as a default log level? Usually error is set as default log level. I guess you want to have something for dev purposes.

If we want to have the log level to be the same for all of the services and configurable, it would be better if we have another envvar. If this envvar is not set from outside the run script will set a default value.

Than you can simply overwrite the default log level with something like this:

```MF_LOG_LEVEL=info ./run.sh```

Another question is if we want to have granular setting of the log level per service, so that the users can set different log levels per service.

drasko:
@darkodraskovic I agree with @chombium's remarks - add `MF_LOG_LEVEL=info ./run.sh` in `Makefile`.

@chombium this `make run` is running microservices natively and is for now for development purposes, this is why @darkodraskovic proposed to add some more logs useful during development stage.

I would propose that we change also `run` target into `rundev` (for development native run with logs), and use `make run` to run docker containers.

chombium:
@drasko I understand that that's for development purposes, but even then, because of the missing flexibility to set the log level from outside, it can happen that someone changes the log levels to and unintentionally commits the changed shell script.

We could also add a new rundev.sh file to the git ignores and everyone can configure the log levels as he/she wants. That would also do the trick

chombium:
@drasko I understand that that's for development purposes, but even then, because of the missing flexibility to set the log level from outside, it can happen that someone changes the log levels to and unintentionally commits the changed shell script.

We could also add a new rundev.sh file to the git ignores and everyone can configure the log levels as he/she wants. That would also do the trick

chombium:
@drasko I understand that that's for development purposes, but even then, because of the missing flexibility to set the log level from outside, it can happen that someone changes the log levels to and unintentionally commits the changed shell script.

drasko:
Closing this one, will be implemented through https://github.com/mainflux/mainflux/pull/427

Issue #426, opened 2018-10-23
Allow custom certificate authorities to be optionally specified in any grpc.Dial


nwest1:
**FEATURE REQUEST**

1. Is there an open issue addressing this request? If it does, please add a "+1" reaction to the
   existing issue, otherwise proceed to step 2.

2. Describe the feature you are requesting, as well as the possible use case(s) for it.

We're running into errors when a grpc client calls into an ssl-enabled grpc things or users endpoint. I was thinking about adding something like an optional `CA_CERTS` environmental variable. When set, will try to read the file path provided, and provide `grpc.credentials.createSsl(root_certs)` to the `grpc.Dial` function. 

A quick search reveals a few places where the insecure connection exists:
- [ ] `cmd/cassadra-reader/main.go`
- [ ] `cmd/http/main.go`
- [ ] `cmd/influxdb-reader/main.go`
- [ ] `cmd/mongodb-reader/main.go`
- [ ] `cmd/things/main.go`
- [ ] `cmd/ws/main.go`
- [ ] `mqtt/mqtt.js`

Open questions:
- Does this seem like a reasonable approach?
- Should I include a pem of CAs by default? If so, where should that come from?

3. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).
For us it's a must-have.

We've tested out these changes successfully, but want to get feedback and make it properly configurable.

drasko:
@nwest1 in current implementation we protect the whole Mainflux system only by NginX edge proxy and we consider that everything behind this proxy is on the private and secure network. I.e. no encryption is needed between Mainflux components themselves.

This is why your approach is totally untested by us. In theory it should work.

@anovakovic01 can we provide some example for `grpc.credentials.createSsl(root_certs)` cert set-up - did we do something similar for SDK?

anovakovic01:
@nwest1 We use these `gRPC` endpoints only for internal communication. That's why we haven't exposed it to the outside world and haven't protected it. If you want to use certs, we recommend doing that on nginx (or some other proxy) level.

drasko:
I agree with this recommendation from @anovakovic01 - let's keep Mainflux code simple and if needed for specific cases where it is deployed over insecure network - then reverse proxy with TLS termination could protect a specific microservice.

@nwest1 would this work for you?

chombium:
[...]

PR #430, opened 2018-10-28
MF-426 - Add optional MF_CA_CERTS env variable to allow GRPC client to use TLS certs

nwest1:
### What does this do?
Allows a user to specify a path to their own trusted certs file via environmental variable `MF_CA_CERTS`. 
related #426 
None
Not yet
Not yet
### Notes
Wanted to get the idea out for feedback before moving onto other components


drasko:
@nwest1 this PR looks good to me.

You will need to sign-off the commit in order for DCO check to pass

Just a question for the reference - is there some special way to deploy `things` service in order for it to provide TLS connection (private/public cert)?

nwest1:
@drasko The SSL termination is currently handled by the ingress and not directly by the service itself. I can add that capability as discussed. Would you prefer all in one PR or leave this one for clients?

drasko:
@nwest1 let's leave this one, making the PRs smaller.

Also, would WS and MQTT adapter be needing similar functionality? I guess yes, but if you are not using them or not capable to test them, then someone else from Mainflux core team can do these.

nwest1:
Yes, MQTT will need to be changed in a similar fashion - I'll work on that soon. Though it looks like similar changes would need to happen in the WS adapter, we're not using it and aren't setup to test it.

cmd/things/main.go
```diff
@@ -53,6 +56,7 @@ const (
 	envDBUser    = "MF_THINGS_DB_USER"
 	envDBPass    = "MF_THINGS_DB_PASS"
 	envDBName    = "MF_THINGS_DB"
+	envCACerts   = "MF_THINGS_CA_CERTS"
```
dborovcanin: Can you, please, update the docs and docker-compose as well (README.md)?

cmd/http/main.go
```diff
@@ -61,7 +66,21 @@ func main() {
...
 	defer nc.Close()
 
-	conn, err := grpc.Dial(cfg.ThingsURL, grpc.WithInsecure())
+	secureOption := grpc.WithInsecure()
+	if cfg.CACerts != "" {
+		thingsURL, err  := url.Parse(cfg.ThingsURL)
+		if err != nil {
+			logger.Error(fmt.Sprintf("Failed to parse ThingsURL: %s", err))
+		}
+		tpc, err := credentials.NewClientTLSFromFile(cfg.CACerts, thingsURL.Host)
```
dborovcanin: If `url.Parse(cfg.ThingsURL)` fails, `thingsURL` is possible an empty value, or nil, in case of the pointer, so calling `thingsURL.Host` is dangerous.

cmd/http/main.go
```diff
@@ -61,7 +66,21 @@ func main() {
 	}
 	defer nc.Close()
 
-	conn, err := grpc.Dial(cfg.ThingsURL, grpc.WithInsecure())
+	secureOption := grpc.WithInsecure()
+	if cfg.CACerts != "" {
+		thingsURL, err  := url.Parse(cfg.ThingsURL)
+		if err != nil {
+			logger.Error(fmt.Sprintf("Failed to parse ThingsURL: %s", err))
```
dborovcanin: Does it make sense to proceed in case of failure (maybe `os.Exit(1)` here)?

dborovcanin (changes requested):

cmd/http/main.go
```diff
@@ -61,7 +66,21 @@ func main() {
...
+		}
+		tpc, err := credentials.NewClientTLSFromFile(cfg.CACerts, thingsURL.Host)
+		if err != nil {
+			logger.Error(fmt.Sprintf("Failed to load certs: %s", err))
+			os.Exit(1)
+		}
+		secureOption = grpc.WithTransportCredentials(tpc)
+	}
```
chombium: @nwest1 I think, It would be good if in case that no CA certs are set, to log a warning so that the user knows that the gRPC communication is not encrypted.

chombium (changes requested):
An update of the documentation with explanation about how this feature works and how to configure it is also needed

cmd/things/main.go
nwest1: I think I've covered where this makes sense - should I leave the docker-compose with the default (blank) or do you want to include a sample CA along with your existing self-signed certs?

cmd/things/main.go
```diff
@@ -148,8 +154,21 @@ func connectToDB(cfg config, logger logger.Logger) *sql.DB {
 	return db
 }
 
-func connectToUsersService(usersAddr string, logger logger.Logger) *grpc.ClientConn {
-	conn, err := grpc.Dial(usersAddr, grpc.WithInsecure())
+func connectToUsersService(cfg config, logger logger.Logger) *grpc.ClientConn {
```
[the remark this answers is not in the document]
nwest1: I've removed the url parsing as well, as the `serverNameOverride` is actually for testing: https://godoc.org/google.golang.org/grpc/credentials#NewClientTLSFromCert

cmd/http/main.go
nwest1: Updated!

cmd/http/main.go
```diff
@@ -61,7 +65,19 @@ func main() {
 	}
 	defer nc.Close()
 
-	conn, err := grpc.Dial(cfg.ThingsURL, grpc.WithInsecure())
+	secureOption := grpc.WithInsecure()
```
manuio: @nwest1 I would put this in a separate function as you did here: https://github.com/mainflux/mainflux/blob/897d4373b15b5fbd7b346086a3552929c3fedef4/cmd/things/main.go#L156

manuio (changes requested):

cmd/http/main.go
chombium: @nwest1 cool, but I personally think logger.Warn might be more appropriate.

@drasko, @dusanb94 what do you think about this?

nwest1: All the other `logger.Warn`s are currently related to some error happening, but the app will continue to run. Though I think security is important, it's not in error, which is why I chose info. 

Happy to change this to whatever makes sense for the project.

cmd/http/main.go
nwest1: Done!

cmd/http/main.go
anovakovic01: @nwest1 I agree with you, we should use warn only for errors. This is valid flow, so I think that we should leave info.

chombium: @nwest1 thanks for checking on other warning usages. I totally agree with you. Let's leave it as an info message.

cmd/things/main.go
chombium: @nwest1 you can reuse the certs and ca from [here](https://github.com/mainflux/mainflux/tree/master/docker/ssl). We use them for nginx for our dev environment .

Regarding the docs, you should add a subsection in the [getting started guide](https://github.com/mainflux/mainflux/blob/master/docs/getting-started.md) with a short description how to setup the encryption and what happens if the certificate configuration is missing ;)

anovakovic01:
@nwest1 @drasko I think that we should add cert support for service providers too (`users` and `things`) in this PR.

cmd/things/main.go
nwest1: I updated the getting started guide with some simple explanation. Unless you'd like me to configure nginx for grpc load balancing, I'm not sure it makes sense to secure this connection.

Another option would be to add a separate ENV var to control the `grpc.WithInsecure()` rather than the current logic. That pattern might scale out better than controlling two options with one parameter.

chombium: @nwest1 it looks good, we can change that in future if there is a need it and if we notice some problems.

I would ask you one more thing and then we are done regarding the docs :)
Please add a note in the getting started guide that if the env vars for setting the ca are omitted, the gRPC communication will not be encrypted.

nwest1:
@anovakovic01 @drasko I'm happy to add cert support in this PR.

nwest1:
I'm testing out the docker-compose portion, but am running into certificate validation errors. They're being eaten at the service level (https://github.com/mainflux/mainflux/blob/master/things/service.go#L149) 

It's looking for a `users` and/or `things` host cert to be presented - where in the current configuration it's `localhost`.

Would you like me to configure this differently or would it be possible to generate two more cert+key combos for the two hostnames expected? The CA key is not available in the repo.

cmd/things/main.go
nwest1: Added :) Let me know if there are any other details!

http/README.md
```diff
@@ -33,6 +34,7 @@ services:
       MF_NATS_URL: [NATS instance URL]
       MF_HTTP_ADAPTER_LOG_LEVEL: [HTTP Adapter Log Level]
       MF_HTTP_ADAPTER_PORT: [Service HTTP port]
+      MF_CA_CERTS: [Path to trusted CAs in PEM format]
```
nwest1: Should this be `MF_HTTP_ADAPTER_CA_CERTS`? Or should `MF_THINGS_CA_CERTS` be `MF_CA_CERTS`?

drasko: I think it should be `MF_THINGS_CA_CERTS`, as CA issues private/public cert pair for server - which in this case is `things`.

nwest1:
SSL changes: using https://github.com/mainflux/proxy/blob/master/certs/generate-CA.sh I recreated the ca and server cert + key combo with `localhost` `users` and `things` as Subject Alternative Names.  

I'm not sure why github thinks there are conflicts in docker-compose as they are only additions, but maybe someone else can resolve.

drasko:
@nwest1 thanks a lot, looks good - we'll take a look over the WE.

manuio (dismissed):

chombium:
Hi @nwest1,

thanks for your engagement and the fast responses to our comments. 
Please merge the latest changes from the master and resolve the conflicts.

Two questions out of curiosity:
1. Why does the `nginx` containter depends on `things` the `users` containers?
2. Why are the following certificate files changed:
*  docker/ssl/certs/ca.crt 
*  docker/ssl/certs/mainflux-server.crt
* docker/ssl/certs/mainflux-server.key

things/README.md
```diff
@@ -25,6 +25,7 @@ default values.
 | MF_THINGS_DB_USER     | Database user                                   | mainflux       |
 | MF_THINGS_DB_PASS     | Database password                               | mainflux       |
 | MF_THINGS_DB          | Name of the database used by the service        | things         |
+| MF_THINGS_CA_CERTS    | Path to trusted CAs in PEM format               |                |
```
anovakovic01, line 28: Please add `MF_THINGS_SERVER_CERT` and `MF_THINGS_SERVER_KEY` to docs.

cmd/users/main.go
```diff
@@ -32,36 +34,42 @@ import (
...
+	envDBHost     = "MF_USERS_DB_HOST"
+	envDBPort     = "MF_USERS_DB_PORT"
+	envDBUser     = "MF_USERS_DB_USER"
+	envDBPass     = "MF_USERS_DB_PASS"
+	envDBName     = "MF_USERS_DB"
+	envHTTPPort   = "MF_USERS_HTTP_PORT"
+	envGRPCPort   = "MF_USERS_GRPC_PORT"
+	envSecret     = "MF_USERS_SECRET"
+	envServerCert = "MF_USERS_SERVER_CERT"
+	envServerKey  = "MF_USERS_SERVER_KEY"
```
anovakovic01, line 58: Add `MF_USERS_SERVER_CERT` and `MF_USERS_SERVER_KEY` to `users/README.md`.

anovakovic01 (changes requested):

dborovcanin (commented):
@nwest1 Great job! Please check if docs are up to date (env variables as well as env var list when starting services) and update the branch, so that we can merge.

dborovcanin (dismissed):

nwest1:
@chombium 
1. I was running into a few errors when configuring, but this is not needed and I've removed.
2. I regenerated the SSL certificates, adding `users` `things` and `localhost` as subject alternative names, so the same certificate can be used by all servers + nginx.

Shall I include the CA key as well? I used https://github.com/mainflux/proxy/blob/master/certs/generate-CA.sh to generate these.

nwest1:
Happy to contribute! This branch should be all fixed up at this point. 

Thanks for all the reviews and comments - I appreciate the quick feedback.

drasko:
@nwest1 please update the branch, we just merged some other stuff so your PR is out of sync now :)

dborovcanin (dismissed):

users/README.md
```diff
@@ -26,6 +26,8 @@ default values.
 | MF_USERS_DB          | Name of the database used by the service        | users        |
 | MF_USERS_HTTP_PORT   | Users service HTTP port                         | 8180         |
 | MF_USERS_GRPC_PORT   | Users service gRPC port                         | 8181         |
+| MF_USERS_SERVER_CERT | Path to server certificate in pem format        | 8181         |
```
drasko: Why is port relevant here?

http/README.md
```diff
@@ -14,6 +14,7 @@ default values.
 | MF_HTTP_ADAPTER_PORT        | Service HTTP port              | 8180                  |
 | MF_NATS_URL                 | NATS instance URL              | nats://localhost:4222 |
 | MF_THINGS_URL               | Things service URL             | localhost:8181        |
+| MF_HTTP_ADAPTER_CA_CERTS    | Path to trusted CAs in PEM format |                    |
```
drasko: Please align tabs/spaces for readability.

drasko (changes requested):
@nwest1 almost there - few minor tweaks left.

drasko (approved):
LGTM

anovakovic01 (approved):
LGTM!

dborovcanin (approved):

drasko:
@chombium if OK with you we would like to proceed merging this one.

chombium (approved):

chombium:
LGTM! Great work @nwest1

drasko:
Merged! Thanks a lot @nwest1 - great contribution!

PR #434, opened 2018-10-31
MF-407 - Values of zero are being omitted 

### LINKED ISSUE #407: Values of zero are being omitted
**(Semi) BUG REPORT**

Noticed this when the value field wasn't being written out onto a kafka topic. 
This may be more of an issue with what protoc generates but I believe Mainflux should account for this. So I wanted to just make sure you all were aware it. 

Sending a message where the value is zero.

That the `value` field does not get omitted and contains a value of zero when the message that is sent contains 0.

The value field gets omitted and is non-existent rather than zero

The way I found easiest to visualize this, was through subscribing to the NATS channel where the normalized messages are written to (out.senml for example) and then decoding the messages.
- Sub to out.senml and capture the output.
- Send a message containing a zero for the value for example: 
```
[  
   {  
      "bn":"MVP-mikes-fridge-sensor:",
      "bt":1537822790,
      "bu":"degC",
      "bver":5,
      "n":"Temperature",
      "u":"degC",
      "v":0
   }
]
```
- Decode that output using protoc and although a value was sent, it does not appear 

What I am seeing when I do this:
example

Kubernetes 

I believe that a Zero value should be valid and not get omitted anywhere in Mainflux.  Since things like refrigerators/freezers can send temperatures that are 0. 

The issue probably truly lies in protoc/go since protoc generates message.pg.go:
```
type Message struct {
        ...
	Value                float64  `protobuf:"fixed64,6,opt,name=Value,proto3" json:"Value,omitempty"`
        ...
}
```
Due to a combination of omitempty getting added and zero being considered empty in go, the `Value` field does not get populated. 

Let me know thoughts about this!

dborovcanin:
### What does this do?
This pull request updates Mainflux Message format.
 
Fix #407.

Tests have been updated alongside with services that use Mainflux normalized Message.

Docs are updated due to migration to [gogoprotobuf](https://github.com/gogo/protobuf) Protobuf compiler .


anovakovic01:
@dusanb94 Can you please try to return code coverage to 91%?

normalizer/normalizer.go
```diff
@@ -32,28 +32,34 @@ func (n normalizer) Normalize(msg mainflux.RawMessage) (NormalizedData, error) {
...
+			m.Value = &mainflux.Message_BoolValue{*v.BoolValue}
+		}
+
+		if v.DataValue != "" {
+			m.Value = &mainflux.Message_DataValue{v.DataValue}
+		}
+
+		if v.StringValue != "" {
+			m.Value = &mainflux.Message_StringValue{v.StringValue}
 		}
```
anovakovic01, line 54: You can use switch case for all of these (you can have only one of these per message, right?).

readers/cassandra/messages.go
```diff
@@ -40,12 +40,37 @@ func (cr cassandraRepository) ReadAll(chanID, offset, limit uint64) []mainflux.M
...
+			msg.Value = &mainflux.Message_StringValue{*strVal}
+		}
+
+		if boolVal != nil {
+			msg.Value = &mainflux.Message_BoolValue{*boolVal}
+		}
+
+		if dataVal != nil {
+			msg.Value = &mainflux.Message_DataValue{*dataVal}
+		}
```
anovakovic01, line 63: If these are exclusive, use switch case here too.

readers/cassandra/messages_test.go
```diff
@@ -68,9 +89,9 @@ func TestReadAll(t *testing.T) {
 		},
 		"read message last page": {
 			chanID:   chanID,
-			offset:   40,
-			limit:    10,
-			messages: messages[40:42],
+			offset:   0,
+			limit:    15,
+			messages: messages,
 		},
```
anovakovic01, line 95: You've removed test cases that are testing offset that is greater than 0. Please add them.

readers/influxdb/messages.go
```diff
@@ -90,11 +117,8 @@ func genMessage(names []string, fields []interface{}) mainflux.Message {
 			u, _ := strconv.ParseUint(fields[i].(string), 10, 64)
 			msgField.SetUint(u)
 		case float64:
-			msgField.SetFloat(parseFloat(fields[i]))
-		case bool:
-			if b, ok := fields[i].(bool); ok {
-				msgField.SetBool(b)
-			}
+			val, _ := strconv.ParseFloat(fields[i].(string), 64)
```
anovakovic01, line 120: Please handle this error.

readers/mongodb/messages.go
```diff
@@ -40,11 +57,42 @@ func (repo mongoRepository) ReadAll(chanID, offset, limit uint64) []mainflux.Mes
...
+			msg.Value = &mainflux.Message_StringValue{*m.StringValue}
+		}
+
+		if m.DataValue != nil {
+			msg.Value = &mainflux.Message_DataValue{*m.DataValue}
+		}
+
+		if m.BoolValue != nil {
+			msg.Value = &mainflux.Message_BoolValue{*m.BoolValue}
+		}
```
anovakovic01, line 85: If these are exclusive, use switch case here too.

writers/cassandra/messages_test.go
```diff
@@ -26,14 +26,34 @@ var (
...
+		case 3:
+			msg.Value = &mainflux.Message_DataValue{"base64data"}
+		case 4:
+			msg.ValueSum = nil
+		case 5:
+			msg.ValueSum = &mainflux.SumValue{Value: 45}
+		}
+
+		err = repo.Save(msg)
+		assert.Nil(t, err, fmt.Sprintf("expected no error, go %s", err))
```
anovakovic01: `got`

anovakovic01 (changes requested):

readers/influxdb/messages.go
dborovcanin: Based on the private discussion we've had, I'll mark this as resolved.
> This obscure way of handling InfluxDB read is caused by strange API of the InfluxDB official Go client and should be updated as soon as the official client gets support for mapping data to custom struct. Also, this query could be optimized, but it can be a part of the aforementioned PR sometime in the future.

anovakovic01 (approved):
LGTM!

drasko (approved):
LGTM

drasko:
Merged! Thanks @dusanb94!

PR #441, opened 2018-11-01
Expose the InfluxDB and Cassandra ports to host

### LINKED ISSUE #437: Expose the InfluxDB HTTP API from the mainflux-influxdb container to localhost
As we all know from the gitter discussions, there are many people who try to access the processed messages saved in the InfluxDB from the outside world. They want to have an easy way to "play" with the data stored in the InfluxDB of the mainflux-influxdb container from localhost. It would be nice if we expose the InfluxDB HTTP to be reachable from localhost.

Therefore we should add port mapping for the `mainflux-influxdb` container in the to the InfluxDB writer docker compose file somewhere after [line 25](https://github.com/mainflux/mainflux/blob/master/docker/addons/influxdb-writer/docker-compose.yml#L25) we should add:
```
ports:
- 8086:8086
```

chombium:
Expose the InfluxDB's HTTP API port 8086 and  Cassandra's CQL port 9042 from the containers
to the host machine, in order to enable easier exploration of the data stored in the databases
from the host machine and development of apps which connect and use the data stored in the databases.

Fixes #437

### DIFF
--- a/docker/addons/cassandra-writer/docker-compose.yml
+++ b/docker/addons/cassandra-writer/docker-compose.yml
@@ -27,8 +27,6 @@ services:
     container_name: mainflux-cassandra-writer
     depends_on:
       - cassandra
-    expose:
-      - 8902
     restart: on-failure
     environment:
       MF_NATS_URL: nats://nats:4222
--- a/docker/addons/cassandra-writer/docker-compose.yml
+++ b/docker/addons/cassandra-writer/docker-compose.yml
@@ -19,6 +19,8 @@ services:
     restart: on-failure
     networks:
       - docker_mainflux-base-net
+    ports:
+      - 9042:9042
 
   cassandra-writer:
     image: mainflux/cassandra-writer:latest
--- a/docker/addons/influxdb-writer/docker-compose.yml
+++ b/docker/addons/influxdb-writer/docker-compose.yml
@@ -23,6 +23,8 @@ services:
       INFLUXDB_ADMIN_PASSWORD: mainflux
     networks:
       - docker_mainflux-base-net
+    ports:
+      - 8086:8086
 
   influxdb-writer:
     image: mainflux/influxdb-writer:latest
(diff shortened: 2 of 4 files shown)


drasko (dismissed):
LGTM

drasko:
@chombium thanks for the PR, just one question - do we need similar for MongoDB?

chombium:
@drasko the ports for mongdb are already exposed [here](https://github.com/mainflux/mainflux/blob/master/docker/addons/mongodb-writer/docker-compose.yml#L25).

nmarcetic (dismissed):

docker/addons/cassandra-writer/docker-compose.yml
```diff
@@ -19,6 +19,10 @@ services:
     restart: on-failure
     networks:
       - docker_mainflux-base-net
+    expose:
```
anovakovic01: We don't need `expose` in combination with `ports`.

docker/addons/influxdb-writer/docker-compose.yml
```diff
@@ -23,6 +23,10 @@ services:
       INFLUXDB_ADMIN_PASSWORD: mainflux
     networks:
       - docker_mainflux-base-net
+    expose:
```
anovakovic01: Same here.

anovakovic01 (changes requested):

anovakovic01:
@chombium Can you remove `expose` where there are `ports` for other services too?

docker/addons/influxdb-writer/docker-compose.yml
chombium: I found that a bit weird too in the [mongodb config](https://github.com/mainflux/mainflux/blob/master/docker/addons/mongodb-writer/docker-compose.yml#L20) , but as it was already there, I added the ports to the other configs the same way for consistency. Reading the docker-compose docs again I found out that ports does the same as expose plus it exposes the port to the host. I will fix the mongodb config as well.

chombium:
@drasko, @anovakovic01 I fixed all the expose and ports doubles. I've checked and compared the docker exposed ports before and after the changes and I've tested locally in dockerized environment all the request based on the getting started guide, everything works properly including the db readers and writers. Tell me if I or some of you guys needs to do some more testing, before merging.

anovakovic01 (approved):
LGTM!

dborovcanin (approved):

PR #449, opened 2018-11-06
MF-448 - Option for Postgres SSL Mode

### LINKED ISSUE #448: Give option to connect to Postgres with SSL (Things & Users)
**FEATURE REQUEST**

Might be related/similar to #442 - Feel free to close this and I can add to that issue if that is prefered.

The postgres connection string for Things and Users currently does not have an option for picking between disable, require, etc while connecting to the DB. This would add a flag such as: 
```
| MF_THINGS_DB_SSL_MODE | SSL mode to connect to the database with | disable        |
```
Allowing someone to pick the ssl mode.

This is a must have for us as any db we can connect to, at minimum, requires ssl making disable not an option. 

I currently have this implemented in a way that supports both disable and require but not verify-ca or verify-full. If it makes sense, I can open a PR now that supports disable and require as a first step towards fully supporting this and we can come back to verify-ca and verify-full. 

 https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters

mfinley3:
### What does this do?
Adds an option to choose the ssl mode when connecting to postgres.
Only supporting `disable` or `require` for now.

Related to #448 

Current functionality is maintained with `disable` being the default for connections.

Updated the tests to pass `disable` to the connect function.

Yep! READMEs for Things & Users both now have the env var documented

### Notes
This does not support verify-ca and verify-full as of yet. I think some more discussion around what that should look like would be good. For example should these be separate options or can (should) the options for certs be leveraged the existing env vars? 
``` 
MF_THINGS_CA_CERTS   
MF_THINGS_SERVER_CERT
MF_THINGS_SERVER_KEY 
```

To support verify-ca and verify-full we'd need to provide:
```
sslcert - Cert file location. The file must contain PEM encoded data.
sslkey - Key file location. The file must contain PEM encoded data.
sslrootcert - The location of the root certificate file. The file must contain PEM encoded data.
```

### DIFF
--- a/cmd/things/main.go
+++ b/cmd/things/main.go
@@ -5,7 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-package main
+package things
 
 import (
 	"database/sql"
@@ -43,6 +43,7 @@ const (
 	defDBUser     = "mainflux"
 	defDBPass     = "mainflux"
 	defDBName     = "things"
+	defDBSSLMode  = "disable"
 	defCACerts    = ""
 	defCacheURL   = "localhost:6379"
 	defCachePass  = ""
@@ -58,6 +59,7 @@ const (
 	envDBUser     = "MF_THINGS_DB_USER"
 	envDBPass     = "MF_THINGS_DB_PASS"
 	envDBName     = "MF_THINGS_DB"
+	envDBSSLMode  = "MF_THINGS_DB_SSL_MODE"
 	envCACerts    = "MF_THINGS_CA_CERTS"
 	envCacheURL   = "MF_THINGS_CACHE_URL"
 	envCachePass  = "MF_THINGS_CACHE_PASS"
@@ -76,6 +78,7 @@ type config struct {
 	DBUser     string
 	DBPass     string
 	DBName     string
+	DBSSLMode  string
 	CACerts    string
 	CacheURL   string
 	CachePass  string
@@ -126,6 +129,7 @@ func loadConfig() config {
 		DBUser:     mainflux.Env(envDBUser, defDBUser),
 		DBPass:     mainflux.Env(envDBPass, defDBPass),
 		DBName:     mainflux.Env(envDBName, defDBName),
+		DBSSLMode:  mainflux.Env(envDBSSLMode, defDBSSLMode),
 		CACerts:    mainflux.Env(envCACerts, defCACerts),
 		CacheURL:   mainflux.Env(envCacheURL, defCacheURL),
 		CachePass:  mainflux.Env(envCachePass, defCachePass),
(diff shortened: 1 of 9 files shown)


mfinley3:
The dashflux build is failing at `npm install`. Not too sure why.
@drasko Mind looking into that? 

https://semaphoreci.com/mainflux/mainflux/branches/pull-request-449/builds/2

anovakovic01:
@mfinley3 This happens sometimes, I triggered rebuild, so it should be OK now.

drasko (commented):
Very good stuff here @mfinley3 . Let's just sync on this package name change.

cmd/things/main.go
```diff
@@ -5,7 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-package main
+package things
```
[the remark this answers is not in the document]
mfinley3: oops, yep shouldn't have changed that I'll fix that up now!

drasko (dismissed):
LGTM

chombium:
@mfinley3 It would be also nice if you add one new "Securing PostgeSQL connections" or a like in the getting started guide in which in few sentences it will be described how to do it.

I also think that we should adjust the things and users  container configs in docker-compose to use these parameters. The encryption by default won't harm us, as one can always connect to the db and see what's in it. @mainflux/contributors any thoughts on this?

dborovcanin (dismissed):
As @chombium suggested, it would be nice to add a section in the docs just to explain this SSL config in a sentence or two.

users/README.md
```diff
@@ -16,19 +16,20 @@ The service is configured using the environment variables presented in the
...
-| MF_USERS_SECRET      | String used for signing tokens                  | users        |
+| Variable             | Description                                      | Default      |
+|----------------------|--------------------------------------------------|--------------|
+| MF_USERS_LOG_LEVEL   | Log level for Users (debug, info, warn, error)   | error        |
+| MF_USERS_DB_HOST     | Database host address                            | localhost    |
+| MF_USERS_DB_PORT     | Database host port                               | 5432         |
+| MF_USERS_DB_USER     | Database user                                    | mainflux     |
+| MF_USERS_DB_PASSWORD | Database password                                | mainflux     |
+| MF_USERS_DB          | Name of the database used by the service         | users        |
+| MF_USERS_DB_SSL_MODE | Database connection SSL mode (disable or require)| disable      |
```
anovakovic01, line 27: Can you add `allow`, `prefer`, `verify-ca` and `verify-full` as options too?

anovakovic01 (changes requested):

users/README.md
mfinley3: The pq package only supports `disable`, `require`, `verify-ca` and `verify-full`

I was a bit hesitant to add `verify-ca` and `verify-full` to the readme right away because these changes don't support them (at least not yet). 

I think we need talk about requirements and best way to implement `verify-ca` and `verify-full` since they will need some extra info in the connection string: 
```
sslcert - Cert file location. The file must contain PEM encoded data.
sslkey - Key file location. The file must contain PEM encoded data.
sslrootcert - The location of the root certificate file. The file must contain PEM encoded data.
```
Do we add more/separate env vars for these or can we leverage the certs vars that were added in that recent PR? 

My initial thought was that this PR supports require/disable then we open up talk about supporting the other two but we can certainly add support for these in this PR is that is prefered! 

Thoughts on this @anovakovic01?

anovakovic01: @mfinley3 I was thinking about adding more env vars, but we can do that in separate PR. Maybe you're right, we should add this in separate PR (it will be easier to review changes). So, just sync with master, and we can merge.

mfinley3:
@chombium & @dusanb94 Good thoughts, added a small section `Securing PostgeSQL connections` in the getting started doc!

drasko (approved):
LGTM

drasko:
This looks ready, nice work again @mfinley3 - many thanks!

dborovcanin (approved):
Great job, thank you.

anovakovic01 (approved):
LGTM!

PR #455, opened 2018-11-09
MF-454 - Use message Time field as a time for InfluxDB points

### LINKED ISSUE #454: Use message Time field as a time for InfluxDB points
In order to provide more meaningful measurements, consider using time value received from the message as the value of "time" column when saving data in InfluxDB.

dborovcanin:
Signed-off-by: Dusan Borovcanin <dusan.borovcanin@mainflux.com>

### What does this do?
Use time from received message to set time value in InfluxDB.  

This pull request closes #454.

### Notes
This pull request slightly modifies the way messages are saved in InfluxDB. Field "Time" is removed and replaced by default "time" column of InfluxDB. Previously, "time" kept information about timestamp when the message is saved in the database, so that info is no longer persistent in InfluxDB. However, this information can be found in service logs if needed. This also means that the size of tag set of the single point is slightly reduced.

### DIFF
--- a/readers/cassandra/messages_test.go
+++ b/readers/cassandra/messages_test.go
@@ -10,6 +10,7 @@ package cassandra_test
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/mainflux/mainflux"
 	readers "github.com/mainflux/mainflux/readers/cassandra"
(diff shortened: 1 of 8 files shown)


drasko (changes requested):
As discussed in the issue - tagset is wrong.

This tagset will form huge number of different series. Only `channelID` and eventually `content-type` should be in a tagset.

drasko (dismissed):
LGTM

nmarcetic (dismissed):

nmarcetic (dismissed):

drasko (approved):
LGTM

PR #459, opened 2018-11-19
Add TLS support to CoAP adapter and all readers

anovakovic01:
### What does this do?
Adds TLS client support for CoAP adapter and all readers.

No.

Yes, I updated `README.md` files.

### DIFF
--- a/cmd/cassandra-reader/main.go
+++ b/cmd/cassandra-reader/main.go
@@ -94,7 +94,7 @@ func main() {
 func loadConfig() config {
 	tls, err := strconv.ParseBool(mainflux.Env(envClientTLS, defClientTLS))
 	if err != nil {
-		tls = false
+		log.Fatalf("Invalid value passed for %s\n", envClientTLS)
 	}
 
 	return config{
--- a/cmd/cassandra-reader/main.go
+++ b/cmd/cassandra-reader/main.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"syscall"
 
@@ -26,6 +27,7 @@ import (
 	thingsapi "github.com/mainflux/mainflux/things/api/grpc"
 	stdprometheus "github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 )
 
 const (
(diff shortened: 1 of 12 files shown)


nwest1 (dismissed):
I do like the use of a flag in the TLS configuration!

drasko:
@nwest1 usually the goal here is for @anovakovic01 to see this picture: 
image

You have been too soft on him, there must be something to request changes ;)

cmd/cassandra-reader/main.go
```diff
@@ -103,13 +118,27 @@ func connectToCassandra(cluster, keyspace string, logger logger.Logger) *gocql.S
 	return session
 }
 
-func connectToThings(url string, logger logger.Logger) *grpc.ClientConn {
-	conn, err := grpc.Dial(url, grpc.WithInsecure())
+func connectToThings(cfg config, logger logger.Logger) *grpc.ClientConn {
+	var opts []grpc.DialOption
+	if cfg.clientTLS {
+		if cfg.caCerts != "" {
+			tpc, err := credentials.NewClientTLSFromFile(cfg.caCerts, "")
```
drasko, line 125: What's `tpc`? If I did not get this on a first sight, would that mean that naming is badly chosen?

drasko (changes requested):

chombium (commented):
The pr looks good. 
I have one small comment. In each strconv.ParseBool error check if the envClientTLS var contains a string which can not be converted to bool (user typo when setting the envvar's value, for example), the TLS encryption is silently turned off.
Later in the code, the user will get a message that the encryption is turned off, but it might be confusing why: both TLS envvars are set, but no TLS is activated.

IMO it would be better if a message is logged sayng that the envvar's value is wrong. The other thing to think about is if we want to throw an error and exit, because of the misconfigured envvar, or we want to do our best and start the component in a way which is not wanted by the user (that's what's currently being done).

What do you think about that?

cmd/cassandra-reader/main.go
anovakovic01: We have this name in every service, and it's short for `TransportCredentials`. Also, this is local var that is used in only one place, so this name is not so important (it is in go spirit to use short names for local vars).

anovakovic01:
@chombium Thanks for the feedback, I was not sure about this myself. Now that I think about this, I think that we should exit with logged error, if TLS env var has invalid value.

chombium:
@anovakovic01 thanks for the quick response. That looks much better

drasko (approved):
LGTM

dborovcanin (approved):

PR #485, opened 2018-11-30
NOISSUE - Add Makefile target to clean old imgs

drasko:

### What does this do?
Add Makefile target to clean old and ghost docker images and containers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,23 @@ clean:
 	rm -rf ${BUILD_DIR}
 	rm -rf mqtt/node_modules
 
+cleandocker: cleanghost
+	# Stop all containers (if running)
+	docker-compose -f docker/docker-compose.yml stop
+	# Remove mainflux containers
+	docker ps -a | grep mainflux | awk '{print $$1}' | xargs -r docker rm
+	# Remove old mainflux images
+	docker images | grep mainflux | awk '{print $$3}' | xargs -r docker rmi
+
+# Clean ghost docker images
+cleanghost:
+	# Remove exited containers
+	docker ps --filter status=dead --filter status=exited -aq | xargs -r docker rm -v
+	# Remove unused images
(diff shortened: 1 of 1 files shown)


chombium (changes requested):
Thumbs up for this PR as I'm doing this manually all of the time. Silly me I could have made this pr :-/

I've played a bit with these commands on on my machine Arch Linux with
```
$ awk --version
GNU Awk 4.2.1, API: 2.0 (GNU MPFR 4.0.1, GNU MP 6.1.2)
```
the awk print commands with double $ return garbage and with single $ the ids of the containers and the images. All the awk print docs and examples also show that single $ should be used.

@drasko have you maybe made some mistake when writing the commands? Could you please check that?

We could also simplify the commands and get rid of the grep-awk combos.
`docker ps -a | grep mainflux | awk '{print $$1}' | xargs -r docker rm` would be
`docker rm $(docker ps -a -f "name=mainflux" -q)`
and
`docker images | grep mainflux | awk '{print $$3}' | xargs -r docker rmi` would be 
`docker rmi $(docker images "mainflux\/*" -q)`

for the `<none>` images as per [this](http://www.projectatomic.io/blog/2015/07/what-are-docker-none-none-images/) blogpost `docker rmi $(docker images -f "dangling=true" -q)` should do the trick, but I don't have any of the so I can test.

@drasko what do you think about this?

drasko:
@chombium `awk` uses single `$`, but `$$` is escape sequence in Makefile - otherwise command would not be correct, as Makefile would treat this as a variable.

Regarding the commands you proposed they make sense, and I will try them and incorporate them if they produce indentical results as they are more elegant.

drasko:
@chombium thanks for advices, looks like it's working and it should be OK now.

anovakovic01 (approved):
LGTM!

dborovcanin (approved):

manuio (approved):

Issue #435, opened 2018-10-31
Enable service configuration by using configuration files


ghost:
**FEATURE REQUEST**

1. Is there an open issue addressing this request? If it does, please add a "+1" reaction to the
   existing issue, otherwise proceed to step 2.

2. Describe the feature you are requesting, as well as the possible use case(s) for it.

Currently, services are configured by setting environment variables, or by defaults if variables are not set.
It would be nice to be able to supply a configuration file to the service as well.

3. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).

This is a nice to have feature, especially in cases when someone runs a service outside a docker container.

ghost:
I would advise checking viper, since it can read environment variables and also read configuration files in several formats

drasko:
Agreed - [Viper](https://github.com/spf13/viper) looks like a nice tool for the job.

drasko:
I am not sure that we need this one. Config files are more difficult to distribute (although not much more), and we have nice solution no with [.env](https://github.com/mainflux/mainflux/blob/master/.env).

@nmarcetic @anovakovic01 opinions?

drasko:
Also - if we leave both solutions - ENV and `.config` files if present (or used with a `--config` switch), this might be the best option.

chombium:
@mainflux/contributors do we really need two ways for configuration? IMHO we should have one and only one way for configuration. I personally don't like systems where a same functionality can be achieved in many different ways as it confuses the users and it makes the implementation more complex

drasko:
@chombium my opinion is that config via ENV variables is the best approach, but it would not bother me if config files and command line switches exist as an alternative.

I just think that this is really not necessary and makes things harder to maintain.

manuio:
@drasko We solved the problem for docker but natively it still complicated to setup the system (if you're not using default one, of course) because there are many duplicated variables all around the code.
How hard it would be to remove all definitions from main.go and copy the values from the same file used for Docker `.env` ?

nmarcetic:
I agree with @srados I would use Viper to read config files and/or ENV var. So would support bought config + env, it would be great!

manuio:
Reopening this one to finish services implementation

cyber-goka:
idk if it is relevant but in microservices architecture, it is very common to use discovery + config servers such as a consul, it can even store passwords if integrated with the vault. provisioning in k8s, docker-compose will be easier. also can implement service discovery with consul

drasko:
[...]

PR #500, opened 2018-12-10
MF-448 - Add option to connect to DB with verify-ca and verify-full

### LINKED ISSUE #448: Give option to connect to Postgres with SSL (Things & Users)

mfinley3:
### What does this do?
Adds the option to connect to the DBs with verify-ca and verify-full
Users can now specify any extra certs and keys they may need.

Resolves #448 

None of the default behavior has changed, the default db ssl connection mode remains `disabled`

I have updated the relevant tests. The pq library tests should cover the changes here since these are just implementing their options

Yes!

### Notes
No other notes

### DIFF
--- a/cmd/users/main.go
+++ b/cmd/users/main.go
@@ -34,45 +34,54 @@ import (
 )
 
 const (
-	defLogLevel   = "error"
-	defDBHost     = "localhost"
-	defDBPort     = "5432"
-	defDBUser     = "mainflux"
-	defDBPass     = "mainflux"
-	defDBName     = "users"
-	defDBSSLMode  = "disable"
-	defHTTPPort   = "8180"
-	defGRPCPort   = "8181"
-	defSecret     = "users"
-	defServerCert = ""
-	defServerKey  = ""
-	envLogLevel   = "MF_USERS_LOG_LEVEL"
-	envDBHost     = "MF_USERS_DB_HOST"
-	envDBPort     = "MF_USERS_DB_PORT"
-	envDBUser     = "MF_USERS_DB_USER"
-	envDBPass     = "MF_USERS_DB_PASS"
-	envDBName     = "MF_USERS_DB"
-	envDBSSLMode  = "MF_USERS_DB_SSL_MODE"
-	envHTTPPort   = "MF_USERS_HTTP_PORT"
-	envGRPCPort   = "MF_USERS_GRPC_PORT"
-	envSecret     = "MF_USERS_SECRET"
-	envServerCert = "MF_USERS_SERVER_CERT"
-	envServerKey  = "MF_USERS_SERVER_KEY"
+	defLogLevel      = "error"
+	defDBHost        = "localhost"
+	defDBPort        = "5432"
+	defDBUser        = "mainflux"
+	defDBPass        = "mainflux"
+	defDBName        = "users"
+	defDBSSLMode     = "disable"
+	defDBSSLCert     = ""
+	defDBSSLKey      = ""
+	defDBSSLRootCert = ""
+	defHTTPPort      = "8180"
+	defGRPCPort      = "8181"
+	defSecret        = "users"
+	defServerCert    = ""
+	defServerKey     = ""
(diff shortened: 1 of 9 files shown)


dborovcanin (dismissed):
Have you considered wrapping database connection parameters to some kind of config entity (map, for example)?

mfinley3:
That actually came up from our end as well @dusanb94. 
CC: @KarthikSubramanian2

9 params being passed seems a bit excessive. A map would make sense, I'm going to see what that looks like and report back!

users/postgres/init.go
```diff
@@ -18,8 +18,8 @@ import (
 // Connect creates a connection to the PostgreSQL instance and applies any
 // unapplied database migrations. A non-nil error is returned to indicate
 // failure.
-func Connect(host, port, name, user, pass, sslMode string) (*sql.DB, error) {
-	url := fmt.Sprintf("host=%s port=%s user=%s dbname=%s password=%s sslmode=%s", host, port, user, name, pass, sslMode)
+func Connect(host, port, name, user, pass, sslMode, sslCert, sslKey, sslRootCert string) (*sql.DB, error) {
```
anovakovic01: I think that it makes sense to create db config structure and use it as parameter here.

chombium: @anovakovic01 what do you mean by "db config structure"? I hope you are not thinking of storing the config in some db table ;)
If we need config lets do it as @srados suggested in #435

ghost: @chombium I think he meant that instead of passing a set of parameters, he should pass a struct containing all parameters

anovakovic01: @chombium yes, I meant passing struct instead of all these parameters.

chombium: @chombium, @srados Thanks for the clarification. I thought that too, but that db caught my eyes :)

mfinley3:
I've rebased and modified the functions to accept a config struct rather than a bunch of params.

@anovakovic01 & @dusanb94 let me know what you think!

xuantinfx (approved):

xuantinfx (dismissed):
ok

dborovcanin (dismissed):

cmd/users/main.go
```diff
@@ -34,40 +34,41 @@ import (
...
-	DBHost     string
-	DBPort     string
-	DBUser     string
-	DBPass     string
-	DBName     string
-	DBSSLMode  string
+	dbConfig   postgres.Config
 	HTTPPort   string
 	GRPCPort   string
 	Secret     string
```
anovakovic01: Please, change other `config` fields to private.

things/postgres/init.go
```diff
@@ -15,11 +15,23 @@ import (
 	migrate "github.com/rubenv/sql-migrate"
 )
 
+type Config struct {
```
anovakovic01, line 19: Add comment to explain meaning of this config struct.

anovakovic01 (changes requested):
@mfinley3 Almost there.

things/postgres/init.go
mfinley3: Added a comment to both this one and users/postgres/init.go

drasko (approved):
LGTM

anovakovic01 (approved):
LGTM!

PR #504, opened 2018-12-13
MF-419 - Use JSON for CLI commands output

### LINKED ISSUE #419: Change CLI commands output
**FEATURE REQUEST**

It would be nice to change the CLI output (probably to JSON) so it's easy to parse it.

Example use case is:

- create user
- extract user token
- create channel (by using user token)
- create thing (by using user token)
- extract thing token
- connect thing to channel (by using user token)
- publish message (using thing token)

The commands above need to have the output which is easy to parse in order to make scripting/command piping simple.

I think this is a should-have feature

manuio:

### What does this do?
Use JSON format for CLI commands output.
Resolves #419
nc
no
no
### Notes
nc


chombium:
I personally don't like that now all the output will be json. Usually CLIs are buit to execite some tasks quickly and give clear human readable and understandable text as output.
I think it would be better to keep the text output as default and add an extra parameter if json output is needed. The other question is what is the motivation for the issue and the pr in first place. Why would someone use cli for automation when we also have a sdk and rest api?

@manuIO, @srados, @drasko what are your thoughts on this?

manuio:
@chombium I think that we are talking about the same thing. This fix it's to print commands output on the console in JSON format. 

Using the CLI you can provision Mainflux, connect things to channels and send messages very quickly. The SDK is very useful to create applications but you need to code. With the CLI you just need to write some commands, which is useful IMHO.

chombium:
Thanks for the clarification @manuIO Of course we are talking about the same thing. I'm worried abou the usability as 80-85+% of the users would use the cli to create things or channels and would expect that they get a readable message like "thing created with id ..." as a response. On the other side, if we say that if someone is using the terminal is a bit of advanced user, json output would also be acceptable :)

ghost:
@chombium the reason for the json output is so you can do something meaningful with it, other than just looking at it on the screen :)  
For example, you can pipe it to jq in order to be able to easily extract the fields you need.
Sometimes you need to quickly script something in bash...

manuio:
@chombium To be exact the output is a JSON only when you fetch channels or things. When you create a user or send a message the output is `ok` and when you create entities the output is just `created: <entityID>`.

ghost (dismissed):

ghost:
@manuIO maybe it would be also good to return user id on user creation (in json of course :))

anovakovic01:
@srados user id is an email that is passed with creation request

cli/utils.go
```diff
@@ -21,18 +22,34 @@ var (
...
-	fmt.Printf(color.YellowString("Usage:  %s\n"), u)
+	fmt.Printf(color.YellowString("\nusage: %s\n\n"), u)
 }
 
 func logError(err error) {
-	fmt.Printf("%s\n", color.RedString(err.Error()))
+	fmt.Printf("\n%s\n\n", color.RedString(err.Error()))
 }
 
 func logOK() {
```
dborovcanin, line 51: Maybe we can join `logOK` and `logCreated` to single variadic `logSuccess(s ...string)` function. Something like:
```go
func logSuccess(s ...string) {
	switch len(s) {
	case 1:
		fmt.Printf(..., s[0])
	case 0:
		fmt.Println("OK")
	}
}

```
It can be used as `logSuccess()` or `logSuccess("message")`.

drasko (dismissed):
LGTM

cli/provision.go
```diff
@@ -230,10 +230,10 @@ var cmdProvision = []cobra.Command{
...
 			}
 
-			flush(user)
-			flush(ut)
-			flush(things)
-			flush(channels)
+			logJSON(user)
+			logJSON(ut)
+			logJSON(things)
+			logJSON(channels)
```
anovakovic01: Maybe you can update `logJSON` function to receive list of JSON objects (objs...), so you can pass `users`, `ut`, `things` and `channels` in one function call.

anovakovic01 (changes requested):

anovakovic01 (approved):
LGTM!

drasko (approved):
LGTM

PR #510, opened 2018-12-16
MF-506 - Fix influx & mongo readers queries to return most recent messages

### LINKED ISSUE #506: Readers should return the newer messages
Readers should return newer messages instead of older ones because of the pagination (`LIMIT` condition in DB queries). Otherwise, without setting the offset it always fetch older messages and to fetch newer messages you need to know how many messages you already stored and adjust the offset, which doesn't make sense.

 It's easy as add `ORDER BY time DESC` in DB queries.

Also, from the UI perspective it will be much more easy to manipulate messages.

manuio:

### What does this do?
Inverse the order of returned messages by the influxdb-reader
Resolves #506
no
no
### Notes
Have to be done all readers


dborovcanin (approved):

anovakovic01 (approved):
LGTM!

PR #511, opened 2018-12-16
 MF-433 - Add method for reading messages from readers to SDK.

manuio:


sdk/go/sdk.go
```diff
@@ -93,6 +96,23 @@ type Channel struct {
 	Metadata string  `json:"metadata,omitempty"`
 }
 
+// Message represents mainflux message.
+type Message struct {
```
drasko, line 100: Why are we duplicating the code instead of having shared libs?

cli/message.go
```diff
@@ -7,7 +7,9 @@
 
 package cli
 
-import "github.com/spf13/cobra"
+import (
```
drasko: When having only one import then please inline (like it was before).

sdk/go/channels.go
```diff
@@ -24,7 +24,7 @@ func (sdk mfSDK) CreateChannel(channel Channel, token string) (string, error) {
 		return "", ErrInvalidArgs
 	}
 
-	url := createURL(sdk.url, sdk.thingsPrefix, channelsEndpoint)
+	url := createURL(sdk.baseURL, sdk.thingsPrefix, channelsEndpoint)
```
drasko, line 27: Have this been tested on both docker and local installation?

drasko (changes requested):

sdk/go/channels.go
manuio: @drasko This have been tested from CLI locally.

sdk/go/sdk.go
manuio: @dusanb94 what's your opinion here?

anovakovic01 (approved):
LGTM!

drasko (approved):
LGTM

PR #527, opened 2018-12-27
NOISSUE - Fix event sourcing client on LoRa adapter

anovakovic01:
Fixes event sourcing client on LoRa adapter. Client was missing events because it couldn't create user group. This was a known issue and it's fixed with new SDK method for creating consumer group. Now we use method that creates empty stream if stream doesn't already exist.

### DIFF
--- a/lora/redis/streams.go
+++ b/lora/redis/streams.go
@@ -69,7 +69,7 @@ func NewEventStore(svc lora.Service, client *redis.Client, consumer string, log 
 }
 
 func (es eventStore) Subscribe(subject string) {
-	es.client.XGroupCreate(stream, group, "$").Err()
+	es.client.XGroupCreateMkStream(stream, group, "$").Err()
 	for {
 		streams, err := es.client.XReadGroup(&redis.XReadGroupArgs{
 			Group:    group,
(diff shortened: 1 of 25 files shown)


lora/redis/streams.go
```diff
@@ -105,7 +105,7 @@ func (es eventStore) Subscribe(subject string) {
 				rce := decodeRemoveChannel(event)
 				err = es.handleRemoveChannel(rce)
 			}
-			if err != nil {
+			if err != nil && err != errMetadataType {
```
drasko, line 108: What happens if `err != nil` and it is not `errMetadataType` (but some other kind of error)? We do not log this at all?

drasko (changes requested):

lora/redis/streams.go
anovakovic01: Yes, we log only that case (case that you just asked about, where error is not `errMetadataType`), and we don't need to log `errMetadataType`, because this means that event is of no interest for our service.

manuio (dismissed):
LGTM

drasko (dismissed):
LGTM

manuio (dismissed):

manuio (approved):

drasko (approved):
LGTM

PR #541, opened 2019-01-10
MF-370 - Simplify and refine CI

### LINKED ISSUE #370: Simplify and refine CI
Now that we do not need `protobuf` build, Semafore CI task should be refined and simplified.

Additionally, Semphore CI open-source plan permits 4 parallel jobs which we are not fully utilizing  - parallelize UI and backend build as much as possible.

dborovcanin:
Signed-off-by: Dusan Borovcanin <dusan.borovcanin@mainflux.com>.

### What does this do?
This pull request adds a script to be executed by a CI tool.

This pull request closes #370.

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -1,41 +1,42 @@
+# This script contains commands to be executed by the CI tool.
+
 setup_protoc() {
-  echo "Setup protoc..."
-  PROTOC_ZIP=protoc-3.6.1-linux-x86_64.zip
-  curl -0L https://github.com/google/protobuf/releases/download/v3.6.1/$PROTOC_ZIP -o $PROTOC_ZIP
-  unzip -o $PROTOC_ZIP -d protoc3
-  sudo mv protoc3/bin/* /usr/local/bin/
-  sudo mv protoc3/include/* /usr/local/include/
-  rm -f PROTOC_ZIP
-  go get -u github.com/golang/protobuf/protoc-gen-go \
-    github.com/gogo/protobuf/protoc-gen-gofast \
-    google.golang.org/grpc
-  export PATH=$PATH:/usr/local/bin/protoc
+	echo "Setup protoc..."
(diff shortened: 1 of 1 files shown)


.semaphore/semaphore.yml
```diff
@@ -0,0 +1,28 @@
...
+            - unzip -o $PROTOC_ZIP -d protoc3
+            - sudo mv protoc3/bin/* /usr/local/bin/
+            - sudo mv protoc3/include/* /usr/local/include/
+            - rm -f PROTOC_ZIP
+            - export PATH=$PATH:/usr/local/bin/protoc
+            - go get -u github.com/golang/protobuf/protoc-gen-go github.com/gogo/protobuf/protoc-gen-gofast google.golang.org/grpc
+            - go get -d github.com/mainflux/mainflux
+            - cd $GOPATH/src/github.com/mainflux/mainflux
+            - make proto
+            - set -e; echo "" > coverage.txt; for d in $(go list ./... | grep -v 'vendor\|cmd'); do GOCACHE=off go test -v -race -tags test -coverprofile=profile.out -covermode=atomic $d; if [ -f profile.out ]; then cat profile.out >> coverage.txt; rm profile.out; fi done
```
chombium: I'd suggest use of multiline string (Block Scalar >) as described [here](https://stackoverflow.com/a/21699210) for better visibility, but I'm not sure if it will work with Semaphore CI. Other than that all looks good to me, but Semaphore CI does not report a status... I guess something else should be adjusted as well. Do we have any logs?

chombium (dismissed):

.semaphore/semaphore.yml
dborovcanin: @chombium We've decided not to switch to Semaphore 2 yet because it's still in Beta phase. This PR will add simple shell script that will be executed by Semaphore so that we can track CI commands changes.

drasko (approved):
LGTM

nmarcetic (approved):

PR #545, opened 2019-01-16
NOISSUE - Add env var for number of concurrent messages

anovakovic01:
Add env var for number of messages that can be concurrently exchanged over one instance of MQTT adapter.


manuio (approved):
LGTM

drasko (approved):
LGTM

PR #547, opened 2019-01-17
NOISSUE - Fix pagination response for empty page

anovakovic01:
Fix pagination response for the case when list is empty. API was returning `null` in some cases, now it will return empty list.


drasko (approved):
LGTM

dborovcanin (approved):

PR #561, opened 2019-01-28
NOISSUE - Fix MQTT adapter by setting subscription queue

anovakovic01:
Fix MQTT adapter by setting subscription queue instead of normal subscription to NATS.


blokovi (approved):
LGTM

drasko (approved):
LGTM

PR #562, opened 2019-01-29
NOISSUE- Improve LoRa doc

manuio:
### What does this do?
Improve lora-adapter docs


drasko (dismissed):
LGTM

drasko (dismissed):
LGTM

docs/lora.md
```diff
@@ -54,7 +62,7 @@ The lora-adapter uses the matadata of provision events emitted by mainflux-thing
     "type": "lora",
     "appID": "<application ID>"
   }
-}
+}a
```
dborovcanin: Is this `a` added accidentally?

docs/lora.md
```diff
@@ -10,40 +16,42 @@ Once everything is installed, execute the following command from the LoRa Server
 docker-compose up
 ```
 
-The Mainflux lora-adapter can do bridging between both systems. Basically, the service subscribe to the [LoRa Gateway Bridge](https://www.loraserver.io/lora-gateway-bridge/overview/), a service which abstracts the [SemTech packet-forwarder UDP protocol](https://github.com/Lora-net/packet_forwarder/blob/master/PROTOCOL.TXT) into JSON over MQTT.
+**Troubleshouting:**  Mainflux and LoRa Server use their own MQTT brokers. By default, those use the standard MQTT port `1883`. If you are running both systems in the same machine you must use different ports. You can fix this on Mainflux side configuring the environment variable `MF_MQTT_ADAPTER_PORT`.
```
dborovcanin: `On` the same machine.

manuio: Thanks!

docs/lora.md
manuio: yes.. good catch

drasko (approved):
LGTM

PR #565, opened 2019-01-30
NOISSUE: add subtopic support for nats messages

beres:
### What does this do?
Nats listener subscribe to "channels.>" to catch all messages to manage also subtopic.

### DIFF
--- a/mqtt/mqtt.js
+++ b/mqtt/mqtt.js
@@ -80,6 +80,12 @@ function startMqtt() {
     return net.createServer(aedes.handle).listen(config.mqtt_port);
 }
 
+function parseMqttTopic(topic) {
+    // allow channels/any-char-allowed/message with 0|+ block /any-char-allowed
+    // the text messages and the last subtopic can not terminate with /
+    return /^channels\/(.+?)\/messages(\/.+[^\/]$)*$/.exec(topic)
+}
+
 nats.subscribe('channel.>', {'queue':'mqtts'}, function (msg) {
     var m = message.RawMessage.decode(Buffer.from(msg)),
         packet;
@@ -99,29 +105,21 @@ nats.subscribe('channel.>', {'queue':'mqtts'}, function (msg) {
 aedes.authorizePublish = function (client, packet, publish) {
     // Topics are in the form `channels/<channel_id>/messages`
     // Subtopic's are in the form `channels/<channel_id>/messages/<subtopic>`
-    var channel = /^channels\/(.+?)\/messages\/?.*$/.exec(packet.topic);
+    var channel = parseMqttTopic(packet.topic)
     if (!channel) {
-        logger.warn('unknown topic');
+        logger.warn('unknown or malformed topic');
         publish(4); // Bad username or password
         return;
     }
+
     var channelId = channel[1],
+        channelTopic = 'channel.' + channelId + (
+            channel[2] !== undefined ? channel[2].replace(/\//g, '.') : ''
+        ),
         accessReq = {
             token: client.password,
             chanID: channelId
         },
-        // Parse unlimited subtopics
-        baseLength = 3, // First 3 elements which represents the base part of topic.
-        elements = packet.topic.split('/').slice(baseLength),
-        baseTopic = 'channel.' + channelId;
-    // Remove empty elements
-    for (var i = 0; i < elements.length; i++) {
-      if (elements[i] === '') {
-        elements.pop(i)
-      }
-    }
-    var channelTopic = elements.length ? baseTopic + '.' + elements.join('.') : baseTopic,
-
         onAuthorize = function (err, res) {
             var rawMsg;
             if (!err) {
@@ -149,7 +147,7 @@ aedes.authorizePublish = function (client, packet, publish) {
 aedes.authorizeSubscribe = function (client, packet, subscribe) {
     // Topics are in the form `channels/<channel_id>/messages`
     // Subtopic's are in the form `channels/<channel_id>/messages/<subtopic>`
-    var channel = /^channels\/(.+?)\/messages\/?.*$/.exec(packet.topic);
+    var channel = parseMqttTopic(packet.topic)
     if (!channel) {
       logger.warn('unknown topic');
       subscribe(4, packet); // Bad username or password
--- a/mqtt/mqtt.js
+++ b/mqtt/mqtt.js
@@ -113,6 +113,8 @@ aedes.authorizePublish = function (client, packet, publish) {
     }
 
     var channelId = channel[1],
+        // channelTopic contains channelId prefixed by 'channel.' and suffiex
+        // by every topic part in the forms .topic
         channelTopic = 'channel.' + channelId + (
             channel[2] !== undefined ? channel[2].replace(/\//g, '.') : ''
         ),
--- a/mqtt/mqtt.js
+++ b/mqtt/mqtt.js
@@ -80,7 +80,7 @@ function startMqtt() {
     return net.createServer(aedes.handle).listen(config.mqtt_port);
 }
 
-nats.subscribe('channel.*', {'queue':'mqtts'}, function (msg) {
+nats.subscribe('channel.>', {'queue':'mqtts'}, function (msg) {
     var m = message.RawMessage.decode(Buffer.from(msg)),
         packet;
     if (m && m.protocol !== 'mqtt') {
--- a/mqtt/mqtt.js
+++ b/mqtt/mqtt.js
@@ -88,12 +88,19 @@ function parseMqttTopic(topic) {
 
 nats.subscribe('channel.>', {'queue':'mqtts'}, function (msg) {
     var m = message.RawMessage.decode(Buffer.from(msg)),
-        packet;
+        packet, channelParts;
     if (m && m.protocol !== 'mqtt') {
+        // rexexp has two element, the first one is the channelId, wile
+        // the second is undefined or the subtopic parts starting from the dot
+        channelParts = /^([\w\-]+)(\..+[^\.]$)*$/.exec(m.channel)
         packet = {
             cmd: 'publish',
             qos: 2,
-            topic: 'channels/' + m.channel + '/messages',
+            // compose the topic with channelId followed by /messages and
+            // all the subtopics in the form /subtopic/subN/
+            topic: 'channels/' + channelParts[1] + '/messages' + (
+                channelParts[2] !== undefined ? channelParts[2].replace(/\./g, '\/') : ''
+            ),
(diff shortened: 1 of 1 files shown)


drasko (commented):
Looks good and very nice fix, I just need to re-check this regexp in little more detail.

@nmarcetic please review

mqtt/mqtt.js
```diff
@@ -80,14 +80,16 @@ function startMqtt() {
...
     var m = message.RawMessage.decode(Buffer.from(msg)),
-        packet;
+        packet, channel;
     if (m && m.Protocol !== 'mqtt') {
+        channel = /^([\w\-]+)(\.(.+))?$/.exec(m.Channel)
         packet = {
             cmd: 'publish',
             qos: 2,
-            topic: 'channels/' + m.Channel + '/messages',
+            topic: 'channels/' + channel[1] + '/messages' + (
```
manuio: Just curious, what is channel[0] ?

manuio (changes requested):

mqtt/mqtt.js
beres: hi,

0 is always all the string if match the regexp
1 channel id
2 the remaing part  of the string or undefined
3 the remaing of the string or undefined

as example is m.Channel is "a22de3db-61df-461a-806f-5c2ed7f4cb10.data" channel var will be
[ 'a22de3db-61df-461a-806f-5c2ed7f4cb10.data', 'a22de3db-61df-461a-806f-5c2ed7f4cb10',  '.data',  'data',  index: 0,  input: 'a22de3db-61df-461a-806f-5c2ed7f4cb10.data' ]

while if m.Channel is "a22de3db-61df-461a-806f-5c2ed7f4cb10" channel var will be
[ 'a22de3db-61df-461a-806f-5c2ed7f4cb10', 'a22de3db-61df-461a-806f-5c2ed7f4cb10',  undefined, undefined,  index: 0,  input: 'a22de3db-61df-461a-806f-5c2ed7f4cb10' ]

manuio: Got it, thank you1

manuio (dismissed):
@beres LGTM. Just update the branch please

manuio (approved):

manuio (dismissed):

nmarcetic:
@beres Can you please update branch?

nmarcetic (commented):
@beres Maybe is a better idea if you can align it with [authorizePublish](https://github.com/mainflux/mainflux/blob/2f3fe5e5e70dbec2fd50e079f55258786ab842f4/mqtt/mqtt.js#L101) maybe even move this parse logic to separate helper function and just use it in both publish and subscribe, its exactly the same. I don't like the idea to repeat the code and do the same thing in 2 different ways, maybe I am wrong but looks like the same logic no? What do you think?

mqtt/mqtt.js
```diff
@@ -101,29 +107,21 @@ nats.subscribe('channel.>', {'queue':'mqtts'}, function (msg) {
 aedes.authorizePublish = function (client, packet, publish) {
     // Topics are in the form `channels/<channel_id>/messages`
     // Subtopic's are in the form `channels/<channel_id>/messages/<subtopic>`
-    var channel = /^channels\/(.+?)\/messages\/?.*$/.exec(packet.topic);
+    var channel = parseMqttTopic(packet.topic)
```
drasko: This name is confusing. Regexp does not return only channel ID in the selection, but subtopic as well from what I see.

Something like `topicParts` would make more sense.

mqtt/mqtt.js
```diff
@@ -101,29 +107,21 @@ nats.subscribe('channel.>', {'queue':'mqtts'}, function (msg) {
...
     if (!channel) {
-        logger.warn('unknown topic');
+        logger.warn('unknown or malformed topic');
         publish(4); // Bad username or password
         return;
     }
+
     var channelId = channel[1],
+        channelTopic = 'channel.' + channelId + (
+            channel[2] !== undefined ? channel[2].replace(/\//g, '.') : ''
```
drasko: Also add a comment here what are these hard-coded array members (what they represent).

drasko (changes requested):

mqtt/mqtt.js
beres: RegExp return and array with channel name and the topic parts, maybe i can raneme it channelParts, what do you thing?

drasko: Yes, this is good also. Just not `channel`, as it is missleadning.

drasko (approved):
LGTM

mqtt/mqtt.js
```diff
@@ -80,14 +80,22 @@ function startMqtt() {
...
+    // the text messages and the last subtopic can not terminate with /
+    return /^channels\/(.+?)\/messages(\/.+[^\/]$)*$/.exec(topic)
+}
+
+nats.subscribe('channel.>', {'queue':'mqtts'}, function (msg) {
     var m = message.RawMessage.decode(Buffer.from(msg)),
-        packet;
+        packet, channel;
     if (m && m.Protocol !== 'mqtt') {
+        channel = /^([\w\-]+)(\.(.+))?$/.exec(m.Channel)
```
drasko: Again, is this `channel` or `channelParts`? Looks like there are two selections here.

drasko (changes requested):

drasko (approved):
LGTM

drasko:
@beres this looks good, please update the branch.

mqtt/mqtt.js
```diff
@@ -80,14 +80,27 @@ function startMqtt() {
...
+}
+
+nats.subscribe('channel.>', {'queue':'mqtts'}, function (msg) {
     var m = message.RawMessage.decode(Buffer.from(msg)),
-        packet;
+        packet, channelParts;
     if (m && m.protocol !== 'mqtt') {
+        // rexexp has two element, the first one is the channelId, wile
+        // the second is undefined or the subtopic parts starting from the dot
+        channelParts = /^([\w\-]+)(\..+[^\.]$)*$/.exec(m.channel)
```
anovakovic01, line 95: @drasko AFAIK earlier we used `channel` only for channel ID. This means that we'll store whole topic in `channel` field, which is not only channel ID. Just want you to be aware of this. Also, it won't work with the current system (I think).

mqtt/mqtt.js
```diff
@@ -80,14 +80,27 @@ function startMqtt() {
...
         packet = {
             cmd: 'publish',
             qos: 2,
-            topic: 'channels/' + m.channel + '/messages',
+            // compose the topic with channelId followed by /messages and
+            // all the subtopics in the form /subtopic/subN/
+            topic: 'channels/' + channelParts[1] + '/messages' + (
```
anovakovic01, line 101: What if `channelParts` has length 1?

mqtt/mqtt.js
```diff
@@ -99,29 +112,23 @@ nats.subscribe('channel.*', {'queue':'mqtts'}, function (msg) {
...
-        logger.warn('unknown topic');
+    var channelParts = parseMqttTopic(packet.topic)
+    if (!channelParts) {
+        logger.warn('unknown or malformed topic');
         publish(4); // Bad username or password
         return;
     }
-    var channelId = channel[1],
+
+    var channelId = channelParts[1],
```
anovakovic01, line 122: What if `channelParts` has length 1.

mqtt/mqtt.js
```diff
@@ -149,13 +156,13 @@ aedes.authorizePublish = function (client, packet, publish) {
...
-    var channel = /^channels\/(.+?)\/messages\/?.*$/.exec(packet.topic);
-    if (!channel) {
+    var channelParts = parseMqttTopic(packet.topic)
+    if (!channelParts) {
       logger.warn('unknown topic');
       subscribe(4, packet); // Bad username or password
       return;
     }
-    var channelId = channel[1],
+    var channelId = channelParts[1],
```
anovakovic01, line 165: Same here, what if `channelParts` has length 1.

anovakovic01 (changes requested):

mqtt/mqtt.js
drasko: Why wouldn't it?

mqtt/mqtt.js
drasko: Good catch there!

drasko (changes requested):

mqtt/mqtt.js
anovakovic01: Sorry, didn't realized that there was a dot there. But still, I didn't know that you've already decided that you're going to pass subtopic here too. Maybe the rest of the team should be in sync with this change too.

drasko: Yep, subtopics will be enabled for all 4 adapters. @nmarcetic - remainder to open the issues.

anovakovic01: That's OK, but there is no spec how will this be done, that's why I don't know if this is how we are going to do it. So we are going to pass `<channel_id>.subtopic` in `channel` field? Why not adding separate field that will be called subtopic, and leave `channel` field as is? Then there will be no need to extract `channel_id` from string.

drasko: Yes, that would make code probably cleaner. But to see exactly what you propose, can you write here a small snippet example please.

anovakovic01: So, instead of using current `channel` field from here, I would add a new field: `subtopic`. So our structure would look like this:
```
message RawMessage {
	string channel     = 1;
        string subtopic = 2;
	string publisher   = 3;
	string protocol    = 4;
	string contentType = 5;
	bytes  payload     = 6;
}
```
In `channel`, we would store just channel ID, like we've done so far, and in `subtopic` we would store everything else (subtopic).

mqtt/mqtt.js
beres: @anovakovic01 afsik if the regexp match it will be always of length 3, first element contains the full string, second the channeid, and the third could be undefined or the subtopic part

beres:
What do you think to reduce  the valid characters in the subtopic to [a-zA-Z0-9_] and - like in channelId?
For example if the secondary argument of mqtt contains points we would have conflicts with nats 

But if we limit the characters as in the case of the channelId we should do two regexp one for publish and one for subscribe, this should also accept the asterisk character.

PS: at the moment the regexp does not accept * in the subtopic

mqtt/mqtt.js
anovakovic01: @beres Sorry, I missed this check later.

nmarcetic:
Duplicated #642

PR #567, opened 2019-02-01
MF-519 - Refine Message

### LINKED ISSUE #519: Refine Message
Update internal Message so that it can be used out-of-box or provide a workaround on a root level. Currently, Protobuf generated message has problems with `Value` fields representation due to the way `oneof` works in Protobuf. JSON tags are not generated for `Value` fields, so we need to make a workaround in order to marshal and unmarshal Message to/from JSON.

dborovcanin:
### What does this do?
This PR provides a workaround for marshaling and unmarshaling Mainflux Message to/from JSON.
 
This PR resolves #519.

There are no such changes.

### Notes
Methods are added to Message type so that it now implements JSON `Marshaler` and `Unmarshaler` interfaces.


message.proto
```diff
@@ -10,33 +10,33 @@ package mainflux;
 
 // RawMessage represents a message emitted by the Mainflux adapters layer.
 message RawMessage {
-	string Channel     = 1;
-	string Publisher   = 2;
-	string Protocol    = 3;
-	string ContentType = 4;
-	bytes  Payload     = 5;
+	string channel     = 1;
```
manuio, line 13: @dusanb94 just curious, why did you removed capital letters here? Somethign with proto?

message.proto
```diff
@@ -10,33 +10,33 @@ package mainflux;
...
+		double floatValue  = 6;
+		string stringValue = 7;
+		bool   boolValue   = 8;
+		string dataValue   = 9;
 	}
-	SumValue ValueSum    = 10;
-	double Time        = 11;
-	double UpdateTime  = 12;
-	string Link        = 13;
+	SumValue valueSum    = 10;
```
drasko: Fix alignment

drasko (changes requested):

message.proto
dborovcanin: @manuio There are 2 major reasons, none of which is too much important:
1) There is no need for these fields to be uppercase.
2) Generated JSON tags will be lowercase this way.

Apart from these cosmetic changes, everything will work the same as before.

drasko (dismissed):
LGTM

manuio (approved):
LGTM

anovakovic01 (approved):
LGTM!

PR #573, opened 2019-02-04
MF-560 - Normalizer is not passing messages that cannot be normalized

anovakovic01:
Fix non-SenML message routing in `normalizer` service.


manuio (approved):
LGTM

nmarcetic (approved):

PR #599, opened 2019-02-14
NOISSUE -update message module

blokovi:
Sending message as selected thing to multiple channels at once

### What does this do?

Update message module to send message as selected thing to multiple channels at once

### DIFF
--- a/ui/src/Channel.elm
+++ b/ui/src/Channel.elm
@@ -53,6 +53,8 @@ type Msg
     = SubmitName String
     | SubmitOffset String
     | SubmitLimit String
+    | SubmitOffsetForThing String String
+    | SubmitLimitForThing String String
     | ProvisionChannel
     | ProvisionedChannel (Result Http.Error Int)
     | RetrieveChannels
@@ -73,6 +75,12 @@ update msg model token =
 
         SubmitLimit limit ->
             updateChannelList { model | limit = limit } token
+
+        SubmitOffsetForThing thingid offset ->
+            updateChannelListForThing { model | offset = offset } token thingid
+
+        SubmitLimitForThing thingid limit ->
+            updateChannelListForThing { model | limit = limit } token thingid
 
         ProvisionChannel ->
             ( { model | name = "" }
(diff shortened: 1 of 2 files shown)


drasko (approved):
LGTM

ui/src/Channel.elm
```diff
@@ -298,3 +305,15 @@ updateChannelList model token =
         )
         token
     )
+
+
+updateChannelListForThing : Model -> String -> String -> ( Model, Cmd Msg )
```
darkodraskovic, line 310: Wondering whether we can merge `updateChannelList` and `updateChannelListForThing` into one func that receives something like `urlPath` arg instead of `(url.thingsPath ++ [ thingid ] ++ url.channelsPath)` or of `url.channelsPath` and, consequently, remove unnecessary code duplication here

```
    | SubmitOffset String
    | SubmitLimit String
    | SubmitOffsetForThing String String
    | SubmitLimitForThing String String
```

It's just a suggestion and probably will do a PR with this change, but let's keep it like you did it for now. So, for now, this solution LGTM.

ui/src/Message.elm
```diff
@@ -101,33 +109,21 @@ update msg model token =
...
             ( { model | channels = updatedChannel }, Cmd.map ChannelMsg channelCmd )
 
-        SelectedThing thingid ->
+        SelectedThing thingid thingkey ->
             let
                 ( updatedChannel, channelCmd ) =
                     Channel.update (Channel.RetrieveChannelsForThing thingid) model.channels token
             in
-            ( { model | thingid = thingid, channels = updatedChannel }, Cmd.map ChannelMsg channelCmd )
+            ( { model | thingid = thingid, channels = updatedChannel, token = thingkey }, Cmd.map ChannelMsg channelCmd )
```
darkodraskovic: Stick to the terminology, i.e. be consistent. If you use `thingid = thingid` in the model, please use also `token = token` or, better `thingkey = thingkey`. Change the model to hold `thingkey` instead of `token` (`token` is ambivalent and can refer to user token).

blokovi: I agree. 
Mainflux docs itself is not consistent in naming this key/token, that's why it's mixed up here. It's referred to as `thing_key` and `thing_token` here: https://mainflux.readthedocs.io/en/latest/messaging/

PR #611, opened 2019-02-22
MF-571 Add pagination to Connection and Message

darkodraskovic:
Add a basic version of pagination to all modules and submodules.


blokovi (approved):
LGTM

PR #615, opened 2019-02-27
NOISSUE - Add icons, drop down user menu

blokovi:
### What does this do?
Add font awesome icons in main menu
Add drop down on user/logout menu


drasko:
@blokovi plrese do not send PR on master

PR #620, opened 2019-03-01
MF-571 - UI Thing module with modals

darkodraskovic:
Add complete thing manipulation via modals.

### DIFF
--- a/ui/src/Thing.elm
+++ b/ui/src/Thing.elm
@@ -2,7 +2,6 @@ module Thing exposing (Model, Msg(..), Thing, initial, subscriptions, update, vi
 
 import Bootstrap.Button as Button
 import Bootstrap.ButtonGroup as ButtonGroup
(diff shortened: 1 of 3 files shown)


blokovi (approved):
Great work @darkodraskovic 
LGTM

ui/src/Thing.elm
```diff
@@ -11,9 +13,9 @@ import Bootstrap.Modal as Modal
 import Bootstrap.Table as Table
 import Bootstrap.Utilities.Spacing as Spacing
 import Debug exposing (log)
-import Dict exposing (Dict)
+import Dict as Dict
```
anovakovic01: @darkodraskovic Why don't you just use `import Dict`? It will have the same effect, right?

darkodraskovic: Fixed in the last commit.

drasko (approved):
LGTM

PR #623, opened 2019-03-02
MF-571 - Fix user login and refactor User

darkodraskovic:
Further modularize code and reuse modules. Fix bugs, like logout button not working.


ui/src/HttpMF.elm
```diff
@@ -8,6 +8,15 @@ import Json.Encode as E
 import Url.Builder as B
 
 
+url =
+    { base = "http://localhost"
```
drasko, line 12: Take care that this will have to be configurable through ENV variable to work in Docker, I think.

drasko (approved):
LGTM

ui/src/HttpMF.elm
darkodraskovic: Ok. Sure. We'll add a variable. We'll probably need some bash script to run the app and to set env vars or something similar.

blokovi (approved):
LGTM

PR #633, opened 2019-03-05
MF-579 Things & Channels returns 404 when not found or ID is malformed, not 500

### LINKED ISSUE #579: Invalid ID format results in Internal Server Error
**BUG REPORT**

I'm trying to fetch a Thing by sending an ID in the wrong format.

I expect to receive 404 or 400. I'd prefer 404 since the Thing with the given ID does not exist. 

I get 500, Internal Server Error.

    - Create the user and log in.
    - Try to fetch a Thing (or Channel) by sending an invalid formatted ID (not in UUID format).

I run Mainflux version 0.7.0 on my local machine.

ghost:
Signed-off-by: Vladimir Starcevic <vstarcevic@gmail.com>

### What does this do?
Retrieving things or channels by Id, returns 404 when not found or id is malformed, not 500
Resolves #579
NA
Yes
NA

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ build
 
 site/
 
+# editor specific
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@ build
 
 site/
 
-# editor specific
-.vscode
--- a/things/postgres/channels_test.go
+++ b/things/postgres/channels_test.go
@@ -157,6 +157,11 @@ func TestSingleChannelRetrieval(t *testing.T) {
 			ID:    c.ID,
 			err:   things.ErrNotFound,
 		},
+		"retrieve thing with malformed ID": {
+			owner: c.Owner,
+			ID:    wrongValue,
+			err:   things.ErrNotFound,
+		},
 	}
 
 	for desc, tc := range cases {
--- a/things/postgres/channels_test.go
+++ b/things/postgres/channels_test.go
@@ -157,7 +157,7 @@ func TestSingleChannelRetrieval(t *testing.T) {
 			ID:    c.ID,
 			err:   things.ErrNotFound,
 		},
-		"retrieve thing with malformed ID": {
+		"retrieve channel with malformed ID": {
 			owner: c.Owner,
 			ID:    wrongValue,
 			err:   things.ErrNotFound,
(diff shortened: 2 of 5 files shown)


manuio:
@vstarcevic this is the expected behavior. `404` is `not found ` `500` is I`nternal server error`

.gitignore
```diff
@@ -2,3 +2,5 @@ build
 
 site/
 
+# editor specific
+.vscode
```
anovakovic01: You should add this to your local `.gitignore`. I think that we don't need it on remote repository.

things/postgres/things.go
```diff
@@ -89,10 +89,7 @@ func (tr thingRepository) RetrieveByID(owner, id string) (things.Thing, error) {
 
 	if err != nil {
 		empty := things.Thing{}
-		if err == sql.ErrNoRows {
-			return empty, things.ErrNotFound
-		}
-		return empty, err
+		return empty, things.ErrNotFound
```
anovakovic01: If we do this, we will ignore every other error (like connection errors). I think that this removed `if` should stay. Maybe problem isn't in this part of code.

anovakovic01 (changes requested):

manuio (dismissed):
LGTM

things/postgres/things.go
dborovcanin: @vstarcevic This way, all read errors will cause the same NotFound error, which is not what we want. I'd prefer to extract error info from SQL error and check if it's invalid ID format error. Something similar to this: https://github.com/mainflux/mainflux/blob/33008140268cc3d9f5edd3394dce8cbc37f31956/things/postgres/things.go#L64

This is far from perfect because in this case, you extract semantics from the error message string, but it's better than shadowing all the errors with NotFound. Also, there is a way to check if the string is a valid UUID without trying to access a database, which is a more efficient way. Maybe adding something like `Validate()` method to IdentityProvider interface (https://github.com/mainflux/mainflux/blob/master/things/uuid/idp.go)...

dborovcanin (changes requested):

manuio (dismissed):

things/postgres/channels_test.go
```diff
@@ -157,6 +157,11 @@ func TestSingleChannelRetrieval(t *testing.T) {
 			ID:    c.ID,
 			err:   things.ErrNotFound,
 		},
+		"retrieve thing with malformed ID": {
```
dborovcanin: Channel, not Thing.

ghost: Fixed!

manuio (approved):

anovakovic01 (approved):
LGTM!

dborovcanin (approved):

PR #644, opened 2019-03-12
MF-639 Split Content-Type header field on semicolon and evaluate all substrings

### LINKED ISSUE #639: error 415 when posting to /tokens with content-type appended charset=utf-8
Sending the following POST request will generate a 415 Unsupported Media Type error and fail

**Steps:**
POST request:
POST http://URL/tokens
accept: application/json
content-type: application/json;charset=utf-8

{"email":"tony@lucidtron.com","password":"123"}

**Expected results:**
HTTP/1.1 201 Created

**Received results:**
HTTP/1.1 415 Unsupported Media Type

**workaround**
do not append charset=utf-8 to content-type:

POST request:
POST http://URL/tokens
accept: application/json
content-type: application/json

{"email":"tony@lucidtron.com","password":"123"}

tony-ward:
Signed-off-by: Tony Ward <tony@lucidtron.com>

Resolves #639 

- update bootstrap, things and user API to support multiple entry comments in Content-Type of HTTP header

No


bootstrap/api/transport.go
```diff
@@ -105,10 +105,17 @@ func MakeHandler(svc bootstrap.Service, reader bootstrap.ConfigReader) http.Hand
 }
 
 func decodeAddRequest(_ context.Context, r *http.Request) (interface{}, error) {
-	if r.Header.Get("Content-Type") != contentType {
-		return nil, errUnsupportedContentType
+	s := r.Header.Get("Content-Type")
```
chombium: Hi @tony-ward what @anovakovic01 meant with [this comment](https://github.com/mainflux/mainflux/issues/639#issuecomment-471440902) was to change the original line:
```
if r.Header.Get("Content-Type") != contentType {
```
with:
```
if !strings.Contains(r.Header.Get("Content-Type"), contentType) {
```

If we do it as in your implementation, to avoid code duplication, it would be better if we make a separate package which will parse the content type header. IMHO, at the moment, there is no need for that as we only work with utf-8 encoded strings and don't care about other character encodings.
If in the future we decide that we will support different encodings we can parse the header and possibly convert the strings to utf-8.

of course: run the test locally with `make test` as well, before committing to make sure that you haven't broken something ;)

chombium (changes requested):
@tony-ward Thanks for you effort.
Please take care about the code formatting use gofmt to format the code. The best way is to configure that gofmt runs automatically on save in your editor ;)

bootstrap/api/transport.go
dborovcanin: Great feedback, @chombium. I agree with you. @tony-ward There is no need for this much code, the simple snippet that @chombium suggested is good enough.

tony-ward (commented):
> @tony-ward Thanks for you effort.
> Please take care about the code formatting use gofmt to format the code. The best way is to configure that gofmt runs automatically on save in your editor ;)

done!

tony-ward:
@chombium there are many instances of 
`if r.Header.Get("Content-Type") != contentType {`

Should we change them all to 
`!strings.Containts()`
?

drasko:
Are we doing this only for HTTP and not for other adapters?

Or do you want to do changes for other adapters in subsequent PR?

tony-ward:
> Are we doing this only for HTTP and not for other adapters?
> 
> Or do you want to do changes for other adapters in subsequent PR?

I do not see this Content-Type in non http services

anovakovic01 (dismissed):
LGTM!

chombium:
@tony-ward for consistency it would be better if every check is adjusted. I agree with @drasko that we should do this change for the other adapters as well. I personally think subsequent PRs are not an option because we'll have inconsistent content-type processing until all of them get merged.

chombium (dismissed):

things/api/http/transport.go
```diff
@@ -149,7 +150,7 @@ func MakeHandler(svc things.Service) http.Handler {
 }
 
 func decodeThingCreation(_ context.Context, r *http.Request) (interface{}, error) {
-	if r.Header.Get("Content-Type") != contentType {
```
dborovcanin, line 152: The same is applicable to `decodeChannelCreation` and `decodeCahnnelUpdate` methods.

bootstrap/api/transport.go
```diff
@@ -105,7 +105,7 @@ func MakeHandler(svc bootstrap.Service, reader bootstrap.ConfigReader) http.Hand
 }
 
 func decodeAddRequest(_ context.Context, r *http.Request) (interface{}, error) {
-	if r.Header.Get("Content-Type") != contentType {
+	if !strings.Contains(r.Header.Get("Content-Type"), contentType) {
```
dborovcanin, line 108: The same is applicable to other places in `transport.go` where `r.Header.Get("Content-Type") != contentType` occurs.

dborovcanin (changes requested):

tony-ward:
@dusanb94 I have now changed a further 5 occurrences

dborovcanin (approved):

dborovcanin (commented):
@tony-ward Please update the branch.

drasko (approved):
LGTM

drasko:
@tony-ward please update the branch if you want this PR in the 0.8 release which is planned for Monday.

tony-ward:
@drasko this is now ready to go.

PR #647, opened 2019-03-13
NOISSUE - Prevent infinite loop in lora-adapter if Redis init fail

manuio:

### DIFF
--- a/cmd/lora/main.go
+++ b/cmd/lora/main.go
@@ -203,7 +203,9 @@ func subscribeToLoRaBroker(svc lora.Service, mc mqtt.Client, logger logger.Logge
 func subscribeToThingsES(svc lora.Service, client *r.Client, consumer string, logger logger.Logger) {
 	eventStore := redis.NewEventStore(svc, client, consumer, logger)
 	logger.Info("Subscribed to Redis Event Store")
-	eventStore.Subscribe("mainflux.things")
+	if err := eventStore.Subscribe("mainflux.things"); err != nil {
+		logger.Warn(fmt.Sprintf("Lora-adapter service failed to subscribe to event sourcing: %s", err))
+	}
 }
 
 func newRouteMapRepositoy(client *r.Client, prefix string, logger logger.Logger) lora.RouteMapRepository {
--- a/lora/redis/streams.go
+++ b/lora/redis/streams.go
@@ -71,7 +71,8 @@ func NewEventStore(svc lora.Service, client *redis.Client, consumer string, log 
 }
 
 func (es eventStore) Subscribe(subject string) error {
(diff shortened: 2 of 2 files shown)


dborovcanin (dismissed):

cmd/lora/main.go
```diff
@@ -203,7 +203,9 @@ func subscribeToLoRaBroker(svc lora.Service, mc mqtt.Client, logger logger.Logge
 func subscribeToThingsES(svc lora.Service, client *r.Client, consumer string, logger logger.Logger) {
 	eventStore := redis.NewEventStore(svc, client, consumer, logger)
 	logger.Info("Subscribed to Redis Event Store")
-	eventStore.Subscribe("mainflux.things")
+	if err := eventStore.Subscribe("mainflux.things"); err != nil {
+		logger.Warn(fmt.Sprintf("Botstrap service failed to subscribe to event sourcing: %s", err))
```
anovakovic01: Not Botstrap service, but LoRa adapter.

lora/redis/streams.go
```diff
@@ -68,8 +70,11 @@ func NewEventStore(svc lora.Service, client *redis.Client, consumer string, log
 	}
 }
 
-func (es eventStore) Subscribe(subject string) {
-	es.client.XGroupCreateMkStream(stream, group, "$").Err()
+func (es eventStore) Subscribe(subject string) error {
+	if err := es.client.XGroupCreateMkStream(stream, group, "$").Err(); err != nil && err.Error() != exists {
+		return err
+	}
```
anovakovic01, line 77: You shouldn't return here. If consumer group already exists, just continue.

anovakovic01 (changes requested):

cmd/lora/main.go
manuio: done

lora/redis/streams.go
manuio: @dusanb94 what we do here? I thought that this is the fix

dborovcanin: You return only if there is an error other than `BUSYGROUP Consumer Group name already exists`. If there is no error or error is exactly this one, you will proceed with your subscription.

manuio: Right, this is correct @anovakovic01

anovakovic01: Didn't see that part. IMHO, this line is too long. Can you break it into multiple lines? Also, is there predefined error for this in redis lib?

manuio: I could not find any meaningful error. Line length fixed.

dborovcanin (approved):

anovakovic01 (approved):
LGTM!

PR #653, opened 2019-03-15
MF-636 - added docs about docker-compose config overriding

### LINKED ISSUE #636: Add example configuration about adding persistent storage for the things-db and users-db containers
Hi,

We had a small discussion with @bergernetch on our Gitter channel about adding persistent storage to the `things-db` and `users-db` and he sent me updated docker-compose configs with added volumes for these contaners plus influxdb and grafana (the influxdb writer).
The changes are practically addition of top level `volumes` definitions and usage of these volumes in the container definitions. The volumes are mapped to the default location host machine, something like `/var/lib/docker/volumes/project-name_volume-name`.

The relevant sections are:
[main docker-compose.yml](https://github.com/mainflux/mainflux/blob/master/docker/docker-compose.yml)
```
volumes:
  mainflux-users-db-volume:
  mainflux-things-db-volume:
...
  users-db:
    image: postgres:10.2-alpine
    container_name: mainflux-users-db
    restart: on-failure
    environment:
      POSTGRES_USER: mainflux
      POSTGRES_PASSWORD: mainflux
      POSTGRES_DB: users
    networks:
      - mainflux-base-net
    volumes:
      - mainflux-users-db-volume:/var/lib/postgresql/data
...
  things-db:
    image: postgres:10.2-alpine
    container_name: mainflux-things-db
    restart: on-failure
    environment:
      POSTGRES_USER: mainflux
      POSTGRES_PASSWORD: mainflux
      POSTGRES_DB: things
    networks:
      - mainflux-base-net
    volumes:
      - mainflux-things-db-volume:/var/lib/postgresql/data
```

and [influxdb-writer docker-compose.yml](https://github.com/mainflux/mainflux/blob/master/docker/addons/influxdb-writer/docker-compose.yml):
```
volumes:
  grafana-volume:
  influxdb-volume:
...
  influxdb:
    image: influxdb:1.6.4-alpine
    container_name: mainflux-influxdb
    restart: on-failure
    environment:
      INFLUXDB_DB: mainflux
      INFLUXDB_ADMIN_USER: mainflux
      INFLUXDB_ADMIN_PASSWORD: mainflux
    networks:
      - docker_mainflux-base-net
    ports:
      - 8086:8086
    volumes:
      - influxdb-volume:/var/lib/influxdb
...
  grafana:
    image: grafana/grafana:5.1.3
    container_name: mainflux-grafana
    depends_on:
      - influxdb
    restart: on-failure
    ports:
      - 3001:3000
    networks:
      - docker_mainflux-base-net
    volumes:
      - grafana-volume:/var/lib/grafana
    environment:
      - PUID=0
      - GUID=0
```

I know that the definitions of external storage/volumes depend on the deployment strategy and can not be defined by us, but I find it useful if we have an example how one can configure a persistent storage for the dev setup, because with every new release someone who's trying out Mainflux is not careful, he or she will lose his/hers data. We already have the warning about the data loss in our [dev guide](https://github.com/mainflux/mainflux/blame/master/docs/dev-guide.md#L68).

I was thinking to add these examples in the dev guide, for our docker setup, but then if someone's using Mainflux with Kubernetes might also find such example configuration useful. A while ago we've spoken with @drasko about some kind of operations guide, but I think that would be to much to start it for two examples.

@mainflux/contributors What do you think about the idea of having an example persistent volume configuration in the dev guide? We could also provide patch files for those who want to use the changes.

Best Regards,
Jovan
P.S. I've also attached the whole configuration files [mainflux-persistence.zip](https://github.com/mainflux/mainflux/files/2939033/mainflux-persistence.zip).

chombium:
### What does this do?
As there are too many questions and complaints about lost data when recreating the Mainflix infrastructure with docker-compose, I've added a small description how the default docker-compose configuration can be overridden and added an example for adding persistent volumes to the users-db, things-db, influxdb and grafana containers.

The two changes are:
* added description about overriding the default docker-compose configuration in the dev guide.
* added the docker-compose.*.yml files in the docker folder and it's sub-folders to the gitignoes

Resolves #636 

None. I've tested the docker composition locally.

No.

Yes. I've added a new section in the dev guide. 

### Notes
Kudos to @bergernetch for his initiative and initial solution and to @beres for his hint how to make the solution more elegant.

### DIFF
--- a/docs/dev-guide.md
+++ b/docs/dev-guide.md
@@ -65,7 +65,7 @@ make docker_http
 
 > N.B. Mainflux creates `FROM scratch` docker containers which are compact and small in size.
 
-> N.B. The `things-db` and `users-db` containers are built from a vanilla PostgreSQL docker image downloaded from docker hub which does not persist the data when these containers are rebuilt. Thus, __rebuilding of all docker containers with `make dockers` or rebuilding the `things-db` and `users-db` containers separately with `make docker_things-db` and `make docker_users-db` respectively, will cause data loss. All your users, things, channels and connections between them will be lost!__ As we use this setup only for development, we don't guarantee any permanent data persistence. If you need to retain the data between the container rebuilds you can attach volume to the `things-db` and `users-db` containers. Check the official docs on how to use volumes [here](https://docs.docker.com/storage/volumes/) and [here](https://docs.docker.com/compose/compose-file/#volumes).
+> N.B. The `things-db` and `users-db` containers are built from a vanilla PostgreSQL docker image downloaded from docker hub which does not persist the data when these containers are rebuilt. Thus, __rebuilding of all docker containers with `make dockers` or rebuilding the `things-db` and `users-db` containers separately with `make docker_things-db` and `make docker_users-db` respectively, will cause data loss. All your users, things, channels and connections between them will be lost!__ As we use this setup only for development, we don't guarantee any permanent data persistence. If you need to retain the data between the container rebuilds you can attach volume to the `things-db` and `users-db` containers. Check the official docs on how to use volumes [here](https://docs.docker.com/storage/volumes/) and [here](https://docs.docker.com/compose/compose-file/#volumes). For examples on how to add persistent volumes check the [Overriding the default docker-compose configuration](#overriding-the-default-docker-compose-configuration) section.
 
 #### Building Docker images for development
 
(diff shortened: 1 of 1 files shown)


manuio (dismissed):

.gitignore
```diff
@@ -1,3 +1,4 @@
 build
 
 site/
+docker/**/docker-compose.*.yml
```
drasko: Why `docker/**/...` and not `docker/*/...` - is this some kind of special syntax of `.gitignore` to have two asterisk?

drasko: Also - IMHO it is confusing to see something like this in .gitignore. I understand why it is here because I read the doc, but just by inspecting the code it would be hard to figure out.

Do we have to put this into ignore, even if we do not currently have files like this in the repo?

I would prefer that we don't, and we document to the piple who create `docker/docker-compose.customX.yml` files how to put them into the their `.gitignore` themselves

drasko (changes requested):

chombium:
@drasko I was also thinking if I should add the files to gitignore or not, and I did so to protect our docker configuration. I completely agree with you with you, so I removed them from the gitignore

drasko (dismissed):
LGTM

nmarcetic (dismissed):

manuio (dismissed):

drasko (approved):
LGTM

manuio (approved):

PR #654, opened 2019-03-15
MF-571 - Add Env.elm to set custom base URL

darkodraskovic:
Add Env.elm so we can set custom base URL and develop locally.

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -11,34 +11,16 @@ then
 
 `docker-compose -f docker/docker-compose.yml up`
 
-if you want to launch a whole Mainflux docker composition or just
+if you want to launch a whole Mainflux docker composition, or just
 
 `docker-compose -f docker/docker-compose.yml up ui`
 
 if you want to launch just GUI.
 
-
 ### Native GUI build
 
-Install Elm (https://guide.elm-lang.org/install.html) and then edit
-*src/Env.elm* by replacing
-
-```
-env =
-    { url = ""
-    }
-```
-
-with
-
-```
-env =
-    { url = "http://localhost:80/"
-    }
-```
-
-So you can send requests with absolute URLs. Dont forget the trailing slash.
-Then run the following commands,
+Install Elm (https://guide.elm-lang.org/install.html) and then run the following
+commands:
 
 ```
 git clone https://github.com/mainflux/mainflux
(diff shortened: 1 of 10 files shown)


ui/README.md
```diff
@@ -3,58 +3,69 @@ Dashboard made with [elm-bootstrap](http://elm-bootstrap.info/).
...
+### Native GUI build
 
-### About Elm `make`
+Install Elm (https://guide.elm-lang.org/install.html) and then edit
+*src/Env.elm* by replacing
 
-`make` does `elm make src/Main.elm`.
+```
+env =
+    { protocol = ""
```
drasko: There should be some default values IMHO.

ui/src/Env.elm
```diff
@@ -0,0 +1,14 @@
...
+-- Mainflux
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+
+module Env exposing (env)
+
+
+env =
+    { protocol = ""
```
drasko: Put some default values here if needed.

I let you decide.

drasko (changes requested):

nmarcetic (dismissed):

blokovi (dismissed):
LGTM

ui/README.md
darkodraskovic: No default value needed, cause we need it to work out of the box in the docker composition. We only need to input some value if it's natively deployed or user wants a special setting.

ui/src/Env.elm
```diff
@@ -0,0 +1,12 @@
...
+-- Mainflux
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+
+module Env exposing (env)
+
+
+env =
+    { url = ""
```
drasko: Please use default value. If the user did not set ENV var, it should fallback to something like `http://localhost:3000`, right?

drasko (changes requested):

ui/README.md
drasko: Why can not it work natively with default value, and be controlled in the Docker via ENV variable?

ui/src/Env.elm
drasko: Probably again - defult should be left empty, because of Docker deployment?

manuio (dismissed):

ui/README.md
```diff
@@ -3,58 +3,52 @@ Dashboard made with [elm-bootstrap](http://elm-bootstrap.info/).
...
+`make run`
 
-Install Docker (https://docs.docker.com/install/) and Docker compose
-(https://docs.docker.com/compose/install/), `cd` to Mainflux root directory and
-then
+and follow the instructions on screen.
 
-`docker-compose -f docker/docker-compose.yml up`
+**NB:** `make` does `elm make src/Main.elm` and `make run` executes `elm
+reactor`. Cf. _Makefile_ for more options.
```
blokovi, line 39: Meaning of Cf. abbreviation is not so common knowlegde

blokovi (dismissed):
LGTM

ui/README.md
darkodraskovic: Let the developers learn a bit of humanities language :)

drasko (approved):
LGTM

manuio (approved):

PR #657, opened 2019-03-18
MF-655 Proper usage of docker volumes

### LINKED ISSUE #655: Correct usage of docker persistent volumes
As I was working on #636 I've noticed that some volumes are generated when executing `docker-compose`.

I've dug a bit deeper and I've found out that we're already using volumes for the users-db, things-db, mainflux-redis, mainflux-things-redis, bootstrap, influxdb-writer, cassandra-writer, mongodb-writer containers, but we delete them with `make cleandocker` and `make cleanghost`.

Here is the test procedure:

1. start from clear state

```
$ make cleandocker

$ docker volume ls
DRIVER              VOLUME NAME
```

2. start Mainflux
```
$	docker-compose -f docker/docker-compose.yml up -d (equivalent to make run)
```
3. the state with the volumes afterwards:
```
$ docker volume ls
DRIVER              VOLUME NAME
local               18f49a276e35cc4ad115ecefd638a1b594538b7abe117d4abb2ee9ce60700231
local               264a2c95f31d7d51291ef9cca2294d017d51c5961e421a2e35797ee070505172
local               c2442c40d8dca66ae9f96ff5c438417d6a12f19f56b3f03501d3b3a129b41c3d
local               d5886127963d2728dd2d3e7b31c4f4ad66ffa135550c2fce633ab0f342177832
```

4. find the containers to which these volumes belong to: `docker ps --filter=volume=<VOLUME_NAME>`
```
$ docker ps --filter=volume=18f49a276e35cc4ad115ecefd638a1b594538b7abe117d4abb2ee9ce60700231 --filter=volume=264a2c95f31d7d51291ef9cca2294d017d51c5961e421a2e35797ee070505172 --filter=volume=c2442c40d8dca66ae9f96ff5c438417d6a12f19f56b3f03501d3b3a129b41c3d --filter=volume=d5886127963d2728dd2d3e7b31c4f4ad66ffa135550c2fce633ab0f342177832
CONTAINER ID        IMAGE                  COMMAND                  CREATED             STATUS              PORTS               NAMES
0e0dd71e5174        postgres:10.2-alpine   "docker-entrypoint.s…"   19 minutes ago      Up 19 minutes       5432/tcp            mainflux-things-db
039b13ffd155        redis:4.0.9-alpine     "docker-entrypoint.s…"   19 minutes ago      Up 19 minutes       6379/tcp            mainflux-redis
2e4188a46327        postgres:10.2-alpine   "docker-entrypoint.s…"   19 minutes ago      Up 19 minutes       5432/tcp            mainflux-users-db
bdde4c8e5f25        redis:5.0-alpine       "docker-entrypoint.s…"   19 minutes ago      Up 19 minutes       6379/tcp            mainflux-things-redis
```

5. The existence of the volumes can be easily confirmed by checking the Dockerfiles from which the docker hub images are build:

* Postgres 10.2-alpine  [Dockerfile](https://github.com/docker-library/postgres/blob/master/10/alpine/Dockerfile#L145)
* Redis 5.0-alpine [Dockerfile](https://github.com/docker-library/redis/blob/60db7082c81f5d457046f199244c5d651f04b3fa/5.0/alpine/Dockerfile#L60)
* Redis 4.0.9-alpine  [Dockerfile](https://github.com/docker-library/redis/blob/8e1acc18156603cc3643bf22ced6069cf8be1622/4.0/alpine/Dockerfile#L60)
* Cassandra [Dockerfile](https://github.com/docker-library/cassandra/blob/064fb4e2682bf9c1909e4cb27225fa74862c9086/3.11/Dockerfile#L156)
* InfluxDB [Dockerfile](https://github.com/influxdata/influxdata-docker/blob/d76a6a184dec6a82978964882641a743adb76989/influxdb/1.6/alpine/Dockerfile#L31)
* MongoDB [Dockerfile](https://github.com/docker-library/mongo/blob/7276b14c1e9b5f83b37f5f452e6a9f293aea2d8b/3.6/Dockerfile#L90)
* Grafana [recommended setup](http://docs.grafana.org/installation/docker/#grafana-container-with-persistent-storage-recommended)

6. cleaning up the docker containers
```
make cleandocker
```
this make task calls the `cleanghost` task which contains a `docker volume ls -f dangling=true -q | xargs -r docker volume rm` commands which deletes all the dangling volumes. __The problem is that it deletes all__ dangling volumes, volumes not connected to any container. IMHO this is totally wrong. We should only delete the containers which we have created and not all of the dangling containers which means we might delete some oder important data from other containers. The __second problem__ is that the deletion of the dangling containers happens before the Mainflux containers are stopped and removed so we leave some "garbage" behind. If a new composition is executed there will be 8 volumes which are created by us. The volumes from ther previous composition are deleted on the second run of `cleandocker` or `cleanghost`
In fact, the whole `cleanghost` target does not check if the containers, images or volumes are created by us or not, but it simply deletes them all.

__The solution__ for this problem would be to use names for the volumes so that we can filter them easier, clearly show which volumes were created by us and delete only these volumes and add filters to all of the `cleanghost` commands. We should also add a switch which will tell if the volumes should be deleted or not. The default behavior should be the same as now, when the switch is not present the `cleandocker` and `cleanghost` make task should delete the volumes.

Also, having proper names for the volumes makes it straightforward to backup and restore them without too much hassle and playing around with to many docker commands ;)

Although this is more an enhancement, I've marked it as as bug as well, as we delete all exited contaniers and unused images instead of deleting only the things that we have created

chombium:
### What does this do?

* adds names for the persistent volumes that we already use them
* adds filters for the `cleanghost` to filter out only the containers which we have created
* adds a switch which when is present the persistent volumes will not be deleted. The default behavior or volume deletion is not changed.

Resolves #655 

none

no

added description about the changes in the dev guide

### Notes
See the issue for more details

### DIFF
--- a/docs/dev-guide.md
+++ b/docs/dev-guide.md
@@ -65,7 +65,7 @@ make docker_http
 
 > N.B. Mainflux creates `FROM scratch` docker containers which are compact and small in size.
 
-> N.B. The `things-db` and `users-db` containers are built from a vanilla PostgreSQL docker image downloaded from docker hub which does not persist the data when these containers are rebuilt. Thus, __rebuilding of all docker containers with `make dockers` or rebuilding the `things-db` and `users-db` containers separately with `make docker_things-db` and `make docker_users-db` respectively, will cause data loss. All your users, things, channels and connections between them will be lost!__ As we use this setup only for development, we don't guarantee any permanent data persistence. If you need to retain the data between the container rebuilds you can attach volume to the `things-db` and `users-db` containers. Check the official docs on how to use volumes [here](https://docs.docker.com/storage/volumes/) and [here](https://docs.docker.com/compose/compose-file/#volumes). For examples on how to add persistent volumes check the [Overriding the default docker-compose configuration](#overriding-the-default-docker-compose-configuration) section.
+> N.B. The `things-db` and `users-db` containers are built from a vanilla PostgreSQL docker image downloaded from docker hub which does not persist the data when these containers are rebuilt. Thus, __rebuilding of all docker containers with `make dockers` or rebuilding the `things-db` and `users-db` containers separately with `make docker_things-db` and `make docker_users-db` respectively, will cause data loss. All your users, things, channels and connections between them will be lost!__ As we use this setup only for development, we don't guarantee any permanent data persistence. If you want to update your Mainflux dockerized installation and want to keep your data, use `make spv=true cleandocker` to clean the containers and images and keep the data and then `make run` to update the images and the containers. Check the [Cleaning up your dockerized Mainflux setup](#cleaning-up-your-dockerized-mainflux-setup) section for details. Please note that this kind of updating might not work if there are database changes.
 
 #### Building Docker images for development
 
@@ -85,56 +85,34 @@ Commands `make dockers` and `make dockers_dev` are similar. The main difference 
 
 ### Overriding the default docker-compose configuration
 Sometimes, depending on the use case and the user's needs it might be useful to override or add some extra parameters to the docker-compose configuration. These configuration changes can be done by specifying multiple compose files with the [docker-compose command line option -f](https://docs.docker.com/compose/reference/overview/) as described [here](https://docs.docker.com/compose/extends/).
-The following format of the `docker-compose` can be used to extend or override the configuration:
+The following format of the `docker-compose` command can be used to extend or override the configuration:
 ```
 docker-compose -f docker/docker-compose.yml -f docker/docker-compose.custom1.yml -f docker/docker-compose.custom2.yml up [-d]
 ```
 In the command above each successive file overrides the previous parameters.
 
-A practical example in our case would be to add persistent volumes to the users-db, things-db, influxdb and grafana containers so that we don't loose data between updates of Mainflux. The volumes are mapped to the default location on the host machine, something like `/var/lib/docker/volumes/project-name_volume-name`:
+A practical example in our case would be to enable debugging and tracing in NATS so that we can see better how are the messages moving around.
 
+`docker-compose.nats-debugging.yml`
 ```yaml
-# docker/docker-compose.persistence.yml
 version: "3"
 
-volumes:
-  mainflux-users-db-volume:
-  mainflux-things-db-volume:
-
 services:
-  users-db:
-    volumes:
-      - mainflux-users-db-volume:/var/lib/postgresql/data
-
-  things-db:
-    volumes:
-      - mainflux-things-db-volume:/var/lib/postgresql/data
+  nats:
+    command: --debug -DV
 ```
 
-```yaml
-# docker/addons/influxdb-writer/docker-compose.persistence.yml
-version: "3"
-
-volumes:
-  influxdb-volume:
-  grafana-volume:
-
-services:
-  influxdb:
-    volumes:
-      - influxdb-volume:/var/lib/influxdb
-
-  grafana:
-    volumes:
-      - grafana-volume:/var/lib/grafana
-```
 When we have the override files in place, to compose the whole infrastructure including the persistent volumes we can execute:
 ```
-docker-compose -f docker/docker-compose.yml -f docker/docker-compose.persistence.yml up -d
-docker-compose -f docker/addons/influxdb-writer/docker-compose.yml -f docker/addons/influxdb-writer/docker-compose.persistence.yml up -d
+docker-compose -f docker/docker-compose.yml -f docker/docker-compose.nats-debugging.yml up -d
 ```
 
-__Note:__ Please store your customizations to some folder outside the Mainflux's source folder and maybe add them to some other git repository. You can always apply your customizations by pointing to the right file using `docker-compose -f ...`. Also be sure not to use the `make cleandocker` and `make cleanghost` as they might delete something which you don't want to delete (i.e. the newly created persistent volumes).
+__Note:__ Please store your customizations to some folder outside the Mainflux's source folder and maybe add them to some other git repository. You can always apply your customizations by pointing to the right file using `docker-compose -f ...`.
+
+
+### Cleaning up your dockerized Mainflux setup
+If you want to clean your whole dockerized Mainflux installation you can use the `make cleandocker` command. Please note that __by default the `make cleandocker` command will stop and delete all of the containers, images and persistent volumes__. If you want to keep gathered data in the system (the persistent volumes) please use the following command `make spv=true cleandocker` (spv = skip persistent volumes). This form of the command will stop and delete the containers and images, but won't delete the persistent volumes.
+
 
 ### MQTT Microservice
 The MQTT Microservice in Mainflux is special, as it is currently the only microservice written in NodeJS. It is not compiled, but node modules need to be downloaded in order to start the service:
--- a/Makefile
+++ b/Makefile
@@ -30,23 +30,25 @@ clean:
 	rm -rf ${BUILD_DIR}
 	rm -rf mqtt/node_modules
 
-cleandocker: cleanghost
+cleandocker:
 	# Stop all containers (if running)
 	docker-compose -f docker/docker-compose.yml stop
 	# Remove mainflux containers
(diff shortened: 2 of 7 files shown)


chombium:
I'm still working on this. I've opened the pull request to get feedback.

The current state is:
```
$ docker volume ls
DRIVER              VOLUME NAME
local               cassandra-writer_mainflux-cassandra-volume
local               docker_mainflux-redis-volume
local               docker_mainflux-things-db-volume
local               docker_mainflux-things-redis-volume
local               docker_mainflux-users-db-volume
local               influxdb-writer_mainflux-grafana-volume
local               influxdb-writer_mainflux-influxdb-volume
local               mongodb-writer_mainflux-mongodb-configdb-volume
local               mongodb-writer_mainflux-mongodb-db-volume
```
the `make cleanghost` make task works only with "mainflux artifacts" and the delete switch works.
I still need to do more testing and eventually adjust the commands in `cleandocker` and `cleanghost` make tasks

chombium:
The fixes which this PR brings are:
* no more "ghost volumes"
* better filtering for the docker containers, images and volumes, so that we are sure that we are only working with the artifacts that we created
* optional data preservation (the persistent volumes are not deleted) when executing `make cleandocker`. The default behavior stays the same; no parameter, everything is deleted

The summary of the changes:
* added persistent volumes in the docker-compose.yml files for all of the components (we were already using persistent volumes anyways)
* removed the `make cleanghost` target as there are no more ghost volumes
* added switch to the `make cleandocker` so that the persistent volumes are not deleted. The command is `make spv=true cleandocker`.
* added documentation of the switch in the dev guide
* added filtering for all of the docker commands, so that we work only with our containers, images and volumes

nmarcetic:
@chombium Can you pls checkout conflicts and update branch. Thanks!

chombium:
@nmarcetic done ;)

docs/dev-guide.md
```diff
@@ -85,56 +85,34 @@ Commands `make dockers` and `make dockers_dev` are similar. The main difference
...
-docker-compose -f docker/docker-compose.yml -f docker/docker-compose.persistence.yml up -d
-docker-compose -f docker/addons/influxdb-writer/docker-compose.yml -f docker/addons/influxdb-writer/docker-compose.persistence.yml up -d
+docker-compose -f docker/docker-compose.yml -f docker/docker-compose.nats-debugging.yml up -d
 ```
 
-__Note:__ Please store your customizations to some folder outside the Mainflux's source folder and maybe add them to some other git repository. You can always apply your customizations by pointing to the right file using `docker-compose -f ...`. Also be sure not to use the `make cleandocker` and `make cleanghost` as they might delete something which you don't want to delete (i.e. the newly created persistent volumes).
+__Note:__ Please store your customizations to some folder outside the Mainflux's source folder and maybe add them to some other git repository. You can always apply your customizations by pointing to the right file using `docker-compose -f ...`.
+
+
+### Cleaning up your dockerized Mainlfux setup
```
anovakovic01: `Mainflux`

docs/dev-guide.md
```diff
@@ -85,56 +85,34 @@ Commands `make dockers` and `make dockers_dev` are similar. The main difference
...
+If you want to clean your whole dockerized Mainlfux installation you can use the `make cleandocker` command. Please note that __by default the `make cleandocker` command will stop and delete all of the containers, images and persistent volumes__. If you want to keep gathered data in the system (the persistent volumes) please use the following command `make spv=true cleandocker` (spv = skip persistent volumes). This form of the command will stop and delete the containers and images, but won't delete the persistent volumes.
```
anovakovic01: `Mainflux`

anovakovic01 (changes requested):
Looks good, just fix these two typos. Great job @chombium!

docs/dev-guide.md
```diff
@@ -65,7 +65,7 @@ make docker_http
...
+> N.B. The `things-db` and `users-db` containers are built from a vanilla PostgreSQL docker image downloaded from docker hub which does not persist the data when these containers are rebuilt. Thus, __rebuilding of all docker containers with `make dockers` or rebuilding the `things-db` and `users-db` containers separately with `make docker_things-db` and `make docker_users-db` respectively, will cause data loss. All your users, things, channels and connections between them will be lost!__ As we use this setup only for development, we don't guarantee any permanent data persistence. If you want to update your Mainflux dockerized installation and want to keep your data, use `make spv=true cleandocker` to clean the containers and images and keep the data and then `make run` to update the images and the containers. Check the [Cleaning up your dockerized Mainlfux setup](#cleaning-up-your-dockerized-mainlfux-setup) section for details. Please note that this kind of updating might not work if there are database changes.
```
anovakovic01: `Mainflux`

anovakovic01 (changes requested):

chombium:
@anovakovic01 really good catch. LOLed to myself. The same type in two consecutive sentences :)

docs/dev-guide.md
```diff
@@ -65,7 +65,7 @@ make docker_http
...
+> N.B. The `things-db` and `users-db` containers are built from a vanilla PostgreSQL docker image downloaded from docker hub which does not persist the data when these containers are rebuilt. Thus, __rebuilding of all docker containers with `make dockers` or rebuilding the `things-db` and `users-db` containers separately with `make docker_things-db` and `make docker_users-db` respectively, will cause data loss. All your users, things, channels and connections between them will be lost!__ As we use this setup only for development, we don't guarantee any permanent data persistence. If you want to update your Mainflux dockerized installation and want to keep your data, use `make spv=true cleandocker` to clean the containers and images and keep the data and then `make run` to update the images and the containers. Check the [Cleaning up your dockerized Mainflux setup](#cleaning-up-your-dockerized-mainlfux-setup) section for details. Please note that this kind of updating might not work if there are database changes.
```
anovakovic01: Just update the link `#cleaning-up-your-dockerized-mainflux-setup`.

chombium: thanks once again. I've missed that one :( it is fixed now

anovakovic01 (dismissed):
@chombium Don't worry, it happens. At least you were consistent. 😄 LGTM!

manuio (dismissed):

anovakovic01 (commented):
@chombium please rebase

chombium:
@anovakovic01 done

anovakovic01 (approved):
LGTM!

manuio (approved):

docs/dev-guide.md
```diff
@@ -85,56 +85,34 @@ Commands `make dockers` and `make dockers_dev` are similar. The main difference
...
+__Note:__ Please store your customizations to some folder outside the Mainflux's source folder and maybe add them to some other git repository. You can always apply your customizations by pointing to the right file using `docker-compose -f ...`.
+
+
+### Cleaning up your dockerized Mainflux setup
+If you want to clean your whole dockerized Mainflux installation you can use the `make cleandocker` command. Please note that __by default the `make cleandocker` command will stop and delete all of the containers, images and persistent volumes__. If you want to keep gathered data in the system (the persistent volumes) please use the following command `make spv=true cleandocker` (spv = skip persistent volumes). This form of the command will stop and delete the containers and images, but won't delete the persistent volumes.
```
drasko, line 114: I think it is better to skip persistent volume by default, but add explicitly extra parameter if you want to delete it also - like `pv=true`.

I think this can help avoinding errors and making idiot-proof solution. I see more a situations where I cleaned dockers, but I fortgot vlumes - so I run the command again with correct param, no big deal. However, the other way around, if I cleaned everything because I forgot a param - it is too late :).

I let you decide on this, but I would personaly do it like this.

chombium: @drasko I completely agree with you from data persistence view point. My idea was to to leave the current `make cleandocker` behavior as it is and offer an option to skip the deletetion of the persistent volumes.

On the second thought, taking in account all the complaints about data loss, I would by default skip the volume deletion.

I hope the possible running of the same command twice by the @mainflux/contributors  won't be a problem. 

I will prepare a new pr

PR #672, opened 2019-03-27
MF-661 - Bootstrap pagination in UI

### LINKED ISSUE #661: Fix pagination
Switch from using custom pagination function to Bootstrap.Pagination module:
https://package.elm-lang.org/packages/rundis/elm-bootstrap/5.1.0/Bootstrap-Pagination

blokovi:
### What does this do?
Switch to bootstrap style pagination

Resolves #661

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -46,7 +46,7 @@ Open the _src/Env.elm_ file and edit the values of the `env` record.
 
 Follow the instructions above to install and run GUI as a native build. In
 _src/Env.elm_ change a `url` field value of the `elm` record to
-`http://localhost:80`. Instead of `make run` you can install `elm-live`
+`http://localhost:80/`. Instead of `make run` you can install `elm-live`
 (https://github.com/wking-io/elm-live) and execute `elm-live src/Main.elm` to
 get a live reload when your `.Elm` files change.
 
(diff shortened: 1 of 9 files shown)


ui/README.md
```diff
@@ -46,7 +46,7 @@ Open the _src/Env.elm_ file and edit the values of the `env` record.
 
 Follow the instructions above to install and run GUI as a native build. In
 _src/Env.elm_ change a `url` field value of the `elm` record to
-`http://localhost:80`. Instead of `make run` you can install `elm-live`
+`http://localhost:80/`. Instead of `make run` you can install `elm-live`
```
manuio: why that?

manuio (changes requested):

ui/README.md
blokovi: UI doesn't work without trailing slash in this variable. I found this error in README while working on this issue but it's not related to pagination. Should I make seperate NOISSUE PR for this?

drasko: Nope, here is OK. Just that this must be noted somewhere.

manuio: ok, I resolve this one then

manuio (dismissed):

ui/README.md
drasko: Make a special note that trailing `/` is mandatory, as it is often source of errors.

> You can use markdown quote like this for this kind of notices.

drasko (changes requested):

dborovcanin (approved):

drasko (approved):
LGTM

PR #693, opened 2019-04-05
NOISSUE - added DOCKER_IMG_NAME var to makefile

beres:
### What does this do?

The patch could be useful if someone want to use custom image name

Added DOCKER_IMG_NAME to make file, default value is mainflux

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,8 @@ DOCKERS_DEV = $(addprefix docker_dev_,$(SERVICES))
 CGO_ENABLED ?= 0
 GOOS ?= linux
 
+DOCKER_IMG_NAME ?= mainflux
+
 define compile_service
 	CGO_ENABLED=$(CGO_ENABLED) GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) go build -ldflags "-s -w" -o ${BUILD_DIR}/mainflux-$(1) cmd/$(1)/main.go
 endef
@@ -34,9 +36,9 @@ cleandocker: cleanghost
 	# Stop all containers (if running)
 	docker-compose -f docker/docker-compose.yml stop
 	# Remove mainflux containers
-	docker ps -f name=mainflux -aq | xargs -r docker rm
+	docker ps -f name=$(DOCKER_IMG_NAME) -aq | xargs -r docker rm
 	# Remove old mainflux images
-	docker images -q mainflux\/* | xargs -r docker rmi
+	docker images -q $(DOCKER_IMG_NAME)\/* | xargs -r docker rmi
 
 # Clean ghost docker images
 cleanghost:
@@ -67,7 +69,7 @@ docker_ui:
 
 docker_mqtt:
 	# MQTT Docker build must be done from root dir because it copies .proto files
-	docker build --tag=mainflux/mqtt -f mqtt/Dockerfile .
+	docker build --tag=$(DOCKER_IMG_NAME)/mqtt -f mqtt/Dockerfile .
 
 dockers: $(DOCKERS) docker_ui docker_mqtt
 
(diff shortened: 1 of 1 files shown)


chombium:
Hi @beres,

could you please explain what kind of added value does this bring? We are using this Makefile for our dev setup and for pushing the created docker images to the docker hub. I would not  say that anyone of us (@mainflux/contributors) would upload docker images with another name.

I guess you want to pack and deliver your own Mainflux flavor. That is all cool and you are allowed to do that, Mainflux is published under Apache 2 license. By the Apache 2 license it is required that you state that your work is derivative of this project ;)

If you want to do your own customization, I would kindly ask you to do them in as separate repo without changing the Mainlfux directly. Now it's this PR about the name of the docker image, someone else might want to some special adaptation for his/hers Mainflux flavor in the future... Please don't get me wrong, it's great that you want to make your own customizations, that means that the project grows and gains adaptation, but I would like that keep our upstream project clean. Of course you can always open pull request with new features based on your experience in using Mainflux to make it better ;)

Now, one technical thing about the pull request: I see that you have two commits and the one is your experiment with `DOCKER_IMG_NAME ?= metav`. Could you please clean up this and add a single commit, so that if this thing get's merged, we'll have a cleaner commit history?

Happy hacking,
Jovan

Makefile
```diff
@@ -10,6 +10,8 @@ DOCKERS_DEV = $(addprefix docker_dev_,$(SERVICES))
 CGO_ENABLED ?= 0
 GOOS ?= linux
 
+DOCKER_IMG_NAME ?= metav
```
chombium: Please clean up this commit, I think we don't need your tests in our commit history ;)

chombium (changes requested):
Please clean up the commits, I don't like your

beres:
Hi @chombium 

It's not a way distribute  mainflux customizations, even if it could be a use case, by the way probably due to may docker knowledge, I give it another name simply to easily cleanup my test images without  touching mainflux ones

Makefile
beres: fixed sorry

drasko:
@beres as @chombium noted, I am not sure that providing this feature should be part of Mainflux. This is dev-ops related thing needed only for the people who fork Mainflux and what to re-package it.

On the other hand if it is for dev purposes, again I think everyone should keep some small part of their own dev tools around Mainflux, as there is no way to satisfy all use-cases.

I let others comment, but I am leaning towards refusing this PR, I do not think it has its place in the mainstram.

chombium:
@beres thank you very much for the clarification. IMHO the way that the deployments are handled or what are the development practices of each developer is completely personal thing. 
If you say that this is the way you easily handle your docker containers that's up to you and you could create a customized Makefile which suits your workflow.

All of us have different workflows and different dev setups, but that does not mean that each customization should be made available through the upstream project. One's personal settings might not suit the needs of the others.

I'm also against merging this.

beres:
Thank you @chombium I understand the point, don't worry to reject the pull :P

drasko:
@beres every line of code included in the mainstream is a potential bug, and we'd like to keep complexity at the minimum.

This looks like a dev-only tool. I suggest that we do not include it for now, until we see the real need from several members of community (we have been developing Mainlfux for years, and never had a need for such a tool so far, so it looks specific).

I will close the PR for now, we can re-open it later if needed.

dborovcanin (changes requested):
I think that we don't need these changes. It makes sense for, let's say, fork, for some modified version of Mainflux services. However, Mainflux code will always be...well, Mainflux. And we'll always push images to Mainflux DockerHub. I think that this PR makes the script more flexible, but I think that we don't need this kind of flexibility now. Maybe in the future, we'll need this. But for now, I would not introduce these changes.

PR #696, opened 2019-04-05
NOISSUE - Remove debugging message from response of handle error function

blokovi:

### What does this do?

Remove debugging message that explains what went wrong with JSON decoder from response of handle error function. Mainflux version on dashboard display that response,  so it's swiched to nice error message instead of huge debug.

No issue

### DIFF
--- a/ui/src/Error.elm
+++ b/ui/src/Error.elm
@@ -25,4 +25,4 @@ handle error =
             "Bad status: " ++ String.fromInt code
 
         Http.BadBody err ->
-            err
+            "Invalid response body"
--- a/ui/src/Error.elm
+++ b/ui/src/Error.elm
@@ -6,6 +6,7 @@
 
 module Error exposing (handle)
 
+import Debug
 import Http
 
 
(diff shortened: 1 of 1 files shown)


ui/src/Error.elm
```diff
@@ -25,4 +25,4 @@ handle error =
             "Bad status: " ++ String.fromInt code
 
         Http.BadBody err ->
-            err
+            "Invalid response body"
```
drasko, line 33: Is the error however printed in a browser console, so we can debug problems?

blokovi: No, but that's a good idea so I print it in console in new commit 4a5f929837a27054ab33b264373a47dc480a3d69 
Because you can't have any side-effects in Elm you will see that I have to do a bit of a trick for logging values that I don't want to return.

darkodraskovic: We'll need to have some kind of better developed failure reporting mechanism to user. For now, user can inspect browser console and MF console messages. LGTM

darkodraskovic (approved):
LGTM

drasko (approved):
LGTM

Issue #698, opened 2019-04-08
Add missing info and docs about sys event sourcing


nmarcetic:
Missing Information regarding sys event sourcing in both [things](https://github.com/mainflux/mainflux/tree/master/things) and [bootstrap](https://github.com/mainflux/mainflux/tree/master/bootstrap) service as well as in [official documentation](https://mainflux.readthedocs.io/en/latest/) section `Events` is missing.
I want to be able to see what kind of events each service produce (in README also) and how I can consume it.

PR #713, opened 2019-04-16
NOISSUE - Update event fields naming

dborovcanin:
### What does this do?
This pull request updates fields names in the Bootstrap service events.

Yes, I've updated tests.

No, event sourcing documentation will be updated in #698.

### DIFF
--- a/bootstrap/redis/producer/events.go
+++ b/bootstrap/redis/producer/events.go
@@ -51,7 +51,7 @@ type createConfigEvent struct {
 
 func (cce createConfigEvent) encode() map[string]interface{} {
 	return map[string]interface{}{
-		"id":          cce.mfThing,
+		"thing_id":    cce.mfThing,
 		"owner":       cce.owner,
 		"name":        cce.name,
 		"channels":    strings.Join(cce.mfChannels, ", "),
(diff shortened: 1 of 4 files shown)


bootstrap/redis/producer/events.go
```diff
@@ -51,7 +51,7 @@ type createConfigEvent struct {
 
 func (cce createConfigEvent) encode() map[string]interface{} {
 	return map[string]interface{}{
-		"id":          cce.mfThing,
+		"thing_id":    cce.mfThing,
```
drasko, line 54: Why do we use snake_case here? Isn't cammelCase something that is Go standard?

drasko (changes requested):

bootstrap/redis/producer/events.go
anovakovic01: This is not `go` event but Redis event. If events were in JSON format, we would use underscore (AFAIK we already use it for JSON in our API).

dborovcanin: I've changed `id` to `thing_id` just to be in sync with Things service. I think it's also more precise. There is no other reason for this change.

drasko (dismissed):
LGTM

nmarcetic (dismissed):

manuio (approved):
LGTM

nmarcetic (approved):

ghost (approved):

anovakovic01 (approved):
LGTM!

PR #723, opened 2019-04-20
NOISSUE - follow up of PR #657

chombium:
This PR is a followup of #657. After the @drasko's comment I remembered myself how many times we had complains that some user lost his/hers data because they were updating their Mainflux installation or recreated some of the `*-db` containers.

This PR, introduces changes to the current behavior of the `make cleandocker` command.
From now on, an extra parameter must be given in order to delete the persistent volumes,  which up until now was done by default. The new command which represents the old behavior (deleting all artifacts created from the docker composition including the persistent volumes) is `make pv=true cleandocker`.

I understand that this might be a slight problem for the @mainflux/contributors as we'll have to use a new form of the command, but I agree with @drasko's comment that is better to be on the safe side and don't delete the data, the persistent volumes by default, although we, the @mainflux/contributors might have to run the command twice if we forget the `pv=true` parameter. A command alias would to the trick ;)


drasko (dismissed):
LGTM

drasko:
@chombium please resolve the conflicts.

chombium:
@drasko done

dborovcanin (approved):

drasko (approved):
LGTM

drasko:
Merged!

Thanks @chombium !

PR #724, opened 2019-04-20
fix to makefile for OSX/Darwin

mandric:


drasko (dismissed):
LGTM

dborovcanin (commented):
Please sign your commits.

mandric:
Sorry, added GPG signature to commit, rebased and pushed again.  Also signed-off on my own commit.

Makefile
```diff
@@ -8,7 +8,7 @@ SERVICES = users things http normalizer ws coap lora influxdb-writer influxdb-re
 DOCKERS = $(addprefix docker_,$(SERVICES))
 DOCKERS_DEV = $(addprefix docker_dev_,$(SERVICES))
 CGO_ENABLED ?= 0
-GOOS ?= linux
+GOOS ?= $(shell go env GOOS || echo linux)
```
manuio: `GOOS ?= $(go env GOOS)` should also work for linux

Can you please also add this ones:
`GOARCH ?= $(go env GOARCH)`
`GOARM ?= $(go env GOARM)`

manuio (changes requested):

Makefile
drasko: @mandric please use these simpler checks that @manuio proposed

mandric: Sure thing! But in my testing it needed `shell` keyword.

manuio: It's working for us on Linux and OS X without this keyword. Why you need it?

mandric: Without the shell keyword I'm pretty sure the shell does not get invoked.  It doesn't work for me without shell keyword, I just get an empty string.

drasko: @mandric might  be that Makefile demands this. Let me re-check.

drasko: Yes, I checked - `shell` keyword is needed, so please just add it.

mandric: Another option that works on OSX here is just remove the `GOOS ?= linux` line.

mandric: Here is where it was added: https://github.com/mainflux/mainflux/pull/196

manuio: Did you tested on Windows? I don't think that this is safe. Please, add again your previous version: 
`GOOS ?= $(shell go env GOOS)`
`GOARCH ?= $(shell go env GOARCH)`
`GOARM ?= $(shell go env GOARM)`

manuio (approved):

drasko (approved):
LGTM

drasko:
Meged! @mandric welcome to Mainflux contributors club ;).

mandric:
Awesome, thank you!

PR #726, opened 2019-04-22
MF-707 - Allow custom Thing key

### LINKED ISSUE #707: Allow custom Thing key's
**FEATURE REQUEST**

No

As a Thing owner, I want to be able to define a Thing key [during provisioning reques](https://mainflux.readthedocs.io/en/latest/provisioning/#provisioning-devices)t. If I don't set a `key` in Create request, the system should generate random `key` for my thing, as its doing now. This way I would be able to set my own custom key, like my MAC address or whatever I want for my thing key, its mine (you can think about this as password, I should be able to choose it).
This will help me with X509 also, my thing key is part of my X509 cert the `CN` so I will be able to do pre-provisioning and generate X509 certs before I create my things in the system.

Next important feature which can be possible with these changes is `Just in time provisioning` where thing with valid X509 try to connect to Mainflux, but does not exist yet, it's not provisioned, MQTT adapter can publish this event then provision service can catch it and do auto provisioning because thing have valid X509 signed by root CA or Intermediate CA. 

should have

anovakovic01:
### What does this do?
Adds update key endpoint and adds `key` field to create thing endpoint. In other words, this PR adds support for a custom key.

Resolves #707.

The current functionality shouldn't be broken. If `key` field is not sent in create thing request, `things` service will generate this key for you in UUID format.

Yes, I've updated existing tests and added new ones.

I've updated swagger docs.

### DIFF
--- a/things/api/http/endpoint_test.go
+++ b/things/api/http/endpoint_test.go
@@ -89,7 +89,9 @@ func TestAddThing(t *testing.T) {
 	ts := newServer(svc)
 	defer ts.Close()
 
-	data := toJSON(thing)
+	th := thing
+	th.Key = "key"
+	data := toJSON(th)
 
 	cases := []struct {
 		desc        string
@@ -106,6 +108,14 @@ func TestAddThing(t *testing.T) {
 			auth:        token,
 			status:      http.StatusCreated,
 			location:    "/things/1",
+		},
+		{
+			desc:        "add thing with existing key",
+			req:         data,
+			contentType: contentType,
+			auth:        token,
+			status:      http.StatusConflict,
+			location:    "",
 		},
 		{
 			desc:        "add thing with empty JSON request",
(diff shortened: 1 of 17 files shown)


manuio (dismissed):
LGTM

things/api/http/endpoint_test.go
```diff
@@ -280,6 +290,126 @@ func TestUpdateThing(t *testing.T) {
...
+			auth:        token,
+			status:      http.StatusOK,
+		},
+		{
+			desc:        "update thing with conflicting key",
+			req:         data,
+			id:          sth.ID,
+			contentType: contentType,
+			auth:        token,
+			status:      http.StatusConflict,
```
chombium: I think that http [status 409](https://www.checkupdown.com/status/E409.html) is not appropriate for this. 
I think that 422 Unprocessable entity would be more appropriate. This error means that the server understood the request but was unable to process it because of an error. In this case setting an existing key value as a new key.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422
https://stackoverflow.com/a/20215807

anovakovic01: @chombium I'm open for suggestions, but can you explain to me why it's not appropriate? Key should be unique and if it's not for me it makes sense that server returns `409 Conflict`. If you think that in this way one user can find a key of a thing that he shouldn't have access to, well, I'm not sure that this is a problem, because he still cannot see other users channels and also doesn't now to which thing specified key belongs to.

anovakovic01: I missed the first reference, sorry, everything is clear.

anovakovic01: But [here](https://httpstatuses.com/409) it says that `409` status should be used in such situations. Not sure what to do.

anovakovic01: OK, final decision, I'll use `422`. `409` is used when request is in a conflict with the current state of a target resource which is not the case here. Thanks @chombium!

anovakovic01: Fixed.

chombium (dismissed):

things/api/http/endpoint_test.go
chombium: 409 would be a good response if two users read the same data record, the first updates it and then the second tries to update it. The second update should get a 409 because the data has changed and he/she has tried to update older version of the data

dborovcanin (approved):

manuio (approved):

PR #728, opened 2019-04-22
Add UI websocket open/close and send/receive

darkodraskovic:

### DIFF
--- a/ui/src/Message.elm
+++ b/ui/src/Message.elm
@@ -48,7 +48,7 @@ type alias Model =
     , thingid : String
     , checkedChannelsIds : List String
     , checkedChannelsIdsWs : List String
-    , websocketData : List String
+    , websocketIn : List String
     }
 
 
@@ -62,7 +62,7 @@ initial =
     , thingid = ""
     , checkedChannelsIds = []
     , checkedChannelsIdsWs = []
-    , websocketData = []
+    , websocketIn = []
     }
 
 
@@ -70,8 +70,8 @@ type Msg
     = SubmitMessage String
     | SendMessage
     | WebsocketSend
-    | WebsocketMsg String
-    | RetrievedWebsocket E.Value
+    | WebsocketIn String
+    | RetrievedWebsockets E.Value
     | SentMessage (Result Http.Error String)
     | ThingMsg Thing.Msg
     | ChannelMsg Channel.Msg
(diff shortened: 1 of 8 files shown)


ui/src/Message.elm
```diff
@@ -146,13 +188,15 @@ view model =
                                     [ Input.text [ Input.id "message", Input.onInput SubmitMessage ]
                                     ]
                                 , Button.button [ Button.secondary, Button.attrs [ Spacing.ml1 ], Button.onClick SendMessage ] [ text "Send" ]
+                                , Button.button [ Button.secondary, Button.attrs [ Spacing.ml1 ], Button.onClick WebsocketSend ] [ text "Websocket" ]
```
drasko: Why the tet is "Websocket"? It would be better that it is something like "Recieve" or "Listen" - user does not have to know really about underlaying technology.

drasko: Also, you need to transform this button in something like "Stop" or something, to stop listening (i.e. unsubscribe, i.e. close this websocket),

ui/src/Websocket.js
```diff
@@ -0,0 +1,54 @@
...
+    MF.log("Websocket is not open. URL: " + url);
+  }
+});
+
+app.ports.disconnectWebsocket.subscribe(function(data) {
+  var url = 'wss://localhost/ws/channels/' + data.channelid + '/messages?authorization=' + data.thingkey
+  if (wss[url]) {
+    wss[url].close();
+  } else {
+    MF.log("Websocket is not open. URL: " + url);
```
drasko: Is this comment OK? Looks like a copy of precious one.

drasko (changes requested):

ui/src/Message.elm
darkodraskovic: Addressed in the last commit.

ui/src/Websocket.js
darkodraskovic: No, it just informs the user that the websocket he's trying to close is not open, so he can't close it. I can reword the text so that it becomes more clear.

blokovi (dismissed):
LGTM

ui/src/Websocket.js
```diff
@@ -0,0 +1,79 @@
...
+}
+
+MF.url = function(data) {
+    return 'wss://localhost/ws/channels/' + data.channelid + '/messages?authorization=' + data.thingkey
+}
+
+app.ports.connectWebsocket.subscribe(function(data) {
+    var url = MF.url(data);
+    if (wss[url]) {
+        MF.log("Websocket already open. URL: " + url );
```
manuio: Can you use `''` instead of `""` here and in the rest of the code?

manuio (changes requested):

ui/src/Websocket.js
darkodraskovic: The idea was to use single quotes for the things that happen under the hood and to use the double quotes for the messages intended for the user.

manuio (dismissed):

drasko (dismissed):
LGTM

manuio (dismissed):

drasko (approved):
LGTM

manuio (approved):

PR #733, opened 2019-04-29
MF-732 - Add PostgreSQL writer

manuio:


drasko (dismissed):
LGTM

cmd/postgres-writer/main.go
```diff
@@ -0,0 +1,165 @@
+//
+// Copyright (c) 2018
```
anovakovic01: Change these to `2019` for this service.

cmd/postgres-writer/main.go
```diff
@@ -0,0 +1,165 @@
...
+	envDBSSLMode     = "MF_POSTGRES_DB_SSL_MODE"
+	envDBSSLCert     = "MF_POSTGRES_DB_SSL_CERT"
+	envDBSSLKey      = "MF_POSTGRES_DB_SSL_KEY"
+	envDBSSLRootCert = "MF_POSTGRES_DB_SSL_ROOT_CERT"
+)
+
+type config struct {
+	natsURL  string
+	logLevel string
+	Port     string
```
anovakovic01: Why public field?

cmd/postgres-writer/main.go
```diff
@@ -0,0 +1,165 @@
...
+	go startHTTPServer(cfg.Port, errs, logger)
+
+	go func() {
+		c := make(chan os.Signal)
+		signal.Notify(c, syscall.SIGINT)
+		errs <- fmt.Errorf("%s", <-c)
+	}()
+
+	err = <-errs
+	logger.Error(fmt.Sprintf("Cassandra writer service terminated: %s", err))
```
anovakovic01: Change `Cassandra` to `Postgres`.

docker/addons/postgres-writer/docker-compose.yml
```diff
@@ -0,0 +1,52 @@
...
+###
+
+version: "3"
+
+networks:
+  docker_mainflux-base-net:
+    external: true
+
+volumes:
+  mainflux-postgres-volume:
```
anovakovic01: Maybe rename this one to `mainflux-postgres-writer-volume`?

writers/postgres/init.go
```diff
@@ -0,0 +1,82 @@
...
+            value         FLOAT,
+            string_value  TEXT,
+            bool_value    BOOL,
+            data_value    TEXT,
+            value_sum     FLOAT,
+            time          FlOAT,
+            update_time   FLOAT,
+            link          TEXT,
+						PRIMARY KEY (channel)
+					)`,
```
anovakovic01, line 72: Please fix schema formatting. More importantly, `channel` cannot be message primary ID (it's not unique).

writers/postgres/messages.go
```diff
@@ -0,0 +1,116 @@
...
+	dbth := toDBMessage(msg)
+
+	_, err := pr.db.NamedExec(q, dbth)
+	if err != nil {
+		pqErr, ok := err.(*pq.Error)
+		if ok {
+			switch pqErr.Code.Name() {
+			case errInvalid:
+				return nil
+			}
```
anovakovic01, line 52: Not sure if this is the right way to handle this error.

anovakovic01 (changes requested):

writers/postgres/messages.go
chombium: That's the way to check for PostgreSQL errors. I'm not sure about ignoring the "invalid_text_representation" errors. The protobuf and the go compiler will protect us from setting wrong string representations for the parameter values, but I think one extra check does not hurt.
@manuio What's the idea with ignoring this kind of errors?

anovakovic01: Yes, that is what I meant, `return nil` is just ignoring it. Also, can there be any other errors from Postgres?

chombium: sure there can be other errors. I also think we should also "repack" the [pq.error](https://github.com/lib/pq/blob/master/error.go) error code and error message and return it as go error.

anovakovic01: Yes, I agree. @manuio you can check out `things` service implementation for example.

manuio: Good catch, I removed the error and then forgot... I'll fix it

manuio: Good catch, I removed and then forgot... I'll fix it

drasko (dismissed):
LGTM

anovakovic01 (changes requested):
Almost done.

anovakovic01 (approved):
LGTM!

drasko (approved):
LGTM

PR #737, opened 2019-05-08
MF-708 - Assign Writer(s) to a channel

### LINKED ISSUE #708: Assign Writer(s) to a channel
No

When messages are sent to Mainflux, and more than one Writer is installed, all of the messages which pass the normalizer are stored via the Writers in all of the activated storage databases.
There should be the possibility to assign Writer(s) to a channel which should be used to write the data to the storage database(s)

I think this is a very important feature. 
Example: If I post geo-location data, I want to store the data only in MongoDB, but not in the InfluxDB. 

Again, thank You for the cool work!

Regards, 
Boris Bokun
strictit d.o.o.
Beograd

anovakovic01:
### What does this do?
Adds the configuration file with the list of channels. Writers will only persist messages from specified channels.

Resolves #708.

Configuration should be specified in order to persist messages from channels.

No.

Yes, I've updated README files.


cmd/cassandra-writer/main.go
```diff
@@ -23,25 +24,28 @@ import (
...
-	envLogLevel = "MF_CASSANDRA_WRITER_LOG_LEVEL"
-	envPort     = "MF_CASSANDRA_WRITER_PORT"
-	envCluster  = "MF_CASSANDRA_WRITER_DB_CLUSTER"
-	envKeyspace = "MF_CASSANDRA_WRITER_DB_KEYSPACE"
+	defNatsURL     = nats.DefaultURL
+	defLogLevel    = "error"
+	defPort        = "8180"
+	defCluster     = "127.0.0.1"
+	defKeyspace    = "mainflux"
+	defChanCfgPath = "/config/channels.yaml"
```
drasko: We use TOML for all our configs in Mainflux.

cmd/influxdb-writer/main.go
```diff
@@ -41,6 +43,7 @@ const (
 	defDBPort       = "8086"
 	defDBUser       = "mainflux"
 	defDBPass       = "mainflux"
+	defChanCfgPath  = "/config/channels.yaml"
```
drasko: TOML

cmd/mongodb-writer/main.go
```diff
@@ -26,62 +27,72 @@ import (
...
-	envDBName   = "MF_MONGO_WRITER_DB_NAME"
-	envDBHost   = "MF_MONGO_WRITER_DB_HOST"
-	envDBPort   = "MF_MONGO_WRITER_DB_PORT"
+	defNatsURL     = nats.DefaultURL
+	defLogLevel    = "error"
+	defPort        = "8180"
+	defDBName      = "mainflux"
+	defDBHost      = "localhost"
+	defDBPort      = "27017"
+	defChanCfgPath = "/config/channels.yaml"
```
drasko: TOML

cmd/postgres-writer/main.go
```diff
@@ -42,6 +44,7 @@ const (
 	defDBSSLCert     = ""
 	defDBSSLKey      = ""
 	defDBSSLRootCert = ""
+	defChanCfgPath   = "/config/channels.yaml"
```
drasko: TOML

drasko (changes requested):
Please use TOML, as it is official config method of Mainflux. Also, please provide some documentation about the feature.

anovakovic01:
@drasko I've updated README files and added doc comments in the configuration itself. I didn't know that we had any config files in Mainflux, especially TOML. Can you send me an example?

docker/addons/cassandra-writer/channels.toml
```diff
@@ -0,0 +1,4 @@
+# If you want to listen on all channels, just pass one element ["*"], otherwise
+# pass the list of channels.
+[channels]
```
drasko, line 3: Why do we need a special table for this config? Would there be any other setting for channels except `filter`? I.e. would just
```
channels = [...]
```
do the job?

Moreover, if all that we need is alist of channels, maybe we do not need even TOML? Maybe a simplle file with one channel ID per line is enough?

anovakovic01: We talked about adding support for configuration for every service (instead of using current env vars). That's why I added it as a separate table (this config will grow, and in the end, this config will create a lot of other fields). We can use simple file, but I would keep standard config format (especially because this config will grow).

drasko: Agreed.

drasko (approved):
LGTM

docker/addons/mongodb-writer/channels.toml
```diff
@@ -0,0 +1,4 @@
+# If you want to listen on all channels, just pass one element ["*"], otherwise
+# pass the list of channels.
```
drasko, line 2: Please add to a comment - "list of channels or any other valid NATS subscription regexp"

drasko (changes requested):

docker/addons/mongodb-writer/channels.toml
anovakovic01: @drasko But this is not true. Writer service always takes everything, and then filters messages using this list and channels field.

anovakovic01: It's done like this because normalizer publishes messages to `out.senml` topic, so I don't have info about channel as part of NATS subject.

drasko: Yes, I get it - writers only subscribe ot `out.senml` topic.

drasko (approved):
LGTM

nmarcetic (approved):

PR #745, opened 2019-05-15
MF-574 - Add missing environment variables to Cassandra writer

### LINKED ISSUE #574: Add missing environment variables to Cassandra writer
**FEATURE REQUEST**

No.

Currently there are no environment variables for username, password and cluster port in Cassandra writer. These should be added ASAP.

This is a must-have.

anovakovic01:
### What does this do?
Adds env vars for DB username, password, and port.

Resolves #574.

There are no such changes.

No.

I've updated existing README files.

### DIFF
--- a/readers/cassandra/README.md
+++ b/readers/cassandra/README.md
@@ -13,6 +13,9 @@ default values.
 | MF_CASSANDRA_READER_PORT        | Service HTTP port                              | 8180           |
 | MF_CASSANDRA_READER_DB_CLUSTER  | Cassandra cluster comma separated addresses    | 127.0.0.1      |
 | MF_CASSANDRA_READER_DB_KEYSPACE | Cassandra keyspace name                        | mainflux       |
+| MF_CASSANDRA_READER_DB_USERNAME | Cassandra DB username                          |                |
+| MF_CASSANDRA_READER_DB_PASSWORD | Cassandra DB password                          |                |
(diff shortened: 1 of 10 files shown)


readers/cassandra/README.md
```diff
@@ -13,6 +13,9 @@ default values.
 | MF_CASSANDRA_READER_PORT        | Service HTTP port                              | 8180           |
 | MF_CASSANDRA_READER_DB_CLUSTER  | Cassandra cluster comma separated addresses    | 127.0.0.1      |
 | MF_CASSANDRA_READER_DB_KEYSPACE | Cassandra keyspace name                        | mainflux       |
+| MF_CASSANDRA_READER_DB_USERNAME | Cassandra DB username                          |                |
```
manuio, line 16: I would put an empty string here and for all empty variables

manuio (changes requested):

readers/cassandra/README.md
anovakovic01: Maybe, but then we have to be consistent and add `"` for all values, which is too much for this PR, should be separate. Also, this will be presented as a table, so I think that it's clear that it's an empty value.

manuio: I should remove it from Postgres reader and writer then

drasko (approved):
LGTM

manuio (approved):

PR #749, opened 2019-05-16
MF-742 - Things to support single user scenario

### LINKED ISSUE #742: Things to support single user scenario
Currently things service can create entities for a specific user, and it authentifies user via Users service.

In a scenarious of deployments on a simple IoT gateway, Users service might be ommitted (dues to the size), as the whole system migh work in a single-user (gateway admin) scenario.

Change the Things service to accept pre-defined user from a config file or ENV variable.

anovakovic01:
### What does this do?
Adds single user mode to `things` service.

Resolves #742.

There are two new env vars for single user mode.

No.

I've updated existing `README.md` file.

### DIFF
--- a/cmd/things/main.go
+++ b/cmd/things/main.go
@@ -30,6 +30,7 @@ import (
 	httpapi "github.com/mainflux/mainflux/things/api/http"
 	"github.com/mainflux/mainflux/things/postgres"
 	rediscache "github.com/mainflux/mainflux/things/redis"
+	localusers "github.com/mainflux/mainflux/things/users"
 	"github.com/mainflux/mainflux/things/uuid"
 	usersapi "github.com/mainflux/mainflux/users/api/grpc"
 	stdprometheus "github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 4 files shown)


cmd/things/main.go
```diff
@@ -196,6 +208,15 @@ func connectToDB(dbConfig postgres.Config, logger logger.Logger) *sqlx.DB {
 	return db
 }
 
+func createUsersClient(cfg config, logger logger.Logger) (mainflux.UsersServiceClient, func() error) {
+	if cfg.singleUserEmail == "" || cfg.singleUserToken == "" {
```
drasko: @anovakovic01 this is good, but I would prefer this check:
- If it's sungle user, then return `NewSingleUserService`
- Otherwise proceed as default

That check is logically easier to grasp, because for us defaut would be multi-user, and we do single user only uder some conditions.

Currently your check is a bit different, putting single-suer in the main light, and multi-user only if the single-user checkd do not pass.

Logic is the same, just written differently to make code more understandable.

drasko (changes requested):

cmd/things/main.go
anovakovic01: Done.

drasko (approved):
LGTM

manuio (approved):

PR #750, opened 2019-05-17
NOISSUE - Add searchable Things name

manuio:


bootstrap/mocks/things.go
```diff
@@ -166,7 +166,7 @@ func (svc *mainfluxThings) UpdateKey(string, string, string) error {
 	panic("not implemented")
 }
 
-func (svc *mainfluxThings) ListThings(string, uint64, uint64) (things.ThingsPage, error) {
+func (svc *mainfluxThings) ListThings(string, uint64, uint64, string) (things.ThingsPage, error) {
```
drasko, line 169: Consider using variadic functions so that we do not have to pass this param every time.

I am not saying it is better, just analyze this possibility.

@anovakovic01 @dusanb94 @nmarcetic comments?

chombium: Instead of using variadic functions and checking the variables like:
```go
package main

import (
	"fmt"
	"reflect"
)

func test(a ...interface{}) {
	for i, param := range a {
		fmt.Printf("%d : %s\n", i, reflect.TypeOf(param))
	}
}

func main() {
	test(1, "my_way", true)
}
```
It might be better if we use some dependency injection framework like [wire](https://github.com/google/wire) ([blog post](https://blog.golang.org/wire)) or one from awesome-go's [di list](https://github.com/avelino/awesome-go#dependency-injection). I've used DI in other languages, but not in Go and I can not say which framework is better.

things/api/http/requests.go
```diff
@@ -121,14 +138,16 @@ type listResourcesReq struct {
...
 }
 
 func (req *listResourcesReq) validate() error {
 	if req.token == "" {
 		return things.ErrUnauthorizedAccess
 	}
 
-	if req.limit == 0 || req.limit > maxLimitSize {
+	if req.limit == 0 || req.limit > maxLimitSize ||
+		len(req.name) > maxNameSize {
```
anovakovic01: Please move the name part to a separate `if` statement in order to keep it simple.

things/api/logging.go
```diff
@@ -81,7 +81,7 @@ func (lm *loggingMiddleware) ViewThing(token, id string) (thing things.Thing, er
 	return lm.svc.ViewThing(token, id)
 }
 
-func (lm *loggingMiddleware) ListThings(token string, offset, limit uint64) (_ things.ThingsPage, err error) {
+func (lm *loggingMiddleware) ListThings(token string, offset, limit uint64, name string) (_ things.ThingsPage, err error) {
 	defer func(begin time.Time) {
 		message := fmt.Sprintf("Method list_things for token %s took %s to complete", token, time.Since(begin))
```
anovakovic01: Think about adding name to log msg.

things/postgres/things.go
```diff
@@ -158,14 +159,21 @@ func (tr thingRepository) RetrieveByKey(key string) (string, error) {
...
 }
 
-func (tr thingRepository) RetrieveAll(owner string, offset, limit uint64) (things.ThingsPage, error) {
-	q := `SELECT id, name, key, metadata FROM things
-	      WHERE owner = :owner ORDER BY id LIMIT :limit OFFSET :offset;`
+func (tr thingRepository) RetrieveAll(owner string, offset, limit uint64, name string) (things.ThingsPage, error) {
+	nq := ""
+	if name != "" {
+		name = fmt.Sprintf(`%%%s%%`, name)
+		nq = fmt.Sprintf(`AND name LIKE :name`)
```
anovakovic01: No need to use this second `fmt.Sprintf`.

things/postgres/things_test.go
```diff
@@ -315,29 +315,36 @@ func TestThingRetrieveByKey(t *testing.T) {
...
+
+		th := things.Thing{
 			Owner: email,
+			ID:    thid,
 			Key:   thkey,
 		}
 
-		thingRepo.Save(t)
+		// Create one thing with name
+		if i == 1 {
```
anovakovic01: Why not using 0 here?

things/postgres/things_test.go
```diff
@@ -358,10 +365,24 @@ func TestMultiThingRetrieval(t *testing.T) {
...
 			size:   0,
 		},
+		"retrieve things with valid name": {
+			owner:  email,
+			offset: 0,
+			limit:  n,
+			name:   name,
+			size:   1,
+		},
+		"retrieve things with invalid name": {
```
anovakovic01: This is not an invalid name. There are no things with this name. Think about adding an invalid name use case (with a name which length is larger than the limit that you've set).

things/redis/streams_test.go
```diff
@@ -186,8 +186,8 @@ func TestListThings(t *testing.T) {
 	require.Nil(t, err, fmt.Sprintf("unexpected error %s", err))
 
 	essvc := redis.NewEventStoreMiddleware(svc, redisClient)
-	esths, eserr := essvc.ListThings(token, 0, 10)
-	ths, err := svc.ListThings(token, 0, 10)
+	esths, eserr := essvc.ListThings(token, 0, 10, "")
+	ths, err := svc.ListThings(token, 0, 10, "lora")
```
anovakovic01: Can use change this to something else. It's not a huge deal, but I don't like having anything protocol specific even in tests.

things/api/http/requests.go
```diff
@@ -27,6 +28,10 @@ func (req addThingReq) validate() error {
 		return things.ErrUnauthorizedAccess
 	}
 
+	if len(req.Name) > maxNameSize {
+		return things.ErrMalformedEntity
+	}
+
```
anovakovic01, line 34: If we're going to introduce this limit, you should add it to DB schema too (use varchar(64)). Also, not sure if 64 is large enough. I would put at least 128. @drasko @nmarcetic @dusanb94 @srados what do you think?

anovakovic01 (changes requested):

bootstrap/mocks/things.go
manuio: I agree with @anovakovic01

things/redis/streams_test.go
manuio: Good catch, I forgot this one

anovakovic01 (dismissed):

anovakovic01 (approved):
LGTM!

anovakovic01 (dismissed):
LGTM!

things/api/http/requests.go
nmarcetic: I agree! Especially with about size 64 not sounds to much.

things/api/http/requests.go
nmarcetic: Yes pls, we are avoiding chaining as much as we can.

anovakovic01 (dismissed):
LGTM!

anovakovic01 (approved):
LGTM!

nmarcetic (approved):

PR #756, opened 2019-06-03
MF-466 - ARM docker deployment

### LINKED ISSUE #466: Mainflux ARM Docker images on dockerhub
**FEATURE REQUEST**

It would be nice to have Mainflux ARM docker images on dockerhub in case someone wants to deploy it on an ARM device

blokovi:

### What does this do?
Docker deployment for arm32v7, compatible with MFX-1 gateway

Resolves #466


docker/Dockerfile.arm
```diff
@@ -0,0 +1,15 @@
+FROM golang:1.10-alpine AS builder
```
drasko: This Dockerfile does not make sense, because it is identical to existing one - just ARGs are passed.

Please change Makefile to pass ARCH args that are set into existing Dockerfile. I.e. use the same Dockerfile for both architectures.

mqtt/Dockerfile.arm
```diff
@@ -0,0 +1,12 @@
+FROM arm32v7/node:10.16.0-stretch-slim
+
+COPY qemu-arm-static /usr/bin
+
+COPY *.proto mqtt/* ./
+
+RUN npm rebuild && npm install
+
+EXPOSE 1883 8880
+
```
drasko, line 10: Please remove `qemu-arm-static` at the end, to keep the image smaller.

drasko (changes requested):

mqtt/Dockerfile.arm
blokovi: Removed, however new image have same size

docker/Dockerfile.arm
blokovi: Dockerfile.arm deleted, in Makefile I made `make docker_arm32v7` that pass this args

manuio (dismissed):
LGTM

docker/Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM golang:1.10-alpine AS builder
 ARG SVC_NAME
+ARG GOARCH
```
drasko, line 3: How is this used in Dockerfile? Looks like it is not.

docker/Dockerfile
```diff
@@ -1,5 +1,7 @@
...
+ARG GOARM
```
drasko, line 4: Same here

docker/Dockerfile
blokovi: It is used in `&& make $SVC_NAME \` which needs this arguments.

docker/Dockerfile
blokovi: It is used in `&& make $SVC_NAME \` which needs this arguments.

Makefile
```diff
@@ -17,6 +18,10 @@ define make_docker
 	docker build --no-cache --build-arg SVC_NAME=$(subst docker_,,$(1)) --tag=mainflux/$(subst docker_,,$(1)) -f docker/Dockerfile .
 endef
 
+define make_docker_arm
+	docker build --no-cache --build-arg GOARCH=arm --build-arg GOARM=7 --build-arg SVC_NAME=$(subst docker_arm_,,$(1)) --tag=mainflux/$(subst docker_arm_,,$(1))-arm32v7 -f docker/Dockerfile .
```
drasko, line 22: ARM version is hardcoded (v7). I would advise that never the less we pass the ARGs into Docer containers via ENV variables.

BTW, I need compiling for various architectures, so please keep cross-compiling generic.

Makefile
```diff
@@ -7,6 +7,7 @@ BUILD_DIR = build
 SERVICES = users things http normalizer ws coap lora influxdb-writer influxdb-reader mongodb-writer mongodb-reader cassandra-writer cassandra-reader postgres-writer postgres-reader cli bootstrap
 DOCKERS = $(addprefix docker_,$(SERVICES))
 DOCKERS_DEV = $(addprefix docker_dev_,$(SERVICES))
+DOCKERS_ARM = $(addprefix docker_arm_,$(SERVICES))
```
drasko, line 10: Not needed.

Makefile
```diff
@@ -63,20 +68,32 @@ $(SERVICES):
 $(DOCKERS):
 	$(call make_docker,$(@))
 
+$(DOCKERS_DEV):
+	$(call make_docker_dev,$(@))
+
+$(DOCKERS_ARM):
```
drasko, line 74: Not needed.

Makefile
```diff
@@ -63,20 +68,32 @@ $(SERVICES):
...
+	# MQTT Docker build must be done from root dir because it copies .proto files
+	docker build --tag=mainflux/mqtt-arm32v7 -f mqtt/Dockerfile.arm .
 
-$(DOCKERS_DEV):
-	$(call make_docker_dev,$(@))
+dockers: $(DOCKERS) docker_ui docker_mqtt
 
 dockers_dev: $(DOCKERS_DEV)
 
+dockers_arm: $(DOCKERS_ARM) docker_ui_arm docker_mqtt_arm
```
drasko, line 95: Not needed.

docker/Dockerfile
drasko: Please set ARGs as:
1. ARCH
2. VER

drasko (changes requested):

docker/Dockerfile
blokovi: This is not new variables, they are already being used and in current docs we have them like `GOARCH` and `GOARM`: https://mainflux.readthedocs.io/en/latest/dev-guide/#cross-compiling-for-arm
If I change it, I will need to change that also. do you still want this change?

manuio (approved):

drasko (approved):
Approving so we can move on. Will be corrected in the future if needed.

PR #782, opened 2019-07-13
MF-399 - Add open tracing support

### LINKED ISSUE #399: Add message UUID so that we can follow e2e flow
Add UUID to each message that flows though Mainflux so we can log and follow the flow.

For [RewMessage](https://github.com/mainflux/mainflux/blob/master/message.proto#L12) `UUID` field with be UUIDv4. For [Message](https://github.com/mainflux/mainflux/blob/master/message.proto#L21) `UUID` can be `<parent_uuid>.<local_increment>` format - for example `<parent_uuid>.0`,  `<parent_uuid>.1` and so on - because RawMessage is de-aggregated locally in `normalizer`.

anovakovic01:
### What does this do?
Adds base open tracing support to `users` and `things` service.

Resolves #399.

There are no such changes.

Existing tests are updated.

`README.md` doc is updated with new env vars.

### Notes
Open tracing support can be improved by adding tags to the existing spans. Also, open tracing support should be added to bootstrap service. One of the next steps would also be to add tracing of messages that are passing through NATS.


drasko (dismissed):
LGTM

cmd/things/main.go
```diff
@@ -1,5 +1,5 @@
 //
```
drasko, line 1: I would suggest dropping a year from headers and contracting them to something like this:
```
// Copyright (c) Mainflux
```

http/adapter.go
```diff
@@ -1,5 +1,5 @@
 //
```
drasko, line 1: Same here

http/api/endpoint.go
```diff
@@ -1,5 +1,5 @@
 //
```
drasko, line 1: And so on...

cmd/things/main.go
anovakovic01: @drasko can we do this in another PR, this one is already huge?

drasko: Yes, sure.

mteodor:
LGTM

mteodor (approved):

nmarcetic (approved):

PR #801, opened 2019-07-22
Add support for Open tracing to be optional

anovakovic01:


drasko (approved):
I will approve, but this is not really consistent with other ENV vars I think, which have default value if not provided.

OK, to be honest this one has also a default value when not provided - empty string :).

nmarcetic (approved):

PR #810, opened 2019-07-29
NOISSUE - Add content type as part of MQTT subscription topic

anovakovic01:
### What does this do?
Adds content type as part of MQTT subscription topic.

There is no issue for this PR.

The default value for content type is not SenML JSON, but an empty string.

No.

I've updated existing docs.


drasko (approved):
LGTM

dborovcanin (approved):

PR #816, opened 2019-08-06
NOISSUE - Add version endpoint to MQTT adapter

dborovcanin:
### What does this do?
This pull request adds `/version` endpoint to the MQTT adapter (over WS/HTTP).

There are no such changes.

No.

No.


drasko (dismissed):
LGTM

docker/docker-compose.yml
```diff
@@ -256,6 +256,7 @@ services:
       MF_JAEGER_URL: ${MF_JAEGER_URL}
     ports:
       - ${MF_COAP_ADAPTER_PORT}:${MF_COAP_ADAPTER_PORT}/udp
+      - ${MF_COAP_ADAPTER_PORT}:${MF_COAP_ADAPTER_PORT}
```
anovakovic01: Why remove udp?

mqtt/aedes/mqtt.js
```diff
@@ -109,6 +111,13 @@ esclient.on('error', function(err) {
 // MQTT over WebSocket
 function startWs() {
     var server = http.createServer();
+    server.on('request', (req, res) => {
+        if (req.url === '/version') {
+            res.statusCode = 200;
+            res.setHeader('Content-Type', 'text/plain; charset=utf-8');
+            res.end(`{"service":"mqtt-adapter","version":"${version}"}`);
+        }
```
anovakovic01, line 119: What if it's not version, but invalid path?

anovakovic01 (changes requested):

docker/docker-compose.yml
dborovcanin: I didn't remove UDP, I just exposed TCP, so that we can access version and metrics endpoints.

mqtt/aedes/mqtt.js
dborovcanin: I think that default is an empty response.

anovakovic01 (approved):
LGTM!

mqtt/aedes/mqtt.js
```diff
@@ -5,6 +5,8 @@
 
 'use strict';
 
+const version = '0.9.0';
```
anovakovic01, line 8: This service is written in ES 5 style. Think about using `var` here. If you don't want to, it's not a problem.

drasko (approved):
LGTM

PR #818, opened 2019-08-12
NOISSUE - fix generating thing certificate for mqtt test tool

mteodor:
### What does this do?
fix generating thing certificate for mqtt test tool when using mtls

MF-295


mteodor:
this is made by mistake, should be remove

PR #837, opened 2019-09-03
MF-832 Invalid Content-Type in MQTT Version Endpoint, MF-822 Return 404 for non `/version` paths on MQTT adapter

### LINKED ISSUE #822: Return 404 for non `/version` paths on MQTT adapter
**BUG REPORT**

Get version endpoint with misspelled `/verision` path.

Status code 404 Not Found.

Status code 204 No Content.

Send an HTTP request to MQTT service path other than `/version`

I'm running latest docker composition.

This does not affect service functionality (therefore this is the low priority issue), but it is not consistent with other services.

### LINKED ISSUE #832: Invalid Content-Type in MQTT Version Endpoint
In the following code, a JSON is returned from the version endpoint, but it has `text/plain` for Content-Type.

https://github.com/mainflux/mainflux/blob/a8836d67e0fa78e4a111b8b63557ca90396ab40b/mqtt/aedes/mqtt.js#L117

I think this cause issue in clients for decoding the results.

1995parham:
### What does this do?
It fixes content-type and not-found issues in MQTT version endpoint. if you use invalid route instead of getting empty response you get 404 error with a not-found message.

Resolve #832 
Resolve #822 

It doesn't break or change current functionality.

No

No

### DIFF
--- a/mqtt/aedes/mqtt.js
+++ b/mqtt/aedes/mqtt.js
@@ -112,11 +112,13 @@ esclient.on('error', function(err) {
 function startWs() {
     var server = http.createServer();
     server.on('request', (req, res) => {
+        res.setHeader('Content-Type', 'application/json');
         if (req.url === '/version') {
             res.statusCode = 200;
-            res.setHeader('Content-Type', 'text/plain; charset=utf-8');
             res.end(`{"service":"mqtt-adapter","version":"${version}"}`);
         }
+        res.statusCode = 404;
+        res.end('{"service":"mqtt-adpater", "message": "not found"}')
     }); 
     websocket.createServer({server: server}, aedes.handle);
     server.listen(config.ws_port);


mqtt/aedes/mqtt.js
```diff
@@ -112,11 +112,13 @@ esclient.on('error', function(err) {
 function startWs() {
     var server = http.createServer();
     server.on('request', (req, res) => {
+        res.setHeader('Content-Type', 'application/json');
         if (req.url === '/version') {
             res.statusCode = 200;
-            res.setHeader('Content-Type', 'text/plain; charset=utf-8');
             res.end(`{"service":"mqtt-adapter","version":"${version}"}`);
```
nmarcetic, line 118: We don't use template literals (e.g mixing with ES6). Just a tip, but we should consider refactoring and using ES6.

nmarcetic (approved):

drasko (approved):
LGTM

drasko:
Please change issue name to something like `MF-832 - <PR title>`

1995parham:
@drasko I have changed the title to issues that this pull request resolves.

PR #841, opened 2019-09-06
NOISSUE - Add docker-compose for MQTT cluster

drasko:

### What does this do?
Introduces docker-compose files for MQTT cluster. Also, modifies `nginx.conf` to improve performance.

`nginx.conf` files have been changed, and this demands testing


drasko:
WIP label removed

:warning: Please pull the branch and test on your machines.

This PR tries to move as much as possible of mutual `nginx.conf` code into the snippets, also introduce cluster support.

Additionally, in `entrypoint.sh` environment (read from `.env`) is substituted in nginx.conf, so proxy URLs are not hadrcoded any more.

`docker-compose.yaml` is broken in several files, as Aedes is pulled out in `aedes.yaml` so it can be replaced with `verne.yaml`. Additionally, `verne-cluster.yaml` can be used instead (which will pull and setup correct `nginx.conf`).

docker/nginx/nginx-key.conf
```diff
@@ -64,155 +53,62 @@ http {
...
-            proxy_pass http://http-adapter:8185/;
-
-            # Allow OPTIONS method CORS
-            if ($request_method = OPTIONS ) {
-                add_header Content-Length 0;
-                add_header Content-Type text/plain;
-                return 200;
-            }
+            include snippets/proxy-headers.conf;
+            proxy_pass http://http-adapter:${MF_HTTP_ADAPTER_PORT};
```
mteodor: I think you missed '/' here

docker/nginx/nginx-key.conf
```diff
@@ -64,155 +53,62 @@ http {
...
-
-            # Allow OPTIONS method CORS
-            if ($request_method = OPTIONS ) {
-                add_header Content-Length 0;
-                add_header Content-Type text/plain;
-                return 200;
-            }
+            include snippets/proxy-headers.conf;
+            include snippets/ws-upgrade.conf;
+            proxy_pass http://ws-adapter:${MF_WS_ADAPTER_PORT};
```
mteodor: I think you missed '/' here

mteodor (dismissed):

.gitignore
```diff
@@ -1,2 +1,17 @@
+# Mainflux
 build
-site/
+.vscode
+
+# MQTT
+mqtt/aedes/node_modules
```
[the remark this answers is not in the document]
drasko: We'll keep it in one place for now to make visible - we can move later.

docker/nginx/nginx-key.conf
drasko: Thanks, good catch. Added this with a comment.

mteodor (dismissed):
LGTM

docker/nginx/nginx-x509.conf
```diff
@@ -69,154 +55,53 @@ http {
...
         }
 
         # Proxy pass to things service
         location ~ ^/(things|channels) {
-            proxy_redirect off;
-            proxy_set_header Host $host;
-            proxy_set_header X-Real-IP $remote_addr;
-            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-            proxy_set_header X-Forwarded-Proto $scheme;
+            include snippets/proxy-headers.conf;
```
nmarcetic, line 64: Do we need ` include snippets/verify-ssl-client.conf;` here also ?

docker/verne-cluster.yml
```diff
@@ -0,0 +1,72 @@
...
+  mqtt-adapter-1:
+    << : *mqtt-adapter
+    container_name: mainflux-mqtt-1
+    ports:
+      - 18831:${MF_MQTT_ADAPTER_PORT}
+      - 8891:${MF_MQTT_ADAPTER_WS_PORT}
+      - 7777:8888
+
+  mqtt-adapter-2:
+    << : *mqtt-adapter
```
nmarcetic, line 45: Nice trick with anchors, love it.

.gitignore
```diff
@@ -1,2 +1,17 @@
+# Mainflux
 build
-site/
+.vscode
```
anovakovic01: This is specific for the local dev env. Please remove this. (You can add it to your global `.gitignore` on your machine).

anovakovic01 (changes requested):
Just fix gitignore

docker/nginx/nginx-key.conf
```diff
@@ -1,18 +1,18 @@
...
 worker_processes auto;
+worker_cpu_affinity auto;
 pid /run/nginx.pid;
 include /etc/nginx/modules-enabled/*.conf;
 
 events {
-    worker_connections 768;
+    # Explanation: https://serverfault.com/questions/787919/optimal-value-for-nginx-worker-connections
+    # We'll keep 1M connections per core (assuming one worker per core)
+    worker_connections 1000000;
```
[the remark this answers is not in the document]
drasko: I know, but with 10k we already hit the limitation even for simple test of 1000 publishers publishing each 10 messages.

This is not a production set-up, but I wanted to avoid at least that this limit us do the tests.

Makefile
```diff
@@ -148,16 +148,18 @@ rundev:
 	cd scripts && ./run.sh
 
 run:
-	docker-compose -f docker/docker-compose.yml up
+	docker-compose -f docker/docker-compose.yml -f docker/aedes.yml up
```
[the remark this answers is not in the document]
drasko: Aedes contains it's Redis. If I left them into docker-compose, then I would not be capable to replace them with additional `verne.yml`. While it would be possible to have generic `mqtt-adapter` that can be "casted" into Aedes by default and Verne by demand, it is impossible that other .yml removes services from previous (i.e. this Aedes Redis service).

docker/nginx/nginx-x509.conf
drasko: Let me check, if there was one before then we do.

docker/verne-cluster.yml
drasko: Unfortunately, Docker dropped support for `extends` in `v3`, they suggest using this approach now.

.gitignore
```diff
@@ -1,2 +1,17 @@
...
+
+# MQTT
```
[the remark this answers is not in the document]
drasko: I prefer this centralized overview and one-place tuning. For now our structure is small enough, and sections are well commented.

If really needed let's do this in a separate issue, not to delay this PR.

docker/nginx/nginx-x509.conf
drasko: Yes, I checked it out - no. SSL verification is used for mTLS for messaging. `/things` and `/channels` are still accessed by user with his JWT.

docker/nginx/nginx-key.conf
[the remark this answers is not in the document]
drasko: Fixed to follow reccomendation, just in case. If needed can be bumped up for tests, but I guess limitations will first come from Linux TCP buffers and otehr stuff...

anovakovic01 (approved):
LGTM!

dborovcanin (approved):

PR #849, opened 2019-09-11
NOISSUE - Search by metadata

mteodor:
### What does this do?
Add searching things by metadata

Yes
Yes


docs/provisioning.md
```diff
@@ -100,6 +100,14 @@ group of things. In that case, your request should look like:
 curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -H "Authorization: <user_auth_token>" https://localhost/things?offset=0&limit=5
 ```
 
+You can specify `name` and `metadata` parameters in order to fetch specific
```
drasko: and/or

things/api/things/http/transport.go
```diff
@@ -380,3 +387,21 @@ func readStringQuery(r *http.Request, key string) (string, error) {
...
+
+func readMetadataQuery(r *http.Request, key string) (interface{}, error) {
+	vals := bone.GetQuery(r, key)
+	if len(vals) > 1 {
+		return "", errInvalidQueryParams
+	}
+
+	if len(vals) == 0 {
+		return nil, nil
+	}
```
drasko, line 399: If you added a whiteline above this if, add below also. Otherwise kill both.

things/postgres/things.go
```diff
@@ -171,22 +171,29 @@ func (tr thingRepository) RetrieveByKey(_ context.Context, key string) (string,
...
 		nq = `AND LOWER(name) LIKE :name`
 	}
+	mq := ""
+	if metadata != nil {
+		mq = `metadata @> :metadata AND`
+	}
 
 	q := fmt.Sprintf(`SELECT id, name, key, metadata FROM things
-	      WHERE owner = :owner %s ORDER BY id LIMIT :limit OFFSET :offset;`, nq)
+		  WHERE %s owner = :owner %s ORDER BY id LIMIT :limit OFFSET :offset;`, mq, nq)
```
drasko, line 192: `go fmt` did this inline?

drasko (changes requested):
No remaks on the logic. just a minor remarks on a code style.

drasko:
@mteodor this starts to look good.

Just one question - does this impact SDK and CLI?

mteodor:
> @mteodor this starts to look good.
> 
> Just one question - does this impact SDK and CLI?

I dont think so, this affects only list things endpoint which is used in
````
func (sdk mfSDK) Things(token string, offset, limit uint64, name string) (ThingsPage, error)
````
but call to that enpoint will remain unchanged and results will be the same

File: things/postgres/init.go [hunk 4]
```diff
@@ -57,14 +57,14 @@ func migrateDB(db *sqlx.DB) error {
 						owner    VARCHAR(254),
 						key      VARCHAR(4096) UNIQUE NOT NULL,
 						name     VARCHAR(1024),
-						metadata JSON,
+						metadata JSONB,
```
dborovcanin: Please consider adding a separate migration for type change, rather than changing data type this way.

dborovcanin (changes requested):

things/postgres/init.go
drasko: @dusanb94 I do not see the need to support `JSON` anymore. Pre `1.0` can stil break APIs and we are not in obligation to drag with us unneeded field.

things/api/things/http/transport.go
```diff
@@ -380,3 +387,22 @@ func readStringQuery(r *http.Request, key string) (string, error) {
 
 	return vals[0], nil
 }
+
+func readMetadataQuery(r *http.Request, key string) (interface{}, error) {
+	vals := bone.GetQuery(r, key)
+	if len(vals) > 1 {
+		return "", errInvalidQueryParams
```
nmarcetic: Why string and not nil like in next lines?

things/postgres/init.go
```diff
@@ -57,14 +57,14 @@ func migrateDB(db *sqlx.DB) error {
...
 						PRIMARY KEY (id, owner)
 					)`,
 					`CREATE TABLE IF NOT EXISTS channels (
 						id       UUID,
 						owner    VARCHAR(254),
 						name     VARCHAR(1024),
-						metadata JSON,
+						metadata JSONB,
```
nmarcetic: Please do this as part of Channels metadata PR.

things/api/things/http/requests.go
```diff
@@ -135,10 +135,11 @@ func (req viewResourceReq) validate() error {
...
 type listResourcesReq struct {
-	token  string
-	offset uint64
-	limit  uint64
-	name   string
+	token    string
+	offset   uint64
+	limit    uint64
+	name     string
+	metadata interface{}
```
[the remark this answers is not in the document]
nmarcetic: I would create a new type `map[string]interface{}` and add it here and also use it to pass as param on other places. Its easier and easy to change.

things/postgres/init.go
dborovcanin: It would be easier for us because we wouldn't need to change schema manually (no need to drop/change in Docker volumes, nor local DB). I agree we can remove that migration before production because this is <1.0, but adding even temporary migration will help (especially since it's done only once).

nmarcetic (dismissed):

docs/provisioning.md
```diff
@@ -100,6 +100,14 @@ group of things. In that case, your request should look like:
...
 ```
 
+You can specify `name` and/or `metadata` parameters in order to fetch specific
+group of things. When specifiying metadata you can specify just a part of the metadata json you want to match
+
+```
+curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -H "Authorization: <user_auth_token>" https://localhost/things?offset=0&limit=5&metadata={"serial":"123456"}
+```
+
+
```
drasko: No need for 2 whitelines - one should be enough.

things/postgres/things.go
```diff
@@ -171,22 +171,33 @@ func (tr thingRepository) RetrieveByKey(_ context.Context, key string) (string,
 	return id, nil
 }
 
-func (tr thingRepository) RetrieveAll(_ context.Context, owner string, offset, limit uint64, name string) (things.ThingsPage, error) {
+func (tr thingRepository) RetrieveAll(_ context.Context, owner string, offset, limit uint64, name string, metadata things.ThingMetadata) (things.ThingsPage, error) {
 	name = strings.ToLower(name)
 	nq := ""
+	var m []byte
```
drasko: Why is this declared here? I see it inicialilzed below, where is can be auto-declared.

drasko: BTW. Go's reccomended/idiomatic way of declaration is using `:=` in favor of `var`, so lean to this approach whenever possible.

drasko (changes requested):

things/postgres/things.go
nmarcetic: Following FP principles use var/const when you are declaring empty values.

nmarcetic (dismissed):

dborovcanin (dismissed):

nmarcetic (dismissed):

things/postgres/init.go
```diff
@@ -57,7 +57,7 @@ func migrateDB(db *sqlx.DB) error {
 						owner    VARCHAR(254),
 						key      VARCHAR(4096) UNIQUE NOT NULL,
 						name     VARCHAR(1024),
-						metadata JSON,
+						metadata JSONB,
```
anovakovic01: Please remove this because you have the migration later on.

anovakovic01 (changes requested):

drasko (commented):
LGTM

drasko (dismissed):

anovakovic01 (approved):
LGTM!

nmarcetic (approved):

PR #852, opened 2019-09-12
NOISSUE - Simplify MQTT benchmarking tool

dborovcanin:
### What does this do?
This pull request simplifies our MQTT benchmarking tool.

There are no such changes.

No.

No.


tools/mqtt-bench/bench.go
```diff
@@ -4,102 +4,27 @@
...
-}
-
-// Config struct holds benchmark configuration
-type Config struct {
-	MQTT mqttConfig   `toml:"mqtt" mapstructure:"mqtt"`
-	Test testConfig   `toml:"test" mapstructure:"test"`
-	Log  logConfig    `toml:"log" mapstructure:"log"`
-	Mf   mainfluxFile `toml:"mainflux" mapstructure:"mainflux"`
-}
+const fixedSize = 411
```
drasko: Please comment this magic number

tools/mqtt-bench/templates/config.toml
```diff
@@ -1,26 +1,26 @@
 [mqtt]
+  timeout = 1000
```
drasko: If this is `config.toml` that makes best performance, I suggest that we rename it to something like `reference.toml` or something, because I am afraid that someone will overwrite this config.

Also - I am not sure that in templates we really need a file called `config.toml` - this one must be produced manually by copying some of the templates for example in the root dir and then custommizing.

drasko (changes requested):

drasko (approved):
LGTM

nmarcetic (approved):

PR #855, opened 2019-09-13
MF-730 - Add digital twin service for things

### LINKED ISSUE #730: Thing digital twin concept
Considering digital twin concept, thinking about benefits and metadata search capabilities, world state vs known state, device status, queued messages/commands,  etc..

References:
[The Role Of Device Twins In Industrial IoT Solutions](https://www.forbes.com/sites/janakirammsv/2017/10/30/the-role-of-device-twins-in-designing-industrial-iot-solutions/#2b5d292c20ac)
[Understand Azure IoT Hub device twins](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-device-twins)

darkodraskovic:
Resolves #730 

We need to change `.env`, `docker-compose.yml` as well as `Makefile`. However, none of this changes should break any of the current functionalities.

Not yet, will be included in the final PR.

Work in progress.


.env
```diff
@@ -152,3 +152,10 @@ MF_POSTGRES_READER_DB_SSL_MODE=disable
 MF_POSTGRES_READER_DB_SSL_CERT=""
 MF_POSTGRES_READER_DB_SSL_KEY=""
 MF_POSTGRES_READER_DB_SSL_ROOT_CERT=""
+
+## Twin
+MF_TWINS_LOG_LEVEL=debug
+MF_TWINS_HTTP_PORT=9021
+MF_TWINS_SECRET=secret
+MF_TWINS_SERVER_CERT=
+MF_TWINS_SERVER_KEY=
```
manuio: Add brackets here to make it clearer please

twins/api/twins/http/requests.go
```diff
@@ -0,0 +1,108 @@
...
+		return twins.ErrUnauthorizedAccess
+	}
+
+	return nil
+}
+
+type addTwinReq struct {
+	token    string
+	Name     string                 `json:"name,omitempty"`
+	Key      string                 `key:"key,omitempty"`
```
manuio: I think that first `key` should be `json`, right?

manuio (changes requested):

cmd/twins/main.go
```diff
@@ -0,0 +1,309 @@
...
+	cfg := loadConfig()
+
+	logger, err := logger.New(os.Stdout, cfg.logLevel)
+	if err != nil {
+		log.Fatalf(err.Error())
+	}
+
+	db, err := twinsmongodb.Connect(cfg.dbCfg, logger)
+	if err != nil {
+		log.Fatalf(err.Error())
```
manuio: You could use logger instead

manuio (changes requested):

cmd/twins/main.go
```diff
@@ -22,8 +22,10 @@ import (
 	mqtt "github.com/eclipse/paho.mqtt.golang"
 	"github.com/mainflux/mainflux"
 	"github.com/mainflux/mainflux/logger"
+	"github.com/mainflux/mainflux/normalizer"
```
drasko: No need for Noramlizer at all. Normalizer is a service that subscribes to NATS and reads messages which are "unnormalized" SenML (by the standard SenML can be an array of JSON structures, called "unnromalized", and can parse SenML content and transform it to "normalized" - by breaking this array in individual messages which is shoots back to NATS.

Writers cn read only "normalized" SenML - i.e. they know to read only individual messages, not an array, so they write these messages one by one into DB.

Normalizer is a common procedure for all Writers, so instaed of putting SenML normalization procedure in each writer, this is separates into individual service.

Twin service will not need to normalize anything, as messages (content) comming over Twin state channel will be in the form of Twin state - i.e. arbitrary JSON that represents new state - and should be written "tel quel" into the Twin's Mongo DB.

twins/uuid/idp.go
```diff
@@ -0,0 +1,40 @@
+//
```
drasko: Why do you need this? there is JWT used for something?

twins/users/users.go
```diff
@@ -0,0 +1,45 @@
+//
```
drasko: What is this? Why users? I mean why Twins servis has a differnet user than Mainflux?

twins/nats/pubsub.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	if err := ps.publish(msg); err != nil {
+		ps.logger.Warn(fmt.Sprintf("Publishing failed: %s", err))
+		return
+	}
+}
+
+func (ps pubsub) publish(msg mainflux.RawMessage) error {
+	output := mainflux.OutputSenML
```
drasko: No need for SenML and normalization.

twins/mocks/users.go
```diff
@@ -0,0 +1,34 @@
+//
```
drasko: Again - why users? This service should use the `user` of Mainflux

twins/mocks/idp.go
```diff
@@ -0,0 +1,47 @@
+//
```
drasko: No need for this file IMHO

twins/idp.go
```diff
@@ -0,0 +1,17 @@
+//
```
drasko: No need for IDP here IMHO

drasko (changes requested):

manuio (changes requested):

twins/twins.go
```diff
@@ -0,0 +1,78 @@
...
+	"time"
+)
+
+type Metadata map[string]interface{}
+
+// Twin represents a Mainflux thing digital twin. Each twin is owned by one thing, and
+// is assigned with the unique identifier and (temporary) access key.
+type Twin struct {
+	ID         string
+	Key        string
```
nmarcetic: What is a key ?

twins/twins.go
```diff
@@ -0,0 +1,78 @@
...
+	Owner      string
+	Name       string
+	thingID    string
+	channelID  string
```
nmarcetic: What is a ChannelID ?

twins/twins.go
```diff
@@ -0,0 +1,78 @@
...
+	created    time.Time
+	updated    time.Time
+	attributes map[string]interface{}
+	state      map[string]interface{}
+	topic      string
+	revision   int
+	Metadata   Metadata
```
nmarcetic: You don't need metadata

twins/twins.go
```diff
@@ -0,0 +1,78 @@
...
+	// error response.
+	Save(context.Context, Twin) (string, error)
+
+	// Update performs an update to the existing twin. A non-nil error is
+	// returned to indicate operation failure.
+	Update(context.Context, Twin) error
+
+	// UpdateKey performs an update key to the existing twin. A non-nil error is
+	// returned to indicate operation failure.
+	UpdateKey(ctx context.Context, owner, id, key string) error
```
nmarcetic: What is a key ?

twins/twins.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// RetrieveByKey retrieves the twin having the provided key.
+	RetrieveByKey(context.Context, string) (string, error)
+
+	// RetrieveAll retrieves the subset of things owned by the specified user.
+	RetrieveAll(context.Context, string, uint64, string, Metadata) (TwinsSet, error)
+
+	// RetrieveByChannel retrieves the subset of twins
+	// connected to specified channel.
+	RetrieveByChannel(context.Context, string, uint64) (TwinsSet, error)
```
nmarcetic: This is unnecessary.

twins/twins.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// UpdateKey performs an update key to the existing twin. A non-nil error is
+	// returned to indicate operation failure.
+	UpdateKey(ctx context.Context, owner, id, key string) error
+
+	// RetrieveByID retrieves the twin having the provided identifier.
+	RetrieveByID(ctx context.Context, owner, id string) (Twin, error)
+
+	// RetrieveByKey retrieves the twin having the provided key.
+	RetrieveByKey(context.Context, string) (string, error)
```
nmarcetic: What is a key ?

twins/twins.go
```diff
@@ -0,0 +1,92 @@
...
+	Created    time.Time
+	Payload    map[string]interface{}
+}
+
+// Twin represents a Mainflux thing digital twin. Each twin is owned by one thing, and
+// is assigned with the unique identifier and (temporary) access key.
+type Twin struct {
+	Owner       string
+	ID          string
+	Key         string
```
nmarcetic: What is Key ?

twins/twins.go
```diff
@@ -0,0 +1,92 @@
...
+type Definition struct {
+	Revision   int
+	Created    time.Time
+	Attributes map[string]interface{}
+}
+
+// State stores actual snapshot of entity's values
+type State struct {
+	Definition int
+	Serial     int
```
nmarcetic: What is Serial?

twins/nats/senml.go
```diff
@@ -0,0 +1,22 @@
+package nats
+
+// Record represents one senML record.
+type Record struct {
```
nmarcetic: You have this in mainflux/senml lib no ? Maybe is better idea to use structure from lib, where you have validation and everything not just copy/paste it.

twins/mongodb/twins.go
```diff
@@ -0,0 +1,229 @@
...
+	if len(tw.Name) > maxNameSize {
+		return "", twins.ErrMalformedEntity
+	}
+
+	coll := tr.db.Collection(twinsCollection)
+
+	if _, err := tr.RetrieveByID(ctx, tw.ID); err == nil {
+		return "", twins.ErrConflict
+	}
+	if _, err := tr.RetrieveByKey(ctx, tw.Key); err == nil {
```
nmarcetic: What exactly you want to accomplish here ? To make sure that ID and Key are unique ? I don't like 2 additional queries in Save().

nmarcetic: Maybe you can check Indexing 
```Inserting an duplicate value for any key that is part of a unique index, such as _id, throws an exception. The following attempts to insert a document with a _id value that already exists:```

**Review by nmarcetic — CHANGES_REQUESTED** [review]

File: docker/docker-compose.yml [hunk 21]
```diff
@@ -109,7 +111,8 @@ services:
       - ./users/emailer/templates/${MF_EMAIL_TEMPLATE}:/${MF_EMAIL_TEMPLATE}
     depends_on:
       - users-db
-      - authn
+    # expose:
```
drasko: We should not really commit commented code like this on master.

twins/api/doc.go
```diff
@@ -0,0 +1,10 @@
+//
+// Copyright (c) 2018
+// Mainflux
```
drasko: Condense this to one line

twins/api/logging.go
```diff
@@ -0,0 +1,122 @@
+//
+// Copyright (c) 2019
```
drasko: Same here - condense and remove year (align with new header style)

twins/api/metrics.go
```diff
@@ -0,0 +1,99 @@
+//
+// Copyright (c) 2019
```
drasko: Same here

twins/api/twins/http/doc.go
```diff
@@ -0,0 +1,9 @@
+//
+// Copyright (c) 2019
+// Mainflux
```
drasko: Same here

drasko (changes requested):

twins/mongodb/twins.go
```diff
@@ -0,0 +1,229 @@
...
+
+// Save persists the twin
+func (tr *twinRepository) Save(ctx context.Context, tw twins.Twin) (string, error) {
+	if len(tw.Name) > maxNameSize {
+		return "", twins.ErrMalformedEntity
+	}
+
+	coll := tr.db.Collection(twinsCollection)
+
+	if _, err := tr.RetrieveByID(ctx, tw.ID); err == nil {
```
mteodor: when is this ID created?
doesnt the mongo driver take care that there is already entity with ID?

twins/api/logging.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+		lm.logger.Info(fmt.Sprintf("%s without errors.", message))
+	}(time.Now())
+
+	return lm.svc.ListStates(ctx, token, offset, limit, id)
+}
+
+func (lm *loggingMiddleware) ViewTwinByThing(ctx context.Context, token, thingid string) (tw twins.Twin, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method view_twin_by_thing for for token %s and thing %s took %s to complete", token, thingid, time.Since(begin))
```
manuio: fix `for for token`

twins/api/logging.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+		lm.logger.Info(fmt.Sprintf("%s without errors.", message))
+	}(time.Now())
+
+	return lm.svc.ViewTwinByThing(ctx, token, thingid)
+}
+
+func (lm *loggingMiddleware) RemoveTwin(ctx context.Context, token, id string) (err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method remove_twin for for token %s and twin %s took %s to complete", token, id, time.Since(begin))
```
manuio: fix `for for token`

manuio (changes requested):

twins/paho/pubsub.go
```diff
@@ -0,0 +1,82 @@
...
+func New(mc paho.Client, topic string) Mqtt {
+	return Mqtt{
+		client: mc,
+		topic:  topic,
+	}
+}
+
+// Connect to MQTT broker
+func Connect(mqttURL, id, key string, logger logger.Logger) paho.Client {
+	opts := paho.NewClientOptions()
```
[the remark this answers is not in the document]
mteodor: you can check here https://github.com/mainflux/export/blob/master/cmd/main.go#L263

twins/api/twins/http/doc.go
```diff
@@ -0,0 +1,9 @@
+//
```
dborovcanin: There is no need for `twins/api/twins/http/`, `twins/api` is sufficient.

dborovcanin (changes requested):

twins/mocks/twins.go
```diff
@@ -0,0 +1,164 @@
...
+	defer trm.mu.Unlock()
+
+	for _, tw := range trm.twins {
+		if tw.ID == twin.ID {
+			return "", twins.ErrConflict
+		}
+	}
+
+	for _, tw := range trm.twins {
+		if tw.Key == twin.Key {
```
drasko: Could not this be checked in the loop above - do we need to loop two times?

twins/mongodb/twins.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	return tw, nil
+}
+
+func decodeTwins(ctx context.Context, cur *mongo.Cursor) ([]twins.Twin, error) {
+	defer cur.Close(ctx)
+	var results []twins.Twin
+	for cur.Next(ctx) {
+		var elem twins.Twin
```
drasko: Create elem var before loop, do not allocate new var for every pass through the loop.

twins/nats/pubsub.go
```diff
@@ -0,0 +1,138 @@
...
+		mqttClient: mc,
+		logger:     logger,
+		twins:      tr,
+		states:     sr,
+	}
+	ps.natsClient.QueueSubscribe(input, queue, ps.handleMsg)
+}
+
+func (ps pubsub) handleMsg(m *nats.Msg) {
+	var err error
```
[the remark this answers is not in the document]
drasko: I also think this is not needed, `err` is special var that is reusable in Go.

twins/service.go
```diff
@@ -0,0 +1,258 @@
...
+		twins:      twins,
+		states:     sr,
+		idp:        idp,
+	}
+}
+
+func (ts *twinsService) AddTwin(ctx context.Context, token string, twin Twin, def Definition) (tw Twin, err error) {
+	id := ""
+	b := []byte{}
+	defer ts.mqttClient.Publish(&id, &err, "create/success", "create/failure", &b)
```
drasko: Are these topics hard-coded? If yes, move them to string constants.

twins/service.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func isZeroOfUnderlyingType(x interface{}) bool {
+	return reflect.DeepEqual(x, reflect.Zero(reflect.TypeOf(x)).Interface())
+}
+
+func (ts *twinsService) UpdateTwin(ctx context.Context, token string, twin Twin, def Definition) (err error) {
+	b := []byte{}
+	id := ""
+	defer ts.mqttClient.Publish(&id, &err, "update/success", "update/failure", &b)
```
drasko: Move to string constants.

twins/service.go
```diff
@@ -0,0 +1,258 @@
...
+
+	id = twin.ID
+	b, err = json.Marshal(tw)
+
+	return nil
+}
+
+func (ts *twinsService) ViewTwin(ctx context.Context, token, id string) (tw Twin, err error) {
+	b := []byte{}
+	defer ts.mqttClient.Publish(&id, &err, "get/success", "get/failure", &b)
```
drasko: Same here, string constants.

twins/service.go
```diff
@@ -0,0 +1,258 @@
...
+	if err != nil {
+		return Twin{}, ErrUnauthorizedAccess
+	}
+
+	return ts.twins.RetrieveByThing(ctx, thingid)
+}
+
+func (ts *twinsService) RemoveTwin(ctx context.Context, token, id string) (err error) {
+	b := []byte{}
+	defer ts.mqttClient.Publish(&id, &err, "remove/success", "remove/failure", &b)
```
drasko: Same here.

drasko (changes requested):

drasko:
Very good job in general, code looks of high quality. PR is huge, so please take our remarks into account and let's get closer to merge.

twins/twins.go
darkodraskovic: It's a unique twin key.

twins/paho/pubsub.go
```diff
@@ -0,0 +1,82 @@
+package paho
```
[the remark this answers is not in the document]
darkodraskovic: Yes, sure, but we call the other stuff nats, etc. and besides, @manuio replaces mqtt by paho in edgent. @dusanb94 what do you think here?

twins/service.go
```diff
@@ -0,0 +1,258 @@
...
+	ErrNotFound = errors.New("non-existent entity")
+
+	// ErrConflict indicates that entity already exists.
+	ErrConflict = errors.New("entity already exists")
+)
+
+// Service specifies an API that must be fullfiled by the domain service
+// implementation, and all of its decorators (e.g. logging & metrics).
+type Service interface {
+	// AddTwin adds new twin to the user identified by the provided key.
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

twins/doc.go
```diff
@@ -0,0 +1,11 @@
+//
+// Copyright (c) 2018
+// Mainflux
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+// Package things contains the domain concept definitions needed to support
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

twins/paho/pubsub.go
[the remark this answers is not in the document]
nmarcetic: nats is the name of the broker that we use and client lib also, paho is the name of the client lib only, not the SW that we use (like VerneMQ for example). Not really the same. 
`replaces mqtt by paho in edgent` IMHO not good idea.

twins/nats/pubsub.go
```diff
@@ -0,0 +1,138 @@
...
+	if err := proto.Unmarshal(m.Data, &msg); err != nil {
+		ps.logger.Warn(fmt.Sprintf("Unmarshalling failed: %s", err))
+		return
+	}
+	if msg.Channel == ps.mqttClient.Topic() {
+		return
+	}
+
+	b := []byte{}
+	id := ""
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

twins/nats/pubsub.go
```diff
@@ -0,0 +1,138 @@
...
+	defer ps.mqttClient.Publish(&id, &err, "state/success", "state/failure", &b)
```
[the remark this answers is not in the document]
darkodraskovic: Fixed with `map[string]string`. Please recheck the solution.

twins/nats/pubsub.go
```diff
@@ -0,0 +1,138 @@
...
+
+	tw, err := ps.twins.RetrieveByThing(context.TODO(), msg.Publisher)
+	if err != nil {
+		ps.logger.Warn(fmt.Sprintf("Retrieving twin for %s failed: %s", msg.Publisher, err))
+		return
+	}
+
+	var recs []Record
```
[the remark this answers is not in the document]
darkodraskovic: I've put it there because I've defined it right before the line where I use it. Seems clearer like this to me. I think that's how we do it generally. What do you think?

twins/nats/pubsub.go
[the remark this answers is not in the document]
darkodraskovic: Yes I need it, because of the mqtt line
```defer ps.mqttClient.Publish(&id, &err, mqttOp["stateSucc"], mqttOp["stateFail"], &b)```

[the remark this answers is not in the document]
Comment (darkodraskovic) [reply twins/nats/pubsub.go]: Ok, found the way to remove it.

File: twins/nats/senml.go (hunk 19 above)
Comment (darkodraskovic) [reply twins/nats/senml.go]: Yes, at the moment of coding, it was not there. Will use from `mainflux/senml`.

File: twins/nats/pubsub.go (hunk 43 above)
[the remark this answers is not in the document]
Comment (nmarcetic) [reply twins/nats/pubsub.go]: Ah I see, I saw also some others var's declared at the fn begin so I was thinking that you defined all var's there and you missed this one :) All good.

File: docker/docker-compose.yml (hunk 21 above)
Comment (darkodraskovic) [reply docker/docker-compose.yml]: I've reverted to the master version of `docker-compose.yml`

File: cmd/twins/main.go [hunk 44]
```diff
@@ -0,0 +1,319 @@
...
+	authapi "github.com/mainflux/mainflux/authn/api/grpc"
+	"github.com/mainflux/mainflux/logger"
+	localusers "github.com/mainflux/mainflux/things/users"
+	"github.com/mainflux/mainflux/twins"
+	"github.com/mainflux/mainflux/twins/api"
+	twinshttpapi "github.com/mainflux/mainflux/twins/api/twins/http"
+	twinsmongodb "github.com/mainflux/mainflux/twins/mongodb"
+	"github.com/mainflux/mainflux/twins/nats"
+	"github.com/mainflux/mainflux/twins/uuid"
+	broker "github.com/nats-io/go-nats"
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

docker/docker-compose.yml
```diff
@@ -48,6 +48,8 @@ services:
     image: nats:1.3.0
     container_name: mainflux-nats
     restart: on-failure
+    ports:
+      - 4222:4222
```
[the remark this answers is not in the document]
darkodraskovic: It's for the testing purposes. Removed it from the pushed `docker-compose.yml`.

twins/mongodb/twins.go
darkodraskovic: We need a separate instance of Twin, because we append it to an array of twins. Can't use the same structure instance for every instance of twin.

twins/twins.go
darkodraskovic: It's a unique twin key. Dunno if we need it at all... @nmarcetic ?

twins/node-red/flows.json
```diff
@@ -0,0 +1,258 @@
+[
```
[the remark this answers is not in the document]
darkodraskovic: No, we don't need this. Will be removed.

twins/nats/senml.go
darkodraskovic: Fixed.

twins/twins.go
nmarcetic: I don't think so. Twin doesn't need a key, because it's just a representation (e.g Twin can't update itself or do anything actually).

twins/mocks/idp.go
darkodraskovic: Needed for tests.

twins/idp.go
darkodraskovic: Needed for tests.

twins/twins.go
darkodraskovic: Removed Key from Twin.

twins/twins.go
darkodraskovic: Removed Key from Twin.

twins/mongodb/twins.go
```diff
@@ -0,0 +1,209 @@
...
+
+// Save persists the twin
+func (tr *twinRepository) Save(ctx context.Context, tw twins.Twin) (string, error) {
+	if len(tw.Name) > maxNameSize {
+		return "", twins.ErrMalformedEntity
+	}
+
+	coll := tr.db.Collection(twinsCollection)
+
+	if _, err := tr.RetrieveByID(ctx, tw.ID); err == nil {
```
nmarcetic: @darkodraskovic Did you conside a [unique index](https://docs.mongodb.com/manual/core/index-unique/)? To avoid this lookup before save.

darkodraskovic: @nmarcetic as per discussion with @drasko I've decided to simply use a uuid for the id in the case we use some other database. mongo's `_id` is still relevant for states, since they need to be ordered as time series.

nmarcetic: But the unique index is not related to _id, it can be used for any field (for example your ID generated by UUID), the `_id` is just an example in the doc so you understand 

```
A unique index ensures that the indexed fields do not store duplicate values; i.e. enforces uniqueness for the indexed fields. By default, MongoDB creates a unique index on the _id field during the creation of a collection.
```

My question is, can you consider using a unique index for ID field (generated by UUID) instead of or manually check uniqueness using `RetrieveByID`

twins/twins.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"context"
+	"time"
+)
+
+// Metadata stores arbitrary twin data
+type Metadata map[string]interface{}
+
+// Attribute stores indivdual attribute data
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

twins/twins.go
```diff
@@ -0,0 +1,85 @@
...
+type Attribute struct {
+	Channel     string
+	Subtopic    string
+	PeristState bool
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

twins/states.go
```diff
@@ -0,0 +1,37 @@
...
+
+// StateRepository specifies a state persistence API.
+type StateRepository interface {
+	// Save persists the state
+	Save(context.Context, State) error
+
+	// Count returns the number of states related to state
+	Count(context.Context, Twin) (int64, error)
+
+	// RetrieveAll retrieves the subset of staes related to twin specified by id
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

twins/service.go
```diff
@@ -0,0 +1,258 @@
...
+		tw.Key = twin.Key
+	}
+
+	if !isZeroOfUnderlyingType(twin.Name) {
+		tw.Name = twin.Name
+	}
+
+	if !isZeroOfUnderlyingType(twin.ThingID) {
+		tw.ThingID = twin.ThingID
+	}
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

cmd/twins/main.go
```diff
@@ -0,0 +1,319 @@
...
+	"github.com/mainflux/mainflux/twins/paho"
+
+	kitprometheus "github.com/go-kit/kit/metrics/prometheus"
+	"github.com/mainflux/mainflux"
+	authapi "github.com/mainflux/mainflux/authn/api/grpc"
+	"github.com/mainflux/mainflux/logger"
+	localusers "github.com/mainflux/mainflux/things/users"
+	"github.com/mainflux/mainflux/twins"
+	"github.com/mainflux/mainflux/twins/api"
+	twinshttpapi "github.com/mainflux/mainflux/twins/api/twins/http"
```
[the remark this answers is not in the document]
darkodraskovic: Maybe `twapi` and `twmongo` rather?

twins/README.md
```diff
@@ -0,0 +1,67 @@
+# Twins
+
+Twins service provides a barebones HTTP API for development of a Mainflux
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

twins/api/twins/http/doc.go
darkodraskovic: Moved to `twins/api/http`.

docker/docker-compose.yml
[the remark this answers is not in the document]
drasko: I am not against leaving this one, our docker-compose.yml in the official repo is a testing one (not suitable for professional deployment).

drasko (changes requested):

twins/nats/pubsub.go
```diff
@@ -0,0 +1,139 @@
...
+		twins:      tr,
+		states:     sr,
+	}
+	ps.natsClient.QueueSubscribe(input, queue, ps.handleMsg)
+}
+
+func (ps pubsub) handleMsg(m *nats.Msg) {
+	var msg mainflux.Message
+	err := proto.Unmarshal(m.Data, &msg)
+	if err != nil {
```
anovakovic01: You can inline this.

twins/node-red/flows.json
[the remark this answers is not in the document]
anovakovic01: Please remove this file.

twins/service.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		return Twin{}, err
+	}
+
+	twin.Owner = res.GetValue()
+
+	twin.Created = time.Now()
+	twin.Updated = time.Now()
+
+	fmt.Println(def)
```
anovakovic01: Remove println.

twins/service.go
```diff
@@ -0,0 +1,247 @@
...
+		def = Definition{}
+		def.Attributes = make(map[string]Attribute)
+	}
+	def.Created = time.Now()
+	def.ID = 0
+	twin.Definitions = append(twin.Definitions, def)
+
+	twin.Revision = 0
+	_, err = ts.twins.Save(ctx, twin)
+	if err != nil {
```
anovakovic01: You can line this.

twins/nats/pubsub.go
```diff
@@ -0,0 +1,139 @@
...
+
+	if save := prepareState(&st, &tw, recs, msg); !save {
+		ps.logger.Info(fmt.Sprintf("No persistent attributes for %s for %s", msg.Subtopic, msg.Publisher))
+		return
+	}
+
+	if err := ps.states.Save(context.TODO(), st); err != nil {
+		ps.logger.Warn(fmt.Sprintf("Updating state for %s failed: %s", msg.Publisher, err))
+		return
+	}
```
dborovcanin: Is it possible to move this logic to Twin service? Also, can you try to simplify this logic? Retrieving twin and state, then storing state for each message is very heavy. You can consider replacing this with a single, more complex update query.

dborovcanin (changes requested):

anovakovic01 (changes requested):
Please consider using Mainflux `errors` package for error handling.

docker/docker-compose.yml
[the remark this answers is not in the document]
manuio: You could use the envar from .env here

manuio (changes requested):

twins/api/http/responses.go
```diff
@@ -0,0 +1,147 @@
...
+func (res viewTwinRes) Headers() map[string]string {
+	return map[string]string{}
+}
+
+func (res viewTwinRes) Empty() bool {
+	return false
+}
+
+type viewStateRes struct {
+	TwinID     string                 `json:"twinid"`
```
manuio: Use `twin_id` please

twins/api/http/requests.go
```diff
@@ -0,0 +1,128 @@
...
+const maxLimitSize = 100
+
+type apiReq interface {
+	validate() error
+}
+
+type addTwinReq struct {
+	token      string
+	Name       string                 `json:"name,omitempty"`
+	ThingID    string                 `json:"thingID"`
```
manuio: Use `thing_id` instead of `thingID`

manuio (changes requested):

twins/api/http/requests.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return nil
+}
+
+type updateTwinReq struct {
+	token      string
+	id         string
+	Name       string                 `json:"name,omitempty"`
+	ThingID    string                 `json:"thingID,omitempty"`
```
manuio: Use `thing_id` instead of `thingID`

twins/api/http/responses.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (res twinRes) Empty() bool {
+	return true
+}
+
+type viewTwinRes struct {
+	Owner       string                 `json:"owner,omitempty"`
+	ID          string                 `json:"id"`
+	ThingID     string                 `json:"thingID"`
```
manuio: Use `thing_id` instead of `thingID`

manuio (changes requested):

twins/mongodb/twins_test.go
```diff
@@ -0,0 +1,375 @@
...
+			total:  2,
+		},
+		"retrieve twins with non-existing name": {
+			offset: 0,
+			limit:  n,
+			name:   "wrong",
+			size:   0,
+			total:  0,
+		},
+		// "retrieve twins with metadata": {
```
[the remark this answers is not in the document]
darkodraskovic: These tests does not pass! I didn't know how to make them work, so I've commented them. Should I remove them completely? @drasko and @anovakovic01

twins/twins.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Attribute stores individual attribute data
+type Attribute struct {
+	Channel      string
+	Subtopic     string
+	PersistState bool
+}
+
+// Definition stores entity's attributes
+type Definition struct {
```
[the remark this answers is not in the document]
darkodraskovic: There are two IDs. One is twin's ID. The other is definition's ID. The definition's ID correspond to the ordinal number of definitions, and bumps up twin's revision number. We want to know when the definition, which consisists and is defined entirely by attributes is created. That is why we do not store ID and timestamp in the attributes - they are dependent entities and share the single creation time.

twins/mqtt/pubsub.go
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) Mainflux
```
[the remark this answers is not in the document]
darkodraskovic: In fact, it's the other way around :) There is only subscription. I never use publish func, so will remove it.

twins/mqtt/pubsub.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	token := mqtt.client.Publish(topic, 0, false, *payload)
+	token.Wait()
+
+	return token.Error()
+}
+
+// Publish sends mqtt message to a predefined topic
+func (mqtt *Mqtt) Publish(id *string, err *error, succOp, failOp string, payload *[]byte) error {
```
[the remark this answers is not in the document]
darkodraskovic: yes, because `defer` func in

```	
var id string
var b []byte
defer ts.mqttClient.Publish(&id, &err, mqttOp["createSucc"], mqttOp["createFail"], &b)
```
is evaluated at the moment of it's definition. The `id` simply taken is the empty string at the moment of defer's definition. It is populated (or not, in the case of error) with the value during the execution of the function.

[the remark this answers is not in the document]
darkodraskovic: Because we want to send a user information that his request failed. It's by design. @nmarcetic ^

twins/nats/pubsub.go
```diff
@@ -0,0 +1,139 @@
...
+var mqttOp = map[string]string{
+	"stateSucc": "state/success",
+	"stateFail": "state/failure",
+}
+
+type pubsub struct {
+	natsClient *nats.Conn
+	mqttClient mqtt.Mqtt
+	logger     log.Logger
+	twins      twins.TwinRepository
```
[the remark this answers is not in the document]
darkodraskovic: Because we do not want to have overhead of using a service when we only need repos. Should I use service rather? I think the latter is not a wise idea, since you update state very often. You don't want to pass by the service overhead. @nmarcetic ^

twins/nats/pubsub.go
darkodraskovic: We have to store the state for each message. That is the business logic of the service. Any single attribute change implies the new state. For now, for the purpose of simplicity, we send only one measurement per message. It's by design for the first version. 

Sure, I can put this in service. Is it a good idea? @nmarcetic ^

For retrieving twin and state - the last state must be retrieved, the twin also. Because we have to compare definition and measurement. Maybe some form of caching of twins and states? @nmarcetic 

@dusanb94 can you give me some pointers towards the simplification of this func. You should keep in mind that we keep twin and its related stated in the separate db collections.

cmd/twins/main.go
```diff
@@ -0,0 +1,309 @@
...
+	defSingleUserToken = ""
+	defClientTLS       = "false"
+	defCACerts         = ""
+	defMqttURL         = "tcp://localhost:1883"
+	defThingID         = ""
+	defThingKey        = ""
+	defChannelID       = ""
+	defNatsURL         = nats.DefaultURL
+
+	defAuthnHTTPPort = "8989"
```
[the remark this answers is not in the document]
darkodraskovic: Removed.

cmd/twins/main.go
```diff
@@ -0,0 +1,309 @@
...
+	if err != nil {
+		logger.Error(err.Error())
+		os.Exit(1)
+	}
+
+	authTracer, authCloser := initJaeger("auth", cfg.jaegerURL, logger)
+	defer authCloser.Close()
+
+	auth, close := createAuthClient(cfg, authTracer, logger)
+	if close != nil {
```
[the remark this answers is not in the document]
darkodraskovic: I've took this from things service. Makes sense what you say. @dusanb94 ^

cmd/twins/main.go
```diff
@@ -0,0 +1,309 @@
...
+
+	return conn
+}
+
+func newService(nc *nats.Conn, ncTracer opentracing.Tracer, mc mqtt.Mqtt, mcTracer opentracing.Tracer, users mainflux.AuthNServiceClient, dbTracer opentracing.Tracer, db *mongo.Database, logger logger.Logger) twins.Service {
+	twinRepo := twmongodb.NewTwinRepository(db)
+	stateRepo := twmongodb.NewStateRepository(db)
+	idp := uuid.New()
+
+	// TODO twinRepo = tracing.TwinRepositoryMiddleware(dbTracer, thingsRepo)
```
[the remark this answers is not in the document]
darkodraskovic: No, in fact, I need a help on this one. @dusanb94

twins/api/http/endpoint.go
```diff
@@ -0,0 +1,217 @@
...
+		return res, nil
+	}
+}
+
+func removeTwinEndpoint(svc twins.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(viewTwinReq)
+
+		err := req.validate()
+		if err == twins.ErrNotFound {
```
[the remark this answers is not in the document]
darkodraskovic: Yes, some left over code from who knows what and when :) Removed.

twins/api/http/endpoint_test.go
```diff
@@ -0,0 +1,461 @@
...
+	thingID     = "5b68df78-86f7-48a6-ac4f-bb24dd75c39e"
+	wrongID     = 0
+	maxNameSize = 1024
+	natsURL     = "nats://localhost:4222"
+	topic       = "topic"
+)
+
+var (
+	invalidName = strings.Repeat("m", maxNameSize+1)
+)
```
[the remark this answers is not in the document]
darkodraskovic: Done.

twins/api/http/transport.go
```diff
@@ -0,0 +1,303 @@
...
+		return nil, errInvalidQueryParams
+	}
+
+	if len(vals) == 0 {
+		return nil, nil
+	}
+
+	m := make(map[string]interface{})
+	err := json.Unmarshal([]byte(vals[0]), &m)
+	if err != nil {
```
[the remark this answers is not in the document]
darkodraskovic: Actually, this is just taken from things service. I've never tested it. I think that @mteodor wrote this. @mteodor ^

twins/mongodb/states.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/mainflux/mainflux/twins"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+const (
+	statesCollection string = "states"
+)
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

twins/mongodb/states.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func decodeStates(ctx context.Context, cur *mongo.Cursor) ([]twins.State, error) {
+	defer cur.Close(ctx)
+
+	var results []twins.State
+	for cur.Next(ctx) {
+		var elem twins.State
+		err := cur.Decode(&elem)
+		if err != nil {
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

twins/mongodb/states.go
```diff
@@ -0,0 +1,143 @@
...
+	filter := bson.D{{"twinid", tw.ID}}
+	total, err := coll.CountDocuments(ctx, filter)
+	if err != nil {
+		return 0, err
+	}
+
+	return total, nil
+}
+
+func decodeStates(ctx context.Context, cur *mongo.Cursor) ([]twins.State, error) {
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

twins/mongodb/states.go
```diff
@@ -0,0 +1,143 @@
...
+		results = append(results, elem)
+	}
+
+	if err := cur.Err(); err != nil {
+		return []twins.State{}, nil
+	}
+	return results, nil
+}
+
+// RetrieveAll retrieves the subset of staes related to twin specified by id
```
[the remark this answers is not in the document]
darkodraskovic: Fixed.

twins/mqtt/pubsub.go
```diff
@@ -0,0 +1,85 @@
...
+
+	return client
+}
+
+func (mqtt *Mqtt) Topic() string {
+	return mqtt.topic
+}
+
+func (mqtt *Mqtt) publish(id, op string, payload *[]byte) error {
+	topic := fmt.Sprintf("channels/%s/messages/%s/%s", mqtt.topic, id, op)
```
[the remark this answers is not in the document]
darkodraskovic: Op is a twin operation (from the CRUD set) that we publish the notification for via mqtt. `id` is in fact a twin id. Fixed the names. Please see now.

twins/mqtt/pubsub.go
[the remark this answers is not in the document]
darkodraskovic: These need to be pointers because of the defer function. Defer func gets evaluated when it's defined. So, when it's defined, id (twinID - fixed), err and payload are zero val. They get populated with values after the defer func definition. Besides, I've tested the methot without pointers and it does not work as expected.

We publish error by design. We want an mqtt twins service user to be aware of the error (not only the twins service operator).

twins/nats/pubsub.go
darkodraskovic: No, I need this err defined for defer func (see below) to work.

twins/mqtt/pubsub.go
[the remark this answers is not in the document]
darkodraskovic: Should I call it then `subscriber.go`? @dusanb94 ^

twins/mqtt/pubsub.go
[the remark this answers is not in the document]
darkodraskovic: @anovakovic01 ^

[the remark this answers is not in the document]
darkodraskovic: @dusanb94 ^

twins/api/http/requests.go
darkodraskovic: Fixed.

twins/api/http/responses.go
darkodraskovic: Fixed.

cmd/twins/main.go
```diff
@@ -0,0 +1,319 @@
...
+
+	envLogLevel        = "MF_TWINS_LOG_LEVEL"
+	envHTTPPort        = "MF_TWINS_HTTP_PORT"
+	envJaegerURL       = "MF_JAEGER_URL"
+	envServerCert      = "MF_TWINS_SERVER_CERT"
+	envServerKey       = "MF_TWINS_SERVER_KEY"
+	envDBName          = "MF_MONGODB_NAME"
+	envDBHost          = "MF_MONGODB_HOST"
+	envDBPort          = "MF_MONGODB_PORT"
+	envSingleUserEmail = "MF_TWINS_SINGLE_USER_EMAIL"
```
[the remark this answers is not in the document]
darkodraskovic: I simply use directly authn service as in things service.

twins/paho/pubsub.go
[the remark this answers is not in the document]
darkodraskovic: I need help on this one. @nmarcetic @mteodor

twins/api/http/requests.go
darkodraskovic: Fixed.

twins/mqtt/pubsub.go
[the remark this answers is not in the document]
dborovcanin: As far as I can see, it only does MQTT publish. Anyway, calling it `pubsub` is misleading, so `publisher.go` or `subscriber.go` is a way to go.

twins/mongodb/twins_test.go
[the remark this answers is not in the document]
drasko: If you do not know how to make them work, please remove them. But better invest some time and make them work or understand why they are not working :).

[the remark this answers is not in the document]
manuio: Can you try `metaQuery := map[string]string{"type": "test"}` instead of `metadata`. Looks like you are searching with the whole metadata.

twins/mqtt/pubsub.go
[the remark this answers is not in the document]
drasko: This is not a good design IMHO. You should log internal errors in an internal log, and send to user only the information suitable for user.

For security reasons, he should not have insight in the whole system errors.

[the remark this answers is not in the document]
drasko: I did not understand this one. What if it was an empty string instead of pointer? Would it change something?

cmd/twins/main.go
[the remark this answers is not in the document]
drasko: If this is so, please remove it in both services (i.e. fix Things service also).

Open an issue first for this, and fix in a subsequent PR, this one is already huge.

twins/paho/pubsub.go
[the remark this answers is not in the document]
darkodraskovic: Added as an issue.

twins/mongodb/twins_test.go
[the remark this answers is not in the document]
darkodraskovic: @manuio Works.

twins/mqtt/pubsub.go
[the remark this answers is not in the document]
darkodraskovic: Done.

twins/mqtt/pubsub.go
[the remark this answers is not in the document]
darkodraskovic: Submitted as an issue.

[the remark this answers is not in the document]
nmarcetic: I agree with @drasko its exactly what we did by design. Same as we do for HTTP errors etc... We never return the error as is,  under any circumstances. We always handle this and return error message, status code that indicates error etc... In this case of MQTT, you need to figure out how this message should look like.

twins/nats/pubsub.go
nmarcetic: @darkodraskovic Maybe you can checkout MongoDB doc and [update method](https://docs.mongodb.com/manual/reference/operator/update/positional/), more complex query which will do the job is more welcome :) I agree with @dusanb94

twins/mqtt/pubsub.go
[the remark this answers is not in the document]
darkodraskovic: Deferred to a separate issue.

twins/nats/pubsub.go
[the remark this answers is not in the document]
nmarcetic: I would also say service, why not.

twins/nats/pubsub.go
nmarcetic: Yep, I agree about moving this into the service. Nats package should contain only nats logic.

darkodraskovic: @nmarcetic and @dusanb94, regarding the more complex query, I think you do not take into account that twin and state are in different databases.

twins/nats/pubsub.go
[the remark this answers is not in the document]
darkodraskovic: We now use the service. The business logic is also moved to service.

cmd/twins/main.go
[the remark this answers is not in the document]
darkodraskovic: Moved to issues.

cmd/twins/main.go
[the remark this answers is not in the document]
darkodraskovic: Removed `if` statement.

twins/README.md
```diff
@@ -0,0 +1,85 @@
...
+    image: mainflux/twins:[version]
+    container_name: [instance name]
+    ports:
+      - [host machine port]:[configured HTTP port]
+    environment:
+      MF_TWINS_LOG_LEVEL: [Twins log level]
+      MF_TWINS_HTTP_PORT: [Service HTTP port]
+      MF_TWINS_SERVER_CERT: [String path to server cert in pem format]
+      MF_TWINS_SERVER_KEY: [String path to server key in pem format]
+      MF_JAEGER_URL: [Jaeger server URL]
```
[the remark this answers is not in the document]
drasko: Please resolve this one so we van merge.

drasko (changes requested):

twins/README.md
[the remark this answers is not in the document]
darkodraskovic: Fixed.

twins/nats/pubsub.go
darkodraskovic: Moved the logic into the twin service.

twins/mqtt/pubsub.go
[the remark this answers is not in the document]
darkodraskovic: This will be probably replaced by NATS.

twins/api/http/requests.go
```diff
@@ -0,0 +1,132 @@
...
+	Metadata   map[string]interface{} `json:"metadata,omitempty"`
+}
+
+func (req addTwinReq) validate() error {
+	if req.token == "" {
+		return twins.ErrUnauthorizedAccess
+	}
+
+	if req.ThingID == "" {
+		fmt.Printf("%s\n", "HERE")
```
drasko: Please remove these pritfs

drasko (changes requested):

cmd/twins/main.go
```diff
@@ -0,0 +1,301 @@
...
+	defCACerts         = ""
+	defMqttURL         = "tcp://localhost:1883"
+	defThingID         = ""
+	defThingKey        = ""
+	defChannelID       = ""
+	defNatsURL         = nats.DefaultURL
+
+	defAuthnGRPCPort = "8181"
+	defAuthnTimeout  = "1" // in seconds
+	defAuthnURL      = "localhost"
```
manuio: `defAuthnURL` should be `localhost:8181`

twins/mongodb/twins.go
```diff
@@ -0,0 +1,205 @@
...
+	return twins.TwinsPage{
+		Twins: results,
+		PageMetadata: twins.PageMetadata{
+			Total:  uint64(total),
+			Offset: offset,
+			Limit:  limit,
+		},
+	}, nil
+}
+
```
manuio, line 133: Add a description for exported functions

twins/mongodb/twins.go
```diff
@@ -0,0 +1,205 @@
...
+	coll := tr.db.Collection(twinsCollection)
+	tw := twins.Twin{}
+	filter := bson.D{{"thingid", thingid}}
+	if err := coll.FindOne(context.Background(), filter).Decode(&tw); err != nil {
+		return tw, twins.ErrNotFound
+	}
+
+	return tw, nil
+}
+
```
manuio, line 90: Add a description for exported functions

twins/mongodb/twins.go
```diff
@@ -0,0 +1,205 @@
...
+	var tw twins.Twin
+
+	filter := bson.D{{"id", id}}
+	if err := coll.FindOne(context.Background(), filter).Decode(&tw); err != nil {
+		return tw, twins.ErrNotFound
+	}
+
+	return tw, nil
+}
+
```
manuio, line 79: Add a description for exported functions

manuio (changes requested):

twins/mongodb/twins.go
```diff
@@ -23,15 +23,13 @@ type twinRepository struct {
...
-// NewTwinRepository instantiates a MongoDB implementation of twin
-// repository.
+// NewTwinRepository instantiates a MongoDB implementation of twin repository.
 func NewTwinRepository(db *mongo.Database) twins.TwinRepository {
 	return &twinRepository{
 		db: db,
 	}
 }
 
-// Save persists the twin
```
drasko: All exported functions must have comments and linter should warn you.

So I think that you should not remove this and fix your linter in your editor so that it warns you about things like this.

drasko (changes requested):

twins/mongodb/twins.go
drasko: Looks that this is not needed when functions are implementation of an interface.

drasko (approved):
LGTM

manuio (approved):
LGTM

PR #861, opened 2019-09-17
MF-858 Users metadata

### LINKED ISSUE #858: adding metadata for the users
**FEATURE REQUEST**

It would be good  to have possibility to add metadata for the user

Nice to have

mteodor:
#196 # What does this do?

 `Resolves #858`


scripts/run.sh
```diff
@@ -22,6 +22,7 @@ function cleanup {
 # NATS
 ###
 gnatsd &
+sleep 3
```
drasko: Try lowering this to `0.1`, `0.5` or `1` - NATS is usually very fast and `0.1` should do. Also add a comment why you are doing this (although seems evident).

users/api/http/requests.go
```diff
@@ -7,7 +7,9 @@
 
 package http
 
-import "github.com/mainflux/mainflux/users"
+import (
```
drasko: Is this done by Go fmt? because you have only one line, I am not sure that these parenthesis are needed. but if gofmt is putting them, then OK.

users/postgres/init.go
```diff
@@ -54,11 +54,17 @@ func migrateDB(db *sqlx.DB) error {
 				Up: []string{
 					`CREATE TABLE IF NOT EXISTS users (
 						email	 VARCHAR(254) PRIMARY KEY,
-						password CHAR(60)	  NOT NULL
+						password CHAR(60)	  NOT NULL,
```
drasko: Is this `,` at the end needed?

users/postgres/users.go
```diff
@@ -66,18 +67,36 @@ func (ur userRepository) RetrieveByID(_ context.Context, email string) (users.Us
...
 	Password string `json:"password"`
+	Metadata string `json:"metadata",omitempty`
 }
 
 func toDBUser(u users.User) dbUser {
+	m := "{}"
+	if len(u.Metadata) > 0 {
+		data, err := json.Marshal(u.Metadata)
+		if err != nil {
+			return dbUser{}
```
drasko: Wee do not log anything and we do not return error to the caller? I suppose we expect logging middleware to log, but it must have some indication that error happened.

users/postgres/users.go
```diff
@@ -66,18 +67,36 @@ func (ur userRepository) RetrieveByID(_ context.Context, email string) (users.Us
...
 		Password: u.Password,
+		Metadata: m,
 	}
 }
 
 func toUser(dbu dbUser) users.User {
+
+	var metadata map[string]interface{}
+	if err := json.Unmarshal([]byte(dbu.Metadata), &metadata); err != nil {
+		return users.User{}
```
drasko: No logging and no error returned?

drasko (changes requested):

things/postgres/things.go
```diff
@@ -334,17 +334,21 @@ type dbThing struct {
...
+		data = string(b)
 	}
 
 	return dbThing{
 		ID:       th.ID,
 		Owner:    th.Owner,
 		Name:     th.Name,
 		Key:      th.Key,
-		Metadata: string(data),
+		Metadata: data,
```
nmarcetic: This should byte[] now with JSONB not a string type.

users/postgres/init.go
chombium: Adding this field here is a breaking change for the existing users. As the table is already there the migration will not be executed. This field addition should be added in the new "users_2"  migration as well.

nmarcetic (dismissed):

scripts/run.sh
```diff
@@ -22,6 +22,7 @@ function cleanup {
 # NATS
 ###
 gnatsd &
+sleep 1
```
chombium: Are we sure that 1 seconds is enough so that NATS is up and running? There is a possibility that we end up in a "works on my machine" situation.
If we need to wait, wouldn't it be better if we query one of the NATS's [monitoring endpoits](https://nats-io.github.io/docs/nats_server/monitoring.html), `/varz` for example and check the response

users/api/http/requests.go
```diff
@@ -17,6 +19,17 @@ type userReq struct {
...
 func (req userReq) validate() error {
 	return req.user.Validate()
 }
+
+func (req viewUserInfo) validate() error {
+	if req.token == "" {
+		return users.ErrUnauthorizedAccess
+	}
+	return nil
+}
```
anovakovic01, line 33: Please keep struct and it's methods grouped together. Also, add Req at the end of the struct name.

users/api/http/responses.go
```diff
@@ -18,6 +18,15 @@ var _ mainflux.Response = (*tokenRes)(nil)
...
 }
+type identityRes struct {
+	Email    string                 `json:"email"`
+	Metadata map[string]interface{} `json:"metadata,omitempty"`
+}
+
+type userInfo struct {
+	Email    string                 `json:"email"`
+	Metadata map[string]interface{} `json:"metadata,omitempty"`
+}
```
anovakovic01: These two are the same. Also, please add `code`, `headers` and `empty` methods to the one that you leave.

users/api/logging.go
```diff
@@ -67,3 +67,16 @@ func (lm *loggingMiddleware) Identify(key string) (id string, err error) {
 
 	return lm.svc.Identify(key)
 }
+
+func (lm *loggingMiddleware) UserInfo(ctx context.Context, key string) (u users.UserInfo, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method get user info for client %s took %s to complete", u.Email, time.Since(begin))
```
anovakovic01: Change `get user` part to `user_info`. We always use the method name here with underscore notation.

users/api/metrics.go
```diff
@@ -59,3 +59,12 @@ func (ms *metricsMiddleware) Identify(key string) (string, error) {
 
 	return ms.svc.Identify(key)
 }
+
+func (ms *metricsMiddleware) UserInfo(ctx context.Context, key string) (users.UserInfo, error) {
+	defer func(begin time.Time) {
+		ms.counter.With("method", "identity").Add(1)
+		ms.latency.With("method", "identity").Observe(time.Since(begin).Seconds())
```
anovakovic01: Update method name from `identity` to `user_info`.

users/postgres/users.go
```diff
@@ -66,18 +71,36 @@ func (ur userRepository) RetrieveByID(_ context.Context, email string) (users.Us
 type dbUser struct {
 	Email    string `json:"email"`
 	Password string `json:"password"`
+	Metadata []byte `json:"metadata",omitempty`
```
anovakovic01: I think that `omitempty` must be inside "". `json:"metadata,omitempty"`

users/users.go
```diff
@@ -24,6 +24,11 @@ var (
 type User struct {
 	Email    string
 	Password string
+	Metadata map[string]interface{}
+}
+type UserInfo struct {
```
anovakovic01: Add an empty line here.

anovakovic01 (changes requested):

scripts/run.sh
mteodor: I've changed checking availability

users/api/http/responses.go
```diff
@@ -30,3 +33,20 @@ func (res tokenRes) Headers() map[string]string {
...
 }
+
+type identityRes struct {
+	Email    string                 `json:"email"`
+	Metadata map[string]interface{} `json:"metadata,omitempty"`
+}
+
+func (res identityRes) Code() int {
+	return http.StatusCreated
+}
```
anovakovic01, line 44: Identity response is a response to the  GET request. Use `http.StatusOK` here.

users/api/http/responses.go
```diff
@@ -30,3 +33,20 @@ func (res tokenRes) Headers() map[string]string {
...
+
+func (res identityRes) Headers() map[string]string {
+	return map[string]string{}
+}
+
+func (res identityRes) Empty() bool {
+	return res.Email == "" && len(res.Metadata) == 0
```
anovakovic01: Just return `false` here. If there is anything missing (if an error has happened) `encodeError` will be called, where this `Empty` method doesn't matter. We use empty only to indicate if there is any content or not (for example when you create user, there is nothing to return on success, so Empty method of that response returns `true`).

users/users.go
```diff
@@ -24,6 +24,12 @@ var (
 type User struct {
 	Email    string
 	Password string
+	Metadata map[string]interface{}
+}
+
+type UserInfo struct {
```
anovakovic01: I just realized I think that you don't need this struct. Just use `User` without a password. On this layer, we always use entity object (in this case `User`).

anovakovic01 (changes requested):

anovakovic01 (dismissed):
LGTM!

anovakovic01 (dismissed):
LGTM!

scripts/run.sh
```diff
@@ -22,6 +22,18 @@ function cleanup {
 # NATS
 ###
 gnatsd &
+lsof_check=`command -v lsof`
```
drasko: This is already added and commented in #867

drasko (changes requested):
I have this remark for the script, otherwise it looks good.

anovakovic01 (dismissed):
LGTM!

users/api/logging.go
```diff
@@ -67,3 +67,16 @@ func (lm *loggingMiddleware) Identify(key string) (id string, err error) {
 
 	return lm.svc.Identify(key)
 }
+
+func (lm *loggingMiddleware) UserInfo(ctx context.Context, key string) (u users.User, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method user_info for client %s took %s to complete", u.Email, time.Since(begin))
```
manuio: I prefer `for user %s` than `for client %s`

manuio (changes requested):

users/api/http/requests.go
dborovcanin: We don't use parentheses in single-line imports.

dborovcanin (changes requested):

manuio (dismissed):

users/postgres/users.go
```diff
@@ -63,21 +70,61 @@ func (ur userRepository) RetrieveByID(_ context.Context, email string) (users.Us
...
 type dbUser struct {
-	Email    string `json:"email"`
-	Password string `json:"password"`
+	Email    string     `db:"email"`
+	Password string     `db:"password"`
+	Metadata dbMetadata `db:"metadata"`
 }
 
-func toDBUser(u users.User) dbUser {
+func toDBUser(u users.User) (dbUser, error) {
```
dborovcanin: No need to return `(dbUser, error)`, but just `dbUser`.

users/postgres/users.go
```diff
@@ -63,21 +70,61 @@ func (ur userRepository) RetrieveByID(_ context.Context, email string) (users.Us
...
+	// if value is nil, return empty
+	if value == nil {
+		m = nil
+		return nil
+	}
+
+	b, ok := value.([]byte)
+	if !ok {
+		m = &dbMetadata{}
+		return errors.New("Failed to scan metadata")
```
dborovcanin: Can you extract error to the `var` section?

dborovcanin (changes requested):

users/postgres/users.go
```diff
@@ -63,21 +67,64 @@ func (ur userRepository) RetrieveByID(_ context.Context, email string) (users.Us
 	return user, nil
 }
 
+// dbMetadata type for handling metadata properly in database/sql
+type dbMetadata map[string]interface{}
+
+// Scan - Implement the database/sql scanner interface
+func (m *dbMetadata) Scan(value interface{}) error {
+	// if value is nil, return empty
```
drasko: Comments start with capital letters

drasko (changes requested):

drasko (approved):
LGTM

PR #866, opened 2019-09-18
MF-538 - Improve logging and API errors

### LINKED ISSUE #538: Improve logging and API errors
**FEATURE REQUEST**

No

We need to improve logging  and  all API response’s, mostly error responses are critical and not useful. 

Regarding logging problem,

We use custom logger which is a wrapper around go-kit logger, providing basic methods for [different log levels](https://github.com/mainflux/mainflux/blob/master/logger/logger.go#L18)
Mostly logging is handled really well, especially in important places where business logic happens and where logging is crucial. The problem arises on API level where login is happening on transport layer and its very poor, usually errors like malformed entity or similar are not logged at all. 
We need to go over all HTTP API logic  and check the logging, make it useful.    

Regarding API responses problem,

Its similar to logging, API responses are mostly unusable because they  returning you just error code with no additional data, no further information, nothing. Error codes are probably the most useful diagnostic element in the API I agree, but there so much cases now where we need to make line between the errors and try to provide a bit more information’s, like a human readable message or even just a context with reference ID. 
There is a lot of good resources which can help us to design the response like [JSON API specification](https://jsonapi.org/format/#error-objects) etc… We should discuss about this, keep it simple, consistent and useful.

Both those problems results with cases where I as API consumer, consume some API end point and get 400 bad request without any explanation and any logs.

blokovi:

### What does this do?
Create errors package and implement it on users service

Resolves #538

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -12,6 +12,7 @@ type Error interface {
 	Error() string
 	Msg() string
 	Contains(error) bool
+	IsEmpty() bool
 }
 
 var _ Error = (*customError)(nil)
@@ -36,6 +37,13 @@ func (err customError) Msg() string {
 	return err.msg
 }
 
+func (err customError) IsEmpty() bool {
+	if err.Msg() == "" {
+		return true
+	}
+	return false
+}
+
 // Contains inspects if Error's message is same as error
 // in argument. If not it continues to examin in next
 // layers of Error until it founds it or unwrap every layers
@@ -43,13 +51,13 @@ func (err customError) Contains(e error) bool {
 	if e == nil {
 		return false
 	}
-
 	if err.msg == e.Error() {
 		return true
 	}
 	if err.err == nil {
 		return false
 	}
+
 	return err.err.Contains(e)
 }
 
@@ -61,15 +69,12 @@ func Wrap(wrapper Error, err Error) Error {
 	}
 }
 
-// Cast returns pointer to Error type with message of given error
+// Cast returns Error type with message of given error
 func Cast(err error) Error {
 	if err == nil {
-		return nil
+		return New("")
 	}
-
-	return customError{
-		msg: err.Error(),
-	}
+	return New(err.Error())
 }
 
 // New returns an Error that formats as the given text.
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -61,19 +61,27 @@ func Contains(ce Error, e error) bool {
 }
 
 // Wrap returns an Error that wrap err with wrapper
-func Wrap(wrapper Error, err Error) Error {
-	return &customError{
-		msg: wrapper.Msg(),
-		err: err.(*customError),
-	}
-}
-
-// Cast returns Error type with message of given error
-func Cast(err error) Error {
+func Wrap(wrapper Error, err error) Error {
 	if err == nil {
 		return nil
 	}
-	return New(err.Error())
+	return &customError{
+		msg: wrapper.Msg(),
+		err: cast(err),
+	}
+}
+
+func cast(err error) *customError {
+	if err == nil {
+		return nil
+	}
+	if e, ok := err.(*customError); ok {
+		return e
+	}
+	return &customError{
+		msg: err.Error(),
+		err: nil,
+	}
 }
 
 // New returns an Error that formats as the given text.
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -18,9 +18,6 @@ type Error interface {
 	// in argument. If not it continues to examin in next
 	// layers of Error until it founds it or unwrap every layers
 	Contains(error) bool
-
-	//IsEmpty check if Error is empty
-	IsEmpty() bool
 }
 
 var _ Error = (*customError)(nil)
@@ -41,13 +38,6 @@ func (ce customError) Error() string {
 
 func (ce customError) Msg() string {
 	return ce.msg
-}
-
-func (ce customError) IsEmpty() bool {
-	if ce.Msg() == "" {
-		return true
-	}
-	return false
 }
 
 func (ce customError) Contains(e error) bool {
@@ -75,7 +65,8 @@ func Wrap(wrapper Error, err Error) Error {
 // Cast returns Error type with message of given error
 func Cast(err error) Error {
 	if err == nil {
-		return Empty()
+		// return Empty()
+		return nil
 	}
 	return New(err.Error())
 }
@@ -86,9 +77,4 @@ func New(text string) Error {
 		msg: text,
 		err: nil,
 	}
-}
-
-// Empty returns a new empty Error
-func Empty() Error {
-	return New("")
 }
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -6,11 +6,9 @@
 
 package errors
 
-import (
-	"fmt"
-)
+import "fmt"
 
-// Error represents a Mainflux error.
+// Error struct represents a Mainflux error
 type Error struct {
 	msg string
 	err *Error
@@ -30,6 +28,9 @@ func (err Error) Msg() string {
 	return err.msg
 }
 
+// Contains inspects if Error's message is same as error
+// in argument. If not it continues to examin in next
+// layers of Error until it founds it or unwrap every layers
 func (err Error) Contains(e error) bool {
 	if e == nil {
 		return false
@@ -44,6 +45,7 @@ func (err Error) Contains(e error) bool {
 	return err.err.Contains(e)
 }
 
+// Wrap returns an Error that wrap err with wrapper
 func Wrap(wrapper Error, err *Error) Error {
 	return Error{
 		msg: wrapper.msg,
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -43,8 +43,8 @@ func (ce *customError) Err() *customError {
 }
 
 // Contains inspects if Error's message is same as error
-// in argument. If not it continues to examin in next
-// layers of Error until it founds it or unwrap every layers
+// in argument. If not it continues further unwrapping
+// layers of Error until it founds it or unwrap all layers
 func Contains(ce Error, e error) bool {
 	if e == nil {
 		return ce == nil
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -15,7 +15,7 @@ type Error interface {
 	Msg() string
 
 	// Err returns wrapped error
-	Err() *customError
+	Err() Error
 }
 
 var _ Error = (*customError)(nil)
@@ -23,22 +23,25 @@ var _ Error = (*customError)(nil)
 // customError struct represents a Mainflux error
 type customError struct {
 	msg string
-	err *customError
+	err Error
 }
 
 func (ce *customError) Error() string {
-	if ce.err != nil {
-		return fmt.Sprintf("%s: %s", ce.msg, ce.err.Error())
-	}
+	if ce != nil {
+		if ce.err != nil {
+			return fmt.Sprintf("%s: %s", ce.msg, ce.err.Error())
+		}
 
-	return ce.msg
+		return ce.msg
+	}
+	return ""
 }
 
 func (ce *customError) Msg() string {
 	return ce.msg
 }
 
-func (ce *customError) Err() *customError {
+func (ce *customError) Err() Error {
 	return ce.err
 }
 
@@ -46,7 +49,7 @@ func (ce *customError) Err() *customError {
 // in argument. If not it continues further unwrapping
 // layers of Error until it founds it or unwrap all layers
 func Contains(ce Error, e error) bool {
-	if e == nil {
+	if ce == nil || e == nil {
 		return ce == nil
 	}
 	if ce.Msg() == e.Error() {
@@ -57,12 +60,11 @@ func Contains(ce Error, e error) bool {
 	}
 
 	return Contains(ce.Err(), e)
-
 }
 
 // Wrap returns an Error that wrap err with wrapper
 func Wrap(wrapper Error, err error) Error {
-	if err == nil {
+	if wrapper == nil || err == nil {
 		return nil
 	}
 	return &customError{
@@ -71,11 +73,11 @@ func Wrap(wrapper Error, err error) Error {
 	}
 }
 
-func cast(err error) *customError {
+func cast(err error) Error {
 	if err == nil {
 		return nil
 	}
-	if e, ok := err.(*customError); ok {
+	if e, ok := err.(Error); ok {
 		return e
 	}
 	return &customError{
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -16,10 +16,6 @@ type Error interface {
 
 	// Err returns wrapped error
 	Err() *customError
-
-	// Contains inspects if Error's message is same as error
-	// in argument. If not it continues to examin in next
-	// layers of Error until it founds it or unwrap every layers
 }
 
 var _ Error = (*customError)(nil)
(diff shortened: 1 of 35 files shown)


users/service.go
```diff
@@ -73,11 +75,13 @@ func (svc usersService) Register(ctx context.Context, user User) error {
...
 	dbUser, err := svc.users.RetrieveByID(ctx, user.Email)
 	if err != nil {
-		return "", ErrUnauthorizedAccess
+		// return "", ErrUnauthorizedAccess
+		return "", fmt.Errorf("%w: %s", ErrUnauthorizedAccess, err)
 	}
 
 	if err := svc.hasher.Compare(user.Password, dbUser.Password); err != nil {
-		return "", ErrUnauthorizedAccess
+		// return "", ErrUnauthorizedAccess
```
drasko: Dead code

users/service.go
```diff
@@ -86,7 +90,8 @@ func (svc usersService) Login(ctx context.Context, user User) (string, error) {
 func (svc usersService) Identify(token string) (string, error) {
 	id, err := svc.idp.Identity(token)
 	if err != nil {
-		return "", ErrUnauthorizedAccess
+		// return "", ErrUnauthorizedAccess
```
drasko: Dead code

drasko (changes requested):

users/service.go
```diff
@@ -73,11 +75,13 @@ func (svc usersService) Register(ctx context.Context, user User) error {
 func (svc usersService) Login(ctx context.Context, user User) (string, error) {
 	dbUser, err := svc.users.RetrieveByID(ctx, user.Email)
 	if err != nil {
-		return "", ErrUnauthorizedAccess
+		// return "", ErrUnauthorizedAccess
+		return "", fmt.Errorf("%w: %s", ErrUnauthorizedAccess, err)
```
drasko: Here you wrap, but somewhere you need to log as well (print for our server logs, not send to client). I guess you do it in `encodeError()`, but I do not see logger priniting there.

users/api/http/transport.go
```diff
@@ -122,4 +123,9 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 			w.WriteHeader(http.StatusInternalServerError)
 		}
 	}
+
+	if err != nil {
+		w.Header().Set("Content-Type", contentType)
+		json.NewEncoder(w).Encode(errorRes{Err: err.Error()})
```
drasko: Not all should be sent to client, only our description.

Also, I do not see the print on our logger (where everything should be printed).

users/service.go
blokovi: Whole error, with all wrapped layers, will go through logger and every of wrapped errors will be displayed in logs. On the other hand, client will see only last (outer) wrapper

users/api/http/transport.go
blokovi: Logger already print error so no need to change anything in logger. Errors printed with logger will now be more descriptive because they will have multiple wrapped layers and all of them will be printed in log

users/postgres/users.go
```diff
@@ -14,11 +14,16 @@ import (
...
-
 	"github.com/lib/pq"
+	"github.com/mainflux/mainflux/errors"
 	"github.com/mainflux/mainflux/users"
 )
 
+var (
+	// ErrDatabase indicates internal database error
+	ErrDatabase = errors.New("Database error")
+)
```
anovakovic01, line 22: Inline this one.

users/service.go
```diff
@@ -69,7 +70,7 @@ func New(users UserRepository, hasher Hasher, idp IdentityProvider) Service {
 func (svc usersService) Register(ctx context.Context, user User) error {
```
anovakovic01: I recommend that all methods return our `Error`. You don't want to do ugly casting every time you want to do something with it.

anovakovic01 (changes requested):

users/service.go
blokovi: I make changes in `Login` method to return `errors.Error` and in all other code to make that possible. Can you please review these changes before I continue with other methods

anovakovic01 (changes requested):

users/api/http/transport.go
```diff
@@ -136,4 +139,8 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 			w.WriteHeader(http.StatusInternalServerError)
 		}
 	}
+	if mfErr.Msg() != "" {
```
anovakovic01: Use `IsEmpty` here.

users/api/http/transport.go
```diff
@@ -105,28 +102,34 @@ func encodeResponse(_ context.Context, w http.ResponseWriter, response interface
...
+			w.WriteHeader(http.StatusConflict)
+		case mfErr.Contains(ErrUnsupportedContentType):
+			w.WriteHeader(http.StatusUnsupportedMediaType)
+			logger.Warn("Invalid or missing content type.")
+		case mfErr.Contains(io.ErrUnexpectedEOF):
+			w.WriteHeader(http.StatusBadRequest)
+		case mfErr.Contains(io.EOF):
+			w.WriteHeader(http.StatusBadRequest)
+		}
+	} else {
```
anovakovic01: Think about how to avoid this `else`.

users/api/logging.go
```diff
@@ -42,10 +43,10 @@ func (lm *loggingMiddleware) Register(ctx context.Context, user users.User) (err
 	return lm.svc.Register(ctx, user)
 }
 
-func (lm *loggingMiddleware) Login(ctx context.Context, user users.User) (token string, err error) {
+func (lm *loggingMiddleware) Login(ctx context.Context, user users.User) (token string, err errors.Error) {
 	defer func(begin time.Time) {
 		message := fmt.Sprintf("Method login for user %s took %s to complete", user.Email, time.Since(begin))
-		if err != nil {
+		if err.Msg() != "" {
```
anovakovic01: Use `IsEmpty`.

users/mocks/hasher.go
```diff
@@ -19,17 +22,17 @@ func NewHasher() users.Hasher {
...
 }
 
-func (hm *hasherMock) Compare(plain, hashed string) error {
+func (hm *hasherMock) Compare(plain, hashed string) errors.Error {
 	if plain != hashed {
 		return users.ErrUnauthorizedAccess
 	}
 
-	return nil
+	return errors.New("")
```
anovakovic01: Think about adding `Empty()` factory method that will call `New("")`.

anovakovic01 (changes requested):

errors/errors.go
```diff
@@ -0,0 +1,97 @@
+//
```
drasko: Fix header

drasko (commented):
This looks OK to me so far, but definitely must be documented in the Developer's Guide.

anovakovic01:
@manuio Go doesn't support method overloading, and that's why it should be called `Empty`.

anovakovic01:
> This looks OK to me so far, but definitely must be documented in the Developer's Guide.

@drasko I completely agree, and that's the plan.

manuio:
@anovakovic01 We can use `NewError()`

anovakovic01:
> @anovakovic01 We can use `NewError()`

@manuio It will be used for an *empty* error value. Also, `NewError()` is redundant because it will be called `errors.NewError()` which is bad practice in go. I'm not sure that you've understood what I've said earlier. This name should be used instead of `errors.New("")`. It's only used for an empty error value.

manuio:
@anovakovic01 I understood. It's just that I would prefer something else than `Empty()`.  If it's not possible then forget it.

users/api/http/transport.go
blokovi: I manage to avoid it with two switches, please check again. 
Also, I think that json error types are not possible anymore, I keep them from earlier version. But if you agree, I would remove that two cases.

blokovi: I removed json errors from type switch, I'm pretty sure they are not possible any more

anovakovic01:
> @manuio @anovakovic01 maybe `errors.NewEmpty()` instead of `errors.Empty()`?

I think that `Empty()` makes the most sense (you're just using empty error, which doesn't necessarily mean that you're creating new one)

users/jwt/idp.go
```diff
@@ -45,27 +45,29 @@ func (idp *jwtIdentityProvider) TemporaryKey(id string) (string, error) {
...
 	token, err := jwt.Parse(key, func(token *jwt.Token) (interface{}, error) {
 		if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
 			return nil, users.ErrUnauthorizedAccess
 		}
 
-		return []byte(idp.secret), nil
+		return []byte(idp.secret), errors.Empty()
 	})
 
 	if err != nil {
```
dborovcanin, line 54: You're comparing to `nil`. The return value of the errors.Empty() is not an error, but it's not `nil`, either, so it will fall under this case.

users/api/http/endpoint.go
```diff
@@ -18,11 +18,14 @@ func registrationEndpoint(svc users.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(userReq)
 
-		if err := req.validate(); err != nil {
+		if err := req.validate(); !err.IsEmpty() {
 			return nil, err
 		}
 
 		err := svc.Register(ctx, req.user)
+		if err.IsEmpty() {
```
anovakovic01: Inline this.

blokovi:
@anovakovic01 I can't inline there, because I need err to return if it's not empty after `if`

manuio:
@blokovi right, but you can do as you did in the rest of the code:
```
		if err := svc.Register(ctx, req.user); !err.IsEmpty() {
			return tokenRes{}, err
		}
		return tokenRes{}, nil
```

errors/errors.go
```diff
@@ -0,0 +1,94 @@
...
+
+var _ Error = (*customError)(nil)
+
+// customError struct represents a Mainflux error
+type customError struct {
+	msg string
+	err Error
+}
+
+func (err customError) Error() string {
```
manuio: Can you rename `err` -> `ce` here and in the rest of the code please. I don't like too much this `err.err`

users/api/grpc/endpoint_test.go
```diff
@@ -51,7 +52,9 @@ func startGRPCServer(svc users.Service, port int) {
 }
 
 func TestIdentify(t *testing.T) {
-	svc.Register(context.Background(), user)
+	// svc.Register(context.Background(), user)
```
manuio: Remove this comment

manuio (changes requested):

users/bcrypt/hasher.go
```diff
@@ -1,10 +1,14 @@
-// Copyright (c) Mainflux
+//
```
drasko: Fix header

drasko (changes requested):

dborovcanin (commented):
I prefer returning `(val, error)` instead of `(val, errors.Error)`. The error handler receives an `error` rather than `errors.Error`, so we need to do type assertions anyway. Also, there is no need to wrap nil error with Empty error but just return nil. I already got used to `(val, error)` and most of the code is like that, so not changing it is will make this PR much less intrusive. Since I'm used to returning error or nil, I'm afraid of accidentally returning `nil` instead of an empty error, which might cause checks such as `err.IsEmpty()` to panic. This would result in something like `if err != nil && err.IsEmpty()` and `err.IsEmpty()` is redundant here, IMHO.

users/bcrypt/hasher.go
```diff
@@ -1,10 +1,14 @@
...
+// Copyright (c) 2018
+// Mainflux
+//
```
manuio: Fix this header

manuio (changes requested):

errors/errors.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (ce *customError) Err() *customError {
+	return ce.err
+}
+
+// Contains inspects if Error's message is same as error
+// in argument. If not it continues to examin in next
+// layers of Error until it founds it or unwrap every layers
+func Contains(ce Error, e error) bool {
```
mteodor, line 51: continues further unwrapping layers, 
every = all

errors/errors.go
```diff
@@ -0,0 +1,93 @@
...
+type Error interface {
+
+	// Error implements the error interface.
+	Error() string
+
+	// Msg returns error message
+	Msg() string
+
+	// Err returns wrapped error
+	Err() *customError
```
dborovcanin: Should `Err()` be exported or we're only using it in `Contains`? If we're using it only in `Contains`, we don't need it at all.

errors/errors.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	if ce.Msg() == e.Error() {
+		return true
+	}
+	if ce.Err() == nil {
+		return false
+	}
+
+	return Contains(ce.Err(), e)
+
```
dborovcanin: Remove an empty line.

users/service.go
```diff
@@ -95,46 +99,48 @@ func New(users UserRepository, hasher Hasher, idp IdentityProvider, m Emailer, t
 	return &usersService{users: users, hasher: hasher, idp: idp, email: m, token: t}
 }
 
-func (svc usersService) Register(ctx context.Context, user User) error {
+func (svc usersService) Register(ctx context.Context, user User) errors.Error {
 	hash, err := svc.hasher.Hash(user.Password)
 	if err != nil {
-		return ErrMalformedEntity
+
```
dborovcanin: Please remove the empty line.

users/users.go
```diff
@@ -24,9 +26,10 @@ type User struct {
 }
 
 // Validate returns an error if user representation is invalid.
-func (u User) Validate() error {
+func (u User) Validate() errors.Error {
 	if u.Email == "" || u.Password == "" {
 		return ErrMalformedEntity
+
```
dborovcanin, line 32: Please remove the empty line.

dborovcanin (changes requested):

errors/errors.go
```diff
@@ -0,0 +1,93 @@
...
+	// Err returns wrapped error
+	Err() *customError
+}
+
+var _ Error = (*customError)(nil)
+
+// customError struct represents a Mainflux error
+type customError struct {
+	msg string
+	err *customError
```
anovakovic01: Maybe we should use `Error` here instead of `*customError`. Later you can chain some other errors that implement this interface here. I think that using interface is more flexible.

errors/errors.go
```diff
@@ -0,0 +1,93 @@
...
+	if err == nil {
+		return nil
+	}
+	return &customError{
+		msg: wrapper.Msg(),
+		err: cast(err),
+	}
+}
+
+func cast(err error) *customError {
```
anovakovic01: Maybe return `Error` here.

anovakovic01 (changes requested):

errors/errors.go
blokovi: Keeping it

anovakovic01 (dismissed):
LGTM!

dborovcanin (dismissed):

dborovcanin (dismissed):

dborovcanin (dismissed):

anovakovic01 (dismissed):
LGTM!

manuio (dismissed):
LGTM

anovakovic01 (dismissed):
LGTM!

drasko (dismissed):
LGTM

dborovcanin (approved):

drasko (approved):
LGTM

PR #867, opened 2019-09-19
MF-859 - Channels metadata search

### LINKED ISSUE #859: Adding search of channels by metadata
**FEATURE REQUEST**

Adding search of channels by metadata

Nice-to-have

mteodor:
### What does this do?
add retrieve channels my metadata criteria

`Resolves #859`


drasko (changes requested):

scripts/run.sh
```diff
@@ -22,6 +22,18 @@ function cleanup {
 # NATS
 ###
 gnatsd &
+lsof_check=`command -v lsof`
```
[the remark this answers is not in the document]
chombium: @drasko yes, the second check is enough if lsof is installed. If we are not sure if `lsof` is installed we could use something else like `netcat` which is installed on most of the modern systems.

We could use something like the following (start with a timeout):
```shell
PORT=4222
counter=1
while ! nc -zv localhost ${PORT} 1>/dev/null 2>&1; do
sleep 1
echo -n "."
((counter++))
if [ ${counter} -gt 60 ]; then
    echo "NATS not listening on port ${PORT} after 60s."
    exit 1
fi
done
echo "NATS is ready"
```

scripts/run.sh
```diff
@@ -22,6 +22,18 @@ function cleanup {
...
+until nc -zv localhost 4222 1>/dev/null 2>&1; 
+do
+    sleep 0.5
+    ((counter++))
+    if [ ${counter} -gt 20 ]
+    then
+        echo "gnatsd failed, exiting"
+        exit 1
+    fi
+    echo "Waiting for gnatsd"
```
chombium: In order not to repeat and reprint the same line over and over until gnats is up and running, I would suggest that you change this line to:
```bash
echo -ne "Waiting for gnatsd"
```
This will practically write the same message over and over on the same line, so that when the script is run it will show:
```
...
Waiting for gnatsd
...
```
instead of:
```
...
Waiting for gnatsd
Waiting for gnatsd
Waiting for gnatsd
Waiting for gnatsd
...
```

Of course this is my subjective view and I think that one message is enough. If we want to show how long does it take for `natsd` to start, than showing multiple messages would be ok as well.
@mteodor feel free to proceed the way you think it would be better.

scripts/run.sh
```diff
@@ -22,6 +22,18 @@ function cleanup {
...
 ###
 gnatsd &
+counter=1
+until nc -zv localhost 4222 1>/dev/null 2>&1; 
+do
+    sleep 0.5
+    ((counter++))
+    if [ ${counter} -gt 20 ]
+    then
+        echo "gnatsd failed, exiting"
```
chombium: I would change the message to "gnatsd failed to start in 10 seconds, exiting" so that the user knows what the problem was and maybe raise the timeout if needed.

drasko: 10 sec is too much, NATS usually starts in a friction of second. I would opt for 3-5 sec, not more.

drasko (changes requested):

chombium:
@drasko 10 seconds is the maximum time which we wait until natsd is started. It's a timeout when we give up.
It is checked if a connection can be made on the port 4222 and if that's not possible, the script sleeps for half a second and tries again. In a case where natsd starts before the timeout is fired(within 10 seconds), the maximum extra time which the user waits is half a second.

drasko:
@chombium I know, but I think there is no need to wait 10 seconds. If NATS does not start in 3-5 sec - there is probably something wrong beyond repair, better break out.

I can accept this, but never the less I think it is not needed...

manuio (changes requested):

chombium:
@drasko I agree with you. If we are sure that 3-5 seconds are enough than we can lower the timeout

users/api/http/requests.go
```diff
@@ -7,7 +7,9 @@
 
 package http
 
-import "github.com/mainflux/mainflux/users"
+import (
+	"github.com/mainflux/mainflux/users"
```
drasko: Is this a linter that makes `import()` when one-liner is enough?

mteodor:
I think yes

users/api/http/requests.go
drasko: Can you please check this?

manuio (dismissed):
LGTM

drasko (dismissed):
LGTM

dborovcanin (changes requested):

things/things.go
```diff
@@ -10,7 +10,7 @@ package things
 import "context"
 
 // Metadata to be used for mainflux thing or channel for customized
-// describing of particular thing or channel
+// describing of particular thing or channel.
```
drasko, line 13: no need for a full-stop at the end of the comment

things/postgres/channels.go
```diff
@@ -333,7 +332,7 @@ func (cr channelRepository) hasThing(chanID, thingID string) error {
 // dbMetadata type for handling metadata properly in database/sql
 type dbMetadata map[string]interface{}
 
-// Scan - Implement the database/sql scanner interface
+// Scan - Implement the database/sql scanner interface.
```
drasko: No need for a full-stop at the end of the comment.

things/postgres/channels.go
```diff
@@ -354,7 +353,7 @@ func (m *dbMetadata) Scan(value interface{}) error {
 	return nil
 }
 
-// Value Implements valuer
+// Value Implements valuer.
```
drasko: Same here

drasko (changes requested):

users/users.go
```diff
@@ -27,6 +27,11 @@ type User struct {
 	Metadata map[string]interface{}
 }
 
+type UserInfo struct {
+	Email    string
+	Metadata map[string]interface{}
+}
```
anovakovic01: You don't need this structure (you don't use it anywhere).

things/postgres/channels.go
anovakovic01: I recommend that we follow go standard. Don't write `-` after method name; just write a full sentence that begins with the method name. For example:
```
// Scan implements the database/sql scanner interface.
```

anovakovic01 (changes requested):

anovakovic01 (approved):
LGTM!

dborovcanin (approved):

drasko (approved):
LGTM

Issue #787, opened 2019-07-15
Enrich existing open tracing spans with additional data


anovakovic01:
**FEATURE REQUEST**

1. Is there an open issue addressing this request? If it does, please add a "+1" reaction to the
   existing issue, otherwise proceed to step 2.
No.

2. Describe the feature you are requesting, as well as the possible use case(s) for it.
After adding base open tracing support, spans should be enriched with additional tags. There is a recommendation of what should be added to the span and you can check it out [here](https://github.com/opentracing/specification/blob/master/semantic_conventions.md). One of the problems that we'll have to solve is how to write some kind of interceptor to the existing DB client in order to fetch SQL statements and passed params. The end goal should be removing separate logging and metrics middleware.

3. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).

nwneisen:
I'd like to take a look at implementing this.  I'm reading up on the Jaeger project and open traces and then should have a plan.

nmarcetic:
Great @nwneisen :)  Take your time for research, then we can sync here and agree about implementation.

drasko:
@nwneisen fantastic!

What would be great also is that - once this is done - you document it through a blogpost (with nice screenshots), so that we have a documentation for other developers how to use OpenTarcing with Mainflux.

Your text will be then featured on Mainflux official blog: https://medium.com/mainflux-iot-platform (you can ping @Sale72 to help you with the [draft](https://help.medium.com/hc/en-us/articles/214874698-Your-drafts-posts) process).

Also, it would be great to add a chapter in the official [documentation](https://github.com/mainflux/mainflux/tree/master/docs) as well.

nwneisen:
After playing around with Jaeger and its examples, I think I've got my head wrapped around traces.

I found were spans are currently being created in mainflux.  One such location is things/tracing/things.go.  Here `span.SetTag()` can be used to add client information such as component name, span kind, and any errors.  The transaction name (Save, Update, etc) also seems like useful information at this point but I don't see anything for it in the conventions document above.  

From here the span will be passed to the database code at things/postgres/things.go via the context parameter.   The span can be pulled out and have tags such as the query string and parameters added here.  There are also some error cases that could be added to the span logs.

anovakovic01:
[...]

Issue #666, opened 2019-03-19
Create pkg/utils for shared code


nmarcetic:
**FEATURE REQUEST**

1. Is there an open issue addressing this request? If it does, please add a "+1" reaction to the
   existing issue, otherwise proceed to step 2.
No

2. Describe the feature you are requesting, as well as the possible use case(s) for it.
In order to share code over the project, we need to introduce pkg/utils module and extract functional stuff like parsing/checking content types, Regex validators for subtopics etc... which are used strictly for operational stuff. But be careful with domain/business logic and bounded context, we don't want to create a coupling & dependency chaos, keep things self-contained but without code duplication.

3. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).
Nice to have.

rugwirobaker:
I thought utils is an anti-pattern in Go??

nmarcetic:
@rugwirobaker Not really, if you package your code well :) Not making circular dependencies and ant-patterns. More like [/internal philosophy](https://github.com/golang-standards/project-layout#internal)

nmarcetic:
Related to #793

drasko:
@nmarcetic in general `utils` is anti-pattern in Go, it avoids calling packages like `utils` because it does not describe much and is too generic. Suggested names for the packages are very specific in Go.

`/pkg` should hold packages with package names (like `things`, `users`, and so on).

drasko:
As described in #844:
Based on [Go project-layout](https://github.com/golang-standards/project-layout), we should reorganize internal file and directory organization.

Move Mainflux services in `/internal`. Also, move Swagger docs in `/api`.

We need to see what packages whold be imported by other apps, and move these in `/pkg`

nmarcetic:
Same as I said to @rugwirobaker here https://github.com/mainflux/mainflux/issues/666#issuecomment-513767656 The idea was to use  `/internal` 
`utils` is more like abstraction, the idea and the general naming :)

manuio:
@nmarcetic Can we close this one? Looks like it was resolved here: https://github.com/mainflux/mainflux/pull/1191

nmarcetic:
@manuio Yes we can, closing.

PR #869, opened 2019-09-20
MF-787 - Add tags to user, thing, and channel spans

nwneisen:
### What does this do?
Adds tags to tracing spans created for users, things, and channels.  The tags added in the database functions are placeholders since the data recommended in the span convention document is not available.

Relates to #787 

None

No

I will update the official documents showing users how to use Jaeger and open tracing once the related issue is completed.

### DIFF
--- a/cmd/things/main.go
+++ b/cmd/things/main.go
@@ -306,7 +306,6 @@ func connectToUsers(cfg config, logger logger.Logger) *grpc.ClientConn {
 }
 
 func newService(users mainflux.UsersServiceClient, dbTracer opentracing.Tracer, cacheTracer opentracing.Tracer, db *sqlx.DB, cacheClient *redis.Client, esClient *redis.Client, logger logger.Logger) things.Service {
-
 	dbMiddleware := postgres.NewDatabaseMiddleware(db)
 
 	thingsRepo := postgres.NewThingRepository(dbMiddleware)
--- a/cmd/things/main.go
+++ b/cmd/things/main.go
@@ -306,12 +306,12 @@ func connectToUsers(cfg config, logger logger.Logger) *grpc.ClientConn {
 }
 
 func newService(users mainflux.UsersServiceClient, dbTracer opentracing.Tracer, cacheTracer opentracing.Tracer, db *sqlx.DB, cacheClient *redis.Client, esClient *redis.Client, logger logger.Logger) things.Service {
-	dbMiddleware := postgres.NewDatabaseMiddleware(db)
+	database := postgres.NewDatabase(db)
 
-	thingsRepo := postgres.NewThingRepository(dbMiddleware)
+	thingsRepo := postgres.NewThingRepository(database)
 	thingsRepo = tracing.ThingRepositoryMiddleware(dbTracer, thingsRepo)
 
-	channelsRepo := postgres.NewChannelRepository(dbMiddleware)
+	channelsRepo := postgres.NewChannelRepository(database)
 	channelsRepo = tracing.ChannelRepositoryMiddleware(dbTracer, channelsRepo)
 
 	chanCache := rediscache.NewChannelCache(cacheClient)
--- a/cmd/things/main.go
+++ b/cmd/things/main.go
@@ -306,7 +306,9 @@ func connectToUsers(cfg config, logger logger.Logger) *grpc.ClientConn {
 }
 
 func newService(users mainflux.UsersServiceClient, dbTracer opentracing.Tracer, cacheTracer opentracing.Tracer, db *sqlx.DB, cacheClient *redis.Client, esClient *redis.Client, logger logger.Logger) things.Service {
-	thingsRepo := postgres.NewThingRepository(db)
+
+	dbMiddleware := postgres.NewDatabaseMiddleware(db)
+	thingsRepo := postgres.NewThingRepository(dbMiddleware)
 	thingsRepo = tracing.ThingRepositoryMiddleware(dbTracer, thingsRepo)
 
 	channelsRepo := postgres.NewChannelRepository(db)
--- a/cmd/things/main.go
+++ b/cmd/things/main.go
@@ -308,10 +308,11 @@ func connectToUsers(cfg config, logger logger.Logger) *grpc.ClientConn {
 func newService(users mainflux.UsersServiceClient, dbTracer opentracing.Tracer, cacheTracer opentracing.Tracer, db *sqlx.DB, cacheClient *redis.Client, esClient *redis.Client, logger logger.Logger) things.Service {
 
 	dbMiddleware := postgres.NewDatabaseMiddleware(db)
+
 	thingsRepo := postgres.NewThingRepository(dbMiddleware)
 	thingsRepo = tracing.ThingRepositoryMiddleware(dbTracer, thingsRepo)
 
-	channelsRepo := postgres.NewChannelRepository(db)
+	channelsRepo := postgres.NewChannelRepository(dbMiddleware)
 	channelsRepo = tracing.ChannelRepositoryMiddleware(dbTracer, channelsRepo)
 
 	chanCache := rediscache.NewChannelCache(cacheClient)
--- a/cmd/users/main.go
+++ b/cmd/users/main.go
@@ -178,7 +178,8 @@ func connectToDB(dbConfig postgres.Config, logger logger.Logger) *sqlx.DB {
 }
 
 func newService(db *sqlx.DB, tracer opentracing.Tracer, secret string, logger logger.Logger) users.Service {
-	repo := tracing.UserRepositoryMiddleware(postgres.New(db), tracer)
+	dbMiddleware := postgres.NewDatabaseMiddleware(db)
+	repo := tracing.UserRepositoryMiddleware(postgres.New(dbMiddleware), tracer)
 	hasher := bcrypt.New()
 	idp := jwt.New(secret)
 
--- a/cmd/users/main.go
+++ b/cmd/users/main.go
@@ -178,8 +178,8 @@ func connectToDB(dbConfig postgres.Config, logger logger.Logger) *sqlx.DB {
 }
 
 func newService(db *sqlx.DB, tracer opentracing.Tracer, secret string, logger logger.Logger) users.Service {
-	dbMiddleware := postgres.NewDatabaseMiddleware(db)
-	repo := tracing.UserRepositoryMiddleware(postgres.New(dbMiddleware), tracer)
+	database := postgres.NewDatabase(db)
+	repo := tracing.UserRepositoryMiddleware(postgres.New(database), tracer)
 	hasher := bcrypt.New()
 	idp := jwt.New(secret)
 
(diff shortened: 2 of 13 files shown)


drasko:
@nwneisen can the WIP be removed or this needs more work?

nwneisen:
Everything is working as expected.  The only issue is what to do about the sql.statement tag.  Current value the raw string without values inserted.

`INSERT INTO things (id, owner, name, key, metadata)
		  VALUES (:id, :owner, :name, :key, :metadata);`

I would suggest appending the values to the string or making them separate tags.

nmarcetic:
@anovakovic01 @srados Can you please do a review?

ghost:
just a note: if sql values are added to tracing, some sensitive data may be leaked.

things/postgres/channels.go
```diff
@@ -34,9 +35,13 @@ func NewChannelRepository(db *sqlx.DB) things.ChannelRepository {
...
 
-func (cr channelRepository) Save(_ context.Context, channel things.Channel) (string, error) {
+func (cr channelRepository) Save(ctx context.Context, channel things.Channel) (string, error) {
+	span := opentracing.SpanFromContext(ctx)
+	defer span.Finish()
+
 	q := `INSERT INTO channels (id, owner, name, metadata)
-        VALUES (:id, :owner, :name, :metadata);`
+		VALUES (:id, :owner, :name, :metadata);`
+	span.SetTag("sql.statement", q)
```
anovakovic01: I suggest that you create interface for the db and write a middleware for that interface that will add these tags. Idea is to not mix repository code with an open tracing code. So if you create this middleware, when you call `NamedExec`, it will under the hood call your middleware, that will add tags, and will call real `NamedExec`. For an example take a look at our logging middlewares, and middlewares in `things/tracing` package.

things/postgres/channels.go
```diff
@@ -34,9 +35,13 @@ func NewChannelRepository(db *sqlx.DB) things.ChannelRepository {
 	}
 }
 
-func (cr channelRepository) Save(_ context.Context, channel things.Channel) (string, error) {
+func (cr channelRepository) Save(ctx context.Context, channel things.Channel) (string, error) {
+	span := opentracing.SpanFromContext(ctx)
+	defer span.Finish()
```
anovakovic01: Isn't this span already closed in `tracing` middleware from `things/tracing` package? Same goes for other spans.

things/tracing/things.go
```diff
@@ -148,12 +148,19 @@ func (tcm thingCacheMiddleware) Remove(ctx context.Context, thingID string) erro
...
 			opentracing.ChildOf(parentSpan.Context()),
 		)
+	} else {
+		span = tracer.StartSpan(opName)
 	}
 
-	return tracer.StartSpan(opName)
+	span.SetTag("span.kind", "client")
+	span.SetTag("peer.service", "postgres")
+	span.SetTag("db.type", "sql")
```
anovakovic01: I think that these tags shouldn't be present here, think about moving these tags to the `thingRepositoryMiddleware` and and `channelRepositoryMiddleware`.

anovakovic01 (changes requested):
I've left you some comments regarding opentracing intrusion in `things` service code. Address same issues in the `users` service.

things/postgres/channels.go
nwneisen: Thank you.  Removed extra span.Finish() calls.

things/postgres/channels.go
nwneisen: Interface added.

things/tracing/things.go
nwneisen: I actually moved these into the new dbMiddleware so that I have access to the query for a tag.  This way all of the tags are set in the same location.

nwneisen:
Tests still need to be adjusted for the new database interface

cmd/things/main.go
```diff
@@ -306,10 +306,13 @@ func connectToUsers(cfg config, logger logger.Logger) *grpc.ClientConn {
 }
 
 func newService(users mainflux.UsersServiceClient, dbTracer opentracing.Tracer, cacheTracer opentracing.Tracer, db *sqlx.DB, cacheClient *redis.Client, esClient *redis.Client, logger logger.Logger) things.Service {
-	thingsRepo := postgres.NewThingRepository(db)
+
```
manuio, line 310: Remove this blank line

manuio (changes requested):

nwneisen:
Tests adjusted to use the new interface.

things/postgres/database.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// NewDatabaseMiddleware creates a ThingDatabase instance
+func NewDatabaseMiddleware(db *sqlx.DB) DatabaseMiddleware {
+	return &databaseMiddleware{
+		db: db,
+	}
+}
+
+func (dm databaseMiddleware) NamedExec(ctx context.Context, query string, args interface{}) (sql.Result, error) {
```
anovakovic01: Rename this method to `NamedExecContext`.

things/postgres/database.go
```diff
@@ -0,0 +1,60 @@
...
+	addSpanTags(ctx, query)
+	return dm.db.NamedExec(query, args)
```
anovakovic01: Use `NamedExecContext` here.

things/postgres/database.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (dm databaseMiddleware) QueryRowx(ctx context.Context, query string, args ...interface{}) *sqlx.Row {
```
anovakovic01: Rename to `QueryRowxContext`.

things/postgres/database.go
```diff
@@ -0,0 +1,60 @@
...
+	addSpanTags(ctx, query)
+	return dm.db.QueryRowx(query, args...)
```
anovakovic01: Use `QueryRowxContext` here.

things/postgres/database.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (dm databaseMiddleware) NamedQuery(ctx context.Context, query string, args interface{}) (*sqlx.Rows, error) {
```
anovakovic01: Rename to `NamedQueryContext`.

things/postgres/database.go
```diff
@@ -0,0 +1,60 @@
...
+	addSpanTags(ctx, query)
+	return dm.db.NamedQuery(query, args)
```
anovakovic01: Use `NamedQueryContext`.

things/postgres/database.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (dm databaseMiddleware) Get(ctx context.Context, dest interface{}, query string, args ...interface{}) error {
```
anovakovic01: Rename to `GetContext`.

things/postgres/database.go
```diff
@@ -0,0 +1,60 @@
...
+	addSpanTags(ctx, query)
+	return dm.db.Get(dest, query, args...)
```
anovakovic01: Use `GetContext`.

things/postgres/database.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+var _ DatabaseMiddleware = (*databaseMiddleware)(nil)
+
+type databaseMiddleware struct {
+	db *sqlx.DB
+}
+
+// DatabaseMiddleware provides a database interface
+type DatabaseMiddleware interface {
```
anovakovic01: Please rename this interface to something like `Database`, or `DB`. Don't want to be aware that this is middleware on an interface level.

users/postgres/database.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+var _ DatabaseMiddleware = (*databaseMiddleware)(nil)
+
+type databaseMiddleware struct {
+	db *sqlx.DB
+}
+
+// DatabaseMiddleware provides a database interface
+type DatabaseMiddleware interface {
```
anovakovic01: Do the same thing for this middleware.

anovakovic01 (changes requested):
Good job! You're almost there.

users/postgres/database.go
nwneisen: These middlewares are currently the same.  I'm wondering if they shouldn't be put in a common location, maybe in a folder next to the `Logger` middleware.  I could see the tag values changing depending on where the interface is used so there would need to be a way to override the `addSpanTags` function.

things/postgres/channels.go
```diff
@@ -15,35 +15,34 @@ import (
...
-	"github.com/jmoiron/sqlx"
 	"github.com/lib/pq"
 	"github.com/mainflux/mainflux/things"
 )
 
 var _ things.ChannelRepository = (*channelRepository)(nil)
 
 type channelRepository struct {
-	db *sqlx.DB
+	dm Database
```
anovakovic01: Please return db name. Same goes for the `users`.

things/postgres/channels.go
```diff
@@ -342,10 +341,10 @@ func (cr channelRepository) hasThing(chanID, thingID string) error {
...
 
 type dbChannel struct {
-	ID       string `db:"id"`
-	Owner    string `db:"owner"`
-	Name     string `db:"name"`
-	Metadata string `db:"metadata"`
+	ID       string `dm:"id"`
+	Owner    string `dm:"owner"`
+	Name     string `dm:"name"`
+	Metadata string `dm:"metadata"`
```
anovakovic01: `db` is not related to the database instance, this is tag that is used in `sqlx` lib. Please these go tags to the old version `db:"id"`, `db:"owner"`, etc. Same goes for the `users`.

anovakovic01 (changes requested):

users/postgres/database.go
anovakovic01: @nwneisen we have an issue for creating the packages for the shared code. I recommend that we leave it for another PR.

drasko: Probably related to https://github.com/mainflux/mainflux/issues/666 which is a placeholder for general package reorganization.

anovakovic01 (approved):
LGTM!

cmd/users/main.go
```diff
@@ -178,7 +178,8 @@ func connectToDB(dbConfig postgres.Config, logger logger.Logger) *sqlx.DB {
 }
 
 func newService(db *sqlx.DB, tracer opentracing.Tracer, secret string, logger logger.Logger) users.Service {
-	repo := tracing.UserRepositoryMiddleware(postgres.New(db), tracer)
+	database := postgres.NewDatabase(db)
+	repo := tracing.UserRepositoryMiddleware(postgres.New(database), tracer)
```
manuio, line 182: Why this New tracer here?

manuio (changes requested):

cmd/users/main.go
nwneisen: The UserReposityoryMiddleware creates the initial span and passes it via context when it calls the db interface.  It needs access to the tracer in order to do so

manuio: My question was more about code structure. Looks like you are passing a `dbTracer` for Things and Channels but not here. Thank you for the answer in any case. I'll resolve this one for one and we will see later.

manuio (approved):

PR #873, opened 2019-09-30
MF-532 - Password reset

### LINKED ISSUE #532: New Feature Request- Need an API to retrieve old Password
Request an API  Interface  so that the User can retrieve old password when forgotten

mteodor:
### What does this do?

adding endpoints for password resetting flow
Adds an endpoint for changing password of currently authenticated user

`Resolves #532`  ( option 1. from https://github.com/mainflux/mainflux/issues/532#issuecomment-450579158)


things/swagger.yaml
```diff
@@ -503,9 +503,9 @@ parameters:
     type: string
     minimum: 0
     required: false
-  Metadata
+  Metadata:
     name: metadata
-    descripton: Metadata filter. Filtering is performed matching the parametar with metadata on top level. Parametar is json.
+    description: Metadata filter. Filtering is performed matching the parametar with metadata on top level. Parametar is json.
```
manuio: fix `parametar`

manuio (changes requested):

users/api/http/transport.go
```diff
@@ -57,13 +60,34 @@ func MakeHandler(svc users.Service, tracer opentracing.Tracer, l log.Logger) htt
 		opts...,
 	))
 
+	mux.Post("/passwd/res-req", kithttp.NewServer(
```
manuio: There is a reason to don't use `password` here? I would use `password/request`.

anovakovic01: Also, this is a request for password change, so I recommend that you use `Patch` instead of post. So based on the REST standards, think about request like this one:
```
PATCH /user
{
    "password": "new_pass"
}
```

mteodor: this res-req endpoint initiates the flow for password password/reset is actually where the password is changed

drasko: @mteodor, please use `PATCH /user`, as @anovakovic01 suggested

users/api/http/endpoint.go
```diff
@@ -27,6 +28,54 @@ func registrationEndpoint(svc users.Service) endpoint.Endpoint {
 	}
 }
 
+// Password reset endpoint serves post request with email of the user
+// for whom password reset flow is to be initiated.
+// If request is succsesfull email with reset link will be sent to the
```
anovakovic01: Correct `successful`.

users/api/http/endpoint.go
```diff
@@ -27,6 +28,54 @@ func registrationEndpoint(svc users.Service) endpoint.Endpoint {
...
+// email specified in the request.
+// Link contains token that has TTL that needs to be verified.
+// When user gets email with reset password link this endpoint can serve
+// that request and return response into the ui form which than can
```
anovakovic01: Correct `then`.

users/api/http/endpoint.go
```diff
@@ -27,6 +28,54 @@ func registrationEndpoint(svc users.Service) endpoint.Endpoint {
...
+// be used to enter email and new password and then request can be submited
```
anovakovic01: Correct `submitted`.

users/api/http/responses.go
```diff
@@ -50,3 +56,20 @@ func (res identityRes) Headers() map[string]string {
 func (res identityRes) Empty() bool {
 	return false
 }
+
+type resetPassRes struct {
+	Msg   string `json:"msg"`
+	Error string `json:"error"`
```
anovakovic01: I'm not sure if you should have an error here. @blokovi is working on the PR that should tackle error handling and HTTP error response.

users/api/http/transport.go
manuio: @mteodor I understood that it initiates the flow. That's why `password/request` makes sense.

users/api/http/endpoint.go
```diff
@@ -27,6 +28,54 @@ func registrationEndpoint(svc users.Service) endpoint.Endpoint {
...
+			res.Error = err.Error()
+			return res, err
+		}
+
+		err = svc.SaveToken(ctx, email, tok)
+		if err != nil {
+			res.Error = err.Error()
+			return res, nil
+		}
+		token.SendToken(req.Host, email, tok)
```
anovakovic01: I think that all of these methods should be under one method in the `Service` interface.

users/postgres/init.go
```diff
@@ -56,6 +56,11 @@ func migrateDB(db *sqlx.DB) error {
 						email	 VARCHAR(254) PRIMARY KEY,
 						password CHAR(60)	  NOT NULL
 					)`,
+					`CREATE TABLE IF NOT EXISTS tokens (
+						user_id    VARCHAR(254),
+						token	 VARCHAR(254),
+						FOREIGN KEY (user_id) REFERENCES users (email) ON DELETE CASCADE ON UPDATE CASCADE
+					)`,
```
anovakovic01: Please create a migration for this (just move `tokens` part into the `users_2`).

users/postgres/users.go
```diff
@@ -66,6 +67,83 @@ func (ur userRepository) RetrieveByID(_ context.Context, email string) (users.Us
...
+
+	return nil
+}
+
+func (ur userRepository) ChangePassword(_ context.Context, email, token, password string) error {
+	q := `UPDATE users SET  password = :password  WHERE email = :email`
+
+	db := struct {
+		Password string
+		Email    string
```
anovakovic01: Please add `db` tags here.

users/postgres/users.go
```diff
@@ -66,6 +67,83 @@ func (ur userRepository) RetrieveByID(_ context.Context, email string) (users.Us
...
+		q = `UPDATE tokens SET token = :token  WHERE user_id = :email`
+	}
+
+	hash, err := token.Hash(tok)
+	if err != nil {
+		return err
+	}
+	db := struct {
+		Email string
+		Token string
```
anovakovic01: Please add `db` tags here.

users/api/http/requests.go
```diff
@@ -31,3 +31,37 @@ func (req viewUserInfoReq) validate() error {
...
+	if req.Host == "" {
+		return users.ErrMalformedEntity
+	}
+	return nil
+}
+
+type resetTokenReq struct {
+	Token    string
+	Email    string
+	Password string
```
anovakovic01: If these are json fields, please add json tags.

users/api/http/requests.go
```diff
@@ -31,3 +31,37 @@ func (req viewUserInfoReq) validate() error {
 	}
 	return nil
 }
+
+type passwResetReq struct {
+	Email string
+	Host  string
```
anovakovic01: If these are json fields, please add json tags.

users/api/http/transport.go
```diff
@@ -57,13 +60,34 @@ func MakeHandler(svc users.Service, tracer opentracing.Tracer, l log.Logger) htt
...
+		kitot.TraceServer(tracer, "res-req")(passwordResetRequestEndpoint(svc)),
+		decodePasswordResetRequest,
+		encodeResponse,
+		opts...,
+	))
+
+	mux.Post("/passwd/reset", kithttp.NewServer(
```
anovakovic01: @mteodor I recommend that you use `PATCH /users` here and for the upper one I agree with @manuio, but instead of `password/request` maybe you should put `/password/reset` and for the successful response return `201` because it's part of an async flow.

users/token/token.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+var (
+	emailLength   = 254
+	ttlLength     = 4
+	secretLength  = 20
+	tokenLength   = ttlLength + emailLength // Max token length is 4 bytes + max email length
+	hashCost      = 10
+	tokenDuration = 5                      // Recovery token TTL in minutes, reperesents token time to live
+	secret        = "fcERNb7KpM3WyAmguJMZ" // Random string for secret key, required for signing
```
anovakovic01: You should probably move this to the env var if you haven't already. Also, I recommend that you use https://github.com/dgrijalva/jwt-go here.

anovakovic01 (changes requested):

users/api/http/transport.go
manuio: @anovakovic01 Then both endpoints would have the same name

anovakovic01: @manuio They wouldn't. I've written that the second one should be `PATCH /users` and that the first one should be `POST /password/reset`.

users/api/http/responses.go
mteodor: I'm not sure either, but something needs to be displayed in frontend. For example if token is expired. We can remove this error. Maybe that "token expired", and "non-existent" user messages can be in Msg field.

users/api/http/endpoint.go
anovakovic01: Looks like I wasn't clear enough. All svc methods that are called here should be reduced to one method call of the `Service` interface. This `token` should not be present here but in the `Service` implementation. It's ok that it's a separate method, but here we should see only one method call to the `svc`. Point is that `http` layer depends only on the `users.Service`, while service itself will then call multiple other methods (it should contain business logic).

users/api/http/transport.go
manuio: @anovakovic01 all clear now. I thought that you were proposing PATCH for both. I agree with you in any case.

users/token/token.go
nmarcetic: IMHO JWT is too big plus Verify() must include DB query to compare stored token and token we get which also means we need to hash JWT and save it in DB.

users/api/http/endpoint_test.go
```diff
@@ -175,3 +176,132 @@ func TestLogin(t *testing.T) {
...
+		assert.Nil(t, err, fmt.Sprintf("%s: unexpected error %s", tc.desc, err))
+		token := strings.Trim(string(body), "\n")
+
+		assert.Equal(t, tc.status, res.StatusCode, fmt.Sprintf("%s: expected status code %d got %d", tc.desc, tc.status, res.StatusCode))
+		assert.Equal(t, tc.res, token, fmt.Sprintf("%s: expected body %s got %s", tc.desc, tc.res, token))
+	}
+}
+
+func TestPasswordReset(t *testing.T) {
+
```
drasko: No need for this empty line

users/api/http/endpoint_test.go
```diff
@@ -175,3 +176,132 @@ func TestLogin(t *testing.T) {
...
+	svc := newService()
+	ts := newServer(svc)
+	defer ts.Close()
+	client := ts.Client()
+	reqData := struct {
+		Email    string
+		Password string
+		Token    string
+	}{}
+	// this probably needs to be moved somwhere else
```
drasko: Comments start with capital letters. You can also skip a line here.

users/api/http/endpoint_test.go
```diff
@@ -175,3 +176,132 @@ func TestLogin(t *testing.T) {
...
+	expected := struct {
+		Msg   string `json:"msg"`
+		Error string `json:"error"`
+	}{
+		"", // it would be good to use actuall error constant but it is private
```
drasko: Comments starts with capital letters

users/api/http/transport.go
```diff
@@ -57,13 +60,34 @@ func MakeHandler(svc users.Service, tracer opentracing.Tracer, l log.Logger) htt
...
 	))
 
+	mux.Post("/password/reset", kithttp.NewServer(
+		kitot.TraceServer(tracer, "res-req")(passwordResetRequestEndpoint(svc)),
+		decodePasswordResetRequest,
+		encodeResponse,
+		opts...,
+	))
+
+	mux.Patch("/passwd/reset", kithttp.NewServer(
```
drasko: `/password`, not `/passwd`

users/mail/mail.go
```diff
@@ -0,0 +1,101 @@
...
+		if err != nil {
+			log.Fatalf(err.Error())
+		}
+
+		instance.log = logger
+	})
+	return instance
+}
+
+// Send sends mail.
```
drasko: No need for fullstop at the end of the comment. Only when there are several sentences.

users/mail/mail.go
```diff
@@ -0,0 +1,101 @@
...
+			driver:      mainflux.Env(envMailDriver, defMailDriver),
+			fromAddress: mainflux.Env(envMailFromAddress, defMailFromAddress),
+			fromName:    mainflux.Env(envMailFromName, defMailFromName),
+			host:        mainflux.Env(envMailHost, defMailHost),
+			port:        mainflux.Env(envMailPort, defMailPort),
+			username:    mainflux.Env(envMailUsername, defMailUsername),
+			password:    mainflux.Env(envMailPassword, defMailPassword),
+		}
+
+		// Set up authentication information.
```
drasko: Please remove fullstop at the end of the comment

users/mocks/users.go
```diff
@@ -12,19 +12,22 @@ import (
...
 
 type userRepositoryMock struct {
-	mu    sync.Mutex
-	users map[string]users.User
+	mu     sync.Mutex
+	users  map[string]users.User
+	tokens map[string]string
 }
 
 // NewUserRepository creates in-memory user repository.
```
drasko: No need for fullstop in one-sentence comments

users/service.go
```diff
@@ -28,8 +31,31 @@ var (
 	// ErrNotFound indicates a non-existent entity request.
 	ErrNotFound = errors.New("non-existent entity")
 
+	// ErrUserNotFound indicates a non-existent user request.
```
drasko: PLease avoid fullstops in one-sentence comments

users/service.go
```diff
@@ -28,8 +31,31 @@ var (
...
+	ErrUserNotFound = errors.New("non-existent user")
+
 	// ErrScanMetadata indicates problem with metadata in db
 	ErrScanMetadata = errors.New("Failed to scan metadata")
+
+	// ErrMissingEmail indicates missing email for password reset request.
```
drasko: Drop fullstop

users/service.go
```diff
@@ -28,8 +31,31 @@ var (
...
+	ErrSavingRecoveryToken = errors.New("error saving recovery token")
+
+	// ErrDeletingRecoveryToken indicates error deleting recovery token
+	ErrDeletingRecoveryToken = errors.New("error deleting recovery token")
+
+	// ErrRetrievingRecoveryToken indicates error deleting recovery token
+	ErrRetrievingRecoveryToken = errors.New("error deleting recovery token")
+
+	// ErrMisingResetToken indicates malformed or missing reset token
+	// for reseting password.
```
drasko: No fullstop

users/service.go
```diff
@@ -28,8 +31,31 @@ var (
...
+	ErrMisingResetToken = errors.New("error mising reset token")
+
+	// ErrGeneratingResetToken indicates error in generating password recovery
+	// token.
```
drasko: No fullstop

users/service.go
```diff
@@ -51,6 +77,13 @@ type Service interface {
 
 	// Get authenticated user info for the given token.
```
drasko: No fullstop

users/token/token.go
```diff
@@ -0,0 +1,99 @@
...
+func Generate(email string, offset int) (string, error) {
+
+	exp := tokenDuration + offset
+	if exp < 0 {
+		exp = 0
+	}
+	expires := time.Now().Add(time.Minute * time.Duration(exp))
+
+	// Create a new token object, specifying signing method and the claims
+	// you would like it to contain.
```
drasko: No fullstop

drasko (changes requested):

users/api/http/endpoint_test.go
```diff
@@ -282,11 +293,13 @@ func TestPasswordReset(t *testing.T) {
 		res         string
 		tok         string
 	}{
-		{"password reset with valid token and mail", dataResExisting, contentType, http.StatusOK, expectedSucces, tok},
+		{"password reset with token not valid token and mail", dataResExisting, contentType, http.StatusOK, expectedNotValidYet, tok},
```
manuio: Fix `token not valid token `

manuio (changes requested):

users/mail/mail.go
manuio: `Send sends e-mail`

manuio (changes requested):

users/api/http/requests.go
mteodor: Host is extracted from the referer header, it is used for construction of the reset link

manuio (changes requested):

users/README.md
```diff
@@ -34,6 +34,16 @@ default values.
...
 | MF_USERS_SECRET           | String used for signing tokens                                          | users          |
 | MF_JAEGER_URL             | Jaeger server URL                                                       | localhost:6831 |
+| MF_MAIL_DRIVER            | Mail server driver, mail server for sending reset password token        | smtp           |
+| MF_MAIL_HOST              | Mail server host                                                        | localhost      |
+| MF_MAIL_PORT              | Mail server port                                                        | 25             |
+| MF_MAIL_USERNAME          | Mail server username                                                    |                |
+| MF_MAIL_PASSWORD          | Mail server password                                                    |                |
+| MF_MAIL_FROM_ADDRESS      | Mail "from" address                                                     |                |
+| MF_MAIL_FROM_NAME         | Mail "from" name                                                        |                |
```
manuio: Same here

users/README.md
```diff
@@ -66,6 +76,16 @@ services:
       MF_USERS_SERVER_CERT: [String path to server certificate in pem format]
       MF_USERS_SERVER_KEY: [String path to server key in pem format]
       MF_JAEGER_URL: [Jaeger server URL]
+      MF_MAIL_DRIVER: [Mail server driver smtp]
+      MF_MAIL_HOST: [MF_MAIL_HOST]
+      MF_MAIL_PORT: [MF_MAIL_PORT]
+      MF_MAIL_USERNAME: [MF_MAIL_USERNAME]
+      MF_MAIL_PASSWORD: [MF_MAIL_PASSWORD]
+      MF_MAIL_FROM_ADDRESS: [MF_MAIL_FROM_ADDRESS]
+      MF_MAIL_FROM_NAME: [MF_MAIL_FROM_NAME]
```
manuio: And here

manuio (changes requested):

.env
```diff
@@ -36,6 +36,20 @@ MF_USERS_DB_PASS=mainflux
...
 MF_USERS_SECRET=secret
 
+### Mail utility
+MF_MAIL_DRIVER=smtp
+MF_MAIL_HOST=smtp.mailtrap.io
+MF_MAIL_PORT=2525
+MF_MAIL_USERNAME=18bf7f70705139
+MF_MAIL_PASSWORD=2b0d302e775b1e
+MF_MAIL_FROM_ADDRESS=from@example.com
+MF_MAIL_FROM_NAME=Example
```
manuio: I would use EMAIL instead of MAIL for all envars

manuio (changes requested):

.env
```diff
@@ -36,6 +36,21 @@ MF_USERS_DB_PASS=mainflux
 MF_USERS_DB=users
 MF_USERS_SECRET=secret
 
+### Mail utility
```
manuio: Email

users/README.md
```diff
@@ -34,6 +34,17 @@ default values.
 | MF_USERS_SERVER_KEY       | Path to server key in pem format                                        |                |
 | MF_USERS_SECRET           | String used for signing tokens                                          | users          |
 | MF_JAEGER_URL             | Jaeger server URL                                                       | localhost:6831 |
+| MF_EMAIL_DRIVER           | Mail server driver, mail server for sending reset password token        | smtp           |
```
manuio, line 37: Email server...

users/README.md
```diff
@@ -66,6 +77,17 @@ services:
       MF_USERS_SERVER_CERT: [String path to server certificate in pem format]
       MF_USERS_SERVER_KEY: [String path to server key in pem format]
       MF_JAEGER_URL: [Jaeger server URL]
+      MF_EMAIL_DRIVER: [Mail server driver smtp]
```
manuio, line 80: Email

users/api/http/endpoint_test.go
```diff
@@ -171,3 +172,116 @@ func TestLogin(t *testing.T) {
...
+
+	cases := []struct {
+		desc        string
+		req         string
+		contentType string
+		status      int
+		res         string
+		tok         string
+	}{
+		{"password reset with valid token and mail", dataResExisting, contentType, http.StatusCreated, expectedSuccess, tok},
```
manuio: email

users/api/http/responses.go
```diff
@@ -12,8 +12,12 @@ import (
 var (
 	_ mainflux.Response = (*tokenRes)(nil)
 	_ mainflux.Response = (*identityRes)(nil)
+	_ mainflux.Response = (*resetPassRes)(nil)
 )
 
+// MailSent message response when link is sent
+const MailSent = "Mail with reset link is sent"
```
manuio: Email

users/mail/mail.go
```diff
@@ -0,0 +1,94 @@
+package mail
```
manuio: Use packege email please

manuio (changes requested):

users/README.md
drasko: Actually, mail server is mail server, not email server: https://en.wikipedia.org/wiki/List_of_mail_server_software

So please keep `mail` here.

users/README.md
drasko: Mail server.

users/mail/mail.go
drasko: Mail is quite OK in this context.

users/service.go
```diff
@@ -109,4 +141,37 @@ func (svc usersService) UserInfo(ctx context.Context, token string) (User, error
 		Password: "",
 		Metadata: dbUser.Metadata,
 	}, nil
+
+}
+
+func (svc usersService) GenerateResetToken(ctx context.Context, email, host string) error {
+
```
drasko: No need for empty line here.

users/mocks/users.go
```diff
@@ -47,3 +48,14 @@ func (urm *userRepositoryMock) RetrieveByID(ctx context.Context, email string) (
 
 	return val, nil
 }
+
+// UpdatePassword(
```
drasko: Correct this comment.

drasko (changes requested):

tools/mqtt-bench/client.go
```diff
@@ -88,7 +88,7 @@ func (c *Client) publish(r chan *runResults) {
 
 	for i := 0; i < c.MsgCount; i++ {
 		wg.Add(1)
-		go func(mut *sync.Mutex, wg *sync.WaitGroup, i int, m message) {
+		go func(mut *sync.Mutex, wg *sync.WaitGroup, t *[]*float64, i int, m message) {
```
drasko: PLease do not change this file, it is out of the scope of your PR

users/users.go
```diff
@@ -44,6 +44,9 @@ type UserRepository interface {
 
 	// RetrieveByID retrieves user by its unique identifier (i.e. email).
 	RetrieveByID(context.Context, string) (User, error)
+
+	// UpdatePassword(
```
drasko: Please correct this comment

users/token/token.go
```diff
@@ -0,0 +1,155 @@
+//Package token provides password recovery token generation with TTL
+//Token is sent by email to user as part of recovery URL
```
drasko: Put a space behind `//`

drasko (changes requested):

users/api/http/endpoint_test.go
```diff
@@ -171,3 +172,117 @@ func TestLogin(t *testing.T) {
...
```
anovakovic01: What about an invalid token?

users/api/http/endpoint.go
```diff
@@ -23,6 +24,56 @@ func registrationEndpoint(svc users.Service) endpoint.Endpoint {
...
+		}
+		res := resetPassRes{}
+		email, err := token.Verify(req.Token)
+		if err != nil {
+			res.Msg = err.Error()
+			return res, nil
+		}
+
+		err = svc.UpdatePassword(ctx, email, req.Password)
+		if err != nil {
```
anovakovic01: You can inline this one.

users/api/http/endpoint.go
```diff
@@ -23,6 +24,56 @@ func registrationEndpoint(svc users.Service) endpoint.Endpoint {
...
+// generated in the password reset request endpoint.
+// Token is verified for the TTL and against generated token saved in DB.
+func passwordResetPutEndpoint(svc users.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(resetTokenReq)
+		if err := req.validate(); err != nil {
+			return nil, err
+		}
+		res := resetPassRes{}
+		email, err := token.Verify(req.Token)
```
anovakovic01: I think that this should be called in the `UpdatePassword` method, not here.

users/api/http/endpoint.go
```diff
@@ -23,6 +24,56 @@ func registrationEndpoint(svc users.Service) endpoint.Endpoint {
...
+		}
+		res.Msg = MailSent
+		return res, nil
+	}
+}
+
+// This is post request endpoint that actually sets new password. It requires a token
+// generated in the password reset request endpoint.
+// Token is verified for the TTL and against generated token saved in DB.
+func passwordResetPutEndpoint(svc users.Service) endpoint.Endpoint {
```
anovakovic01: You don't have to put method in the endpoint name (so you can name it just `passwordResetEndpoint`)

users/api/http/transport.go
```diff
@@ -53,13 +56,34 @@ func MakeHandler(svc users.Service, tracer opentracing.Tracer, l log.Logger) htt
 		opts...,
 	))
 
+	mux.Post("/password/request", kithttp.NewServer(
```
anovakovic01: Can you use `/password/reset-request` or something like that? I think that it's not clear from this path that this will start password reset flow.

users/api/http/transport.go
```diff
@@ -53,13 +56,34 @@ func MakeHandler(svc users.Service, tracer opentracing.Tracer, l log.Logger) htt
...
+
 	mux.Post("/tokens", kithttp.NewServer(
 		kitot.TraceServer(tracer, "login")(loginEndpoint(svc)),
 		decodeCredentials,
 		encodeResponse,
 		opts...,
 	))
 
+	mux.Get("/users", kithttp.NewServer(
+		kitot.TraceServer(tracer, "register")(userInfoEndpoint(svc)),
```
anovakovic01: I think that you should change `register` string to `user_info` or something like that.

users/api/http/transport.go
```diff
@@ -114,6 +182,8 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 		w.WriteHeader(http.StatusBadRequest)
 	case users.ErrUnauthorizedAccess:
 		w.WriteHeader(http.StatusForbidden)
+	case users.ErrUserNotFound:
+		w.WriteHeader(http.StatusBadRequest)
```
anovakovic01: Change error name, or send different status code. For not found it makes sense to return `404`.

users/mail/mail.go
```diff
@@ -0,0 +1,94 @@
...
+	defMailFromName    = ""
+
+	envMailDriver      = "MF_MAIL_DRIVER"
+	envMailHost        = "MF_MAIL_HOST"
+	envMailPort        = "MF_MAIL_PORT"
+	envMailUsername    = "MF_MAIL_USERNAME"
+	envMailPassword    = "MF_MAIL_PASSWORD"
+	envMailFromAddress = "MF_MAIL_FROM_ADDRESS"
+	envMailFromName    = "MF_MAIL_FROM_NAME"
+	envMailLogLevel    = "MF_MAIL_LOG_LEVEL"
```
anovakovic01: I think that these should be in the `cmd/users/main.go`

users/mail/mail.go
```diff
@@ -0,0 +1,94 @@
...
+
+		logLevel := mainflux.Env(envMailLogLevel, defMailLogLevel)
+		logger, err := logger.New(os.Stdout, logLevel)
+		if err != nil {
+			log.Fatalf(err.Error())
+		}
+
+		a.log = logger
+	})
+	return a
```
anovakovic01: All of the env vars should be loaded in the main. You can create a separate mail config structure here and pass this mail config as method parameter.

users/service.go
```diff
@@ -110,3 +142,32 @@ func (svc usersService) UserInfo(ctx context.Context, token string) (User, error
...
+	if err != nil || user.Email == "" {
+		return ErrUserNotFound
+	}
+
+	tok, err := token.Generate(email, 0)
+	if err != nil {
+		return ErrGeneratingResetToken
+	}
+
+	token.SendToken(host, email, tok)
```
anovakovic01: Think about using Tokenizer interface that is implemented by the tokenizer structure in the `token` package in order to be able to mock it.

users/service.go
```diff
@@ -110,3 +142,32 @@ func (svc usersService) UserInfo(ctx context.Context, token string) (User, error
...
+	if err != nil || u.Email == "" {
+		return ErrUserNotFound
+	}
+
+	password, err = svc.hasher.Hash(password)
+	if err != nil {
+		return err
+	}
+	err = svc.users.UpdatePassword(ctx, email, password)
+	return err
```
anovakovic01: You can return `svc.users.UpdatePassword(ctx, email, password)` directly.

users/token/token.go
```diff
@@ -0,0 +1,154 @@
...
+func (t *tokenizer) buildBody(host, email, token string) []byte {
+	msg := []byte(fmt.Sprintf("To: %s\r\n"+
+		"Subject: Reset Password!\r\n"+
+		"\r\n"+
+		"You have initiated password reset.\r\n"+
+		"Follow the link below to reset password.\r\n"+
+		"%s%s?token=%s", email, host, t.url, token))
+
+	return msg
+}
```
anovakovic01, line 82: Here you shouldn't have anything related to the mail or password reset. Only token logic.

anovakovic01 (changes requested):

users/mail/mail.go
mteodor: OK, but this way mail can be reused in any other service with just calling mail.Send and providing environment variables,  no need to handle configs in main of the service

users/README.md
manuio: right

users/README.md
manuio: right

users/token/token.go
```diff
@@ -0,0 +1,122 @@
...
+	envTokenSecret   = "MF_TOKEN_SECRET"
+	envTokenDuration = "MF_TOKEN_DURATION"
+	envTokenLogLevel = "MF_TOKEN_DEBUG_LEVEL"
+)
+
+var once sync.Once
+
+type tokenizer struct {
+	hmacSampleSecret []byte // secret for signing token
+	tokenDuration    int    //token in duration in min
```
drasko: Please add space after `//`

drasko (changes requested):

cmd/users/main.go
```diff
@@ -36,39 +38,41 @@ import (
...
+	defDBSSLCert          = ""
+	defDBSSLKey           = ""
+	defDBSSLRootCert      = ""
+	defHTTPPort           = "8180"
+	defGRPCPort           = "8181"
+	defSecret             = "users"
+	defServerCert         = ""
+	defServerKey          = ""
+	defJaegerURL          = ""
+	defTokenResetEndpoint = "/password/reset"
```
drasko: Why is this configurable?

users/api/http/transport.go
```diff
@@ -47,19 +50,47 @@ func MakeHandler(svc users.Service, tracer opentracing.Tracer, l log.Logger) htt
...
 	))
 
+	mux.Post("/password/reset-request", kithttp.NewServer(
+		kitot.TraceServer(tracer, "res-req")(passwordResetRequestEndpoint(svc)),
+		decodePasswordResetRequest,
+		encodeResponse,
+		opts...,
+	))
+
+	mux.Put("/password/reset", kithttp.NewServer(
```
drasko, line 66: This here looks like hard-coded - do we use something that was configures in `main`, and if not why is this configurable in `main`?

drasko (changes requested):
Getting close. I just need clarifications about this configurable URL before I approve.

users/api/http/transport.go
mteodor: this is the flow
1. User opens a page for password reset in a browser where he enters email and clicks submit ( that request goes as POST to /password/reset-request )
2. User gets a link in email, link is with token, clicks on a link making a GET request and lands on UI page where he can enter new and confirm password ( url for creating link to this page is configurable since it may be convenient depending on the front being used )
3. User clicks submit in previously opened page and then makes a PUT request to `/password/reset` and if everything is OK password is changed.

users/mocks/mail.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Mainflux
```
manuio: This file should be email.go

manuio (changes requested):

cmd/users/main.go
mteodor: https://github.com/mainflux/mainflux/pull/873#discussion_r334329934

users/api/http/endpoint_test.go
```diff
@@ -52,15 +53,21 @@ func (tr testRequest) make() (*http.Response, error) {
 	if tr.contentType != "" {
 		req.Header.Set("Content-Type", tr.contentType)
 	}
+
+	if tr.contentType != "" {
+		req.Header.Set("Referer", "http://localhost")
+	}
```
anovakovic01: Do you need this if here? Think about just setting the header.

users/api/metrics.go
```diff
@@ -64,3 +64,30 @@ func (ms *metricsMiddleware) UserInfo(ctx context.Context, key string) (users.Us
...
+		ms.latency.With("method", "generate_reset_token").Observe(time.Since(begin).Seconds())
+	}(time.Now())
+
+	return ms.svc.GenerateResetToken(ctx, email, host)
+}
+
+func (ms *metricsMiddleware) UpdatePassword(ctx context.Context, email, password string) error {
+	defer func(begin time.Time) {
+		ms.counter.With("method", "change_password").Add(1)
+		ms.latency.With("method", "change_password").Observe(time.Since(begin).Seconds())
```
anovakovic01, line 80: Use `update_password` instead of `change_password` (this should metch method name).

users/api/metrics.go
```diff
@@ -64,3 +64,30 @@ func (ms *metricsMiddleware) UserInfo(ctx context.Context, key string) (users.Us
...
+	}(time.Now())
+
+	return ms.svc.UpdatePassword(ctx, email, password)
+}
+
+func (ms *metricsMiddleware) SendToken(ctx context.Context, host, email, token string) error {
+	defer func(begin time.Time) {
+		ms.counter.With("method", "change_password").Add(1)
+		ms.latency.With("method", "change_password").Observe(time.Since(begin).Seconds())
```
anovakovic01: Use `send_token` instead of `change_password`.

users/email/email.go
```diff
@@ -0,0 +1,96 @@
...
+	defEmailFromName    = ""
+
+	envEmailDriver      = "MF_EMAIL_DRIVER"
+	envEmailHost        = "MF_EMAIL_HOST"
+	envEmailPort        = "MF_EMAIL_PORT"
+	envEmailUsername    = "MF_EMAIL_USERNAME"
+	envEmailPassword    = "MF_EMAIL_PASSWORD"
+	envEmailFromAddress = "MF_EMAIL_FROM_ADDRESS"
+	envEmailFromName    = "MF_EMAIL_FROM_NAME"
+	envEmailLogLevel    = "MF_EMAIL_LOG_LEVEL"
```
anovakovic01: All env vars and their default values should be in the `main.go`. But the email configuration structure can be kept in the email package.

users/email/email.go
```diff
@@ -0,0 +1,96 @@
...
+	addr string
+	log  logger.Logger
+}
+
+var a *Agent
+var once sync.Once
+
+// Instance - Thread safe creation of mail agent
+func Instance() *Agent {
+	once.Do(func() {
```
anovakovic01: I'm just curious, why do you use `once.Do`?

drasko (dismissed):
LGTM

users/email/email.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Send sends e-mail
+func (a *Agent) Send(to []string, msg []byte) {
+	go func() {
+		err := smtp.SendMail(a.addr, a.auth, a.conf.fromAddress, to, msg)
+		if err != nil {
+			a.log.Error(fmt.Sprintf("Failed to send email:%s", err.Error()))
+		}
+	}()
```
anovakovic01: I recommend that you don't spawn `go routine` here. Instead, spawn it when using `Send`. This is the best practice. You leave to the user of your method to decide if it should be run sync or async.

users/email/email.go
```diff
@@ -0,0 +1,96 @@
...
+	auth smtp.Auth
+	addr string
+	log  logger.Logger
+}
+
+var a *Agent
+var once sync.Once
+
+// Instance - Thread safe creation of mail agent
+func Instance() *Agent {
```
anovakovic01: This about naming it `New()` like in other packets (we use `New` for factory methods).

users/service.go
```diff
@@ -45,21 +71,39 @@ type Service interface {
...
+
+	// GenerateResetToken email where mail will be sent.
+	// host is used for generating reset link.
+	GenerateResetToken(_ context.Context, email, host string) error
+
+	// UpdatePassword change users password in reset flow
+	UpdatePassword(_ context.Context, token, password string) error
+
+	//SendToken sends reset password link to email
+	SendToken(_ context.Context, host, email, token string) error
```
anovakovic01: Add open tracing support for these methods, or add an issue, so that we don't forget about it.

users/service.go
```diff
@@ -109,4 +153,51 @@ func (svc usersService) UserInfo(ctx context.Context, token string) (User, error
...
+
+// SendToken sends password recovery link to user
+func (svc usersService) SendToken(_ context.Context, host, email, token string) error {
+	url := svc.email.ResetURL
+	msg := []byte(fmt.Sprintf("To: %s\r\n"+
+		"Subject: Reset Password!\r\n"+
+		"\r\n"+
+		"You have initiated password reset.\r\n"+
+		"Follow the link below to reset password.\r\n"+
+		"%s%s?token=%s", email, host, url, token))
```
anovakovic01: I think that this message should be constructed in the `email` package as part of the emailer.

users/token/token.go
```diff
@@ -0,0 +1,119 @@
...
+	hmacSampleSecret []byte // secret for signing token
+	tokenDuration    int    // token in duration in min
+	logger           logger.Logger
+}
+
+var t *tokenizer
+
+// Instance - Thread safe creation singleton instance of tokenizer.
+// Used for creating password reset token.
+func Instance() users.Tokenizer {
```
anovakovic01: Think about renaming it to `New()`. Also, do you need to force singleton for these structures? I think that there is no need for this (you will instantiate them in the `main.go` one anyway).

users/token/token.go
```diff
@@ -0,0 +1,119 @@
...
+	// you would like it to contain
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.MapClaims{
+		"email": email,
+		"exp":   expires.Unix(),
+		"nbf":   nbf.Unix(),
+	})
+
+	// Sign and get the complete encoded token as a string using the secret
+	tokenString, err := token.SignedString(t.hmacSampleSecret)
+	fmt.Println(tokenString)
```
anovakovic01: Remove `Println`

users/token/token.go
```diff
@@ -0,0 +1,119 @@
...
+	if claims, ok := token.Claims.(jwt.MapClaims); ok && token.Valid {
+		if claims.VerifyExpiresAt(time.Now().Unix(), false) == false {
+			t.logger.Error(ErrExpiredToken.Error())
+			return "", ErrExpiredToken
+		}
+		email = claims["email"].(string)
+
+	} else {
+		return email, err
+	}
```
anovakovic01, line 80: You can remove `else` and just `return email, err`.

users/token/token_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+func TestVerify(t *testing.T) {
+	tok, err := token.Instance().Generate(email, 0)
+	if err != nil {
+		t.Errorf("Token generation failed.")
+	}
+
+	e, err := token.Instance().Verify(tok)
+	if err != nil {
+		fmt.Println("Here is a error:", err)
```
anovakovic01: Remove `Println`

users/token/token_test.go
```diff
@@ -0,0 +1,38 @@
...
+		t.Errorf("Token verification failed.")
+	}
+	if e != email {
+		t.Errorf("Token verification failed.")
+	}
```
anovakovic01: Use testify library instead of `if`s.

users/service.go
```diff
@@ -45,21 +71,39 @@ type Service interface {
...
+	SendToken(_ context.Context, host, email, token string) error
 }
 
 var _ Service = (*usersService)(nil)
 
+// Emailer used for sending messages to user
+type Emailer struct {
+	ResetURL string
+	Agent    *email.Agent
+}
```
anovakovic01: `Emailer` can be in the `email` package, and here you can create `Emailer` interface with `Send` method.

users/service.go
```diff
@@ -14,18 +17,41 @@ var (
 	ErrConflict = errors.New("email already taken")
 
 	// ErrMalformedEntity indicates malformed entity specification (e.g.
-	// invalid username or password).
+	// invalid username or password)
```
anovakovic01: I think that we should have `.` at the end because godoc will generate documentation from these comments. And documentation won't look nice without normal sentences.

anovakovic01 (changes requested):

users/email/email.go
mteodor: ok, but this is like singleton and thats whye once.Do is used

users/service.go
mteodor: Idea is that email package can be reusable in other services ResetURL is users specific so that is why I put it here. Same comment goes for the message text.

users/email/email.go
mteodor: to ensure singleton

internal/email/email.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Set up authentication information
+	a.auth = smtp.PlainAuth("", c.Username, c.Password, c.Host)
+	a.addr = fmt.Sprintf("%s:%s", c.Host, c.Port)
+	if t != nil {
+		a.tmpl = t
+		return a
+	}
+
+	tmpl, _ := template.ParseFiles(c.Template)
```
anovakovic01: Don't ignore error.

internal/email/email.go
```diff
@@ -0,0 +1,100 @@
...
+	a.tmpl = tmpl
+	return a
+}
+
+// Init initializes mailing agent
+func (a *Agent) SetTemplate(t *template.Template) {
+	a.tmpl = t
+}
```
anovakovic01, line 64: Do you need to be able to set a template after Agent has been initialized? I would say that there is no need for this method (`New` will initialize everything that's needed).

internal/email/email.go
```diff
@@ -0,0 +1,100 @@
...
+		Content: Content,
+		Footer:  Footer,
+	}
+
+	if From == "" {
+		tmpl.From = a.conf.FromName
+	}
+
+	err := a.tmpl.Execute(email, tmpl)
+	if err != nil {
```
anovakovic01: You can inline this.

users/api/http/endpoint.go
```diff
@@ -23,6 +23,52 @@ func registrationEndpoint(svc users.Service) endpoint.Endpoint {
...
+		req := request.(passwResetReq)
+
+		if err := req.validate(); err != nil {
+			return nil, err
+		}
+
+		res := passwChangeRes{}
+		email := req.Email
+		err := svc.GenerateResetToken(ctx, email, req.Host)
+		if err != nil {
```
anovakovic01: You can inline this.

users/api/http/endpoint.go
```diff
@@ -40,6 +86,38 @@ func userInfoEndpoint(svc users.Service) endpoint.Endpoint {
...
+		u, err := svc.UserInfo(ctx, req.Token)
+		if err != nil {
+			res.Msg = err.Error()
+			return res, nil
+		}
+
+		u.Password = req.OldPassword
+
+		_, err = svc.Login(ctx, u)
+		if err != nil {
```
anovakovic01: You can inline this.

users/api/http/endpoint.go
```diff
@@ -40,6 +86,38 @@ func userInfoEndpoint(svc users.Service) endpoint.Endpoint {
...
+			res.Msg = err.Error()
+			return res, nil
+		}
+
+		err = svc.UpdatePassword(ctx, req.Token, req.Password)
+		if err != nil {
```
anovakovic01: Same here, but maybe here it would be too long.

users/api/http/requests.go
```diff
@@ -27,3 +27,53 @@ func (req viewUserInfoReq) validate() error {
...
+	if req.Password != req.ConfPass {
+		return users.ErrMalformedEntity
+	}
+	return nil
+}
+
+type passwChangeReq struct {
+	Token       string `json:"token,omitempty"`
+	Password    string `json:"password,omitempty"`
+	OldPassword string `json:"oldPassword,omitempty"`
```
anovakovic01: Just to check, do you need all of these `omitempty`? The same question goes for the other request.

users/jwt/idp.go
```diff
@@ -55,7 +55,10 @@ func (idp *jwtIdentityProvider) Identity(key string) (string, error) {
 	}
 
 	if claims, ok := token.Claims.(jwt.MapClaims); ok && token.Valid {
-		return claims["sub"].(string), nil
+		sub := claims["sub"]
+		if sub != nil {
```
anovakovic01: You can inline this.

users/mocks/users.go
```diff
@@ -47,3 +47,13 @@ func (urm *userRepositoryMock) RetrieveByID(ctx context.Context, email string) (
 
 	return val, nil
 }
+
+func (urm *userRepositoryMock) UpdatePassword(_ context.Context, email, password string) error {
+	urm.mu.Lock()
+	defer urm.mu.Unlock()
+	_, ok := urm.users[email]
+	if !ok {
```
anovakovic01: You can inline this.

users/token/token.go
```diff
@@ -0,0 +1,81 @@
...
+	// ErrWrongSignature wrong signature
+	ErrWrongSignature = errors.New("Wrong token signature")
+)
+
+type tokenizer struct {
+	hmacSampleSecret []byte // secret for signing token
+	tokenDuration    int    // token in duration in min
+}
+
+var t *tokenizer
```
anovakovic01: Do you need this?

anovakovic01 (changes requested):

cmd/users/main.go
```diff
@@ -99,7 +139,7 @@ func main() {
 	dbTracer, dbCloser := initJaeger("users_db", cfg.jaegerURL, logger)
 	defer dbCloser.Close()
 
-	svc := newService(db, dbTracer, cfg.secret, logger)
+	svc := newService(db, dbTracer, cfg.secret, cfg.resetURL, cfg, logger)
```
manuio: If you are passing `cfg` as argument it's redundant to also pass  `cfg.secret, cfg.resetURL`

cmd/users/main.go
```diff
@@ -128,16 +168,36 @@ func loadConfig() config {
...
+
+	tokConf := tokConfig{
+		hmacSampleSecret: []byte(mainflux.Env(envTokenSecret, defTokenSecret)),
+		tokenDuration:    mainflux.Env(envTokenDuration, defTokenDuration),
+	}
+
 	return config{
 		logLevel:   mainflux.Env(envLogLevel, defLogLevel),
 		dbConfig:   dbConfig,
+		emailConf:  emlConf,
```
manuio: You could use `emailConf` instead of `emlConf`

cmd/users/main.go
```diff
@@ -128,16 +168,36 @@ func loadConfig() config {
...
+		tokConf:    tokConf,
```
manuio: You could use `tokenConf` instead of `tokConf`

internal/email/email.go
```diff
@@ -0,0 +1,100 @@
...
+	"errors"
+	"fmt"
+	"html/template"
+	"net/smtp"
+
+	"github.com/mainflux/mainflux/logger"
+)
+
+var (
+	// ErrMissingEmailTmpl missing email template file
```
manuio: I would use `ErrMissingEmailTemplate`

internal/email/email.go
```diff
@@ -0,0 +1,100 @@
...
+	ErrMissingEmailTmpl = errors.New("Missing email template file")
```
manuio: `e-mail`

manuio (changes requested):

internal/email/email.go
mteodor: this is new method  it shuldnt retur error and I cant log the error and anyways if thereis an error tmpl will be nil so it will be reported when trying to send an email

anovakovic01: OK, but you will wait for that call in order to fail. IMHO if a template is invalid, the whole service should fail with an error.

mteodor: what about situation when service restarts and template somehow was deleted, whole service will be unavailable cause 10% of functionality is not available?

manuio (dismissed):
LGTM

drasko (dismissed):
LGTM

dborovcanin (changes requested):

docker/Dockerfile
```diff
@@ -11,5 +11,6 @@ RUN apk update \
     && mv build/mainflux-$SVC /exe
 
 FROM scratch
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```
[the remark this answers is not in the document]
mteodor: sure

users/README.md
```diff
@@ -34,6 +34,17 @@ default values.
...
+| MF_EMAIL_PORT             | Mail server port                                                        | 25             |
+| MF_EMAIL_USERNAME         | Mail server username                                                    |                |
+| MF_EMAIL_PASSWORD         | Mail server password                                                    |                |
+| MF_EMAIL_FROM_ADDRESS     | Email "from" address                                                     |                |
+| MF_EMAIL_FROM_NAME        | Email "from" name                                                        |                |
+| MF_TOKEN_SECRET           | Token secret for signing token used for password reset                  |                |
+| MF_TOKEN_DURATION         | Token duration in minutes                                               | 5              |
+| MF_TOKEN_DEBUG_LEVEL      | Token util debug level                                                  | info           |
+| MF_TOKEN_RESET_ENDPOINT   | Password request reset endpoint, for constructing link                  | /password/reset|
```
manuio: Please fix this to `/reset-request`

manuio (changes requested):

internal/email/email.go
```diff
@@ -0,0 +1,96 @@
...
+type Agent struct {
+	conf *Config
+	auth smtp.Auth
+	addr string
+	log  logger.Logger
+	tmpl *template.Template
+}
+
+// New creates new email agent
+func New(c *Config, t *template.Template) (*Agent, error) {
```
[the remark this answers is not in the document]
manuio: I agree with @dusanb94 I would add that you don't need a pointer here.

manuio (changes requested):

users/api/http/endpoint.go
```diff
@@ -23,6 +23,56 @@ func registrationEndpoint(svc users.Service) endpoint.Endpoint {
...
+// Password reset request endpoint.
+// When successful password reset link is generated.
+// Link is generated using MF_TOKEN_RESET_ENDPOINT env.
+// and value from Referer header for host.
+// {Referer}+{MF_TOKEN_RESET_ENDPOINT}+{token=TOKEN}
+// http://mainflux.com/reset-request?token=xxxxxxxxxxx.
+// Email with a link is being sent to the user.
+// When user clicks on a link it should get the ui with form to
+// enter new password, when form is submitted token and new password
+// must be sent as PUT request to 'password/reset' passwordResetEndpoint
```
dborovcanin, line 35: There is no need to mention UI here. In order to recover the password, issue the PUT request containing the token from the link and new password to the `password/reset`.

users/service.go
```diff
@@ -45,8 +68,22 @@ type Service interface {
...
-	// Get authenticated user info for the given token.
+	// Get authenticated user info for the given token
 	UserInfo(ctx context.Context, token string) (User, error)
+
+	// GenerateResetToken email where mail will be sent.
+	// host is used for generating reset link.
+	GenerateResetToken(_ context.Context, email, host string) error
+
+	// ChangePassword change users password for authenticated user.
+	ChangePassword(_ context.Context, authToken, password, oldPassword string) error
```
dborovcanin, line 70: Consider renaming to `UpdatePassword` (`UpdateThing`, `UpdateChannel`).

users/service.go
```diff
@@ -109,4 +148,69 @@ func (svc usersService) UserInfo(ctx context.Context, token string) (User, error
...
+
+	password, err = svc.hasher.Hash(password)
+	if err != nil {
+		return err
+	}
+	return svc.users.UpdatePassword(ctx, email, password)
+}
+
+func (svc usersService) ChangePassword(ctx context.Context, authToken, password, oldPassword string) error {
+
```
dborovcanin: Please remove the blank line.

users/service.go
mteodor: I've considered it and somehow this is best for since UpdatePassword is in repo and it is like primitive action that is part of ChangePassword and ResetPassword

dborovcanin (dismissed):

users/api/http/requests.go
```diff
@@ -27,3 +27,53 @@ func (req viewUserInfoReq) validate() error {
...
+	if req.Email == "" || req.Host == "" {
+		return users.ErrMalformedEntity
+	}
+	return nil
+}
+
+type resetTokenReq struct {
+	Token    string `json:"token"`
+	Password string `json:"password"`
+	ConfPass string `json:"confirmPassword"`
```
manuio: Use `confirm_password` please

users/api/http/requests.go
```diff
@@ -27,3 +27,53 @@ func (req viewUserInfoReq) validate() error {
...
+	if req.Password != req.ConfPass {
+		return users.ErrMalformedEntity
+	}
+	return nil
+}
+
+type passwChangeReq struct {
+	Token       string `json:"token"`
+	Password    string `json:"password"`
+	OldPassword string `json:"oldPassword"`
```
manuio: Use `old_password` please

users/api/http/endpoint_test.go
```diff
@@ -171,3 +178,216 @@ func TestLogin(t *testing.T) {
...
+	tokenizer := token.New([]byte("secret"), 1)
+	resData := struct {
+		Msg string `json:"msg"`
+	}{
+		"",
+	}
+	reqData := struct {
+		Token    string `json:"token,omitempty"`
+		Password string `json:"password,omitempty"`
+		ConfPass string `json:"confirmPassword,omitempty"`
```
manuio: Use `confirm_password` please

users/api/http/endpoint_test.go
```diff
@@ -171,3 +178,216 @@ func TestLogin(t *testing.T) {
...
+		Msg string `json:"msg"`
+	}{
+		"",
+	}
+	expectedSuccess := toJSON(resData)
+
+	reqData := struct {
+		Token    string `json:"token,omitempty"`
+		Password string `json:"password,omitempty"`
+		OldPassw string `json:"oldPassword,omitempty"`
```
manuio: Use `old_password` please

manuio (changes requested):

sdk/go/users_test.go
```diff
@@ -28,8 +29,10 @@ func newUserService() users.Service {
 	repo := mocks.NewUserRepository()
 	hasher := mocks.NewHasher()
 	idp := mocks.NewIdentityProvider()
+	tok := mocks.NewTokenizer()
+	emailer := mocks.NewEmailer()
 
-	return users.New(repo, hasher, idp)
+	return users.New(repo, hasher, idp, emailer, tok)
```
manuio, line 35: You are using `email, token` instead of `emailer, tok` in the rest of the code

manuio (changes requested):

manuio (dismissed):
LGTM

dborovcanin (dismissed):

dborovcanin (dismissed):

configs/resetPasswEmail.tmpl
```diff
@@ -0,0 +1,7 @@
+To: {{.To}}
+From: {{.From}}
+Subject: {{.Subject}}
+{{.Header}}
+{{.Content}}
```
nmarcetic, line 5: Can you move the content here? 
Idea was to use a template's because we don't want cont's with text in code, We want the static text in template's. 
Move 
```
const (
	message = `
You have initiated password reset.
Follow the link below to reset password.`
)
```
Here and pass reset URL as a variable in the template.

nmarcetic: Also, I don't like the location of templates `config/<template>` 
I would put it in `/users/emailer/templates` and call `mailer` not `emailer`, details ;)

users/api/http/endpoint.go
```diff
@@ -23,6 +23,56 @@ func registrationEndpoint(svc users.Service) endpoint.Endpoint {
...
+		if err := req.validate(); err != nil {
+			return nil, err
+		}
+		res := passwChangeRes{}
+
+		if err := svc.ResetPassword(ctx, req.Token, req.Password); err != nil {
+			res.Msg = err.Error()
+			return res, nil
+		}
+		res.Msg = ""
```
nmarcetic, line 71: What's the purpose of res.Msg to indicate an error? I don't see this msg if res is successful.

users/emailer/emailer.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/mainflux/mainflux/internal/email"
+	"github.com/mainflux/mainflux/users"
+)
+
+const (
+	message = `
+You have initiated password reset.
+Follow the link below to reset password.`
+)
+
```
nmarcetic: Move this to template.

anovakovic01 (dismissed):
LGTM!

users/service.go
```diff
@@ -13,19 +13,42 @@ var (
...
+	// ErrUserNotFound indicates a non-existent user request
+	ErrUserNotFound = errors.New("non-existent user")
+
 	// ErrScanMetadata indicates problem with metadata in db
 	ErrScanMetadata = errors.New("Failed to scan metadata")
+
+	// ErrMissingEmail indicates missing email for password reset request
+	ErrMissingEmail = errors.New("missing email for password reset")
+
+	// ErrSavingRecoveryToken indicates error saving recovery token
```
nmarcetic: We are not saving recovery token, we switched to JWT, right?

configs/resetPasswEmail.tmpl
mteodor: mailer was my initial changed to emailer per @manuio request

nmarcetic (approved):

manuio (approved):
LGTM

PR #889, opened 2019-10-16
MF-484 - Add bulk provisioning for things and channels

### LINKED ISSUE #484: Add bulk provisioning via CSV files
An API endpoint should be added so that we can send a CSV file and do automatic provisioning of a series of entities.

This provisioning should be done in a transaction, so that we can roll-back the changes in the case of error.

nwneisen:
### What does this do?
Adds an endpoint for provisioning multiple things or channels.  A transaction is used and will rollback changes if there are any errors.  SDK function allows a JSON or CSV file to be provided for specifying entities.

Resolves #484.

An initial provisioning method was already implemented in the CLI by making multiple calls to create a thing.  This moves that functionality out of the CLI to a JSON endpoint, adds transactions, and provides a convenient method to specify a file of entities.

Yes

Yes


ghost:
@nwneisen thanks for the contribution, but may I ask, do we need to update docs as well. End users should be aware of this great functionality being available to them. @drasko what do you think?

drasko:
@srados yes, of course. I am a big fan of documentation ;).

More seriously - I would say that PR is not complete without including the tests and documentaion.

@nwneisen thanks a lot for this great contribution!

sdk/go/things.go
```diff
@@ -48,6 +51,74 @@ func (sdk mfSDK) CreateThing(thing Thing, token string) (string, error) {
...
+	}
+
+	fmt.Println(string(body))
+
+	var p provisionThingsRes
+	if err := json.Unmarshal(body, &p); err != nil {
+		return []Thing{}, err
+	}
+
+	// fmt.Println(p)
```
drasko: Remove dead code.

things/api/things/http/endpoint.go
```diff
@@ -29,9 +29,48 @@ func addThingEndpoint(svc things.Service) endpoint.Endpoint {
...
-			id:      saved.ID,
+			ID:      saved.ID,
+			created: true,
+		}
+		
+		return res, nil
+	}
+}
+
+func provisionThingsEndpoint(svc things.Service) endpoint.Endpoint {
```
drasko: Needs to be added to Swagger.

drasko (changes requested):

nwneisen:
@srados I will definetly update the docs to include this.  I was thinking I found something previously for provisioning in the CLI but now I'm not seeing it.

This PR is a WIP to get some feedback on how I did things.  There will be more changes coming.

ghost:
here is the cli doc https://mainflux.readthedocs.io/en/latest/cli/
this part in particular deals with the provisioning https://mainflux.readthedocs.io/en/latest/cli/#system-provisioning
I guess you should update that one and also the cli built in help itself

ghost:
@nwneisen the PR seems to be open for review. Please note that you can open the PR in Draft mode, thus indicating to maintainers it's not ready to be merged. Just a tip..

nwneisen:
Thanks @srados!

I wondered if there was a better way I should be doing it.

drasko:
@nwneisen yes, GitHub lets you open PR in a "Draft" mode, settable during PR creation.

I have attached "WIP" label to this PR, let us know when PR is ready (maintainers usually do not do in-depth reviews on WIP PRs due the lack of time).

manuio:
@nwneisen also, remove `WIP` from the PR title please. The label is enough.

nwneisen:
@drasko @manuio @srados This is now ready for review.

manuio (changes requested):

things/api/things/http/responses.go
```diff
@@ -63,6 +63,33 @@ func (res thingRes) Empty() bool {
...
+		return http.StatusCreated
+	}
+
+	return http.StatusOK
+}
+
+func (res thingsRes) Headers() map[string]string {
+	if res.created {
+		return map[string]string{
+			"Location": "/things/provision",
```
manuio: I would use something else than `/things/provision` to set the Location.

sdk/go/channels.go
```diff
@@ -235,3 +289,29 @@ func (sdk mfSDK) DeleteChannel(id, token string) error {
 
 	return nil
 }
+
+func channelsFromFile(path string, channels []Channel) error {
```
manuio: Can you please modify this functions into `func channelsFromFile(path string) ([]Channel, error) {`

manuio (changes requested):

things/api/things/http/responses.go
nwneisen: The other responses that return a list of values have an empty location.  I'll go ahead and change this to match.

sdk/go/sdk.go
```diff
@@ -152,6 +155,9 @@ type SDK interface {
 	// CreateChannel creates new channel and returns its id.
 	CreateChannel(channel Channel, token string) (string, error)
 
+	// ProvisionChannels registers new channels and returns their ids.
+	ProvisionChannels(path string, token string) ([]Channel, error)
```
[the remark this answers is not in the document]
nwneisen: The functions should probably be `CreateChannel` and `BulkCreateChannels` or `ProvisionChannels` and `BulkProvisionChannels`.  The same terminology should be used within the docs.

nwneisen:
@manuio Pointed out that that "create" and "provision" are being used interchangeably both in code and the documentation.  It would be good to decide on one term and be consistent throughout.  

It doesn't make much difference to me but I believe provisioning typically implies creating something with initial values while creating could have initial values or be empty.  The current endpoints do match this behavior as calling `CreateThing` without a payload will return a `Thing` with only an ID while calling `ProvisionThings` without a payload will return an error.  

It's also worth noting that with the exception of the blank payload, the bulk provisioning could be used to replace creating a single thing.  This would be a breaking change but might be worth considering in  the future to reduce maintenance.  If sending a blank payload to create a thing is desired, then this could also be added to the bulk provisioning endpoint.

Thoughts?

sdk/go/sdk.go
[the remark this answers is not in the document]
manuio: I prefer `CreateChannel` and `BulkCreateChannels`
@drasko what do you think?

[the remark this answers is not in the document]
drasko: I'd say `Create` as well, in order to stay consistent with what we have so far.

docs/provisioning.md
```diff
@@ -56,6 +56,28 @@ Date: Tue, 10 Apr 2018 10:02:59 GMT
 Content-Length: 0
 ```
 
+### Bulk Provisioning things
```
drasko: Either capitalize all words or just first one.

docs/provisioning.md
```diff
@@ -119,9 +141,9 @@ In order to remove you own thing you can send following request:
 curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X DELETE -H "Authorization: <user_auth_token>" https://localhost/things/<thing_id>
 ```
 
-### Provisioning channels
+### Creating channels
```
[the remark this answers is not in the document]
drasko: I prefer `Provisioning` here.

docs/provisioning.md
```diff
@@ -35,10 +35,10 @@ Response should look like this:
 Before proceeding, make sure that you have created a new account, and obtained
 an authorization key.
 
-### Provisioning things
+### Creating things
```
drasko: I prefer `Provisioning` here.

drasko (changes requested):

manuio:
@nwneisen IMO having such endpoints would be confusing. You can already create `Things` with or without payload. It will return the ID in the `Location` header response.
iMO `Create` and `Provision` means the same. I would say that `provison` can be used for docs and `create` for the code.

sdk/go/sdk.go
```diff
@@ -79,10 +79,10 @@ type User struct {
...
 // Thing represents mainflux thing.
 type Thing struct {
-	ID       string                 `json:"id,omitempty"`
-	Name     string                 `json:"name,omitempty"`
-	Key      string                 `json:"key,omitempty"`
-	Metadata map[string]interface{} `json:"metadata,omitempty"`
+	ID       string                 `json:"id,omitempty" csv:"id,omitempty"`
+	Name     string                 `json:"name,omitempty" csv:"name,omitempty"`
+	Key      string                 `json:"key,omitempty" csv:"key,omitempty"`
+	Metadata map[string]interface{} `json:"metadata,omitempty" csv:"-"`
```
anovakovic01: Why are you ignoring metadata? Doesn't look nice in the CSV file? If so, maybe use JSON instead of CSV.

sdk/go/things.go
```diff
@@ -48,6 +52,60 @@ func (sdk mfSDK) CreateThing(thing Thing, token string) (string, error) {
...
+	if err != nil {
+		return []Thing{}, err
+	}
+
+	data, err := json.Marshal(things)
+	if err != nil {
+		return []Thing{}, ErrInvalidArgs
+	}
+
+	endpoint := fmt.Sprintf("%s/%s", thingsEndpoint, "bulkCreate")
```
anovakovic01: Maybe this endpoint should just be `/things/bulk`. You already have `POST` method from which is clear that you're creating things.

things/api/things/http/transport.go
```diff
@@ -52,6 +52,13 @@ func MakeHandler(tracer opentracing.Tracer, svc things.Service) http.Handler {
 		opts...,
 	))
 
+	r.Post("/things/bulkCreate", kithttp.NewServer(
```
dborovcanin: Using `/things/bulk` is clear enough, IMHO. Also, you can consider adding a query param such as `bulk=true` to the existing endpoint for single Thing creation (especially since it already is in the plural - `/things`). The same goes for Channels.

dborovcanin (changes requested):

things/postgres/channels.go
```diff
@@ -51,6 +51,44 @@ func (cr channelRepository) Save(ctx context.Context, channel things.Channel) (s
...
+					return []things.Channel{}, things.ErrConflict
+				}
+			}
+
+			return []things.Channel{}, err
+		}
+	}
+
+	err = tx.Commit()
+	if err != nil {
```
anovakovic01: You can inline this and use `:=` instead of `=`.

things/redis/streams.go
```diff
@@ -132,6 +156,30 @@ func (es eventStore) CreateChannel(ctx context.Context, token string, channel th
 	return sch, err
 }
 
+func (es eventStore) BulkCreateChannels(ctx context.Context, token string, channels []things.Channel) ([]things.Channel, error) {
```
anovakovic01: Not sure about these events. Maybe we should break these into multiple events. @drasko any opinion on this?

anovakovic01 (changes requested):

anovakovic01:
> Why not version?

@nmarcetic because there are no tags and versions for this lib.

Gopkg.toml
```diff
@@ -122,3 +122,7 @@
 [[constraint]]
   revision = "ced62fe5104b907b6c16cb7e575c17b2e62ceddd"
   name = "gonum.org/v1/gonum"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/gocarina/gocsv"
```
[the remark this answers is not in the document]
nwneisen: I would've liked to stay with `encoding/csv` as well but I was surprised to see that it did not have a marshal/unmarshal function.  Without marshaling, there will need to be a requirement on the order of the csv's columns and additional checking that the column exists.  I'm not against this method as this library is very young.

things/service.go
```diff
@@ -35,6 +35,9 @@ type Service interface {
 	// AddThing adds new thing to the user identified by the provided key.
 	AddThing(context.Context, string, Thing) (Thing, error)
 
+	// BulkCreateThings adds a list of things to the user identified by the provided key.
+	BulkCreateThings(context.Context, string, []Thing) ([]Thing, error)
```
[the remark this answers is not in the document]
nwneisen: I'm actually going to make this `AddThings` to be consistent with `AddThing` above.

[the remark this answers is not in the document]
nwneisen: My only reason to keep the `bulk` is that the s alone is a subtle difference.  I already caught myself selecting `AddThing` when I meant `AddThings` with autocomplete.

things/api/things/http/transport.go
nwneisen: I mentioned something about combining the endpoints above but I proposed the single endpoint use the bulk endpoint with only 1 thing in the array.  The single endpoint currently accepts 1 JSON object rather than an array so this would need to be changed as well.

Gopkg.toml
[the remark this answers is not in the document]
nwneisen: There are a number of comments about this library.  I'm debating if it should be used...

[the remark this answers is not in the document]
drasko: @nwneisen do we need this library (or any CSV support) at all, when we said that this endpoint should be JSON-only?

sdk/go/sdk.go
nwneisen: Good catch.  The CSV library doesn't seem to support `map` types ATM.  I planned on looking into writing a marshal/unmarshal function for it.  This reminded me to do so.

Gopkg.toml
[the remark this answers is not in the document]
nwneisen: @drasko I was going to comment and make the same suggestion.  I think the best option would be to only expect JSON, even in the SDK, and let people make their own decisions on how to convert CSV or other file types to JSON.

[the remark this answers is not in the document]
drasko: @nwneisen yes, please go for this. We have a consensus for this to be JSON-only, and let apps (or CLI) do the transformation CSV -> JSON eventually.

docs/provisioning.md
```diff
@@ -138,6 +160,28 @@ Date: Tue, 10 Apr 2018 11:30:07 GMT
...
+```
+HTTP/2 201 
+server: nginx/1.16.0
+date: Fri, 18 Oct 2019 03:45:00 GMT
+content-type: application/json
+content-length: 107
+location: ""
+access-control-expose-headers: Location
+
+{"channels":[{"ID":"abcd038e-28ab-492e-b46e-b23d53804f0c"},{"ID":"341cde5e-52b9-49ee-89e1-00b58733afc2"}]}
```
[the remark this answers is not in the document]
nwneisen: I missed adding structure tags in things/api/things/http/response.go.  They have now been added.

things/redis/streams.go
nwneisen: Looking at this again, I'm wondering if it shouldn't be one event for the entire bulk transaction.

nwneisen:
@anovakovic01 @nmarcetic @dusanb94 The `gocarina/gocsv` library was removed due to it being a fairly young library.  The CLI can still handle a CSV file using `encoding/csv` but in general users should make their own decisions on how to handle a CSV file and send a resulting JSON request to the endpoint.

things/redis/streams.go
anovakovic01: Not sure. I'm afraid that we'll have an event that is too large to be sent as a single event. Even processing of this event can take a lot of time. I'm not sure what we should do with this. @nmarcetic @drasko any opinions?

things/api/things/http/requests.go
```diff
@@ -3,7 +3,9 @@
 
 package http
 
-import "github.com/mainflux/mainflux/things"
+import (
+	"github.com/mainflux/mainflux/things"
```
mteodor: this should be one line

things/things.go
```diff
@@ -32,6 +32,10 @@ type ThingRepository interface {
 	// error response.
 	Save(context.Context, Thing) (string, error)
 
+	// BulkSave persistest multiple things. Successful operation is indicated by non-nil
```
mteodor, line 35: persistest -> 'persists'?

nmarcetic:
@nwneisen Great! Especially because of [this](https://github.com/mainflux/mainflux/pull/889#issuecomment-544567855) but in general native `encoding/csv` is perfect.

nmarcetic:
@nwneisen I was thinking that maybe you can consider merging AddThing() one thing and AddThings on a service, repository level. So you will have one method that accepts a slice of things, same for channels. This way you will avoid code duplication (what is a difference actually between bulk and single create, just a slice vs single thing, everything else is the same, you will drop at least hundred LOC), the duplicate test's also, etc... Different API endpoints (for sake of simplicity and easy explanation), or even the same with query param as [suggested here](https://github.com/mainflux/mainflux/pull/889#discussion_r336987662), but same service/repository logic. Does those this make sense to you?

things/api/things/http/requests.go
nwneisen: `gofmt` seems to prefer this method (multi-line parenthesis).  I switched it back as that is how it was originally.

things/things.go
nwneisen: Good catch.  Thanks

nwneisen:
@nmarcetic I agree.  This is along the same lines as what I was thinking with [this](https://github.com/mainflux/mainflux/pull/889#issuecomment-543821259) comment.  I like your idea of keeping the endpoints the same.  I think it makes the endpoints easier to understand.  

I'll add this change to the PR.  I'm also going to switch `AddThings` back to `CreateThings` that that it matches `CreateChannels`.

things/redis/streams.go
drasko: I'd say that premature optimisation is mother of all evil.

Let's make it functional now, we'll optimize later.

nmarcetic: I agree with @anovakovic01 Firing event for each device sounds OK to me. It will cause a high load on ES and I/O but this a cost for bulk actions. This is a true state in the system.

drasko: @nmarcetic @anovakovic01 agreed.

@nwneisen please proceed according to Aleksandar's and Nikola's guidelines.

nwneisen:
@nmarcetic I'm actually going to remove the single `AddThing`/`CreateChannel` in a followup PR.  There are a lot of tests that use these to initialize test data and it will be a decent sized code change.  This PR is already pretty big.

things/redis/streams.go
nwneisen: @anovakovic01 @nmarcetic @drasko Just making sure I understood correctly, the current change I have iterates the slice of things/channels and creates a new event for each, so no change is needed.  Or did I miss something with how events are working?

anovakovic01: You're right, that's what we want, an event per thing. Marking this one as resolved.

drasko (dismissed):
LGTM

bootstrap/mocks/things.go
```diff
@@ -51,6 +51,26 @@ func (svc *mainfluxThings) AddThing(_ context.Context, owner string, thing thing
 	return thing, nil
 }
 
+func (svc *mainfluxThings) CreateThings(_ context.Context, owner string, ths []things.Thing) ([]things.Thing, error) {
+	svc.mu.Lock()
+	defer svc.mu.Unlock()
+
+	for idx := range ths {
```
manuio: I would prefer `for i := range things {`

manuio (changes requested):

bootstrap/mocks/things.go
```diff
@@ -178,6 +198,25 @@ func (svc *mainfluxThings) CreateChannel(context.Context, string, things.Channel
 	panic("not implemented")
 }
 
+func (svc *mainfluxThings) CreateChannels(_ context.Context, owner string, channels []things.Channel) ([]things.Channel, error) {
+	svc.mu.Lock()
+	defer svc.mu.Unlock()
+
+	for idx := range channels {
```
manuio: Same here, I would use `i` instead of `idx` to stay consistent

manuio (changes requested):

Gopkg.lock
```diff
@@ -204,6 +204,14 @@
   revision = "fd0aebc74e908868b09ac140fb5a53cb363884c1"
   version = "1.2"
 
+[[projects]]
+  branch = "master"
+  digest = "1:77e13fb4b86c85c80be600735851c209420438ed8d463f1d0d366d9a096ced83"
+  name = "github.com/gocarina/gocsv"
```
anovakovic01: If you're not using `gocsv` anymore, remove it from `lock` file.

things/api/things/http/responses.go
```diff
@@ -52,7 +55,7 @@ func (res thingRes) Code() int {
 func (res thingRes) Headers() map[string]string {
 	if res.created {
 		return map[string]string{
-			"Location": fmt.Sprintf("/things/%s", res.id),
+			"Location": fmt.Sprintf("/things/%s", res.ID),
```
anovakovic01, line 58: There is no point in `Location` header if you're going to return `id` as part of JSON response. Think about splitting responses for creating a single thing, and multiple things.

things/api/things/http/responses.go
```diff
@@ -116,7 +142,7 @@ func (res channelRes) Code() int {
 func (res channelRes) Headers() map[string]string {
 	if res.created {
 		return map[string]string{
-			"Location": fmt.Sprintf("/channels/%s", res.id),
+			"Location": fmt.Sprintf("/channels/%s", res.ID),
```
anovakovic01, line 145: I recommend doing the same thing as for things.

things/postgres/database.go
```diff
@@ -49,6 +50,16 @@ func (dm database) GetContext(ctx context.Context, dest interface{}, query strin
 	return dm.db.GetContext(ctx, dest, query, args...)
 }
 
+func (dm database) BeginTxx(ctx context.Context, opts *sql.TxOptions) (*sqlx.Tx, error) {
+	span := opentracing.SpanFromContext(ctx)
+	if span != nil {
+		span.SetTag("span.kind", "client")
+		span.SetTag("peer.service", "postgres")
+		span.SetTag("db.type", "sql")
+	}
```
anovakovic01, line 59: Do you need to add tags here? They will be added as part of `NamedExecContext`?

anovakovic01 (changes requested):

cli/provision.go
```diff
@@ -240,3 +178,97 @@ func NewProvisionCmd() *cobra.Command {
...
+
+	file, err := os.OpenFile(path, os.O_RDONLY, os.ModePerm)
+	if err != nil {
+		return []mfxsdk.Thing{}, err
+	}
+	defer file.Close()
+
+	things := []mfxsdk.Thing{}
+	switch filepath.Ext(path) {
+	case ".csv":
```
dborovcanin: You can extract ".json" and ".csv" to const.

cli/provision.go
```diff
@@ -240,3 +178,97 @@ func NewProvisionCmd() *cobra.Command {
...
+			l, err := reader.Read()
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				return []mfxsdk.Thing{}, err
+			}
+
+			if len(l) < 1 {
+				return []mfxsdk.Thing{}, err
```
dborovcanin: You should return a custom error since `err` is probably nil here.

cli/provision.go
```diff
@@ -240,3 +178,97 @@ func NewProvisionCmd() *cobra.Command {
...
+			l, err := reader.Read()
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				return []mfxsdk.Channel{}, err
+			}
+
+			if len(l) < 1 {
+				return []mfxsdk.Channel{}, err
```
dborovcanin: Same here.

things/postgres/channels.go
```diff
@@ -51,6 +51,43 @@ func (cr channelRepository) Save(ctx context.Context, channel things.Channel) (s
 	return channel.ID, nil
 }
 
+func (cr channelRepository) BulkSave(ctx context.Context, channels []things.Channel) ([]things.Channel, error) {
+
```
dborovcanin: Please remove the leading blank line.

things/postgres/things.go
```diff
@@ -62,6 +62,46 @@ func (tr thingRepository) Save(ctx context.Context, thing things.Thing) (string,
 	return dbth.ID, nil
 }
 
+func (tr thingRepository) BulkSave(ctx context.Context, ths []things.Thing) ([]things.Thing, error) {
+
```
dborovcanin: Same here.

dborovcanin (changes requested):

Gopkg.lock
nwneisen: Thank you

things/api/things/http/responses.go
nwneisen: @anovakovic01 I left this in place to keep the current behavior for a single thing creation.  Right now it will return a `thingRes` with the `id` in the`Location` header without a JSON response.  Multiple thing creation will return a `thingsRes` with an empty `Location` header and a JSON response.

manuio (dismissed):

anovakovic01 (approved):
LGTM!

bootstrap/mocks/things.go
```diff
@@ -51,6 +51,25 @@ func (svc *mainfluxThings) AddThing(_ context.Context, owner string, thing thing
 	return thing, nil
 }
 
+func (svc *mainfluxThings) CreateThings(_ context.Context, owner string, ths []things.Thing) ([]things.Thing, error) {
```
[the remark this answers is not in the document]
drasko: @manuio `things` are import, then this would have to be renamed, and I do not like this.

I also feel that `ths` would be more Go idiomatic in this case. Let's keep it as it is.

drasko (approved):
LGTM. @nwneisen please update the branch and let's merge this one!

bootstrap/mocks/things.go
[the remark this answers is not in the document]
manuio: It's not consistent with `for i := range channels {`. I would use `ths` and `chs` if you want to be more Go idiomatic

dborovcanin (approved):

bootstrap/mocks/things.go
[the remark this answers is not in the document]
drasko: @manuio `ths` and `chs` would be better fit then IMHO, as this refers to an array, several things, not just one.

nmarcetic (approved):

nwneisen:
@drasko Updated!

drasko:
Merged! Great work there @nwneisen!

PR #894, opened 2019-10-21
MF-785 - Change CanAccess to CanAccessByKey

### LINKED ISSUE #785: Rename CanAccess endpoint to CanAccessByKey
Currently, `things` service has two gRPC endpoints that are very similar. The first one is called `CanAccess` while the second one is called `CanAccessByID`. The first one should be renamed to `CanAccessByKey` in order to be consistent.

nwneisen:
### What does this do?
Change `CanAccess` code to use `CanAccessByKey`

Resolves #785 

Endpoint `/channels/<channel id>/access` was changed to `/channels/<channel id>/access-by-key`.  This could cause issues if it was being access outside of the SDK.  It didn't seem to be exposed.

No new tests were created.  Current tests were verified to still be passing.

No, new functionality was not added and this endpoint is not documented.


drasko:
@nwneisen great initiative.

However this change will probably impact Aedes and VerneMQ MQTT adapters, as they depend on protobuf (which was changed).

Can you please verify if this is the case, and do the necessary changes over there as well (at least for Aedes, and I can do Erlang stuff for VerneMQ later).

drasko:
@dusanb94 these seem to be commited.

Also I double-checked that Verne does not use `CanAccessByKey()` (it uses only `CanAccessByID()`, which was the reason for introducing it and making these distinctions).

However, I think that Aedes [calls](https://github.com/mainflux/mainflux/blob/master/mqtt/aedes/mqtt.js#L229) need to be adjusted.

drasko:
On double-checking PR Aedes code seems to be handled as well.

I have good impression about this PR ;).

nmarcetic (dismissed):

drasko (dismissed):
LGTM

drasko:
@nwneisen I tried updating your branch via GitHUb UI button (as you have maintainer approvals, and we would like to merge this PR). However, CI is falling - probably because wrong protoc version.

Can you try updating the branch manually to see if we can fix this issue?

Our CI script seems to use [this version](https://github.com/mainflux/mainflux/blob/master/scripts/ci.sh#L32) of `protoc`. We should either bump it, remove this check, or you must for now align with what CI does.

@dusanb94 can we take a look at bumping this `protoc` version, looks like [3.10](https://github.com/protocolbuffers/protobuf/releases) is actual one.

nwneisen:
@drasko I checked my protoc version and I was indeed on 3.9.1.  It looks like the build is now successful and I just needed to push with a signed commit.

drasko (approved):
LGTM

manuio (approved):
LGTM

drasko:
Merged! Thanks @nwneisen!

PR #907, opened 2019-10-24
NOISSUE - Make event sourcing optional

anovakovic01:
### What does this do?
Add the env var flag for event sourcing.

There is no issue for this PR.

There are no such changes. There are only additional env vars.

No, I've tested it manually.

Yes, I've updated the existing documentation.


manuio (approved):
LGTM

.env
```diff
@@ -61,6 +61,7 @@ MF_THINGS_DB_USER=mainflux
 MF_THINGS_DB_PASS=mainflux
 MF_THINGS_DB=things
 MF_THINGS_SECRET=secret
+MF_THINGS_ES_ACTIVE=false
```
drasko, line 64: Do we need an extra ENV var?

Can't we say: if you did not provide value for ES server URL and/or if it is an empty string - just consider that ES is not active.

drasko (changes requested):

.env
drasko: After chat with @anovakovic01 it was concluded that empty strings can be confusing and all other defaults changes in that case (for consistency).

So we stick with this approach.

drasko (approved):
LGTM

PR #915, opened 2019-10-28
MF-780 - Use Normalizer as a lib

### LINKED ISSUE #780: Consider message normalization design changes
**FEATURE REQUEST**

No

Currently, there is a Normalizer service which is responsible for message normalization senlML -> Mainflux message. This service is [subscribed to all channels](https://github.com/mainflux/mainflux/blob/master/normalizer/nats/pubsub.go#L22) on NATS, taking senML as input and publishing Mainflux message as output to `out.<content_type>`. The problem here is double load, we are sending same messages (there are not really same, we did serialization) twice and this has a big impact on broker load. 
Consider moving message normalization to some other place or just avoid double traffic.   
 
Must have.

dborovcanin:

### What does this do?
This pull request removes Normalizer service in favor of using normalizer as a lib on the message consumer side.

Resolves #780.

Message consumers (writers) are modified to listen to all the message topics and _normalize and persist_ messages. 

Yes.

Yes.


chombium:
@dusanb94 Don't forget to add documentation with description of the library and it's api. This would be helpful for the others who will write custom components/writers in some other programming languages ;)

drasko (commented):
I scanned through the PR. Looks like we are really on good track here - I like the simplification!

topics.go
```diff
@@ -7,3 +7,4 @@ package mainflux
 
 // OutputSenML represents subject SenML messages will be published to.
 const OutputSenML = "out.senml"
+const OutputChannels = "channel.>"
```
anovakovic01: I don't think that you can publish using the wild card as part of the topic.

writers/cassandra/messages.go
```diff
@@ -20,37 +20,44 @@ func New(session *gocql.Session) writers.MessageRepository {
 	return &cassandraRepository{session}
 }
 
-func (cr *cassandraRepository) Save(msg mainflux.Message) error {
+func (cr *cassandraRepository) Save(messages []mainflux.Message) error {
```
anovakovic01: Think about updating the interface to have a variadic function.

internal/normalizer/normalizer.go
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) Mainflux
```
anovakovic01: I think that the normalizer shouldn't be in the `internal` package. AFAIK, everything that's in the `internal` cannot be used from the outside and as far as I understood, we want this lib to be usable from the outside too. At least think about creating an interface for the normalizer. Maybe this interface should have:
```
Normalize(msg mainflux.RawMessage) ([]interface{}, error)
```

You'll later have to cast this `interface{}` but that's part of the language (if it had generics maybe we could have done something about it). This way you'll be able to create an interface for any other message type.

anovakovic01 (dismissed):

readers/api/logging.go
```diff
@@ -28,9 +28,14 @@ func LoggingMiddleware(svc readers.MessageRepository, logger logger.Logger) read
 	}
 }
 
-func (lm *loggingMiddleware) ReadAll(chanID string, offset, limit uint64, query map[string]string) (readers.MessagesPage, error) {
+func (lm *loggingMiddleware) ReadAll(chanID string, offset, limit uint64, query map[string]string) (page readers.MessagesPage, err error) {
 	defer func(begin time.Time) {
-		lm.logger.Info(fmt.Sprintf(`Method read_all for offset %d and limit %d took %s to complete without errors.`, offset, limit, time.Since(begin)))
+		message := fmt.Sprintf("Method read_all for offset %d and limit %d took %s to complete", offset, limit, time.Since(begin))
```
manuio: You could log the Channel ID here

docker/addons/normalizer/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+
```
drasko, line 3: Add Mainflux header insted of this blank line at the top.

drasko (changes requested):

Makefile
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 BUILD_DIR = build
-SERVICES = users things http normalizer ws coap lora influxdb-writer influxdb-reader mongodb-writer mongodb-reader cassandra-writer cassandra-reader postgres-writer postgres-reader cli bootstrap opcua
+SERVICES = users things http ws coap lora normalizer influxdb-writer influxdb-reader mongodb-writer mongodb-reader cassandra-writer cassandra-reader postgres-writer postgres-reader cli bootstrap opcua
```
nmarcetic, line 5: Service will be removed, right ?

dborovcanin: Yes, it will be removed as a part of #919 and we'll switch to importing Normalizer as a package on the consumer side and completely remove service that publishes normalized messages back to the internal broker.

topics.go
```diff
@@ -7,3 +7,6 @@ package mainflux
 
 // OutputSenML represents subject SenML messages will be published to.
 const OutputSenML = "out.senml"
+
+// InputChannels represents subject all messages will be published to.
```
nmarcetic, line 12: I wouldn't be mean on words here, its very important to explain clear and simple. I would explain what is input channel, how input message looks like (reference), what is output channel and what is payload on output, etc..

nmarcetic (dismissed):

nmarcetic (approved):

drasko (approved):
LGTM

PR #924, opened 2019-10-30
MF-919 - Mainflux message updates

### LINKED ISSUE #919: Consider Mainflux message updates
**FEATURE REQUEST**
Is there an open issue addressing this request? If it does, please add a "+1" reaction to the
existing issue, otherwise, proceed to step 2.
No.

Currently, the Mainflux uses two types of messages: RawMessage and Message. RawMessages are exchanged internally over NATS and they represent the main data carrier, while Messages represent standardized messages. Working on #780 and extracting Normalizer to the separate package made me think about Message. Message type represents nothing else but a SenML record that can be consumed on the message consumer side. The decision to move message normalization to the separate package that will be utilized by the consumer (see #915) results in Normalizer service to become optional. This also opens an opportunity to make the normalization process more generic and less dependent on SenML specification. We can expose Normalizer as an interface and provide basic SenML implementation, but also leave the space for implementing different standards or using a custom message format passed as RawMessage Payload. This also implies removing Messages from the root level and keep them in SenML normalizer. The proposed interface declaration would be:

```go
type Normalizer interface {
    Normalize(mainflux.RawMessage) (interface{}, error)
}
```
and package structure would be similar to readers and writers:
- normalizer
        - doc.go
        - normalizer.go // Normalizer interface
    - senml
         - normalizer.go // SenML normalizer impl
         - message.go // SenML message struct
    - other_specification
        - normalizer.go // Other_specification normalizer impl
        - message.go // Other_specification message
    - custom
        - normalizer.go // Custom normalizer impl
        - message.go // Custom message

We could also rename `RawMessage` to `Message`, which sounds more convenient to me.

Pros:
1. Simplification and unification of Message
2. Generic Normalization
3. Decoupling from SenML

Cons:
1. If you want to serialize messages to Protobuf from Normalizer, you'll need to write your `.proto` files and compile them, which some developers might find complex.

@mainflux/contributors What are your thoughts on this?

This is should-have.

dborovcanin:
### What does this do?
This pull request removes RawMessage, replacing it with Message. As a consequence of this, different message types are supported using the message `Transformer`. The concept is similar to what used to be Normalizer service but more generic and cleaner. For more information, please read the corresponding issue (#919).

Resolves #919.

The code is reorganized in such a way that RawMessage is renamed to Message. What used to be Message is moved to `trasnformer/senml` package because it represents SenML package.

Yes.

Yes.


Makefile
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 BUILD_DIR = build
-SERVICES = users things http ws coap lora normalizer influxdb-writer influxdb-reader mongodb-writer mongodb-reader cassandra-writer cassandra-reader postgres-writer postgres-reader cli bootstrap opcua
+SERVICES = users things http ws coap lora influxdb-writer influxdb-reader mongodb-writer mongodb-reader cassandra-writer cassandra-reader postgres-writer postgres-reader cli bootstrap opcua
```
drasko: If all the addons are added here, probably it would be good to add `transformer` here as well, so that CI can check it and build the docker (for DoeckerHub usage).

drasko (changes requested):
This looks good to me, misses only VerneMQ changes.

Please add them and I will approve.

drasko (changes requested):
I would be good to rename also [RawMessage record](https://github.com/mainflux/mainflux/blob/818ddafdde283ae09b06e2328e867ee7ab570dd0/mqtt/verne/src/mfx_auth.erl#L118) into something like `Message` or similar.

mqtt/verne/src/mfx_auth.erl
```diff
@@ -123,7 +123,7 @@ auth_on_publish(UserName, {_MountPoint, _ClientId} = SubscriberId, QoS, Topic, P
                 contentType => ContentType,
                 payload => Payload
             },
-            mfx_nats:publish(NatsSubject, message_pb:encode_msg(RawMessage, raw_message)),
+            mfx_nats:publish(NatsSubject, message_pb:encode_msg(RawMessage, message)),
```
manuio: We want to keep `RawMessage` here?

manuio (changes requested):

Makefile
dborovcanin: > If all the addons are added here, probably it would be good to add `transformer` here as well, so that CI can check it and build the docker (for DoeckerHub usage).

Normalizer is removed as a service and we're now using Transformers as packages imported by Message consumers to transform messages. Since there is no standalone Transformer service, we don't have it as an add-on, either.

cmd/cassandra-writer/main.go
```diff
@@ -75,7 +75,7 @@ func main() {
 	defer session.Close()
 
 	repo := newService(session, logger)
-	norm := normalizer.New()
+	norm := senml.New()
```
manuio: This `norm` doesn't make sense anymore

manuio (changes requested):

anovakovic01 (dismissed):
LGTM!

manuio (dismissed):
LGTM

nmarcetic (dismissed):

anovakovic01 (dismissed):
LGTM!

nmarcetic (approved):

drasko (approved):
LGTM

anovakovic01 (approved):
LGTM!

PR #927, opened 2019-10-31
MF-906 - Change single creation endpoints to use bulk service calls

### LINKED ISSUE #906: Change single creation endpoints to use bulk service calls
**FEATURE REQUEST**

As discussed [here](https://github.com/mainflux/mainflux/pull/889#issuecomment-544913349), the single `createThingsEndpoint`/`createChannelsEndpoint` endpoints should be changed to use bulk service functions.  Functions used to create a single thing/channel after the service call should be removed.

should-have

nwneisen:
### What does this do?
Changes the single things and channels creation endpoints to use bulk service calls.  All code for single creation is removed and tests updated to use bulk creation calls.

Resolves #906 

Single creation endpoints behave the same from a user's perspective.

Tests were updated to use bulk creation methods.

NA


things/postgres/things_test.go
```diff
@@ -197,8 +127,8 @@ func TestThingUpdate(t *testing.T) {
 		Key:   thkey,
 	}
 
-	id, _ := thingRepo.Save(context.Background(), thing)
-	thing.ID = id
+	sths, _ := thingRepo.BulkSave(context.Background(), []things.Thing{thing})
```
drasko: If every `Save()` i `BulkSave()` now, i.e. if there I would prefer that we name this `Save()` and know that Save() actually saves an array of things.

things/postgres/channels_test.go
```diff
@@ -659,15 +621,17 @@ func TestHasThing(t *testing.T) {
...
 	}
-	thingID, _ := thingRepo.Save(context.Background(), thing)
+	sths, _ := thingRepo.BulkSave(context.Background(), []things.Thing{thing})
+	thingID := sths[0].ID
 
 	chanRepo := postgres.NewChannelRepository(dbMiddleware)
 	chid, err := uuid.New().ID()
 	require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
-	chanID, _ := chanRepo.Save(context.Background(), things.Channel{
+	schs, _ := chanRepo.BulkSave(context.Background(), []things.Channel{{
```
drasko: Same here.

drasko (changes requested):

things/postgres/things_test.go
nwneisen: That sounds good to me

nmarcetic (commented):
@nwneisen I would prefer to keep method naming simple and clean by removing `Bulk` word :) So I would prefer Create() vs BulkCreate() etc... even if method accept a slice of things/channels 
Word bulk adds complexity and confusion IMHO, keeps things simple as they are and documenting interface (good comments) is perfect. 
I am approving PR, good job really! This is just my personal feeling about naming, which you can accept or not :) It's your call.

nmarcetic (dismissed):

drasko:
@nwneisen you have a message from the SemaphoreCI:
```coverage: 95.3% of statements
ok  	github.com/mainflux/mainflux/things/api/things/http	3.091s	coverage: 95.3% of statements
?   	github.com/mainflux/mainflux/things/mocks	[no test files]
# github.com/mainflux/mainflux/things/postgres_test [github.com/mainflux/mainflux/things/postgres.test]
things/postgres/channels_test.go:624:22: thingRepo.SaveSave undefined (type things.ThingRepository has no field or method SaveSave)
FAIL	github.com/mainflux/mainflux/things/postgres [build failed]
FAIL```

;)

File: things/service.go [hunk 3]
```diff
@@ -32,9 +32,6 @@ var (
 // Service specifies an API that must be fullfiled by the domain service
 // implementation, and all of its decorators (e.g. logging & metrics).
 type Service interface {
-	// AddThing adds new thing to the user identified by the provided key.
-	AddThing(context.Context, string, Thing) (Thing, error)
-
 	// CreateThings adds a list of things to the user identified by the provided key.
 	CreateThings(context.Context, string, []Thing) ([]Thing, error)
```
anovakovic01: I recommend that you use a variadic function here (with `...`).

nmarcetic: Yes agree, looks much nicer.

nwneisen: Good idea

nwneisen:
@nmarcetic I agree completely.  When I do a search for remaining "bulk" word usages, I only see it still used with the single vs bulk endpoints.  Let me know if there is somewhere I am missing.

I will say that I am still in favor of depreciating this single endpoint as well.  If only one thing/channel is needed, then a user would simply send an array of size one.  This would make it so the user only needs to understand one method to create a thing/channel.

drasko:
> I will say that I am still in favor of depreciating this single endpoint as well. If only one thing/channel is needed, then a user would simply send an array of size one.

@nwneisen exactly, this would be the best approach.

nwneisen:
@manuio 

@drasko Making sure everyone is on board as this was my [previous suggestion](https://github.com/mainflux/mainflux/pull/889#issuecomment-543821259).

>It's also worth noting that with the exception of the blank payload, the bulk provisioning could be used to replace creating a single thing. This would be a breaking change but might be worth considering in the future to reduce maintenance. If sending a blank payload to create a thing is desired, then this could also be added to the bulk provisioning endpoint.

The single endpoint will create a thing/channel when is passed an empty payload.  It also returns the created thing's/channel's ID in the location header.

drasko:
> The single endpoint will create a thing/channel when is passed an empty payload. It also returns the created thing's/channel's ID in the location header.

I am interested what would the others say, but personally this looks OK for me - passing an empty call means "create one thing without metadata", while passing a call with array of length 1 in a parameter might mean "create one thing with this meatadata".

I am however thinking that we might be interested in creating for example 100 things without metadata, and I do not know how we can do it with this API...

@nmarcetic @anovakovic01 @dusanb94 ^

nwneisen:
@drasko Using bulk create, an empty JSON payload (`""` or `"[]"`) will not create the thing/channel.  A payload with an empty JSON object (`"[{}]"`) will.  To create 100 things without metadata, a JSON array with 100 empty objects could be used.

drasko:
@nwneisen yep, I see.

It looks OK to me, I think there is no particular need to complicate the code and add other options - especially that most of the time interaction with the REST API will be via CLI or SDK (meaning that CLI and SDK must be changed as well in this PR, if the API changes).

Never the less, I'd like to hear the opinion of other maintainers.

things/things.go
```diff
@@ -28,13 +28,10 @@ type ThingsPage struct {
...
-	// Save persists the thing. Successful operation is indicated by non-nil
+	// Save persists multiple things. Things are saved using a transaction. If one thing
+	// fails then none will be saved. Successful operation is indicated by non-nil
 	// error response.
-	Save(context.Context, Thing) (string, error)
-
-	// BulkSave persistest multiple things. Successful operation is indicated by non-nil
-	// error response.
-	BulkSave(context.Context, []Thing) ([]Thing, error)
+	Save(context.Context, []Thing) ([]Thing, error)
```
anovakovic01: Why are you not using variadic functions here too? (Same goes for channels)

anovakovic01 (changes requested):

anovakovic01:
@drasko I don't think that we need an endpoint for the creation of a single thing.

nmarcetic:
> @drasko I don't think that we need an endpoint for the creation of a single thing.

I agree with @anovakovic01 
@nwneisen You can merge those two endpoints and keep just on which will accept a slice of things/channels. This will keep things simpler but will require changes in SDK, UI also, etc... its breaking change. Maybe is better idea to keep it like this, for now, 2 endpoints like you did, then in the next release merge it and make an announcement.

What do you think @nwneisen ?

dborovcanin:
I agree with @nmarcetic. @nwneisen Please resolve remaining remark (variadic Save), so that we can approve and merge.

things/things.go
nwneisen: I was thinking at this point the things/channels would all be in a slice already and it wouldn't be needed.

nwneisen:
@nmarcetic I was thinking leave the single endpoint for now but look at adding a depreciation warning and/or announcement for when it would be completely removed.

nmarcetic:
> @nmarcetic I was thinking leave the single endpoint for now but look at adding a deprecation warning and/or announcement for when it would be completely removed.

@nwneisen `c I was thinking to leave the single endpoint for now ` Now is not a single endpoint, there is 2 endpoints POST on `/things` and `things/bulk` 
Let's leave it for now and as you said, make `a deprecation warning and/or announcement` and remove it in the next milestone.

nmarcetic (dismissed):

nwneisen:
@nmarcetic 
> I was thinking leave the single endpoint for now

I was thinking leave the single creation endpoint (/things) for now*

drasko:
@nmarcetic @nwneisen can we open an issue for this merging of endpoints not to forget?

drasko (dismissed):
LGTM

nwneisen:
@drasko @nmarcetic I will add the depreciation warning and place a warning in the docs before this PR gets merged

anovakovic01 (approved):
LGTM!

drasko (approved):
LGTM

drasko:
Thanks @nwneisen, again a great contribution!

I like PRs that remove code :).

PR #929, opened 2019-11-03
NOISSUE - Remove ARM multi-arch images

drasko:

### What does this do?

Removes multi-arch image push to DockerHub.
Only `amd64` images are produced and pushed.

It should be noted that manifest currently was hard-coded to ARMv7, and besides that it is suitable for RPi, it does not have too much benefit, nor it is generic. It should be left to developers to produce Docker images for their particular architectures.

Also removes CI build of ARM images.

Local build of ARM images can still be executed by adding just ENV vars to Makefile prefix:

```
GOARCH=arm GOARM=7 make docker_<service>
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,8 +26,8 @@ define make_docker
 
 	# If ARCH is `amd64` then retag image for convinience of devlopment
 	if [ $(GOARCH) = amd64 ]; then \
-            docker tag mainflux/$(svc)-$(2) mainflux/$(svc); \
-    fi
+		docker tag mainflux/$(svc)-$(2) mainflux/$(svc); \
+	fi
 endef
 
 define make_docker_dev
--- a/Makefile
+++ b/Makefile
@@ -21,13 +21,8 @@ define make_docker
 		--build-arg SVC=$(svc) \
 		--build-arg GOARCH=$(GOARCH) \
 		--build-arg GOARM=$(GOARM) \
-		--tag=mainflux/$(svc)-$(2) \
+		--tag=mainflux/$(svc) \
 		-f docker/Dockerfile .
-
-	# If ARCH is `amd64` then retag image for convinience of devlopment
-	if [ $(GOARCH) = amd64 ]; then \
-		docker tag mainflux/$(svc)-$(2) mainflux/$(svc); \
-	fi
 endef
 
 define make_docker_dev
@@ -91,11 +86,7 @@ docker_ui:
 
 docker_mqtt:
 	# MQTT Docker build must be done from root dir because it copies .proto files
-ifeq ($(GOARCH), arm)
-	docker build --tag=mainflux/mqtt-arm -f mqtt/aedes/Dockerfile.arm .
-else
-	docker build --tag=mainflux/mqtt-amd64 -f mqtt/aedes/Dockerfile .
-endif
+	docker build --tag=mainflux/mqtt -f mqtt/aedes/Dockerfile.arm .
 
 docker_mqtt_verne:
 	docker build --tag=mainflux/mqtt-verne -f mqtt/verne/Dockerfile .
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 BUILD_DIR = build
-SERVICES = users things http ws coap lora normalizer influxdb-writer influxdb-reader mongodb-writer mongodb-reader cassandra-writer cassandra-reader postgres-writer postgres-reader cli bootstrap opcua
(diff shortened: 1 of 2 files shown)


blokovi (dismissed):

Makefile
```diff
@@ -13,26 +14,35 @@ define compile_service
...
 		--no-cache \
-		--build-arg SVC=$(subst docker_,,$(1)) \
+		--build-arg SVC=$(svc) \
 		--build-arg GOARCH=$(GOARCH) \
 		--build-arg GOARM=$(GOARM) \
-		--tag=mainflux/$(subst docker_,,$(1))-$(2) \
+		--tag=mainflux/$(svc)-$(2) \
 		-f docker/Dockerfile .
+
+	# If ARCH is `amd64` then retag image for convinience of devlopment
```
blokovi: two typos: convenience of development

Makefile
```diff
@@ -13,26 +14,35 @@ define compile_service
...
+	$(eval svc=$(subst docker_,,$(1)))
+
 	docker build \
 		--no-cache \
-		--build-arg SVC=$(subst docker_,,$(1)) \
+		--build-arg SVC=$(svc) \
 		--build-arg GOARCH=$(GOARCH) \
 		--build-arg GOARM=$(GOARM) \
-		--tag=mainflux/$(subst docker_,,$(1))-$(2) \
+		--tag=mainflux/$(svc)-$(2) \
```
blokovi: I would remove this arch suffix and leave just `--tag=mainflux/$(svc) \`

Makefile
```diff
@@ -13,26 +14,35 @@ define compile_service
...
 		-f docker/Dockerfile .
+
+	# If ARCH is `amd64` then retag image for convinience of devlopment
+	if [ $(GOARCH) = amd64 ]; then \
+		docker tag mainflux/$(svc)-$(2) mainflux/$(svc); \
+	fi
```
blokovi: If you remove arch suffix as I propose above, then no need for retagging

Makefile
drasko: We can do this, but then if you start generation of ARM images on your PC to copy them on your gateway, your current x86 images will be overwritten.

MAybe this can be useful, as you do not have to retag images once you copy them to the gateway, but them compiling for different architectures can be somehow more difficult...

blokovi: After `make dockers`  you will have double all images with this approach (one with and one without arch suffix). This could be confusing for most users.

drasko: Size is not doubled, just the tags. Same docker layers are used.

Let me think if we should remove this, but imagine the situation in which we have to build for several architectures (ARM. MIPS, ...). We would have to backup all images prior to build, as they would be overwritten...

blokovi (dismissed):

Makefile
```diff
@@ -80,11 +86,7 @@ docker_ui:
 
 docker_mqtt:
 	# MQTT Docker build must be done from root dir because it copies .proto files
-ifeq ($(GOARCH), arm)
-	docker build --tag=mainflux/mqtt-arm -f mqtt/aedes/Dockerfile.arm .
-else
-	docker build --tag=mainflux/mqtt-amd64 -f mqtt/aedes/Dockerfile .
-endif
+	docker build --tag=mainflux/mqtt -f mqtt/aedes/Dockerfile.arm .
```
dborovcanin, line 90: Should we use `mqtt/aedes/Dockerfile` instead of `mqtt/aedes/Dockerfile.arm`?

dborovcanin (changes requested):

Makefile
drasko: Damn, good catch!

dborovcanin (approved):

Makefile
```diff
@@ -101,46 +107,28 @@ mqtt:
 
 define docker_push
 	for svc in $(SERVICES); do \
-		docker push mainflux/$$svc-$(1):$(2); \
+		docker push mainflux/$$svc:$(1); \
 	done
-	docker push mainflux/ui-$(1):$(2)
-	docker push mainflux/mqtt-$(1):$(2)
+	docker push mainflux/ui:$(1)
+	docker push mainflux/mqtt:$(1)
```
anovakovic01, line 113: Hopefully, no one will push arm images instead of amd64. :)

anovakovic01 (approved):
LGTM!

manuio (approved):
LGTM

Makefile
drasko: This is why I had in the beginning image retagging, but well... Not really needed, maintainers just neet to take care then they are doing the release (and CI is set to push only x86 images).

PR #943, opened 2019-11-08
MF-942 - Simplify email template mapping

### LINKED ISSUE #942: E-mail template usage is confusing
**BUG REPORT**

Use e-mail template, change path, make is usable on Windows Server

E-mail template usage should be clear to se from the code (seld documented)

Code is confusing and not self-documenting:
- `docker-compose` uses: https://github.com/mainflux/mainflux/blob/841b9437184015415181c23f689f2cc38e33b8ef/docker/docker-compose.yml#L70 and not the one from the [configs](https://github.com/mainflux/mainflux/tree/master/configs) - even more, these two are different.
- What is the purpose than of one in the configs?
- ENV variable in `.env` is set like [this](https://github.com/mainflux/mainflux/blob/841b9437184015415181c23f689f2cc38e33b8ef/.env#L4- In the [default value](https://github.com/mainflux/mainflux/blob/841b9437184015415181c23f689f2cc38e33b8ef/cmd/users/main.go#L68) it is again set differently, and even more confusing with `../<something>`. Having in mind that in containers binaries are put into root, this `..` does not make sense.

Can we use just filename here and let user map this file wherever he wants?

drasko:

Resolves #942

### DIFF
--- a/.env
+++ b/.env
@@ -44,7 +44,7 @@ MF_EMAIL_USERNAME=18bf7f70705139
 MF_EMAIL_PASSWORD=2b0d302e775b1e
 MF_EMAIL_FROM_ADDRESS=from@example.com
 MF_EMAIL_FROM_NAME=Example
-MF_EMAIL_TEMPLATE=/resetPasswEmail.tmpl
+MF_EMAIL_TEMPLATE=resetPasswEmail.tmpl
 
 ### Token utility
 MF_TOKEN_SECRET=some_random_bytes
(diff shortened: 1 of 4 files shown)


chombium:
@drasko, @mteodor it looks to me that this PR overlaps with #944. Could we maybe close this as the changes introduced in this one are already in #944?

dborovcanin (approved):

drasko:
Duplicate of #944

PR #944, opened 2019-11-09
MF-942 - Fix email template logic 

### LINKED ISSUE #942: E-mail template usage is confusing

mteodor:
Resolves #942

Yes


blokovi (dismissed):

users/README.md
```diff
@@ -41,9 +41,9 @@ default values.
 | MF_EMAIL_PASSWORD         | Mail server password                                                    |                |
 | MF_EMAIL_FROM_ADDRESS     | Email "from" address                                                    |                |
 | MF_EMAIL_FROM_NAME        | Email "from" name                                                       |                |
-| MF_TOKEN_SECRET           | Token secret for signing token used for password reset                  |                |
+| MF_EMAIL_TEMPLATE         | Email template for sending emails with password reset link              | email.tmpl     |
+| MF_TOKEN_SECRET           | Password reset token signing secret                    |                |
```
drasko: Align this properly

users/README.md
```diff
@@ -105,9 +105,11 @@ make users
...
 ```
 
+if 'MF_EMAIL_TEMPLATE' doesn't point to any file service will function but password reset functionality will not work
```
drasko: Sentences start with capital letter.

Use backticks (``) rather than single qoutes.

drasko (changes requested):

manuio (approved):

anovakovic01 (approved):
LGTM!

nmarcetic (approved):

manuio (approved):
LGTM

drasko (approved):
LGTM

dborovcanin (approved):

PR #947, opened 2019-11-12
MF-916 - Fix Things and Channels counters

### LINKED ISSUE #916: Total Things or Channels are wrong if filtered by metadata
**FEATURE REQUEST**

Total Things or Channels with a specific metadata.
Must-have
**BUG REPORT**

Know the total
Total Things or Channels containing a specific name or metadata
Total Things or Channels owned by the user
Fetch Things or Channels with name or metadata query parameters in the HTTP request
In any

manuio:

Closes #916

### DIFF
--- a/things/postgres/channels.go
+++ b/things/postgres/channels.go
@@ -22,6 +22,12 @@ type channelRepository struct {
 	db Database
 }
 
+type dbConnection struct {
+	Channel string `db:"channel"`
+	Thing   string `db:"thing"`
+	Owner   string `db:"owner"`
+}
+
 // NewChannelRepository instantiates a PostgreSQL implementation of channel
 // repository.
 func NewChannelRepository(db Database) things.ChannelRepository {
@@ -149,35 +155,27 @@ func (cr channelRepository) RetrieveAll(ctx context.Context, owner string, offse
 		items = append(items, ch)
 	}
 
-	total := uint64(0)
-	cq := `SELECT COUNT(*) FROM channels WHERE owner = $1`
+	cq := `SELECT COUNT(*) FROM channels WHERE owner = :owner`
 	switch name {
 	case "":
 		switch metadata {
 		case nil:
 			cq = fmt.Sprintf("%s;", cq)
-			if err := cr.db.GetContext(ctx, &total, cq, owner); err != nil {
-				return things.ChannelsPage{}, err
-			}
 		default:
-			cq = fmt.Sprintf("%s %s", cq, "AND metadata @> $2;")
-			if err := cr.db.GetContext(ctx, &total, cq, owner, m); err != nil {
-				return things.ChannelsPage{}, err
-			}
+			cq = fmt.Sprintf("%s %s", cq, "AND metadata @> :metadata;")
 		}
 	default:
 		switch metadata {
 		case nil:
-			cq = fmt.Sprintf("%s %s", cq, "AND name LIKE $2;")
-			if err := cr.db.GetContext(ctx, &total, cq, owner, name); err != nil {
-				return things.ChannelsPage{}, err
-			}
+			cq = fmt.Sprintf("%s %s", cq, "AND name LIKE :name;")
 		default:
-			cq = fmt.Sprintf("%s %s", cq, "AND name LIKE $2 AND metadata @> $3;")
-			if err := cr.db.GetContext(ctx, &total, cq, owner, name, m); err != nil {
-				return things.ChannelsPage{}, err
-			}
+			cq = fmt.Sprintf("%s %s", cq, "AND name LIKE :name AND metadata @> :metadata;")
 		}
+	}
+
+	total, err := cr.total(ctx, cq, params)
+	if err != nil {
+		return things.ChannelsPage{}, err
 	}
 
 	page := things.ChannelsPage{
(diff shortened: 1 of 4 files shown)


things/postgres/channels.go
```diff
@@ -149,22 +149,34 @@ func (cr channelRepository) RetrieveAll(ctx context.Context, owner string, offse
...
-
 	total := uint64(0)
+	cq := `SELECT COUNT(*) FROM channels WHERE owner = $1`
 	switch name {
 	case "":
-		if err := cr.db.GetContext(ctx, &total, q, owner); err != nil {
-			return things.ChannelsPage{}, err
+		switch metadata {
+		case nil:
+			cq = fmt.Sprintf("%s %s", cq, ";")
```
anovakovic01: You can write it like this: `cq = fmt.Sprintf("%s;", cq)`

things/postgres/things.go
```diff
@@ -218,22 +218,34 @@ func (tr thingRepository) RetrieveAll(ctx context.Context, owner string, offset,
...
-
 	total := uint64(0)
+	cq := `SELECT COUNT(*) FROM things WHERE owner = $1`
 	switch name {
 	case "":
-		if err := tr.db.GetContext(ctx, &total, q, owner); err != nil {
-			return things.ThingsPage{}, err
+		switch metadata {
+		case nil:
+			cq = fmt.Sprintf("%s %s", cq, ";")
```
anovakovic01: Same here, you can write it as `cq = fmt.Sprintf("%s;", cq)`

anovakovic01 (changes requested):

blokovi (dismissed):

things/postgres/channels.go
```diff
@@ -149,22 +149,34 @@ func (cr channelRepository) RetrieveAll(ctx context.Context, owner string, offse
...
-
-	q = fmt.Sprintf(`SELECT COUNT(*) FROM channels WHERE owner = $1 %s;`, cq)
-
 	total := uint64(0)
+	cq := `SELECT COUNT(*) FROM channels WHERE owner = $1`
 	switch name {
 	case "":
-		if err := cr.db.GetContext(ctx, &total, q, owner); err != nil {
-			return things.ChannelsPage{}, err
+		switch metadata {
```
mteodor: this looks like lot of cases 
cannot this be done like here
https://github.com/mainflux/mainflux/blob/22f1e851cd334f30c33f314eff76f6af23088425/things/postgres/channels.go#L125 using getNameQuery and getMetadataQuery
is total different than number of rows we get from the query above?

manuio: @mteodor The first one returns you one page after to apply filters. The second one counts the total of Things with a certain filter in all pages.

mteodor: and this would not work
```cq := fmt.Sprintf(`SELECT COUNT(*) FROM things WHERE owner = $1 %s%s`, nq, mq)``` ?

Comment (manuio) [reply things/postgres/channels.go]: I don't see how here: https://godoc.org/github.com/jmoiron/sqlx#NamedQueryContext

File: things/postgres/channels.go [hunk 4]
```diff
@@ -438,8 +448,18 @@ func getMetadataQuery(m things.Metadata) ([]byte, string, error) {
 	return mb, mq, nil
 }
 
-type dbConnection struct {
-	Channel string `db:"channel"`
-	Thing   string `db:"thing"`
-	Owner   string `db:"owner"`
+func (cr channelRepository) total(ctx context.Context, query string, params map[string]interface{}) (uint64, error) {
```
anovakovic01: Can you create single helper function for this? (maybe this method can receive `db` as parameter as well)

things/postgres/channels.go
```diff
@@ -149,25 +155,29 @@ func (cr channelRepository) RetrieveAll(ctx context.Context, owner string, offse
...
 
-	cq := ""
-	if name != "" {
-		cq = `AND LOWER(name) LIKE $2`
-	}
-
-	q = fmt.Sprintf(`SELECT COUNT(*) FROM channels WHERE owner = $1 %s;`, cq)
-
-	total := uint64(0)
+	cq := `SELECT COUNT(*) FROM channels WHERE owner = :owner`
```
mteodor: I would do it similar like above since we alread have subqueries for metadata and name
```
cq := fmt.Sprintf(`SELECT COUNT(*) FROM channels WHERE owner = :owner %s%s`, nq, mq )
```
and no need for case at all

manuio: You're right! I would even say that we don't need to do this checks since this name and metadata queries are returned by  `getNameQuery` and `getMetadatQuery`

things/postgres/channels.go
```diff
@@ -418,7 +412,7 @@ func getNameQuery(name string) (string, string) {
 	nq := ""
 	if name != "" {
 		name = fmt.Sprintf(`%%%s%%`, name)
-		nq = ` AND LOWER(name) LIKE :name`
+		nq = ` AND name LIKE :name`
```
mteodor: actually we need LOWER here, sorry

mteodor (approved):
LGTM

anovakovic01 (approved):
LGTM!

Issue #898, opened 2019-10-21
Add bulk connect


nmarcetic:
**FEATURE REQUEST**

1. Is there an open issue addressing this request? If it does, please add a "+1" reaction to the
   existing issue, otherwise proceed to step 2.
No

2. Describe the feature you are requesting, as well as the possible use case(s) for it.
As described [here](https://github.com/mainflux/mainflux/issues/484#issuecomment-544544684)
  Now when we have bulk creation of things and channels, thanks to #889  it would be nice to have an API that can connect things to channels in bulk.

3. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).
Nice to have

nwneisen:
I'm going to start taking a look at this.  I think it will go quickly after doing the bulk create for things and channels.  I'd like to do this is multiple small PRs.  I think it will make for an easier and faster review process for everyone.

drasko:
@nwneisen fantastic. It is scheduled for next release, so it would be great if you could take it.

nmarcetic:
@nwneisen Totally agree we should keep smaller PRs in order to make reviews faster.

nwneisen:
I'm working on the endpoint for this and I'd like to get some opinions of what that should look like.  
@nmarcetic @drasko @manuio 

The current endpoint for a single connection is a PUT at `/channels/:chanId/things/:thingId`.  My initial thought was to create a POST at `/channels/:chanId` since a list of things will be included in the payload.  My issues with this are that I can include the `:chanId` in the payload, this path is already used for GET, and it isn't clear what this is doing.

I'm more in favor using a POST to `/connect` which includes the channel ID and a list of thing IDs in the payload.  `/disconnect` could be used later if bulk disconnect is also desired.

manuio:
@nwneisen I would use POST `/channels/:id/things` with a payload of Things ID. This endpoint is already used to list Things connected to a Channel.

`/connect` could be used to connect multiple channels to multiple things.

nwneisen:
@manuio I've thought about whether this change should be for connecting multiple channels to multiple things.  Having that endpoint would eliminate the need for multiple things for one channel.  It would follow the same reasoning for eliminating the single endpoints.

- Simplify the API by having only one way of doing something
- If the users only have one channel, then they only send one channel

PR #948, opened 2019-11-12
MF-898 - Add bulk connections endpoint

nwneisen:
Signed-off-by: nwneisen <nwneisen@gmail.com>

### What does this do?
Adds an endpoint to create bulk connections.

Relates to #898.

None

Yes

Yes

### DIFF
--- a/bootstrap/mocks/things.go
+++ b/bootstrap/mocks/things.go
@@ -70,7 +70,7 @@ func (svc *mainfluxThings) ViewThing(_ context.Context, owner, id string) (thing
 	return things.Thing{}, things.ErrNotFound
 }
 
-func (svc *mainfluxThings) Connect(_ context.Context, owner string, chIDs []string, thIDs []string) error {
+func (svc *mainfluxThings) Connect(_ context.Context, owner string, chIDs, thIDs []string) error {
 	svc.mu.Lock()
 	defer svc.mu.Unlock()
 
--- a/bootstrap/mocks/things.go
+++ b/bootstrap/mocks/things.go
@@ -70,7 +70,7 @@ func (svc *mainfluxThings) ViewThing(_ context.Context, owner, id string) (thing
 	return things.Thing{}, things.ErrNotFound
 }
 
-func (svc *mainfluxThings) Connect(_ context.Context, owner, chID string, thIDs ...string) error {
+func (svc *mainfluxThings) Connect(_ context.Context, owner string, chIDs []string, thIDs []string) error {
 	svc.mu.Lock()
 	defer svc.mu.Unlock()
 
@@ -78,11 +78,13 @@ func (svc *mainfluxThings) Connect(_ context.Context, owner, chID string, thIDs 
 	if err != nil {
 		return things.ErrUnauthorizedAccess
 	}
-	if svc.channels[chID].Owner != userID.Value {
-		return things.ErrUnauthorizedAccess
-	}
-	for _, thID := range thIDs {
-		svc.connections[chID] = append(svc.connections[chID], thID)
+	for _, chID := range chIDs {
+		if svc.channels[chID].Owner != userID.Value {
+			return things.ErrUnauthorizedAccess
+		}
+		for _, thID := range thIDs {
+			svc.connections[chID] = append(svc.connections[chID], thID)
+		}
 	}
 
 	return nil
--- a/docs/provisioning.md
+++ b/docs/provisioning.md
@@ -256,7 +256,7 @@ curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X PUT -H
 To connect multiple things to a channel, you can send the following request:
 
 ```bash
-curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X POST -H "Content-Type: application/json" -H "Authorization: <user_auth_token>" https://localhost/connect -d '{"chanID":"<channel_id>","thingIDs":["<thing_id>", "<thing_id>"]}'
+curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X POST -H "Content-Type: application/json" -H "Authorization: <user_auth_token>" https://localhost/connect -d '{"channel_ids":["<channel_id>", "<channel_id>"],"thingIDs":["<thing_id>", "<thing_id>"]}'
 ```
 
 You can observe which things are connected to specific channel:
--- a/docs/provisioning.md
+++ b/docs/provisioning.md
@@ -256,7 +256,7 @@ curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X PUT -H
 To connect multiple things to a channel, you can send the following request:
 
 ```bash
-curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X POST -H "Content-Type: application/json" -H "Authorization: <user_auth_token>" https://localhost/connect -d '{"channel_ids":["<channel_id>", "<channel_id>"],"thingIDs":["<thing_id>", "<thing_id>"]}'
+curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X POST -H "Content-Type: application/json" -H "Authorization: <user_auth_token>" https://localhost/connect -d '{"channel_ids":["<channel_id>", "<channel_id>"],"thing_ids":["<thing_id>", "<thing_id>"]}'
 ```
 
 You can observe which things are connected to specific channel:
--- a/docs/provisioning.md
+++ b/docs/provisioning.md
@@ -245,10 +245,18 @@ Only user, who is the owner of a channel and of the things, can connect the
 things to the channel (which is equivalent of giving permissions to these things
 to communicate over given communication group).
 
-To connect thing to the channel you should send following request:
+To connect a thing to the channel you should send following request:
 
-```
+> This endpoint will be depreciated in 0.11.0.  It will be replaced with the bulk endpoint found at /connect.
+
+```bash
 curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X PUT -H "Authorization: <user_auth_token>" https://localhost/channels/<channel_id>/things/<thing_id>
+```
+
+To connect multiple things to a channel, you can send the following request:
+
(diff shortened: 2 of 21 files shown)


things/api/things/http/requests.go
```diff
@@ -240,3 +240,27 @@ func (req connectionReq) validate() error {
 
 	return nil
 }
+
+type createConnectionsReq struct {
+	token    string
+	ChanID   string   `json:"chanID,omitempty"`
+	ThingIDs []string `json:"thingIDs,omitempty"`
```
manuio: I would use `things_id` instead of `thingIDs` here.

manuio (changes requested):

things/api/things/http/requests.go
manuio: I would even use
```
ChannelsID  []string  `json:"channels_id,omitempty"`
ThingsID []string `json:"things_id,omitempty"
```

nwneisen:
@manuio I started switching this code over for multiple channels and I realized that I can only have one variadic parameter.  I don't see another way besides making the channelIDs parameter a slice but I thought I'd see if you had any ideas before changing everything over.

manuio:
@nwneisen I'm not sure either... In any case the service endpoint can accept two slices of strings. But for the service method I'm not sure... Maybe a for loop here `svc.Connect(ctx, cr.token, cr.ChanID, cr.ThingIDs...); `? 
@drasko @nmarcetic what do you think?

nwneisen:
@manuio I was thinking a for loop would be the way to go without getting caught up in too much optimization.  The downside is a for loop on the service method would create multiple transactions; one for each channel.  This behavior might not be terrible but I think if one transaction fails then it should be recorded to be returned to the user and the rest allowed to continue.

anovakovic01:
I vote for a single transaction and two slices.

bootstrap/mocks/things.go
```diff
@@ -70,19 +70,21 @@ func (svc *mainfluxThings) ViewThing(_ context.Context, owner, id string) (thing
 	return things.Thing{}, things.ErrNotFound
 }
 
-func (svc *mainfluxThings) Connect(_ context.Context, owner, chID string, thIDs ...string) error {
+func (svc *mainfluxThings) Connect(_ context.Context, owner string, chIDs []string, thIDs []string) error {
```
manuio: Can you please use `chsID` and `thsID`

docs/provisioning.md
```diff
@@ -256,7 +256,7 @@ curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X PUT -H
 To connect multiple things to a channel, you can send the following request:
 
 ```bash
-curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X POST -H "Content-Type: application/json" -H "Authorization: <user_auth_token>" https://localhost/connect -d '{"chanID":"<channel_id>","thingIDs":["<thing_id>", "<thing_id>"]}'
+curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X POST -H "Content-Type: application/json" -H "Authorization: <user_auth_token>" https://localhost/connect -d '{"channel_ids":["<channel_id>", "<channel_id>"],"thingIDs":["<thing_id>", "<thing_id>"]}'
```
manuio: Use `channels_id` instead of `channel_ids` please

things/api/logging.go
```diff
@@ -201,17 +201,17 @@ func (lm *loggingMiddleware) RemoveChannel(ctx context.Context, token, id string
 	return lm.svc.RemoveChannel(ctx, token, id)
 }
 
-func (lm *loggingMiddleware) Connect(ctx context.Context, token, chID string, thIDs ...string) (err error) {
+func (lm *loggingMiddleware) Connect(ctx context.Context, token string, chIDs []string, thIDs []string) (err error) {
```
manuio: Same here, `chsID` and `thsID`

things/api/metrics.go
```diff
@@ -148,13 +148,13 @@ func (ms *metricsMiddleware) RemoveChannel(ctx context.Context, token, id string
 	return ms.svc.RemoveChannel(ctx, token, id)
 }
 
-func (ms *metricsMiddleware) Connect(ctx context.Context, token, chanID string, thIDs ...string) error {
+func (ms *metricsMiddleware) Connect(ctx context.Context, token string, chIDs []string, thIDs []string) error {
```
manuio: Same here, `chsID` and `thsID`

things/api/things/http/endpoint_test.go
```diff
@@ -1857,85 +1857,93 @@ func TestCreateConnections(t *testing.T) {
...
 
 	cases := []struct {
-		desc     string
-		chanID   string
-		thingIDs []string
-		auth     string
-		status   int
+		desc       string
+		channelIDs []string
+		thingIDs   []string
```
manuio: Use `channelsID` and `thingsID` please

things/api/things/http/requests.go
```diff
@@ -242,20 +242,21 @@ func (req connectionReq) validate() error {
 }
 
 type createConnectionsReq struct {
-	token    string
-	ChanID   string   `json:"chanID,omitempty"`
-	ThingIDs []string `json:"thingIDs,omitempty"`
+	token      string
+	ChannelIDs []string `json:"channel_ids,omitempty"`
+	ThingIDs   []string `json:"thing_ids,omitempty"`
```
manuio, line 247: `ChannelsID` and `ThingsID`, `channels_id` and `things_id`

manuio (changes requested):

bootstrap/mocks/things.go
nwneisen: That would read as channels ID which sounds like multiple channels sharing one ID.  This reads as channel IDs which sounds like multiple channel IDs, which is what this variable contains.  I'll change it if you want but I think it's misleading to what this variable contains.  Same goes for the following comments.

anovakovic01: @nwneisen I agree with you

manuio: `channel IDs` sounds like multiple IDs of one channel. I would say that `channels IDs` is the best 😄. As you prefer guys, I let you vote.

drasko: Agreed with @nwneinsen also. Keeping current implementation is more accurate IMHO.

manuio: @nwneisen people voted. Approving this one!

manuio (dismissed):
LGTM

things/service.go
```diff
@@ -283,13 +283,13 @@ func (ts *thingsService) RemoveChannel(ctx context.Context, token, id string) er
 	return ts.channels.Remove(ctx, res.GetValue(), id)
 }
 
-func (ts *thingsService) Connect(ctx context.Context, token, chID string, thIDs ...string) error {
+func (ts *thingsService) Connect(ctx context.Context, token string, chIDs []string, thIDs []string) error {
```
manuio: This could be `owner string, chIDs, thIDs []string`

nwneisen: There's a few of these.  I'll switch them over.

manuio (dismissed):

docs/provisioning.md
```diff
@@ -245,12 +245,20 @@ Only user, who is the owner of a channel and of the things, can connect the
...
+> This endpoint will be depreciated in 0.11.0.  It will be replaced with the bulk endpoint found at /connect.
+
+```bash
 curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X PUT -H "Authorization: <user_auth_token>" https://localhost/channels/<channel_id>/things/<thing_id>
 ```
 
+To connect multiple things to a channel, you can send the following request:
+
+```bash
+curl -s -S -i --cacert docker/ssl/certs/mainflux-server.crt --insecure -X POST -H "Content-Type: application/json" -H "Authorization: <user_auth_token>" https://localhost/connect -d '{"channel_ids":["<channel_id>", "<channel_id>"],"thingIDs":["<thing_id>", "<thing_id>"]}'
```
drasko: `thingIDs` should be `thing_ids` in this JSON payload

drasko (changes requested):

drasko (dismissed):
LGTM

things/api/things/http/responses.go
```diff
@@ -220,13 +220,29 @@ func (res connectionRes) Code() int {
 }
 
 func (res connectionRes) Headers() map[string]string {
-	return map[string]string{}
+	return map[string]string{
+		"Warning-Deprecated": "This endpoint will be depreciated in 0.11.0. It will be replaced with the bulk endpoint found at /connect.",
```
nmarcetic: We will release v1.0.0  just after 0.10.0 you can see the plan [here](https://github.com/mainflux/mainflux/milestones)

nmarcetic (dismissed):

things/api/things/http/responses.go
nwneisen: I'll switch these messages to 1.0.0.  My thoughts were that the new bulk endpoints would be released in 0.10.0 and the single endpoints removed in the following release.  This would give any users time to switch over.

things/api/logging.go
```diff
@@ -201,17 +201,17 @@ func (lm *loggingMiddleware) RemoveChannel(ctx context.Context, token, id string
 	return lm.svc.RemoveChannel(ctx, token, id)
 }
 
-func (lm *loggingMiddleware) Connect(ctx context.Context, token, chID string, thIDs ...string) (err error) {
+func (lm *loggingMiddleware) Connect(ctx context.Context, token string, chIDs, thIDs []string) (err error) {
 	defer func(begin time.Time) {
-		message := fmt.Sprintf("Method connect for token %s, channel %s and things %s took %s to complete", token, chID, thIDs, time.Since(begin))
+		message := fmt.Sprintf("Method connect for token %s, channel %s and things %s took %s to complete", token, chIDs, thIDs, time.Since(begin))
```
dborovcanin: Use the plural: `...channels %s and...`.

dborovcanin (dismissed):

nmarcetic (approved):

nmarcetic:
LGTM! Good job @nwneisen

dborovcanin (approved):

PR #959, opened 2019-11-20
NOISSUE - Fix Update User

manuio:

### DIFF
--- a/users/service_test.go
+++ b/users/service_test.go
@@ -153,8 +153,8 @@ func TestUpdateUser(t *testing.T) {
 		token string
 		err   error
 	}{
-		"valid token update user":    {user, key, nil},
-		"invalid token's update use": {user, "non-existent", users.ErrUnauthorizedAccess},
+		"update user with valid token":   {user, key, nil},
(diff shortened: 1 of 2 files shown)


users/service_test.go
```diff
@@ -141,12 +141,12 @@ func TestUserInfo(t *testing.T) {
 	}
 }
 
-// UpdateUser updates the user metadata
-
 func TestUpdateUser(t *testing.T) {
 	svc := newService()
 	svc.Register(context.Background(), user)
 	key, _ := svc.Login(context.Background(), user)
+	nonExistingKey, _ := svc.Login(context.Background(), nonExistingUser)
```
mteodor: you cant use login here, you havent registered user, this will fail
to create key
```
j := jwt.New("secret")
key, _ := j.TemporaryKey(user.Email)
```

manuio: @mteodor I agree with you but let's just remove this test. No need to create a valid token for this test.

anovakovic01 (approved):
LGTM!

mteodor (approved):

dborovcanin (approved):

PR #980, opened 2019-12-14
NOISSUE - Use opcua server timestamp in opcua-adapter messages

manuio:


drasko (approved):
LGTM

opcua/gopcua/subscribe.go
```diff
@@ -161,38 +172,38 @@ func (c client) runHandler(sub *opcuaGopcua.Subscription, cfg opcua.Config) erro
...
 
-// Publish forwards messages from OPC-UA MQTT broker to Mainflux NATS broker
-func (c client) Publish(ctx context.Context, token string, m opcua.Message) error {
+// Publish forwards messages from the OPC-UA Server to Mainflux NATS broker
+func (c client) publish(token string, m opcua.Message) error {
 	// Get route-map of the OPC-UA ServerURI
 	chanID, err := c.channelsRM.Get(m.ServerURI)
 	if err != nil {
-		return opcua.ErrNotFoundServerURI
+		return errNotFoundServerURI
```
nmarcetic, line 180: You are not wrapping error messages with our new [errors package](https://github.com/mainflux/mainflux/blob/master/errors/errors.go)? This is why we built it.

nmarcetic (changes requested):

opcua/gopcua/subscribe.go
manuio: Yes I am, as you can see here: https://github.com/mainflux/mainflux/blob/4912abca987302a0a6aaa188226407c795b10655/opcua/gopcua/subscribe.go#L69

I just found unnecessary to return Redis error here.

nmarcetic: I see. `just found unnecessary to return Redis error here.` But this is exactly why we did this wrapping, for example, to see  Redis error, like not found, error connecting, etc... whatever there is plenty of errors from Redis API.

Anyway, thanks for the clarification.

manuio: I understand but in this case the error will be always "not found", which is redundant with my custom error.

nmarcetic: Ok if you are sure that error is always "not found", then yea redundant I agree.

nmarcetic (approved):

PR #1007, opened 2020-01-16
MF-1006 - Password recovery endpoint is not working

### LINKED ISSUE #1006: Password recovery endpoint is not working
**BUG REPORT**

I forgot my password and I'm trying to recover.

I expect an email with a recovery link.

I got status code 400 with info that the user is not found.

Run Docker composition. Register. Try to recover the forgotten password.

I use Docker-compose version 1.25.0, the `latest` version of Mainflux containers. I also faced the same bug with tagged version `0.10.0`.

Issuing a recovery token with `Authentication` service seems to be the problem.

dborovcanin:
### What does this do?
This pull request fixes password recovery.

Fixes #1006.

There are no such changes.

Yes.

Yes.

### DIFF
--- a/authn/service.go
+++ b/authn/service.go
@@ -10,9 +10,9 @@ import (
 )
 
 const (
-	loginDuration = 10 * time.Hour
-	resetDuration = 5 * time.Minute
-	issuerName    = "mainflux.authn"
+	loginDuration    = 10 * time.Hour
+	recoveryDuration = 5 * time.Minute
+	issuerName       = "mainflux.authn"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


drasko (approved):
LGTM

blokovi (approved):

authn/service.go
```diff
@@ -10,9 +10,9 @@ import (
 )
 
 const (
-	loginDuration = 10 * time.Hour
-	resetDuration = 5 * time.Minute
-	issuerName    = "mainflux.authn"
+	loginDuration    = 10 * time.Hour
+	recoveryDuration = 5 * time.Minute
+	issuerName       = "mainflux.authn"
 )
```
nmarcetic, line 16: Can we use this as default values and provide env variables? IMHO 5min is to short for recovery token, I would set at least 15+ min

PR #1026, opened 2020-02-03
MF-998 - Add Twins service to Makefile and docker-compose.yml

darkodraskovic:
Integrates the twins service in the overall mainflux docker composition system.

### DIFF
--- a/docker/addons/opcua-adapter/docker-compose.yml
+++ b/docker/addons/opcua-adapter/docker-compose.yml
@@ -14,6 +14,7 @@ networks:
 
 volumes:
   mainflux-opcua-adapter-volume:
+  mainflux-opcua-redis-volume:
 
 services:
   opcua-redis:
@@ -22,6 +23,8 @@ services:
     restart: on-failure
     networks:
       - docker_mainflux-base-net
+    volumes:
+      - mainflux-opcua-redis-volume:/data
 
   opcua-adapter:
     image: mainflux/opcua:latest


.env
```diff
@@ -190,3 +190,7 @@ MF_POSTGRES_READER_DB_SSL_MODE=disable
 MF_POSTGRES_READER_DB_SSL_CERT=""
 MF_POSTGRES_READER_DB_SSL_KEY=""
 MF_POSTGRES_READER_DB_SSL_ROOT_CERT=""
+
+### Twins
+MF_TWINS_HTTP_PORT=9021
+MF_TWINS_MQTT_URL=tcp://mqtt-adapter:${MF_MQTT_ADAPTER_PORT}
```
manuio: There is no new line at the end of the file

docker/docker-compose.yml
```diff
@@ -260,3 +260,31 @@ services:
...
+      MF_TWINS_THING_ID: ${MF_TWINS_THING_ID}
+      MF_TWINS_THING_KEY: ${MF_TWINS_THING_KEY}
+      MF_TWINS_CHANNEL_ID: ${MF_TWINS_CHANNEL_ID}
+      MF_NATS_URL: ${MF_NATS_URL}
+      MF_TWINS_MQTT_URL: ${MF_TWINS_MQTT_URL}
+    ports:
+      - ${MF_TWINS_HTTP_PORT}:${MF_TWINS_HTTP_PORT}
+    networks:
+      - mainflux-base-net
+    env_file:
```
manuio: This is necessary?

manuio (changes requested):

docker/docker-compose.yml
nmarcetic: I would also avoid additional .env file. If you must put it in global .env twins section

nmarcetic (changes requested):

docker/docker-compose.yml
manuio: It's not an additional one, it's the same. My question is if it really works if you run the composition from a different location than where the .env is.

nmarcetic: @manuio Its called `.env-twins` its a new file. He used both .env and new .env-twins no ?

manuio: Right, I didn't see that he use both. In any case no need to specify the `.env` path

docker/docker-compose.yml
```diff
@@ -260,3 +260,31 @@ services:
       - ${MF_COAP_ADAPTER_PORT}:${MF_COAP_ADAPTER_PORT}/tcp
     networks:
       - mainflux-base-net
+
+  twins:
+    image: mainflux/twins:latest
+    container_name: mainflux-twins
+    depends_on:
+      - authn
+      - nats
```
bbokun: please be aware that the twins service depends on a mongo database, what means that either the dependency has to be to the standard mongodb from the writer, or - better, a separate twins-db has to be created and added to the docker-compose file

.env
bbokun: is the protocol "tcp" neccessary? it works also without it.

bbokun: looking at other sections in the .env file, all the possible variables are listed
here we miss the following:
MF_TWINS_LOG_LEVEL=debug
MF_TWINS_SERVER_CERT=""
MF_TWINS_SERVER_KEY=""
MF_TWINS_DB_NAME=mainflux
MF_TWINS_DB_HOST=twins-db
MF_TWINS_DB_PORT=27017
MF_TWINS_SINGLE_USER_EMAIL=""
MF_TWINS_SINGLE_USER_TOKEN=""
MF_TWINS_CLIENT_TLS=""
MF_TWINS_CA_CERTS=""

bbokun (changes requested):

PR #1029, opened 2020-02-04
NOISSUE - Add opc-ua type handling and unsubscription

manuio:


opcua/gopcua/browser.go
```diff
@@ -82,7 +85,7 @@ func (c browser) Browse(serverURI, nodeID string) ([]opcua.BrowsedNode, error) {
 }
 
 func browse(n *opcuaGopcua.Node, path string, level int) ([]NodeDef, error) {
-	if level > 10 {
+	if level > maxChildrens {
 		return nil, nil
```
mteodor, line 89: no error here?

manuio: No need, it's the end of the iteration.

opcua/gopcua/subscribe.go
```diff
@@ -99,15 +108,15 @@ func (c client) Subscribe(cfg opcua.Config) error {
...
+	if err := c.runHandler(sub, cfg.ServerURI, cfg.NodeID); err != nil {
+		c.logger.Warn(fmt.Sprintf("Unsubscribed from OPC-UA node %s.%s: %s", cfg.ServerURI, cfg.NodeID, err))
 	}
 
 	return nil
 }
 
-func (c client) runHandler(sub *opcuaGopcua.Subscription, cfg opcua.Config) error {
-	nodeID, err := uaGopcua.ParseNodeID(cfg.NodeID)
+func (c client) runHandler(sub *opcuaGopcua.Subscription, uri, node string) error {
```
mteodor, line 118: this method maybe should be split in smaller methods

opcua/gopcua/subscribe.go
```diff
@@ -145,32 +154,43 @@ func (c client) runHandler(sub *opcuaGopcua.Subscription, cfg opcua.Config) erro
...
+						msg.Data = item.Value.Value.Time()
 					default:
 						msg.Data = 0
 					}
 
 					if err := c.publish(token, msg); err != nil {
-						c.logger.Warn(fmt.Sprintf("failed to publish: %s", err))
+						switch err {
+						case errNotFoundServerURI, errNotFoundNodeID, errNotFoundConn:
+							return err
```
mteodor, line 190: I dont understand how is this error returned?
you are waiting on the channel

manuio: This function runs in a go routine that will stop if there is an error with the routemap

opcua/gopcua/subscribe.go
manuio: I agree, it's just that for now I'm trying to keep this code similar to an example that I'm using

mteodor (approved):

opcua/gopcua/subscribe.go
mteodor: so you call it here https://github.com/manuio/mainflux/blob/opcua/opcua/gopcua/subscribe.go#L111
and it stays in runHandler until error occurs?

manuio: yes

drasko (approved):
LGTM

PR #1035, opened 2020-02-11
MF-998 - Add Twins service to Makefile and docker-compose.yml

darkodraskovic:


manuio (dismissed):
LGTM

docker/docker-compose.yml
```diff
@@ -263,3 +265,47 @@ services:
       - ${MF_COAP_ADAPTER_PORT}:${MF_COAP_ADAPTER_PORT}/tcp
     networks:
       - mainflux-base-net
+
+  twins-db:
+    image: mongo:3.6-jessie
```
drasko: Although I have nothing against, I would advise using latest stable - 4.2: https://docs.mongodb.com/manual/release-notes/

Also, it looks like for some reason they prefer Ubuntu (Bionic) then Debian Jessie by default: https://hub.docker.com/_/mongo

.env
```diff
@@ -196,3 +196,20 @@ MF_POSTGRES_READER_DB_SSL_MODE=disable
...
+MF_TWINS_SERVER_KEY=""
+MF_TWINS_DB_NAME=mainflux
+MF_TWINS_DB_HOST=twins-db
+MF_TWINS_DB_PORT=27017
+MF_TWINS_SINGLE_USER_EMAIL=""
+MF_TWINS_SINGLE_USER_TOKEN=""
+MF_TWINS_CLIENT_TLS=""
+MF_TWINS_CA_CERTS=""
+MF_TWINS_MQTT_URL=tcp://mqtt-adapter:1883
+MF_TWINS_THING_ID=cfc93379-d09a-438b-806e-e3cda2180124
```
drasko: Do we need these default values, or is it better to leave this empty?

drasko (changes requested):

docker/docker-compose.yml
darkodraskovic: Will be fixed.

.env
darkodraskovic: Unfotunatelly, these values are necessary. Therefore, we should leave them empty. What is more, I will move twins service to addons.

darkodraskovic: I'm working towards eliminating these env vars alltogether.

docker/nginx/entrypoint.sh
```diff
@@ -14,6 +14,7 @@ envsubst '
     ${MF_THINGS_HTTP_PORT}
     ${MF_THINGS_HTTP_PORT}
     ${MF_HTTP_ADAPTER_PORT}
+    ${MF_TWINS_HTTP_PORT}
```
manuio: Remove this

manuio (changes requested):

manuio (dismissed):
LGTM

.env
mteodor: some vars are not used in docker-compose

mteodor (dismissed):

docker/addons/twins/docker-compose.yml
```diff
@@ -0,0 +1,55 @@
...
+  docker_mainflux-base-net:
+    external: true
+
+volumes:
+  mainflux-twins-db-volume:
+  mainflux-twins-db-configdb-volume:
+
+services:
+  twins-db:
+    image: mongo:3.6-jessie
```
drasko: Use new Mongo release, test with it. Use Ubuntu, as they switched to Ubuntu (Bionic I think).

.env
drasko: Is this still ongoing?

drasko (changes requested):

drasko (approved):
LGTM

docker/addons/twins/docker-compose.yml
drasko: Is there a particular reason for sticking with old MondoDB release?

drasko (changes requested):

drasko (approved):
LGTM

PR #1046, opened 2020-02-26
NOISSUE - Add default subscription nodeID and Interval ENVAR

manuio:


opcua/gopcua/subscribe.go
```diff
@@ -101,7 +101,7 @@ func (c client) Subscribe(cfg opcua.Config) error {
 	defer oc.Close()
 
 	sub, err := oc.Subscribe(&opcuaGopcua.SubscriptionParameters{
-		Interval: 2000 * time.Millisecond,
+		Interval: 1000 * time.Millisecond,
```
bbokun: Does it make sense to use an env variable from outside of the image?

opcua/api/endpoint.go
```diff
@@ -19,7 +19,7 @@ func browseEndpoint(svc opcua.Service) endpoint.Endpoint {
 			return nil, err
 		}
 
-		nodeID := fmt.Sprintf("ns=%s;%s", req.Namespace, req.Identifier)
+		nodeID := fmt.Sprintf("%s;%s", req.Namespace, req.Identifier)
```
bbokun: Didn't we say that when the req.Namespace and req.Identifier are empty, the defaults of ns=1, i=84 are automatically used, as the user definitely can not know this information.

bbokun (changes requested):

opcua/gopcua/subscribe.go
drasko: In any case, move constants to the top. But yes, this should probably be configurable via ENV var.

drasko (changes requested):

cmd/opcua/main.go
```diff
@@ -46,6 +47,7 @@ const (
 
 	envHTTPPort       = "MF_OPCUA_ADAPTER_HTTP_PORT"
 	envLogLevel       = "MF_OPCUA_ADAPTER_LOG_LEVEL"
+	envOPCIntervalMs  = "MF_OPCUA_ADAPTER_INTERVAL_MS"
```
drasko, line 50: Should this be added to README?

drasko (changes requested):

darkodraskovic (dismissed):
LGTM

drasko (approved):
LGTM

PR #1072, opened 2020-03-11
NOISSUE - Adding subtopics filtering in writer services

### LINKED ISSUE #925: Writers - default value for channel list when toml file is not found
**FEATURE REQUEST**

No

When the Channel Config Path is not set then the writers exit with an error - that the file was not found. It would be better that the channel list is set to [*] - meaning all channels.

jonathandreyer:
### What does this do?
Adding subtopic's filtering for each writer service. It is strongly based on channel's filtering.

Also resolves #925.

Update readme of each writer service with new

### Notes
I have also refactored a little part of the main code of every writer services to centralize duplicate code into writer package.


cmd/cassandra-writer/main.go
```diff
@@ -31,33 +29,36 @@ const (
...
-	logLevel string
-	port     string
-	dbCfg    cassandra.DBConfig
-	channels map[string]bool
+	natsURL   string
+	logLevel  string
+	port      string
+	dbCfg     cassandra.DBConfig
+	channels  map[string]bool
+	subtopics map[string]bool
```
manuio: Please, can you create a separate `filtersCfg` for these filters?  You can also move the new structure at the `writers` level.

manuio (changes requested):

cmd/cassandra-writer/main.go
```diff
@@ -76,7 +77,7 @@ func main() {
 
 	repo := newService(session, logger)
 	st := senml.New()
-	if err := writers.Start(nc, repo, st, svcName, cfg.channels, logger); err != nil {
+	if err := writers.Start(nc, repo, st, svcName, cfg.channels, cfg.subtopics, logger); err != nil {
```
manuio: Pass `cfg.filtersCfg`  instead of `cfg.channels, cfg.subtopics` here

manuio (changes requested):

writers/writer.go
```diff
@@ -76,3 +92,70 @@ func (c *consumer) channelExists(channel string) bool {
...
+func (c *consumer) subtopicExists(subtopic string) bool {
+	if _, ok := c.subtopics["*"]; ok {
+		return true
+	}
+
+	_, found := c.subtopics[subtopic]
+	return found
+}
+
+type filterConfig struct {
```
manuio: I think that this should be `subtopicsConfig` or even better `subtopicsFilter`

writers/writer.go
```diff
@@ -58,10 +67,17 @@ func (c *consumer) consume(m *nats.Msg) {
 	var msgs []senml.Message
 	for _, v := range norm {
 		if c.channelExists(v.Channel) {
-			msgs = append(msgs, v)
+			if c.subtopicExists(v.Subtopic) {
```
dborovcanin: Please use `if c.channelExists(v.Channel) && c.subtopicExists(v.Subtopic) {...}`.

writers/writer.go
```diff
@@ -76,3 +92,70 @@ func (c *consumer) channelExists(channel string) bool {
...
+	if _, ok := c.subtopics["*"]; ok {
+		return true
+	}
+
+	_, found := c.subtopics[subtopic]
+	return found
+}
+
+type filterConfig struct {
+	List []string `toml:"filter"`
```
manuio: `Subtopics` instead of `List`

writers/writer.go
```diff
@@ -76,3 +92,70 @@ func (c *consumer) channelExists(channel string) bool {
...
+}
+
+type channelsConfig struct {
```
manuio: Use `channelsFilter`

docker/addons/postgres-writer/subtopics.toml
```diff
@@ -0,0 +1,4 @@
+# If you want to listen on all subtopics, just pass one element ["*"], otherwise
+# pass the list of subtopics.
+[subtopics]
+filter = ["*"]
```
manuio: add a blank line at the end of the file

docker/addons/mongodb-writer/subtopics.toml
```diff
@@ -0,0 +1,4 @@
+# If you want to listen on all subtopics, just pass one element ["*"], otherwise
+# pass the list of subtopics.
+[subtopics]
+filter = ["*"]
```
manuio: add a blank line at the end of the file

writers/writer.go
```diff
@@ -76,3 +92,70 @@ func (c *consumer) channelExists(channel string) bool {
...
+	}
+
+	return channels
+}
+
+type subtopicsConfig struct {
+	Subtopics filterConfig `toml:"subtopics"`
+}
+
+func LoadSubtopicsConfig(subtopicConfigPath string) map[string]bool {
```
manuio: Return `(map[string]bool, error)` here.

writers/writer.go
```diff
@@ -76,3 +92,70 @@ func (c *consumer) channelExists(channel string) bool {
...
+}
+
+type channelsConfig struct {
+	Channels filterConfig `toml:"channels"`
+}
+
+func LoadChannelsConfig(chanConfigPath string) map[string]bool {
+	data, err := ioutil.ReadFile(chanConfigPath)
+	if err != nil {
+		log.Print(fmt.Sprintf("Failed to open configuration file: %s", err))
```
manuio: For any log use `mainflux/logger` package instead of `"log"`. But here better create an error for this message and return `errors.Wrap(errOpenConfFile, err)` (using `mainflux/errors`).

writers/writer.go
```diff
@@ -76,3 +92,70 @@ func (c *consumer) channelExists(channel string) bool {
...
+		return getDefaultFilter()
+	}
+
+	subtopics := map[string]bool{}
+	for _, ch := range subtopicCfg.Subtopics.List {
+		subtopics[ch] = true
+	}
+
+	return subtopics
+}
```
dborovcanin: Can you replace Channels and Subtopics config and filter with one `Filter` (`FilterConfig`, `LoadFilter`, as well) that contains both channels and subtopics filters?

manuio (changes requested):

dborovcanin (changes requested):

writers/writer.go
```diff
@@ -76,3 +92,70 @@ func (c *consumer) channelExists(channel string) bool {
...
+
+type filterConfig struct {
+	List []string `toml:"filter"`
+}
+
+type channelsConfig struct {
+	Channels filterConfig `toml:"channels"`
+}
+
+func LoadChannelsConfig(chanConfigPath string) map[string]bool {
```
manuio: Return (map[string]bool, error) here.

manuio (changes requested):

writers/writer.go
jonathandreyer: Can I merge both files (channels.toml & subtopics.toml) into one file?

jonathandreyer: @manuio What do you think about that?

manuio: @jonathandreyer I think that 2 files it's easier to understand. I also prefer something retro-compatible.

writers/writer.go
jonathandreyer: I have return the error message but I don't know how print the message because at this point the logger (`logger, err := logger.New(os.Stdout, cfg.logLevel)`) isn't already loaded.

writers/writer.go
```diff
@@ -76,3 +90,70 @@ func (c *consumer) channelExists(channel string) bool {
...
+	_, found := c.subtopics[subtopic]
+	return found
+}
+
+type filterConfig struct {
+	List []string `toml:"filter"`
+}
+
+type channelsConfig struct {
+	Channels filterConfig `toml:"channels"`
```
manuio: Use `[]string` instead of `filterConfig`

cmd/influxdb-writer/main.go
```diff
@@ -108,7 +108,9 @@ func main() {
 }
 
 func loadConfigs() (config, influxdata.HTTPConfig) {
-	chanCfgPath := mainflux.Env(envChanCfgPath, defChanCfgPath)
+	channelsCfgPath := mainflux.Env(envChannelsCfgPath, defChannelsCfgPath)
+	subtopicsCfgPath := mainflux.Env(envSubtopicsCfgPath, defSubtopicsCfgPath)
+	filters, _ := writers.LoadFiltersConfig(channelsCfgPath, subtopicsCfgPath)
```
manuio: Don't ignore errors please

manuio (changes requested):

writers/writer.go
```diff
@@ -76,3 +90,70 @@ func (c *consumer) channelExists(channel string) bool {
...
+}
+
+type subtopicsConfig struct {
+	Subtopics filterConfig `toml:"subtopics"`
```
manuio: same here

cmd/postgres-writer/main.go
```diff
@@ -114,40 +113,17 @@ func loadConfig() config {
 		SSLRootCert: mainflux.Env(envDBSSLRootCert, defDBSSLRootCert),
 	}
 
+	channelsCfgPath := mainflux.Env(envChannelsCfgPath, defChannelsCfgPath)
+	subtopicsCfgPath := mainflux.Env(envSubtopicsCfgPath, defSubtopicsCfgPath)
+	filters, _ := writers.LoadFiltersConfig(channelsCfgPath, subtopicsCfgPath)
```
manuio: same here

cmd/mongodb-writer/main.go
```diff
@@ -105,45 +105,20 @@ func main() {
 }
 
 func loadConfigs() config {
-	chanCfgPath := mainflux.Env(envChanCfgPath, defChanCfgPath)
+	channelsCfgPath := mainflux.Env(envChannelsCfgPath, defChannelsCfgPath)
+	subtopicsCfgPath := mainflux.Env(envSubtopicsCfgPath, defSubtopicsCfgPath)
+	filters, _ := writers.LoadFiltersConfig(channelsCfgPath, subtopicsCfgPath)
```
manuio: same here

cmd/cassandra-writer/main.go
```diff
@@ -108,43 +108,18 @@ func loadConfig() config {
 		Port:     dbPort,
 	}
 
-	chanCfgPath := mainflux.Env(envChanCfgPath, defChanCfgPath)
+	channelsCfgPath := mainflux.Env(envChannelsCfgPath, defChannelsCfgPath)
+	subtopicsCfgPath := mainflux.Env(envSubtopicsCfgPath, defSubtopicsCfgPath)
+	filters, _ := writers.LoadFiltersConfig(channelsCfgPath, subtopicsCfgPath)
```
manuio: Don't ignore errors please

manuio (changes requested):

manuio:
@jonathandreyer update your branch please

writers/writer.go
```diff
@@ -76,3 +90,70 @@ func (c *consumer) channelExists(channel string) bool {
...
+}
+
+type FiltersCfg struct {
+	channels  map[string]bool
+	subtopics map[string]bool
```
manuio: Use `[]string` instead of `map[string]bool`

cmd/influxdb-writer/main.go
jonathandreyer: Yes, you have fully right. This is why I have ask this question: https://github.com/mainflux/mainflux/pull/1072#discussion_r391892220

jonathandreyer: @manuio I understand your response but I don't know what I need to do with this error because the loader isn't already loaded.

writers/writer.go
```diff
@@ -76,3 +90,70 @@ func (c *consumer) channelExists(channel string) bool {
...
+	var channelsCfg channelsConfig
+	if err := toml.Unmarshal(data, &channelsCfg); err != nil {
+		return filters, errors.Wrap(errParseConfFile, err)
+	}
+
+	channels := map[string]bool{}
+	for _, ch := range channelsCfg.Channels.List {
+		channels[ch] = true
+	}
+	filters.channels = channels
```
manuio: This could be simplified by:
```
filters.channels = channelsCfg.Channels.List
```

writers/writer.go
```diff
@@ -76,3 +90,70 @@ func (c *consumer) channelExists(channel string) bool {
...
 }
+
+func (c *consumer) subtopicExists(subtopic string) bool {
+	if _, ok := c.subtopics["*"]; ok {
+		return true
+	}
+
+	_, found := c.subtopics[subtopic]
+	return found
+}
```
manuio:
```
for _, s := range c.subtopics {
    if s == subtopic ||  s == "*" {
        return true
    }
}
return false
```

writers/writer.go
```diff
@@ -76,3 +90,70 @@ func (c *consumer) channelExists(channel string) bool {
...
+	var subtopicCfg subtopicsConfig
+	if err := toml.Unmarshal(data, &subtopicCfg); err != nil {
+		return filters, errors.Wrap(errParseConfFile, err)
+	}
+
+	subtopics := map[string]bool{}
+	for _, ch := range subtopicCfg.Subtopics.List {
+		subtopics[ch] = true
+	}
+	filters.subtopics = subtopics
```
manuio: same here `filters.subtopics = subtopicCfg.Subtopics.List`

writers/writer.go
```diff
@@ -57,11 +66,16 @@ func (c *consumer) consume(m *nats.Msg) {
...
 	var msgs []senml.Message
 	for _, v := range norm {
-		if c.channelExists(v.Channel) {
+		if c.channelExists(v.Channel) && c.subtopicExists(v.Subtopic) {
 			msgs = append(msgs, v)
 		}
 	}
 
+	if msgs == nil {
+		c.logger.Debug("No message to saved.")
```
manuio: Fix or remove this log

manuio (changes requested):

cmd/postgres-writer/main.go
jonathandreyer: @manuio To fix this, I need your feedback about that: https://github.com/mainflux/mainflux/pull/1072#discussion_r391903974

writers/writer.go
jonathandreyer: I can't use `s.Key`. Have you idea how to fix that?

jonathandreyer: @manuio I can't use `s.Key`. Have you idea how to fix that?

cmd/cassandra-writer/main.go
```diff
@@ -31,33 +29,35 @@ const (
...
+	defNatsURL          = nats.DefaultURL
+	defLogLevel         = "error"
+	defPort             = "8180"
+	defCluster          = "127.0.0.1"
+	defKeyspace         = "mainflux"
+	defDBUsername       = ""
+	defDBPassword       = ""
+	defDBPort           = "9042"
+	defChannelsCfgPath  = "/config/channels.toml"
+	defSubtopicsCfgPath = "/config/subtopics.toml"
```
drasko: I am not sure that we need to split this in 2 different configs. It would be better to have just one - like `subjects.toml` which would actually list the whole NATS subject (channel + subtopic)

drasko (changes requested):

drasko:
I did not yet go into details, but at the first look this looks over-complicated to me. We must brainstorm to find a simpler solution. This introduces impact, and is not something that I'd like to maintain (at least at the first look).

@jonathandreyer can you please describe us your use-case and requirement here in greater detail. From what I understood, you needed to redirect messages to 2 different DBs. Why could not you do it on the channel level?

writers/writer.go
jonathandreyer: Edit: I just need to remove `.Key` and it is OK. But there is another interrogation:
```go
for _, s := range c.subtopics {
	if s == subtopic ||  s == "*" {
		return true
	}
	return false
}
// Missing return call
```
Do I need to `return false` at the place of comment?

jonathandreyer: `for _, s := range c.subtopics {
	if s == subtopic ||  s == "*" {
		return true
	}
	return false
}
// Missing return call`

jonathandreyer:
> I did not yet go into details, but at the first look this looks over-complicated to me. We must brainstorm to find a simpler solution. This introduces impact, and is not something that I'd like to maintain (at least at the first look).
> 
> @jonathandreyer can you please describe us your use-case and requirement here in greater detail. From what I understood, you needed to redirect messages to 2 different DBs. Why could not you do it on the channel level?

I fully understand the first point.

Sure, I will try to explain the reason. The explain that, I will take as an example the subtopics in the documentation (https://mainflux.readthedocs.io/en/latest/messaging/#subtopics). In my usage, I have two channels by device (which are provided by bootstrap service) and subtopics for each kind of sended values. I would like store numerical values (related to specific topics, like temperature) in influxDB and string values in another database (mongoDB) for more general information about device, like wake-up, shutdown, etc. With this function (filtering by subtopics), I can select which information I would like to store in which database.

drasko:
OK, this is something what I thought.

Now second question: what about having just one config file - for example `subjects.toml` which would list full NATS subject to which Writer should subscribe (instead of having separate `channels.toml` and `subtopics.toml` which IMHO complicate the things, especially that it is not clear to me which is a colleration with a certain channel and its subtopics).

cmd/influxdb-writer/main.go
manuio: @jonathandreyer I didn't see this, sorry. What I propose you is that you call `writers.LoadFiltersConfig` outside `loadConfigs` and after to instantiate the logger. You would be able to check the error and eventually log it with mainflux logger. If there is no error you can do `cfg.filterConfig = filters`

manuio (changes requested):

writers/writer.go
manuio: @jonathandreyer No, you are right. I copy pasted a piece of code but it was not complete. I just edited my previous comment, now it should be ok.

jonathandreyer:
> Now second question: what about having just one config file - for example `subjects.toml` which would list full NATS subject to which Writer should subscribe (instead of having separate `channels.toml` and `subtopics.toml` which IMHO complicate the things, especially that it is not clear to me which is a colleration with a certain channel and its subtopics).

@drasko For me, both alternatives are ok. As discussed here (https://github.com/mainflux/mainflux/pull/1072#discussion_r391604923), @manuio think is better to keep both files for retro-compatible reason and also easier to understand. With the resolution of #925, I think the version with one file is OK.

cmd/cassandra-writer/main.go
manuio: @drasko I proposed to @jonathandreyer to start by adding a new file. It's simplier and retro-compatible but if you think that it's ok to merge both in this PR I'm also ok

writers/writer.go
```diff
@@ -57,22 +66,82 @@ func (c *consumer) consume(m *nats.Msg) {
...
+	var subtopicCfg subtopicsConfig
+	if err := toml.Unmarshal(data, &subtopicCfg); err != nil {
+		return filters, errors.Wrap(errParseConfFile, err)
+	}
+
+	filters.subtopics = subtopicCfg.Subtopics.List
 
-	_, found := c.channels[channel]
-	return found
+	return filters, err
```
manuio: Sorry again, but this can be simplified with:
```
return FiltersCfg{
    channels: channelsCfg.Channels.List,
    subtopics: subtopicsCfg.Subtopics.List,
  }, err
```

If you have errors you can do: `return FiltersCfg{}, errors.Wrap(errOpenConfFile, err)`

manuio (changes requested):

writers/writer.go
dborovcanin: @jonathandreyer I don't mind if you merge files. I think that we can keep backward compatibility even if we merge files because we'll only be adding another TOML table.

cmd/cassandra-writer/main.go
jonathandreyer: @dusanb94 What do you think about that (rename file to `subjects.toml`)?

drasko: My proposal for `subjects.toml` came from the fact that we can make Writers independent of the channels, and provide an arbitrary NATS subject to which they can subscribe and consume the data.

This proved as a beneficial option especially on the gateway where we had some applications that are not necessarily going through adapters, but are sending data directly on NATS, because they are internal to Mainflux Edge system (for example Influx Telegraf does that). Of course - these applications should send data in SenML format for now in order for Writers to parse it, but for now we create these applications (so we can control their format), and in the future we can thing on generalizing Writers.

In any case - I think that approach with NATS subjects is the most generic one, and then user can put any subject he wants - with channel only or with channel + subtopic, or something totally different that is meaningful to his application.

drasko:
> @drasko For me, both alternatives are ok. As discussed here ([#1072 (comment)](https://github.com/mainflux/mainflux/pull/1072#discussion_r391604923)), @manuio think is better to keep both files for retro-compatible reason and also easier to understand. With the resolution of #925, I think the version with one file is OK.

I replayed to this here: https://github.com/mainflux/mainflux/pull/1072#discussion_r392160561

cmd/cassandra-writer/main.go
dborovcanin: @drasko Makes sense to me. Initially, I thought that knowing NATS topics is an unnecessary burden for the user, but the user needs to know topics to which messages are published anyways.

jonathandreyer: @drasko @dusanb94 We merge both files into one file named `subjects.toml`? If it is ok for you, can we take this template of file?
```
[channels]
filter = ["*"]

[subtopics]
filter = ["*"]
```

dborovcanin: @jonathandreyer Yes, that looks fine.

drasko: Do we need a separate table?

Why not just
```toml
subjects = [...]
```

jonathandreyer: @drasko Technically, it is not in the same field we need to check. For channels, it is the `Channel` field and for subtopics, it is the `Subtopics` field (https://github.com/jonathandreyer/mainflux/blob/01f1e4547f0b38deedc09db547dbc9a88be34e5d/writers/writer.go#L69)

drasko: @jonathandreyer I am OK with the previous proposal also, because probably conf file parser would expect something like this. Please keep previous form then, and we'll se how it looks like in a corrected PR.

jonathandreyer: @drasko What do you mean by: 
> Please keep previous form then, and we'll se how it looks like in a corrected PR.

jonathandreyer: I keep the older file `channels.toml` and create a new one `subject.toml`?

drasko: I ment you can keep:
```toml
[subjects]
filter = ["*"]
```

All you need is an array of NATS subjects to which Writer will subscribe to.

jonathandreyer: So, I don't merge the two configuration files? Actually, the PR is in this state without the merge of two configuration files.
PS: I am sorry, but I did not really understand what you would like to do.

drasko: No, no merging. Rather implementation of this design idea - generic NATS subject list.

jonathandreyer: Ok I understand your idea, but I have no idea how to proceed. Have you an idea what you would to have? Can you describe a example of what you would like to have?
When you say `generic NATS subject list`, you would like replace the actual mechanism (channels.toml or channels.toml+subtopics.toml) by something like:
```
subject = ["*.bedroom.temperature"]
```

drasko: Either by:
```toml
subjects = ["*.bedroom.temperature", "something.else", ...]
```
Or by:
```toml
[subjects]
filter = ["*"]
```

But it boils down to the same thing - TOML file should provide a list of NATS subjects to which Writer will subscribe.

manuio (approved):
LGTM

drasko:
@manuio this cannot be approved until remarks are resolved and design is changed not to use 2 files but only one, `subjects.toml`.

drasko:
@jonathandreyer are you working on this one or it has been abandoned?

jonathandreyer:
@drasko I have not abandoned the development. I have taken a break of some days.
About your idea https://github.com/mainflux/mainflux/pull/1072#discussion_r392613992 of only one file, I do not know how modify to implement this feature.

jonathandreyer:
For me, the big question is about the call of the method `QueueSubscribe()` of NATS service (here: https://github.com/jonathandreyer/mainflux/blob/01f1e4547f0b38deedc09db547dbc9a88be34e5d/writers/writer.go#L46). I have not skill in NATS. I don't know if it is possible to subscribe to many channels with rules.

drasko:
@jonathandreyer just crate on file `subjects.toml` and list over there full NATS subjects to which Writes subscribes.

writers/writer.go
```diff
@@ -57,22 +66,79 @@ func (c *consumer) consume(m *nats.Msg) {
...
-		if c.channelExists(v.Channel) {
+		if c.channelExists(v.Channel) && c.subtopicExists(v.Subtopic) {
 			msgs = append(msgs, v)
 		}
 	}
 
+	if msgs == nil {
+		return
+	}
+
```
manuio: This check should be removed from here and inside `Start` method you should subscribe to NATS subjects that are in the config file as channel + subtopic. If channel or/and subtopic are empty `"*"` there is no filter to apply and you can subscribe to subject `channel.>`. That way you don't have to check in the consumer if messages that are passing have to be filtered or not because it would be previously done subscribing to the proper subjects.

manuio (changes requested):

cmd/cassandra-writer/main.go
manuio: Can we use `defSubjectsCfgPath  = "/config/subjects.toml"` and: 
```
[subjects]
channels=["*"]
subtopics=["*"]
```

manuio (changes requested):

cmd/cassandra-writer/main.go
drasko: No need for `channels` and `subtopics` anymore - `subject` is made by concatenating `channel` and `subtopic` (example `subtopic = channels/<chan_id>/messages/a/b/c`)

writers/writer.go
jonathandreyer: Ok, I will see what I can did. I will probably modifying in first inside main's writer (cassandra, influx, ..) and after see how we can manage this point.

jonathandreyer:
@manuio Ok, I see what you propose (https://github.com/mainflux/mainflux/pull/1072#discussion_r396348524) but it is in conflit with @drasko idea (https://github.com/mainflux/mainflux/pull/1072#discussion_r396617040). What is the best choice?

manuio:
@jonathandreyer Use @drasko 's proposition. It's easier to understand the associations using the whole subject, channel and subtopic are defined together.

manuio:
@jonathandreyer The only thing to add is that you probably have to replace `*` by `>` . I let you check

jonathandreyer:
Hi @manuio & @drasko, I have merged both filters files into one file as explain before but now it has no effect. My knowledge in NATS is pretty poor to subscribe in only selected subjects (here: https://github.com/jonathandreyer/mainflux/blob/4c6576193344b82f067a25f22abf9d4ebbc680b5/writers/writer.go#L45). Have you an idea how proceed?

writers/writer.go
```diff
@@ -5,35 +5,43 @@ package writers
...
 	c := consumer{
 		nc:          nc,
-		channels:    channels,
+		subjects:    filters.subjects,
 		repo:        repo,
 		transformer: transformer,
 		logger:      logger,
 	}
 
+	// TODO subscribe with only selected subjects
```
manuio: I think that here you can pass `subjectsConfigPath` as argument and  call `LoadSubjectsConfig` to avoid the duplication in all writers.

writers/writer.go
```diff
@@ -5,35 +5,43 @@ package writers
...
 	_, err := nc.QueueSubscribe(mainflux.InputChannels, queue, c.consume)
```
manuio: Here you should use `mainflux.InputChannels` if `LoadSubjectsConfig` return null or `[*]`. If it returns one or more subjects then you should use a for loop to subscribe to all returned filters. Keep in mind that your subsjects should start by `channel.` and followed by your filter.

manuio (changes requested):

writers/writer.go
jonathandreyer: Ok.

docker/addons/cassandra-writer/subjects.toml
```diff
@@ -0,0 +1,4 @@
+# If you want to listen on all subjects, just pass one element ["*"], otherwise
+# pass the list of subjects.
```
drasko: I think it would be a good to leave here (in a comment) an example of this list (maybe with and without subtopics), so that we have a documentation how this `filter` field can be specified.

drasko (changes requested):
@jonathandreyer this starts to look good. On the first look just a minor suggestion regarding the docs, but I will analyze with more attention tomorrow.

We will be needing approvals from several maintainers, as this touches lot of code.

manuio (changes requested):

writers/writer.go
```diff
@@ -5,37 +5,58 @@ package writers
...
-	return err
+	subjects, err := LoadSubjectsConfig(subjectsCfgPath)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Failed to load subjects: %s", err))
+	}
+
+	if reflect.DeepEqual([]string{"*"}, subjects) {
+		_, err := nc.QueueSubscribe(mainflux.InputChannels, queue, c.consume)
+		return err
+	} else {
```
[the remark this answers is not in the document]
jonathandreyer: Personally, I think it is a good idea to do something with `mainflux.InputChannels`. Probably, we can also keep `mainflux.InputChannels` but only with `channel.` and use `>` as default filter.

writers/writer.go
```diff
@@ -5,36 +5,48 @@ package writers
...
+	"github.com/BurntSushi/toml"
 	"github.com/gogo/protobuf/proto"
 	"github.com/mainflux/mainflux"
-	log "github.com/mainflux/mainflux/logger"
+	"github.com/mainflux/mainflux/errors"
+	"github.com/mainflux/mainflux/logger"
 	"github.com/mainflux/mainflux/transformers"
 	"github.com/mainflux/mainflux/transformers/senml"
 	nats "github.com/nats-io/nats.go"
+	"io/ioutil"
```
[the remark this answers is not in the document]
jonathandreyer: Sure, I am going to wait feedback from @drasko other points to fix altogether.

writers/writer.go
```diff
@@ -5,37 +5,58 @@ package writers
...
 		repo:        repo,
 		transformer: transformer,
 		logger:      logger,
 	}
 
-	_, err := nc.QueueSubscribe(mainflux.InputChannels, queue, c.consume)
-	return err
+	subjects, err := LoadSubjectsConfig(subjectsCfgPath)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Failed to load subjects: %s", err))
```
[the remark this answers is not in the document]
jonathandreyer: Good idea.

writers/writer.go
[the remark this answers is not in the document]
manuio: Ok for `channel.`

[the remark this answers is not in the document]
jonathandreyer: Ok.

docker/addons/cassandra-writer/subjects.toml
jonathandreyer: Good point. I have planned to edit the section filter of the writers in the documentation. I didn't know if it is better to explain here or in the documentation.

jonathandreyer: Personally, I think it better to add documentation inside the `docs` repository.

drasko: This should be done as well. But I prefer that this config file holds one comment as well about the format that it accepts.

jonathandreyer: Maybe something like that can be enough:
`If you want to listen on all subjects, just pass one element [">"], otherwise pass the list of subjects (e.g ["*.bedroom.temperature" "*.livingroom.*"]).`
What do you think?

writers/writer.go
```diff
@@ -5,36 +5,52 @@ package writers
...
 	}
 
-	_, err := nc.QueueSubscribe(mainflux.InputChannels, queue, c.consume)
+	subjects, err := LoadSubjectsConfig(subjectsCfgPath)
+	if err != nil {
+		logger.Warn(fmt.Sprintf("Failed to load subjects: %s", err))
+	}
+
+	for _, subject := range subjects {
+		_, err := nc.QueueSubscribe(mainflux.InputChannels + subject, queue, c.consume)
```
manuio: Please, proceed as proposed in the mainflux/docs review. Use the subject here without prefix and move this prefix into the config file. Also, don't forget to use `mainflux.InputChannels` as default value if `LoadSubjectsConfig` fails.

manuio (changes requested):

writers/writer.go
jonathandreyer: @manuio @drasko OK, I understand. When we set filters in `subjects.toml`, it is OK for me to set the full path.
But I have some interrogations about the default value:
- First point, do I need to restore the constant `mainflux.InputChannels` to `channel.>`. to keep as default value? If I did that, I can simply use this value without concatenation.
- Second point, what is the default value in the file `subjects.toml`? Need I restore `["*"]` or we create an empty array?

manuio: * yes to restore the constant `mainflux.InputChannels`
* default value should be `["channels.>"]`

jonathandreyer: OK, I understand. Personally both versions are OK and I can use both versions. But I still have a preference for the "current version" with relative path for the following reasons:
- As I have understood the operation of Mainflux, all messages are sent on one channel and it is not possible to send messages without `/channels/<channel_id>`. So for me, it is better to set (in a constant) this prefix.
- It is a little bit more complicated but that is more in line with how work Mainflux.
Maybe a good night's sleep will clear this mess :-)

docker/addons/cassandra-writer/subjects.toml
manuio: @jonathandreyer please use version with full subject in the config file.

writers/writer.go
manuio: @jonathandreyer please use version with full subject in the config file.

manuio (dismissed):
LGTM

docker/addons/cassandra-writer/subjects.toml
```diff
@@ -0,0 +1,4 @@
+# If you want to listen on all subjects, just pass one element ["channel.>"], otherwise
+# pass the list of subjects.
```
drasko: As I mention, an example of subjects, for example one with and one without subtopic should be added.

Example: `["channel.<channel_ID>.messages, channel.<channel_id>.messages/sub/topic/x, ...]`

drasko (changes requested):

docker/addons/cassandra-writer/subjects.toml
jonathandreyer: @drasko Something like below?
```
# If you want to listen on all subjects, just pass one element [">"], otherwise
# pass the list of subjects (e.g ["channel.<channel_ID>.messages", "channel.<channel_id>.messages/sub/topic/x", ...]).
```

drasko: `channel.<channel_id>.messages/sub/topic/x` should be `channel.<channel_id>.messages.sub.topic.x`. Also - do we use `messages` on NATS? Maybe, I am not sure, needs to be checked (I know it is used in adapter API, but I am not sure for NATS).

jonathandreyer: @drasko You are right, each topic are separated by dots. You are also right about `messages`. The explanation in PR of documentation repository is correct. The edited comment is:
```
# If you want to listen on all subjects, just pass one element [">"], otherwise
# pass the list of subjects (e.g ["channel.<channel_ID>", "channel.<channel_id>.sub.topic.x", ...]).
```
If it is OK for you, I will edit four `subjects.toml` files.

nmarcetic:
LGTM!

drasko (approved):
LGTM

blokovi (approved):

dborovcanin (approved):

drasko:
Merged! Thanks @jonathandreyer and @manuio

PR #1075, opened 2020-03-16
NOISSUE - Update state based on SenML time value

darkodraskovic:
Currently, every SenML record triggers new state save. Compare record timestamp with state timestamp to decide whether to save or update state.

### DIFF
--- a/twins/mongodb/states.go
+++ b/twins/mongodb/states.go
@@ -45,8 +45,7 @@ func (sr *stateRepository) Update(ctx context.Context, st twins.State) error {
 
 	filter := bson.D{{"id", st.ID}, {"twinid", st.TwinID}}
 	update := bson.D{{"$set", st}}
-	_, err := coll.UpdateOne(context.Background(), filter, update)
-	if err != nil {
+	if _, err := coll.UpdateOne(context.Background(), filter, update); err != nil {
 		return err
 	}
 
--- a/twins/mongodb/states.go
+++ b/twins/mongodb/states.go
@@ -43,8 +43,8 @@ func (sr *stateRepository) Save(ctx context.Context, st twins.State) error {
 func (sr *stateRepository) Update(ctx context.Context, st twins.State) error {
 	coll := sr.db.Collection(statesCollection)
 
-	filter := bson.D{{"id", st.ID}, {"twinid", st.TwinID}}
-	update := bson.D{{"$set", st}}
+	filter := bson.M{"id": st.ID, "twinid": st.TwinID}
+	update := bson.M{"$set": st}
 	if _, err := coll.UpdateOne(context.Background(), filter, update); err != nil {
 		return err
 	}
--- a/twins/mongodb/states.go
+++ b/twins/mongodb/states.go
@@ -39,6 +39,20 @@ func (sr *stateRepository) Save(ctx context.Context, st twins.State) error {
 	return nil
 }
 
+// Update persists the state
+func (sr *stateRepository) Update(ctx context.Context, st twins.State) error {
+	coll := sr.db.Collection(statesCollection)
+
+	filter := bson.D{{"id", st.ID}, {"twinid", st.TwinID}}
+	update := bson.D{{"$set", st}}
+	_, err := coll.UpdateOne(context.Background(), filter, update)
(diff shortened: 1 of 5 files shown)


twins/mongodb/states.go
```diff
@@ -39,6 +39,20 @@ func (sr *stateRepository) Save(ctx context.Context, st twins.State) error {
...
+// Update persists the state
+func (sr *stateRepository) Update(ctx context.Context, st twins.State) error {
+	coll := sr.db.Collection(statesCollection)
+
+	filter := bson.D{{"id", st.ID}, {"twinid", st.TwinID}}
+	update := bson.D{{"$set", st}}
+	_, err := coll.UpdateOne(context.Background(), filter, update)
+	if err != nil {
+		return err
+	}
```
manuio: Inline

twins/service.go
```diff
@@ -326,20 +330,29 @@ func prepareState(st *State, tw *Twin, rec senml.Record, msg *mainflux.Message)
 		}
 	}
 
-	save := false
+	recTime := (rec.BaseTime + rec.Time) * 1000
```
manuio: Comment or const for this 1000

manuio (changes requested):

twins/mongodb/states.go
```diff
@@ -39,6 +39,20 @@ func (sr *stateRepository) Save(ctx context.Context, st twins.State) error {
 	return nil
 }
 
+// Update persists the state
+func (sr *stateRepository) Update(ctx context.Context, st twins.State) error {
+	coll := sr.db.Collection(statesCollection)
+
+	filter := bson.D{{"id", st.ID}, {"twinid", st.TwinID}}
```
manuio: It's normal `twinid` ?

twins/service.go
```diff
@@ -326,20 +330,29 @@ func prepareState(st *State, tw *Twin, rec senml.Record, msg *mainflux.Message)
...
+	action := 0 // 0 - do nothing, 1 - update, 2 - save
```
manuio: You could add this action in the State to don't have to manipulate it as a separate variable

twins/mongodb/states.go
darkodraskovic: Yes. Automatically (un)marshalled (from)to `TwinID`.

twins/service.go
darkodraskovic: This is the action ON the state and not OF the state. State does not know whether it needs to be updated or saved and does not need to keep this information.

twins/service.go
```diff
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
```
manuio: This should be `mainflux/errors`

twins/mongodb/states.go
```diff
@@ -39,6 +39,19 @@ func (sr *stateRepository) Save(ctx context.Context, st twins.State) error {
...
+	update := bson.D{{"$set", st}}
```
golangcibot: composites: `go.mongodb.org/mongo-driver/bson/primitive.E` composite literal uses unkeyed fields (from `govet`)

drasko (dismissed):
LGTM

drasko:
I will approve this, but I think that real resolution of the state update should be solved on the architectural level. We can do it later, but it would be good to open an issue for this work.

twins/service.go
```diff
@@ -293,12 +306,18 @@ func (ts *twinsService) saveState(msg *mainflux.Message, id string) error {
 	}
 
 	for _, rec := range recs {
-		if save := prepareState(&st, &tw, rec, msg); !save {
+		action := prepareState(&st, &tw, rec, msg)
+		switch action {
+		case 0:
```
mteodor: you have constants for `action` , use them

darkodraskovic: Done.

manuio (dismissed):
LGTM

twins/mongodb/states.go
drasko: @darkodraskovic please address this issue

drasko (approved):
LGTM

dborovcanin (approved):

PR #1077, opened 2020-03-18
NOISSUE - Implement errors in Things service

blokovi:

### DIFF
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -35,8 +35,6 @@ do
     echo -ne "Waiting for gnatsd"
 done
 
-
-
 ###
 # Users
 ###
@@ -67,8 +65,10 @@ MF_MQTT_ADAPTER_LOG_LEVEL=info MF_THINGS_URL=localhost:8183 $BUILD_DIR/mainflux-
 ###
 MF_COAP_ADAPTER_LOG_LEVEL=info MF_COAP_ADAPTER_PORT=5683 MF_THINGS_URL=localhost:8183 $BUILD_DIR/mainflux-coap &
 
+###
+# AUTHN
+###
 MF_AUTHN_LOG_LEVEL=debug MF_AUTHN_HTTP_PORT=8189 MF_AUTHN_GRPC_PORT=8181 MF_AUTHN_DB_PORT=5432 MF_AUTHN_DB_USER=mainflux MF_AUTHN_DB_PASS=mainflux MF_AUTHN_DB=authn MF_AUTHN_SECRET=secret $BUILD_DIR/mainflux-authn &
-
 
 trap cleanup EXIT
 
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -35,6 +35,8 @@ do
     echo -ne "Waiting for gnatsd"
 done
 
+
+
 ###
 # Users
 ###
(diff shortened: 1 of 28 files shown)


things/postgres/things.go
```diff
@@ -50,10 +66,10 @@ func (tr thingRepository) Save(ctx context.Context, ths ...things.Thing) ([]thin
 			return []things.Thing{}, err
 		}
 
-		_, err = tx.NamedExecContext(ctx, q, dbth)
-		if err != nil {
+		_, err2 := tx.NamedExecContext(ctx, q, dbth)
```
drasko: err2? Why? Errors are re-assignable.

drasko (changes requested):

scripts/run.sh
```diff
@@ -35,6 +35,8 @@ do
     echo -ne "Waiting for gnatsd"
 done
 
+
```
drasko: Too much white lines.

things/api/things/http/transport.go
```diff
@@ -382,35 +383,62 @@ func encodeResponse(_ context.Context, w http.ResponseWriter, response interface
...
 		}
-	}
+		if errorVal.Msg() != "" {
+			json.NewEncoder(w).Encode(errorRes{Err: errorVal.Msg()})
+		}
+	default:
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+
+	// switch err {
```
drasko: Dead code

things/redis/things.go
```diff
@@ -16,6 +17,9 @@ const (
 	idPrefix  = "thing"
 )
 
+// ErrThingRedis indicates error in redis
+var ErrThingRedis = errors.New("Thing Redis error")
```
drasko: It would be good to know more about this error. Maybe also which operation failed.

drasko (changes requested):

things/postgres/things.go
blokovi: err and err2 are diffrent types here (generic error and errors.Error) so it can't reassign

things/postgres/channels.go
```diff
@@ -365,6 +384,7 @@ func (m *dbMetadata) Scan(value interface{}) error {
 
 	if err := json.Unmarshal(b, m); err != nil {
 		m = &dbMetadata{}
+		// return errors.Wrap(ErrScan, err)
```
manuio: rm this

things/postgres/channels.go
```diff
@@ -379,8 +399,10 @@ func (m dbMetadata) Value() (driver.Value, error) {
 
 	b, err := json.Marshal(m)
 	if err != nil {
+		// return nil, errors.Wrap(ErrValue, err)
```
manuio: and this

things/postgres/channels.go
```diff
@@ -379,8 +399,10 @@ func (m dbMetadata) Value() (driver.Value, error) {
...
 		return nil, err
 	}
+	// return b, errors.Wrap(ErrValue, err)
```
manuio: and this

things/postgres/things.go
```diff
@@ -11,9 +11,25 @@ import (
 
 	"github.com/gofrs/uuid"
 	"github.com/lib/pq" // required for DB access
+	"github.com/mainflux/mainflux/errors"
 	"github.com/mainflux/mainflux/things"
 )
 
+var (
+	// ErrSaveDb indicates error while saving to database
+	ErrSaveDb = errors.New("Save Thing to DB error")
```
manuio: I think that we don't use upercase in errors

things/postgres/things.go
manuio: Here you can probably inline the function and keep `err`. But if you can't do it for other errors then maybe just inline here and find something better than `err2`. Same for others `err2`

manuio (changes requested):

things/postgres/things.go
blokovi: You're right, I inlined here. There is just one more place where I couldn't do it, because result of function is used out of scope of `if` block. I renamed it from `err2` to `errdb`

things/api/things/http/endpoint_test.go
```diff
@@ -584,28 +584,28 @@ func TestViewThing(t *testing.T) {
 			id:     strconv.FormatUint(wrongID, 10),
 			auth:   token,
 			status: http.StatusNotFound,
-			res:    "",
+			res:    toJSON(errorRes{things.ErrNotFound.Error()}),
```
drasko, line 587: This repeating of the procedure where we create error struct, then convert it to JSON is the only suspicious part to me. I have an impression that this can be achieved more elegantly. Maybe by changing `TestViewThing{}`'s `res` field type, IDK.

@nmarcetic @manuio @dusanb94 @mteodor opinions?

manuio: @drasko I agree, if they are used in many places better create a variable for this.

blokovi:
We are introducing improved error package, so there is no need to merge Things service with current error implementation. In this PR we have both new error package and we implemented it in Things service: #1086

PR #1098, opened 2020-04-04
MF-1091 - Use channels. as broker prefix

manuio:

* Use "channels" instead of "channel" as subject prefix of the NATS message broker.

### DIFF
--- a/broker/nats.go
+++ b/broker/nats.go
@@ -28,9 +28,10 @@ type Nats interface {
 }
 
 const (
-	prefix = "channel"
+	chansPrefix = "channels"
+
 	// SubjectAllChannels define the subject to subscribe to all channels messages
-	SubjectAllChannels = "channel.>"
+	SubjectAllChannels = "channels.>"
 )
 
 var (
@@ -64,7 +65,7 @@ func (b broker) Publish(_ context.Context, _ string, msg Message) error {
 		return err
 	}
 
-	subject := fmt.Sprintf("%s.%s", prefix, msg.Channel)
+	subject := fmt.Sprintf("%s.%s", chansPrefix, msg.Channel)
 	if msg.Subtopic != "" {
 		subject = fmt.Sprintf("%s.%s", subject, msg.Subtopic)
 	}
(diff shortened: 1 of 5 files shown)


broker/nats.go
```diff
@@ -64,9 +65,9 @@ func (b broker) Publish(_ context.Context, _ string, msg Message) error {
 		return err
 	}
 
-	subject := fmt.Sprintf("%s.%s", prefix, msg.Channel)
+	subject := fmt.Sprintf("%s.%s", chansPrefix, msg.Channel)
 	if msg.Subtopic != "" {
-		subject = fmt.Sprintf("%s.%s", subject, msg.Subtopic)
+		subject = fmt.Sprintf("%s.%s", chansPrefix, msg.Subtopic)
```
drasko: Looks like we are changing the logic - we used previously defined `subject` (on the line above), now we are using chan prefix, and this looks incorrect - as subtopic is actually added to a topic, not to a prefix.

drasko (changes requested):

drasko:
This is an API breaking change for all applications that are using data directly from the NATS, and this should be documented for the new release.

manuio:
@drasko docs updated: https://github.com/mainflux/docs/pull/22

drasko (approved):
LGTM

broker/nats.go
```diff
@@ -28,9 +28,10 @@ type Nats interface {
 }
 
 const (
-	prefix = "channel"
+	chansPrefix = "channels"
+
 	// SubjectAllChannels define the subject to subscribe to all channels messages
-	SubjectAllChannels = "channel.>"
+	SubjectAllChannels = "channels.>"
```
nmarcetic, line 34: Can you please describe this PR, the purpose and reason.

nmarcetic (changes requested):

broker/nats.go
drasko: @nmarcetic this is more or less evident. `channels.<chan_id>` is to make it consistent with the public API (which is `channels/<chan_id>`, and not `channel/<chan_id>`.

manuio: In case I updated the description of the PR

dborovcanin (approved):

nmarcetic (approved):

PR #1106, opened 2020-04-08
MF-1078 - Add timestamp to published messages and use it in Transformer

### LINKED ISSUE #1069: Adapters should insert timestamp in the case it is missing
**FEATURE REQUEST**

SenML message can be sent without timestamp, because:
- Low-power devices might not have RTC
- Clocks are not synchronized via NTP
- Some other reason

Protocol adapters - who are the first recepients of the message - should insert a timestamp in the case it is missing.

Even more - we should consider of an another fiels in Mainflux `Message{}` like `received`.

Writers will then use this timestamp, and this will help with implementing queue and workers in https://github.com/mainflux/mainflux/issues/1068 and assuring that messages are written in order, although processed by several parallel writers.

Must-have

### LINKED ISSUE #1078: Replace zero time with "now"
**FEATURE REQUEST**

Mainflux Transformer does not modify the zero time in SenML messages. As the timestamp is usually unique for the time-series databases and using time-series DB for storing SenML just makes sense, zero timestamps should be replaced with "now". The [spec](https://tools.ietf.org/html/rfc8428#section-4.5.3) is ambiguous here, and for that reason, SenML modification won't be included in SenML lib itself, but rather in Transformer.
 
manuio:

Resolves #1078 
Resolves #1069


broker/nats.go
```diff
@@ -59,7 +60,10 @@ func New(url string) (Nats, error) {
 	}, nil
 }
 
-func (b broker) Publish(_ context.Context, _ string, msg Message) error {
+func (b broker) Publish(_ context.Context, token string, msg Message) error {
+	// Add publisher timestamp
+	msg.Time = time.Now().Unix()
```
dborovcanin: This should be added at the time of message creation. I agree that it's simpler to create timestamp here rather than in each adapter separately, but the Publisher should not modify the message.

dborovcanin (changes requested):

broker/nats.go
nmarcetic: I don't like the name also `Time`, if you follow @dusanb94 logic that adapter should add this time on msg creation then `Created` timestamp is more logical to me. Even if you put `Published` timestamp here as @manuio did, for me it is OK, even if I agree with @dusanb94, its a bit dirty if we modify msg here, but not to dirty for me. But `Time` timestamp says nothing to me :)

drasko: Messages must be timestamped in protocol adapters, not here.

Agreed with @nmarcetic that `time` is a bad choice for a field name. I would expect `timestamp`, as traditionally used `created` in this context is not really adequate.

drasko (dismissed):
LGTM

drasko:
Great @manuio, this is exactly what we needed I think!

coap/adapter.go
```diff
@@ -32,17 +32,7 @@ const (
...
-	// ErrFailedMessagePublish indicates that message publishing failed.
-	ErrFailedMessagePublish = errors.New("failed to publish message")
-
-	// ErrFailedSubscription indicates that client couldn't subscribe to specified channel.
-	ErrFailedSubscription = errors.New("failed to subscribe to a channel")
-
-	// ErrFailedConnection indicates that service couldn't connect to message broker.
-	ErrFailedConnection = errors.New("failed to connect to message broker")
-)
+var errBadOption = errors.New("bad option")
```
golangcibot: var `errBadOption` is unused (from `unused`)

drasko:
Finally after consultation with @nmarcetic and @dusanb94 there is a prevailing opinion that this field should be called `created` instead of `timestamp`.

drasko (dismissed):
LGTM

http/adapter.go
```diff
@@ -44,5 +45,7 @@ func (as *adapterService) Publish(ctx context.Context, token string, msg broker.
 	}
 	msg.Publisher = thid.GetValue()
 
+	msg.Created = time.Now().Unix()
```
dborovcanin: I think that this field should be created when we create a Message. In this case, that would be [here](https://github.com/mainflux/mainflux/blob/57f7ee25950548aa4a1e4cfddd35db2f7e5ed051/http/api/transport.go#L112).
```go
	msg := broker.Message{
		Protocol:    protocol,
		ContentType: ct,
		Channel:     chanID,
		Subtopic:    subtopic,
                Created:     time.Now().Unix(),
		Payload:     payload,
	}
```

ws/adapter.go
```diff
@@ -93,15 +94,9 @@ func New(broker broker.Nats, log logger.Logger) Service {
...
-	if err := as.broker.Publish(ctx, token, msg); err != nil {
-		switch err {
-		case nats.ErrConnectionClosed, nats.ErrInvalidConnection:
-			return ErrFailedConnection
-		default:
-			return ErrFailedMessagePublish
-		}
-	}
-	return nil
+	msg.Created = time.Now().Unix()
```
dborovcanin: Same as in the case of HTTP adapter.

dborovcanin (changes requested):

http/adapter.go
manuio: I think that the message creation should be done in service layer like it's done in most of the services. Plus, here we also set the publisher.

dborovcanin: We set publisher here because it's not available before. If we can set fields at the time of creation, I don't see any reason not to do so. We're doing that for all the other fields.

manuio: @drasko @nmarcetic what do you think about?

nmarcetic: I don't have strong preferences, really. Maybe I agree with @dusanb94 more, I mean why not add it in the time of creation if we can, it's just easier. I don't understand @manuio 
`I think that the message creation should be done in service layer like it's done in most of the services` Its not like that now, you want to remove message creation from transport and put it in service?

manuio: @nmarcetic Yes I think that `broker.Message{}` should be created in the service layer and that all the data should be passed using `publishReq` (Currently we create the message and pass it as publishReq parameter). Also it's consistent with MQTT, Lora and OPC-UA where there is no transport layer for messaging.

nmarcetic: @manuio I understand now. I agree with you, but :) All this data that makes message actually is decoded on transport layer here (it's go-kit specific, can't compare with other no go-kit services) and this data is made from things like request body, request URL etc... everything we have in the transport layer. If we do propagation of request to service, it's not really clear to me. I really do agree with you, but think this is really specific and its to messy to align it with other no go-kit services or just not HTTP oriented.

drasko (dismissed):
LGTM

http/api/transport.go
```diff
@@ -115,6 +116,7 @@ func decodeRequest(ctx context.Context, r *http.Request) (interface{}, error) {
 		Channel:     chanID,
 		Subtopic:    subtopic,
 		Payload:     payload,
+		Created:     time.Now().Unix(),
```
blokovi: This value will be in whole seconds. And I think that I had problem when I sent more message with same time that they were overwritten in time series databases.
Maybe consider to make this float here: `float64(time.Now().UnixNano()) / float64(1e9)` and in proto file also declare is as `double` to make this value more precise.

drasko: I agree on this!

I think we even set up InfluxDB by default on microsecond precision...

I would suggest using Unix time (integer), but as precise as possible (`UnixNano()` for example)

nmarcetic (dismissed):

drasko (dismissed):
Ummm... Don't you have to add `Created` field to the Mainflux [`Message{}`](https://github.com/mainflux/mainflux/blob/master/broker/message.proto#L8)?

manuio:
@drasko Isn't done?

broker/message.proto
```diff
@@ -12,4 +12,5 @@ message Message {
 	string protocol    = 4;
 	string contentType = 5;
 	bytes  payload     = 6;
+	int64  created     = 7;
```
drasko: Instead of moving Unix nanoseconds though this field, I would suggest protobuf v3 [Timestamp](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/timestamp), which is self-documenting and can be decoded in all protobuf-supported languages.

transformers/senml/transformer.go
```diff
@@ -39,14 +39,20 @@ func (n transformer) Transform(msg broker.Message) (interface{}, error) {
 
 	msgs := make([]Message, len(normalized.Records))
 	for i, v := range normalized.Records {
+		// Use reception timestamp if SenML messsage Time is missing
+		time := v.Time
+		if time == 0 {
+			time = float64(msg.Created) / float64(1e9)
```
drasko: Is this time in seconds (i.e. is this what Transformer uses)? If yes, a comment that explains this should be placed in front of this line.

drasko (changes requested):

coap/api/transport.go
```diff
@@ -237,6 +238,10 @@ func receive(svc coap.Service, msg *gocoap.Message) *gocoap.Message {
 		ContentType: ct,
 		Protocol:    protocol,
 		Payload:     msg.Payload,
+		Created: &timestamp.Timestamp{
```
drasko: I think you should use a function for this: https://github.com/golang/protobuf/blob/84668698ea25b64748563aa20726db66a6b8d299/ptypes/timestamp.go#L113

drasko: Something like `timestamp.TimestampProto(time.Now())`

drasko (changes requested):

Makefile
```diff
@@ -71,7 +71,7 @@ test:
 
 proto:
 	protoc --gofast_out=plugins=grpc:. *.proto
-	protoc --gofast_out=plugins=grpc:. broker/*.proto
+	protoc --gogo_out=plugins=grpc:. broker/*.proto
```
blokovi, line 74: Just a reminder if it's necessary to update this in docs also:
https://github.com/mainflux/docs/blob/master/docs/dev-guide.md#protobuf

blokovi (approved):

Makefile
drasko: @blokovi please send the PR

blokovi: Ok, after merging this PR

nmarcetic (approved):

Makefile
manuio: right!

manuio: @blokovi https://github.com/mainflux/docs/pull/25

drasko (approved):
LGTM

blokovi (approved):

broker/message.proto
```diff
@@ -3,13 +3,15 @@
...
-	string protocol    = 4;
-	string contentType = 5;
-	bytes  payload     = 6;
+	string                    channel     = 1;
+	string                    subtopic    = 2;
+	string                    publisher   = 3;
+	string                    protocol    = 4;
+	string                    contentType = 5;
+	bytes                     payload     = 6;
+	google.protobuf.Timestamp created     = 7;
```
dborovcanin, line 16: While I agree with what @drasko said (Timestamp is self-explanatory), I'm not sure that we should use it. Simple timestamp (Unix nanoseconds or milliseconds) is just fine, I think. The benefits of Timestamp are more user-friendly JSON encoding, which translates to a string rather than a number. This can do more harm than good in our case. Also, there is a burden of the imported package. Timestamp itself does not explain the precision, which kinda undoes self-explanatory argument.

PR #1110, opened 2020-04-13
NOISSUE - Create func to encode SDK errors

manuio:

### DIFF
--- a/sdk/go/sdk.go
+++ b/sdk/go/sdk.go
@@ -240,3 +240,18 @@ func createURL(baseURL, prefix, endpoint string) string {
 
 	return fmt.Sprintf("%s/%s/%s", baseURL, prefix, endpoint)
 }
+
+func encodeError(statusCode int) error {
+	switch statusCode {
+	case http.StatusBadRequest:
+		return ErrInvalidArgs
+	case http.StatusForbidden:
+		return ErrUnauthorized
+	case http.StatusNotFound:
+		return ErrNotFound
+	case http.StatusConflict:
+		return ErrConflict
+	default:
+		return nil
+	}
+}
--- a/sdk/go/things.go
+++ b/sdk/go/things.go
@@ -35,14 +35,10 @@ func (sdk mfSDK) CreateThing(thing Thing, token string) (string, error) {
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusCreated {
-		switch resp.StatusCode {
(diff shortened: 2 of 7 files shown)


drasko (dismissed):
LGTM

sdk/go/sdk.go
```diff
@@ -240,3 +240,18 @@ func createURL(baseURL, prefix, endpoint string) string {
 
 	return fmt.Sprintf("%s/%s/%s", baseURL, prefix, endpoint)
 }
+
+func encodeError(statusCode int) error {
```
mteodor, line 244: do we need this function isn't `http.StatusText(statusCode)` enough?

sdk/go/things.go
```diff
@@ -192,17 +176,13 @@ func (sdk mfSDK) Thing(id, token string) (Thing, error) {
...
-		default:
-			return Thing{}, ErrFetchFailed
+		if err := encodeError(resp.StatusCode); err != nil {
+			return Thing{}, err
 		}
+		return Thing{}, ErrFetchFailed
 	}
 
-	var t Thing
+	t := Thing{}
```
nmarcetic: I think its cleaner if you leave it as before `var t Thing` instead of a short declaration. You don't need to declare anything here, zero values are just fine.

sdk/go/users.go
```diff
@@ -61,12 +57,10 @@ func (sdk mfSDK) User(token string) (User, error) {
...
-			return User{}, ErrUnauthorized
-		default:
-			return User{}, ErrFetchFailed
+		if err := encodeError(resp.StatusCode); err != nil {
+			return User{}, err
 		}
+		return User{}, ErrFetchFailed
 	}
 
 	var u User
```
nmarcetic, line 66: Here you left `var`, no short declaration (which is better  IMHO).

sdk/go/users.go
```diff
@@ -97,14 +91,10 @@ func (sdk mfSDK) CreateToken(user User) (string, error) {
...
-			return "", ErrUnauthorized
-		default:
-			return "", ErrFailedCreation
+		if err := encodeError(resp.StatusCode); err != nil {
+			return "", err
 		}
+		return "", ErrFailedCreation
 	}
 
 	var t tokenRes
```
nmarcetic: Here also `var`

nmarcetic (approved):

mteodor (approved):

PR #1112, opened 2020-04-14
NOISSUE - Add `Expire` to running containers via dockertest

sprql:
If a test failed by a panic, purging a container would fail too.
To prevent it, we can use `Expire()` for a container. In this case, docker daemon will auto-remove it after a specified time.

### DIFF
--- a/authn/postgres/setup_test.go
+++ b/authn/postgres/setup_test.go
@@ -18,6 +18,7 @@ import (
 )
 
 const wrong string = "wrong-value"
+const expInterval = 60 // in seconds
 
 var db *sqlx.DB
 
@@ -35,6 +36,11 @@ func TestMain(m *testing.M) {
 	container, err := pool.Run("postgres", "10.2-alpine", cfg)
 	if err != nil {
 		log.Fatalf("Could not start container: %s", err)
+	}
+
+	// Force remove the container after the interval
+	if err := container.Expire(expInterval); err != nil {
+		log.Printf("Could not expire container: %s", err)
 	}
 
 	port := container.GetPort("5432/tcp")
--- a/bootstrap/postgres/setup_test.go
+++ b/bootstrap/postgres/setup_test.go
@@ -16,8 +16,9 @@ import (
 )
 
 const (
-	wrongID    = "0"
-	wrongValue = "wrong-value"
+	wrongID     = "0"
+	wrongValue  = "wrong-value"
+	expInterval = 60 // in seconds
 )
 
 var (
@@ -39,6 +40,11 @@ func TestMain(m *testing.M) {
 	container, err := pool.Run("postgres", "10.2-alpine", cfg)
 	if err != nil {
 		log.Fatalf("Could not start container: %s", err)
+	}
+
+	// Force remove the container after the interval
+	if err := container.Expire(expInterval); err != nil {
+		log.Printf("Could not expire container: %s", err)
 	}
 
 	port := container.GetPort("5432/tcp")
--- a/bootstrap/postgres/setup_test.go
+++ b/bootstrap/postgres/setup_test.go
@@ -15,11 +15,7 @@ import (
 	dockertest "gopkg.in/ory/dockertest.v3"
 )
 
-const (
-	wrongID     = "0"
-	wrongValue  = "wrong-value"
-	expInterval = 60 // in seconds
-)
+const expInterval = 60 // in seconds
 
 var (
 	testLog, _ = logger.New(os.Stdout, logger.Info.String())
--- a/bootstrap/redis/producer/setup_test.go
+++ b/bootstrap/redis/producer/setup_test.go
@@ -13,11 +13,7 @@ import (
 	dockertest "gopkg.in/ory/dockertest.v3"
 )
 
-const (
-	wrongID     = 0
-	wrongValue  = "wrong-value"
-	expInterval = 60 // in seconds
-)
+const expInterval = 60 // in seconds
 
 var redisClient *redis.Client
 
--- a/bootstrap/redis/producer/setup_test.go
+++ b/bootstrap/redis/producer/setup_test.go
@@ -14,8 +14,9 @@ import (
 )
 
 const (
-	wrongID    = 0
-	wrongValue = "wrong-value"
+	wrongID     = 0
+	wrongValue  = "wrong-value"
+	expInterval = 60 // in seconds
 )
 
 var redisClient *redis.Client
@@ -29,6 +30,11 @@ func TestMain(m *testing.M) {
 	container, err := pool.Run("redis", "5.0-alpine", nil)
 	if err != nil {
 		log.Fatalf("Could not start container: %s", err)
+	}
+
+	// Force remove the container after the interval
+	if err := container.Expire(expInterval); err != nil {
+		log.Printf("Could not expire container: %s", err)
 	}
 
 	if err := pool.Retry(func() error {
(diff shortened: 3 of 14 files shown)


authn/postgres/setup_test.go
```diff
@@ -37,6 +37,7 @@ func TestMain(m *testing.M) {
 		log.Fatalf("Could not start container: %s", err)
 	}
 
+	container.Expire(60) // force remove the container in 60 seconds
```
golangcibot: Error return value of `container.Expire` is not checked (from `errcheck`)

writers/postgres/setup_test.go
```diff
@@ -43,6 +43,7 @@ func TestMain(m *testing.M) {
 		log.Fatalf("Could not start container: %s", err)
 	}
 
+	container.Expire(60) // force remove the container in 60 seconds
```
golangcibot: Error return value of `container.Expire` is not checked (from `errcheck`)

writers/influxdb/setup_test.go
```diff
@@ -29,6 +29,7 @@ func TestMain(m *testing.M) {
 		testLog.Error(fmt.Sprintf("Could not start container: %s", err))
 	}
 
+	container.Expire(60) // force remove the container in 60 seconds
```
golangcibot: Error return value of `container.Expire` is not checked (from `errcheck`)

authn/postgres/setup_test.go
```diff
@@ -37,6 +37,12 @@ func TestMain(m *testing.M) {
 		log.Fatalf("Could not start container: %s", err)
 	}
 
+	// force remove the container in 60 seconds
+	err = container.Expire(60)
+	if err != nil {
```
drasko: Embed `err` definition in `if`:

```go
if err := container.Expire(expInterval); err != nil {
...
}
```

authn/postgres/setup_test.go
```diff
@@ -37,6 +37,12 @@ func TestMain(m *testing.M) {
 		log.Fatalf("Could not start container: %s", err)
 	}
 
+	// force remove the container in 60 seconds
+	err = container.Expire(60)
```
drasko: Consts should be declared on the top of the file.

drasko (changes requested):

writers/postgres/setup_test.go
```diff
@@ -18,8 +18,9 @@ import (
 )
 
 const (
-	wrongID    = "0"
-	wrongValue = "wrong-value"
+	wrongID     = "0"
```
golangcibot: `wrongID` is unused (from `varcheck`)

writers/postgres/setup_test.go
```diff
@@ -18,8 +18,9 @@ import (
...
+	wrongValue  = "wrong-value"
```
golangcibot: `wrongValue` is unused (from `varcheck`)

twins/mongodb/setup_test.go
```diff
@@ -16,8 +16,9 @@ import (
 )
 
 const (
-	wrongID    = "0"
-	wrongValue = "wrong-value"
+	wrongID     = "0"
```
golangcibot: `wrongID` is unused (from `varcheck`)

things/postgres/setup_test.go
```diff
@@ -18,8 +18,9 @@ import (
 )
 
 const (
-	wrongID    = "0"
-	wrongValue = "wrong-value"
+	wrongID     = "0"
```
golangcibot: `wrongID` is unused (from `varcheck`)

bootstrap/postgres/setup_test.go
```diff
@@ -16,8 +16,9 @@ import (
 )
 
 const (
-	wrongID    = "0"
-	wrongValue = "wrong-value"
+	wrongID     = "0"
+	wrongValue  = "wrong-value"
```
golangcibot: `wrongValue` is unused (from `varcheck`)

bootstrap/redis/producer/setup_test.go
```diff
@@ -14,8 +14,9 @@ import (
 )
 
 const (
-	wrongID    = 0
-	wrongValue = "wrong-value"
+	wrongID     = 0
+	wrongValue  = "wrong-value"
```
golangcibot: `wrongValue` is unused (from `varcheck`)

things/redis/setup_test.go
```diff
@@ -14,8 +14,9 @@ import (
 )
 
 const (
-	wrongID    = 0
-	wrongValue = "wrong-value"
+	wrongID     = 0
```
golangcibot: `wrongID` is unused (from `varcheck`)

bootstrap/redis/producer/setup_test.go
```diff
@@ -14,8 +14,8 @@ import (
 )
 
 const (
-	wrongID    = 0
-	wrongValue = "wrong-value"
+	wrongID     = 0
```
golangcibot: `wrongID` is unused (from `varcheck`)

sprql:
@drasko  I just noticed this [GolangCI.com is closing](https://medium.com/golangci/golangci-com-is-closing-d1fc1bd30e0e), maybe it would be useful for you.

drasko:
> @drasko, @mandric I just noticed this [GolangCI.com is closing](https://medium.com/golangci/golangci-com-is-closing-d1fc1bd30e0e), maybe it would be useful for you.

Indeed. It's a pity, very nice and useful tool.

manuio:
@mandric This one looks ok to me but please update your branch

manuio (approved):
LGTM

authn/postgres/setup_test.go
```diff
@@ -18,6 +18,7 @@ import (
 )
 
 const wrong string = "wrong-value"
+const expInterval = 60 // in seconds
```
drasko, line 21: Just a curiosity - how did we get to this interval and is it enough for test to be run in all cases?

sprql: This is a good question because it depends on each case, on the other hand, you can set the value with a large margin. I just had to start with a certain value and it seemed to me that a minute would be fine. Maybe you have a suggestion?

drasko:
TBH I am not sure that we need this. Actually - this 60 sec interval makes me uncertain and I think that this can bring us more problems then benefits.

These tests are usually launched by CI on some temporary VMs, and I am not really aware that our tests panic.

process is explained [here](https://github.com/ory/dockertest#removing-old-containers), but I feel uncomfortable limiting docker test execution to 60 sec.

@dusanb94 @nmarcetic @manuio opinions?

sprql:
@drasko, I agree with your concerns, this approach is most helpful to the local development process when some feature is created and tests can panic from time to time, which leads to creating a lot of zombie containers. In theory, we can add a temporary line with `container.Expire` while developing.

nmarcetic (approved):

dborovcanin (commented):
If the code panics (tests included), I would say that having dangling containers are the least problem. I also agree that setting this expiration to something like 1h or 2h will be safe in almost all the cases. What concerns me is that this can cause very nasty bugs that are hard to discover. If there is there a way to notify the user if the container is expired, I vote for accepting this PR with an increased expiration period. Otherwise, I would avoid this change for the aforementioned reasons.

Also, the warning at the beginning of the test executions would be fine.

drasko:
Based on the prevailing opinion that this might introduce more problems (because of mysterious test killing on CI machines) then benefits (because if tests are done on the local machine, this is done by a developer, and he will know how to deal with dangling containers), I will close this PR.

@sprql thanks for the proposed PR in any case, this stays as a reference/doc for the others who might run into the similar problems.

PR #1118, opened 2020-04-15
MF-1115 - Improve the SDK error encoding

manuio:

### DIFF
--- a/sdk/go/bootstrap.go
+++ b/sdk/go/bootstrap.go
@@ -61,7 +61,7 @@ func (sdk mfSDK) AddBootstrap(key string, cfg BoostrapConfig) (string, error) {
 		if err := encodeError(resp.StatusCode); err != nil {
 			return "", err
 		}
-		return "", ErrFailedCreation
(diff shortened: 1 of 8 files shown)


drasko (approved):
LGTM

sdk/go/bootstrap.go
```diff
@@ -61,7 +61,7 @@ func (sdk mfSDK) AddBootstrap(key string, cfg BoostrapConfig) (string, error) {
 		if err := encodeError(resp.StatusCode); err != nil {
 			return "", err
```
nmarcetic, line 62: What if there is an error we don't need to wrap anything? I am not sure how SDK error looks like

manuio: When there is an HTTP error with a mainflux equivalent we only use the custom error. Otherwise we wrap the HTTP message with the custom mainflux error for this kind of function.

mteodor (approved):

dborovcanin (approved):

PR #1121, opened 2020-04-17
MF-1088 - Remove message payload content type

### LINKED ISSUE #1088: Remove `/ct` content-type topic extension
**FEATURE REQUEST**

Currently MQTT has a topic extension of `/ct`, which signifies content-type. This was needed because WS adapter had to know in advance which type of socket needs to be opened, and then influenced API of all other adapters.

It would be better approach to offload this content-type awareness to an application, and make the infrastructure (Mainflux) content-type agnostic.

After https://github.com/mainflux/mainflux/issues/1087 this will be possible and should be implemented.

Must-have

manuio:

Resolves #1088


drasko (approved):
This looks good, but it will be mandatory that @dusanb94 and @nmarcetic review as well.

Docs will need update, as this breaks the API.

drasko (dismissed):
LGTM

transformers/senml/transformer.go
```diff
@@ -20,20 +27,19 @@ var formats = map[string]senml.Format{
...
 // New returns transformer service implementation for SenML messages.
-func New() transformers.Transformer {
-	return transformer{}
+func New(contentType string) transformers.Transformer {
+	return transformer{
+		format: formats[contentType],
+	}
 }
 
 func (n transformer) Transform(msg broker.Message) (interface{}, error) {
```
dborovcanin: You can rename `func (n transformer) Transform...` to `func (t transformer) Transform...`. Weird receiver name `n` is a legacy from `Normalizer`.

transformers/senml/transformer.go
```diff
@@ -20,20 +27,19 @@ var formats = map[string]senml.Format{
...
+	format senml.Format
+}
 
 // New returns transformer service implementation for SenML messages.
-func New() transformers.Transformer {
-	return transformer{}
+func New(contentType string) transformers.Transformer {
+	return transformer{
+		format: formats[contentType],
+	}
```
dborovcanin: It makes sense to have a default format (also, please rename `contentType` to `contentFormat`):
```go
format, ok := formats[contentFormat]
if !ok {
    format = formats[JSON]
}

return transformer {
    format: format,
} 
```

dborovcanin (dismissed):

nmarcetic (dismissed):

drasko (approved):
LGTM

mteodor (approved):

PR #1140, opened 2020-04-23
NOISSUE - Use `StatusUnauthorized` instead of `StatusForbidden`

sprql:
Use `401 Unauthorized` instead of `403 Forbidden` for errors related to 
authentication. 

According to [https://tools.ietf.org/html/rfc6750#section-3](https://tools.ietf.org/html/rfc6750#section-3)

### DIFF
--- a/authn/api/http/endpoint_test.go
+++ b/authn/api/http/endpoint_test.go
@@ -121,9 +121,10 @@ func TestIssue(t *testing.T) {
 			status: http.StatusBadRequest,
 		},
 		{
-			desc: "issue user key wrong content type",
-			req:  toJSON(uk),
-			ct:   "", token: userKey.Secret,
+			desc:   "issue user key wrong content type",
+			req:    toJSON(uk),
+			ct:     "",
+			token:  userKey.Secret,
 			status: http.StatusUnsupportedMediaType,
 		},
 		{
@@ -134,11 +135,11 @@ func TestIssue(t *testing.T) {
 			status: http.StatusUnsupportedMediaType,
 		},
 		{
-			desc:   "issue key unauthorized",
+			desc:   "issue key with an unauthorized token",
 			req:    toJSON(ak),
 			ct:     contentType,
 			token:  "wrong",
-			status: http.StatusForbidden,
+			status: http.StatusUnauthorized,
 		},
 		{
 			desc:   "issue recovery key with empty token",
@@ -217,10 +218,10 @@ func TestRetrieve(t *testing.T) {
 			status: http.StatusNotFound,
 		},
 		{
-			desc:   "retrieve a key unauthorized",
+			desc:   "retrieve an unauthorized token",
 			id:     k.ID,
 			token:  "wrong",
-			status: http.StatusForbidden,
+			status: http.StatusUnauthorized,
 		},
 	}
 
@@ -269,10 +270,10 @@ func TestRevoke(t *testing.T) {
 			status: http.StatusNoContent,
 		},
 		{
-			desc:   "revoke a key unauthorized",
+			desc:   "revoke an unauthorized token",
 			id:     k.ID,
 			token:  "wrong",
-			status: http.StatusForbidden},
+			status: http.StatusUnauthorized},
 	}
 
 	for _, tc := range cases {
--- a/authn/api/http/transport.go
+++ b/authn/api/http/transport.go
@@ -103,7 +103,9 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 	switch {
 	case errors.Contains(err, authn.ErrMalformedEntity):
 		w.WriteHeader(http.StatusBadRequest)
-	case errors.Contains(err, authn.ErrUnauthorizedAccess):
+	case errors.Contains(err, authn.ErrUnauthenticated):
+		w.WriteHeader(http.StatusUnauthorized)
+	case errors.Contains(err, authn.ErrUnauthorized):
 		w.WriteHeader(http.StatusForbidden)
 	case errors.Contains(err, authn.ErrNotFound):
 		w.WriteHeader(http.StatusNotFound)
--- a/authn/swagger.yaml
+++ b/authn/swagger.yaml
@@ -54,8 +54,10 @@ paths:
             $ref: "#/definitions/Key"
         400:
           description: Failed due to malformed query parameters.
-        403:
+        401:
           description: Missing or invalid access token provided.
+        403:
+          description: Unauthorized access.
         500:
           $ref: "#/responses/ServiceError"
     delete:
@@ -74,8 +76,10 @@ paths:
       responses:
         204:
           description: Key revoked.
-        403:
+        401:
           description: Missing or invalid access token provided.
+        403:
+          description: Unauthorized access.
         500:
           $ref: "#/responses/ServiceError"
 
--- a/bootstrap/api/transport.go
+++ b/bootstrap/api/transport.go
@@ -262,7 +262,9 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 			w.WriteHeader(http.StatusBadRequest)
 		case errors.Contains(errorVal, bootstrap.ErrNotFound):
 			w.WriteHeader(http.StatusNotFound)
-		case errors.Contains(errorVal, bootstrap.ErrUnauthorizedAccess):
+		case errors.Contains(errorVal, bootstrap.ErrUnauthenticated):
+			w.WriteHeader(http.StatusUnauthorized)
+		case errors.Contains(err, bootstrap.ErrUnauthorized):
 			w.WriteHeader(http.StatusForbidden)
 		case errors.Contains(errorVal, bootstrap.ErrConflict):
 			w.WriteHeader(http.StatusConflict)
--- a/readers/api/transport.go
+++ b/readers/api/transport.go
@@ -27,10 +27,11 @@ const (
 )
 
 var (
-	errInvalidRequest     = errors.New("received invalid request")
-	errUnauthorizedAccess = errors.New("missing or invalid credentials provided")
-	auth                  mainflux.ThingsServiceClient
-	queryFields           = []string{"subtopic", "publisher", "protocol", "name", "value", "v", "vs", "vb", "vd"}
+	errInvalidRequest  = errors.New("received invalid request")
+	errUnauthenticated = errors.New("missing credentials provided")
+	errUnauthorized    = errors.New("unauthorized access")
+	auth               mainflux.ThingsServiceClient
+	queryFields        = []string{"subtopic", "publisher", "protocol", "name", "value", "v", "vs", "vb", "vd"}
 )
 
 // MakeHandler returns a HTTP handler for API endpoints.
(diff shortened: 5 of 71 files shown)


sprql:
As far as I understand, it's better to use 401 instead of 403 for authentication-related errors, or maybe I'm missing something.

manuio:
@sprql sounds good. I would maybe keep forbidden for readers because they use a thing key.

sprql:
@manuio good point, I took this into account and updated the code

drasko:
https://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses

manuio:
@sprql After consulting with maintainers we think that this is correct. Can you update the swaggers, please?

authn/api/http/endpoint_test.go
```diff
@@ -219,7 +219,7 @@ func TestRetrieve(t *testing.T) {
 			desc:   "retrieve a key unauthorized",
```
manuio: If you have time I would appreciate that you also fix this logs. It should be `retrieve a token unauthorized` or even `retrieve an unauthorized token`

readers/api/transport.go
```diff
@@ -28,7 +28,8 @@ const (
 
 var (
 	errInvalidRequest     = errors.New("received invalid request")
-	errUnauthorizedAccess = errors.New("missing or invalid credentials provided")
+	errUnauthorizedAccess = errors.New("missing credentials provided")
+	errAccessDenied       = errors.New("invalid credentials provided")
```
drasko: This should be kept consistent all over the code. If there is a difference between missing and invalid credentials (AuthN/AuthZ), then it should be detected in other places, and consequently 401/403 (both of them) codes should be used.

drasko (changes requested):

readers/api/transport.go
sprql: There are 3 cases: 

- Missing token — 401
- Expired token (or maybe invalid by another reason) — 401
- Valid token, but access to some resource with this token is denied — 403

In this example, the current error message ("invalid credentials provided") maybe is a little bit confusing and should be changed to something else (like "permission denied")

Also maybe I should rename `errAccessDenied` to `errPermissionDenied`

authn/api/http/endpoint_test.go
sprql: Ok, I'll do

readers/api/transport.go
drasko: > There are 3 cases:
>
>Missing token — 401
>Expired token (or maybe invalid by another reason) — 401
>Valid token, but access to some resource with this token is denied — 403

@sprql exactly - well understood. Name of the error does not bother me much - `errAccessDenied` is OK and it relates to `403` rather than `401`, i.e. it is self-documenting enough.

What I wanted to check with you is did you cover all the code code base and identified where one of these 3 cases might appear and well differentiated between them - because here I can see `403` practically appearing just at one place in the whole Mainflux, according to your changes.

I did not review whole PR in details - it's bit difficult to review, this is why I am asking (looks to me you know your way areound Mainflux now and you know what you are doing, which is very good).

Anyway - thanks for this PR, it is very nice and think that you are really on a good track here for a nice improvement. Just please do not forget to update Swagger files as well.

@nmarcetic and @dusanb94 should review this one as well.

dborovcanin: I agree with this approach. This will later translate to handling errors from AuthN and AuthZ services (at the moment, only AuthN exists), where errors from AuthN will cause `Forbidden`, and errors from AuthZ will be `Unauthorized`.

drasko: Great, we have general alignment of maintainers then.

@sprql please double-check the PR, ad final touches prepare it for meging. Please do not forget to add Swagger changes.

After that PR will be merged.

sprql: @drasko, currently, the code mainly uses the `errUnauthorizedAccess`, both for AuthN & AuthZ errors, therefore, we will need to enter a new error type in these places. 

One option is to use error `errUnauthorizedAccess` for AuthN, and enter `errPermissionDenied` for AuthZ. But 'errUnauthorizedAccess' is misleading because of its name. 

Alternatively, we can enter `errUnauthenticated' for AuthN errors (like in gRPC).

In other words, the following options are possible:
`errUnauthorizedAccess` (401), `errPermissionDenied` (403)
or `errUnauthenticated` (401), `errUnauthorizedAccess` (403)
or `errUnauthenticated` (401), `errPermissionDenied` (403)

What do you think about it?

drasko: Good question... I would avoid option 2, as this can be misleading (`errUnauthorizedAccess` might be related to 401 when here should be related to 403).

Between options 1 and 3, probably 3 is the least ambigous, although we could probably understant 1 as well now that we know what this is about...

I would personally vote for 3, to remove any abiguity, but I am also interested what others think. @manuio @nmarcetic @dusanb94 @mteodor @blokovi

manuio: I also vote for 3

sprql: I updated PR using option 3

drasko: @sprql please do ot hate me, but after looking at PR I realized that actually option 2 would be more apropriate for Mainflux.

Actually, in Mainflux we have AuthN and AuthZ services, and I think that we should align errors to these functionalities, and not to outdated and missleading HTTP status names.

So I propose option 2, with a slight change: instad of `errUnauthorizedAccess` for `403` just use name `errUnauthorized`, to align it to it's 401 counterpart (`errUnauthenticated`).

Sorry for this, but I think now that everything is almost done, it will mostly be find-replace job :).

sprql: @drasko I've updated, could you look at this again, I also updated error messages

authn/api/http/transport.go
```diff
@@ -103,8 +103,8 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 	switch {
 	case errors.Contains(err, authn.ErrMalformedEntity):
 		w.WriteHeader(http.StatusBadRequest)
-	case errors.Contains(err, authn.ErrUnauthorizedAccess):
-		w.WriteHeader(http.StatusForbidden)
+	case errors.Contains(err, authn.ErrUnauthenticated):
+		w.WriteHeader(http.StatusUnauthorized)
```
dborovcanin, line 107: Should there be 2 cases (in this particular order):
```go
case errors.Contains(err, authn.ErrUnauthenticated):
case errors.Contains(err, authn.ErrUnauthorized):
```
The service itself won't return ErrUnauthenticated, but our JWT implementation (or transport layer in the case `key==""`) might.

authn/swagger.yaml
```diff
@@ -54,7 +54,7 @@ paths:
             $ref: "#/definitions/Key"
         400:
           description: Failed due to malformed query parameters.
-        403:
+        401:
```
dborovcanin, line 57: If we include both 401 and 403 (see my previous comment), ew should update API docs accordingly.

bootstrap/api/transport.go
```diff
@@ -289,8 +289,8 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 			w.WriteHeader(http.StatusBadRequest)
 		case errors.Contains(errorVal, bootstrap.ErrNotFound):
 			w.WriteHeader(http.StatusNotFound)
-		case errors.Contains(errorVal, bootstrap.ErrUnauthorizedAccess):
-			w.WriteHeader(http.StatusForbidden)
+		case errors.Contains(errorVal, bootstrap.ErrUnauthenticated):
+			w.WriteHeader(http.StatusUnauthorized)
```
dborovcanin, line 266: The same remark from Authn service for adding `Unauthorized` case applies here.
Not to repeat myself and spam this thread, please take a look at all the other services and consider adding both cases both in the transport layer (`svc/api/transport.go`) and API docs (Swagger files).

dborovcanin (changes requested):

dborovcanin:
@sprql Please don't address my comments yet. @mainflux/maintainers Need to agree on this. We'll do that internally, bring the final decision and let you know. We apologize for our back and forth request changes. I find both 401 and 403 code verbose values confusing and we had an internal misunderstanding between team members.

sprql:
@dusanb94, do you have any updates?

dborovcanin:
> @dusanb94, do you have any updates?

Unfortunately, not yet. Mainflux team is overloaded with some other urgent tasks, but we will get back to this PR as soon as we finish them.

drasko:
I would like that we can back to this next week, right after `0.11` release.

drasko:
@sprql can you please resolve the conflicts and update the branch. This starts to look good, and I would like that we proceed with this big PR.

@manuio @nmarcetic @dusanb94 please review after update.

blokovi (dismissed):

bootstrap/api/transport.go
sprql: > @sprql Please don't address my comments yet. @mainflux/maintainers Need to agree on this. We'll do that internally, bring the final decision and let you know. 

@dusanb94  Have you made a decision on this? Should I add this cases?

dusanb94: We should probably add these cases, but please don't address this comment yet. We didn't decide what to do with this PR yet, and there is no point that you continue working on this before we give you feedback.

dborovcanin: @sprql Let's do it like that. Please update the branch, add these cases, and let's review and merge.

mteodor:
`Vault` API has only `403` code 

```
204 - Success, no data returned.
400 - Invalid request, missing or invalid data.
403 - Forbidden, your authentication details are either incorrect, you don't have access to this feature, or - if CORS is enabled - you made a cross-origin request from an origin that is not allowed to make such requests.
404 - Invalid path. This can both mean that the path truly doesn't exist or that you don't have permission to view a specific path. We use 404 in some cases to avoid state leakage.

```
https://www.vaultproject.io/api-docs#http-status-codes

nmarcetic:
@sprql I am checking PR , for example 

```
case errors.Contains(err, authn.ErrUnauthenticated):
case errors.Contains(err, authn.ErrUnauthorized):
```

 Why not follow go HTTP lib and use Forbidden for 403 (as its status code Forbidden by spec) and Unautrhozed for 401 (as its status code for Unauthorized by spec)?
You added ErrUnauthenticated for 401, why "custom "naming?  And from the same link, the [spec you are addressing](https://tools.ietf.org/html/rfc6750#section-3) it says 

```
  For example, in response to a protected resource request without
   authentication:

     HTTP/1.1 401 Unauthorized
     WWW-Authenticate: Bearer realm="example"


```

IMHO `ErrUnauthorized` makes more sense than `ErrUnauthenticated` for status code 401. What do you think?

sprql:
> Why not follow go HTTP lib and use Forbidden for 403 (as its status code Forbidden by spec) and Unautrhozed for 401 (as its status code for Unauthorized by spec)?

@nmarcetic 
I suggested 3 different options https://github.com/mainflux/mainflux/pull/1140#discussion_r418028870

nmarcetic:
> > Why not follow go HTTP lib and use Forbidden for 403 (as its status code Forbidden by spec) and Unautrhozed for 401 (as its status code for Unauthorized by spec)?
> 
> @nmarcetic
> I suggested 3 different options [#1140 (comment)](https://github.com/mainflux/mainflux/pull/1140#discussion_r418028870)

@sprql I saw that you voted, but too late, sorry for that. @manuio explained to me but I already left my comment. 
I still think it's confusing and that option 3 is worst, but we have consensus already :) 
So let's see if we can approve and merge this @drasko @dusanb94 You must approve (there are your remarks blocking merge)
Also @sprql can you try to update, a branch is out of date. Thanks.

nmarcetic:
Can we update this branch and resolve conflicts @sprql

sprql:
@nmarcetic I've updated branch

darkodraskovic:
Finally, we've decided to use `http.StatusUnauthorized` for the authentication and `http.StatusForbidden` for authorization.

PR #1152, opened 2020-04-30
MF-739 - Add ID to the User entity

manuio:


users/postgres/init.go
```diff
@@ -49,16 +49,24 @@ func migrateDB(db *sqlx.DB) error {
 				Id: "users_1",
 				Up: []string{
 					`CREATE TABLE IF NOT EXISTS users (
-						email	 VARCHAR(254) PRIMARY KEY,
-						password CHAR(60)	  NOT NULL
-					)`,
+					email VARCHAR(254) PRIMARY KEY,
+					password CHAR(60) NOT NULL)`,
```
drasko, line 53: Move closing brackets in the next line, to mark clearly the block closing.

users/postgres/init.go
```diff
@@ -49,16 +49,24 @@ func migrateDB(db *sqlx.DB) error {
...
 				},
 				Down: []string{"DROP TABLE users"},
 			},
 			{
 				Id: "users_2",
 				Up: []string{
-					`ALTER TABLE IF EXISTS users ADD COLUMN IF NOT EXISTS metadata JSONB`,
+					`ALTER TABLE IF EXISTS users ADD COLUMN IF NOT EXISTS
+					 metadata JSONB`,
```
drasko: No need to break in two lines IMHO.

users/uuid/idp.go
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Mainflux
```
drasko: Is the same one used for things (like UUIDv4 generator)? Maybe then can be moved to project root and mutualized between two packages?

things/postgres/init.go
```diff
@@ -83,16 +83,14 @@ func migrateDB(db *sqlx.DB) error {
 				Id: "things_2",
 				Up: []string{
 					`ALTER TABLE IF EXISTS things ALTER COLUMN
-					 metadata TYPE JSONB using metadata::text::jsonb
-					`,
+					 metadata TYPE JSONB using metadata::text::jsonb`,
```
drasko, line 86: I kinda liked that this comma was in the new line. I am not saying that this is good solution, but I think it would be worth to check best practices when writing SQL strings in the code...

drasko (changes requested):

users/uuid/idp.go
manuio: I agree, I wanted to propose this also. Especially because there are 3 services that use idp

manuio: I'll fix this for other services in a separate PR: https://github.com/mainflux/mainflux/issues/1154

things/postgres/init.go
manuio: I would follow the standard which is to don't add a new line. We are also following this standard in other services.

drasko: @manuio I agree, I just suggested to check where this standard commens and if it widely accepted, not something just invented by us. How other big Go projects handle this problem?

manuio: i think that there are 2 versions:
```
				Up: []string{
					`ALTER TABLE IF EXISTS channels ALTER COLUMN
					 metadata TYPE JSONB using metadata::text::jsonb`,
				},
```
or 
```
				Up: []string{`
                                         ALTER TABLE IF EXISTS channels ALTER COLUMN
					 metadata TYPE JSONB using metadata::text::jsonb`,
				},
```

And I like the one that we are using

drasko: I was expecting some research links... And my remark was about comma (`,`) at the end of the statement, not backticks.

manuio: @drasko  the `,` still there

drasko: Is it, but it is not in the new line. So this is what I am trying to figure out - should it be in a new line or not.

manuio: In this case it's not a good proposition since it doesn't build, Lines are closed with `,`

drasko: How did it build before then?

manuio: Because of the `,

uuid/idp.go
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Mainflux
```
drasko: Will we call this top leve package `uuid`? It does not really implement `uuid`, but rather our IdP...

I do not know, does not bother me much, but I'd like to her form @mainflux/maintainers

drasko (commented):
Generally looks, good. I will let others review as well it is importnant, but I think we are close.

mteodor (dismissed):
LGTM

users/postgres/users.go
```diff
@@ -71,7 +71,7 @@ func (ur userRepository) UpdateUser(ctx context.Context, user users.User) error
 }
 
 func (ur userRepository) RetrieveByID(ctx context.Context, email string) (users.User, error) {
```
nmarcetic: If its `RetrieveByID` why param is `email` ?

uuid/idp.go
nmarcetic: This is not an [Identity provider](https://en.wikipedia.org/wiki/Identity_provider).

users/postgres/users.go
manuio: @nmarcetic because in Users service we use email as unique ID. I can maybe rename the function

uuid/idp.go
manuio: maybe we should rename it to `uuid/uuid.go` ?

nmarcetic: Make more sense to me @manuio or put it under `/pkg` its really shared lib

users/postgres/users.go
nmarcetic: Yea because we didn't have `ID` but now when we introduce `ID`, this should be changed.

users/users.go
```diff
@@ -53,16 +54,16 @@ func (u User) Validate() error {
...
+	Save(ctx context.Context, u User) error
 
 	// Update updates the user metadata.
-	UpdateUser(context.Context, User) error
+	UpdateUser(ctx context.Context, u User) error
 
-	// RetrieveByID retrieves user by its unique identifier (i.e. email).
-	RetrieveByID(context.Context, string) (User, error)
+	// RetrieveByEmail retrieves user by its unique identifier (i.e. email).
+	RetrieveByEmail(ctx context.Context, email string) (User, error)
```
mteodor, line 63: do we need `RetrieveByID` too?

mteodor (dismissed):

sprql:
Sorry for off topic, but do you have any plans to use this ID in references in relations instead of email (owner)?

nmarcetic:
> Sorry for off topic, but do you have any plans to use this ID in references in relations instead of email (owner)?

This makes sense to me, now when we have `ID` we should use it as unique resource identification, as we are doing for other entities in the system.  @mainflux/maintainers what do you think?

manuio:
@sprql @nmarcetic yes, that's the idea. But we have to think about the backward compatibility of Things and Channels owner.

nmarcetic:
> @sprql @nmarcetic yes, that's the idea. But we have to think about the backward compatibility of Things and Channels owner.

Yea it will be a breaking change, make sense. Not sure how we can get backward compatibility, except migration scripts. Which can be tricky for us to do it. Let's think about this

blokovi (approved):

mteodor (approved):
LGTM

PR #1153, opened 2020-04-30
NOISSUE - Fix default nats pubsub subject

manuio:

### DIFF
--- a/transformers/senml/transformer.go
+++ b/transformers/senml/transformer.go
@@ -59,7 +59,7 @@ func (t transformer) Transform(msg messaging.Message) (interface{}, error) {
 		// Use reception timestamp if SenML messsage Time is missing
 		t := v.Time
 		if t == 0 {
-			// Convert the timestamp to float64 with nanoseconds precision
+			// Convert the Unix timestamp in nanoseconds to float64
 			t = float64(msg.Created) / float64(1e9)
 		}
 
--- a/transformers/senml/transformer.go
+++ b/transformers/senml/transformer.go
@@ -4,8 +4,6 @@
 package senml
 
 import (
-	"time"
-
 	"github.com/mainflux/mainflux/errors"
 	"github.com/mainflux/mainflux/messaging"
 	"github.com/mainflux/mainflux/transformers"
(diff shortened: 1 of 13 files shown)


blokovi (dismissed):

mteodor (dismissed):

drasko (changes requested):
I agree that sending `topic` when we already have `msg.channel` in `msg` is redundant, but @dusanb94 had a rationale in developing the interface and he must explain.

mteodor (dismissed):

transformers/senml/transformer.go
```diff
@@ -59,9 +57,10 @@ func (t transformer) Transform(msg messaging.Message) (interface{}, error) {
 	msgs := make([]Message, len(normalized.Records))
 	for i, v := range normalized.Records {
 		// Use reception timestamp if SenML messsage Time is missing
-		t := v.Time
-		if t == 0 {
-			t = float64(time.Now().UnixNano())
+		time := v.Time
```
drasko: I would avoid the name `time`, as it relates to Go library name... I prefer `t` in this case.

transformers/senml/transformer.go
```diff
@@ -59,9 +57,10 @@ func (t transformer) Transform(msg messaging.Message) (interface{}, error) {
...
+		if time == 0 {
+			// Convert the timestamp into float64 with nanoseconds precision
+			time = float64(msg.Created.GetSeconds()) + float64(msg.Created.GetNanos())/float64(1e9)
```
drasko: This looks like precision in seconds to me. If you wanted nanoseconds, you should probably do something like:
```go
float64(msg.Created.GetSeconds() * 1e9 + msg.Created.GetNanos())
```

drasko (changes requested):

drasko (changes requested):

dborovcanin (dismissed):

mteodor (dismissed):

dborovcanin:
> I agree that sending `topic` when we already have `msg.channel` in `msg` is redundant, but @dusanb94 had a rationale in developing the interface and he must explain.

I decided to switch to topic-based pub-sub so that we can have a clean `Pub` and `Sub` interfaces. Also, we don't have to necessarily publish to the `msg.Channel` in the Publisher implementation, but rather let the one who uses publisher decide where to publish. This way, we keep specification (interface), implementation (NATS Pub and PubSub), and usage clean (the one who uses publisher chooses where to publish). Also, we should avoid modifying topic in the implementation, because of possible problems such as the one this PR resolves. We still have it in the `Publish` method implementation (adding channels prefix), though.

dborovcanin (approved):

drasko (approved):
LGTM

PR #1158, opened 2020-05-03
MF-1157 - HTTP Forwarder

### LINKED ISSUE #1157: Forwarding messages by HTTP
**FEATURE REQUEST**

I use the Mainflux platform as a backend of my application and I need to transfer data to another microservices entity (without throughout a database). Actually, it is not possible to transfer some message directly to another application by HTTP. In fact with the Export service, it is possible to do something similar but it is with MQTT protocol. With HTTP, we can easily add circuit breakers, load balancers, rules, etc.
With @drasko, we have already discussed that on the Gitter (beginning of conversation here: https://gitter.im/Mainflux/mainflux?at=5e8e397c9316f34b8d8c5691).

This is definitely a should-have feature.

jonathandreyer:
Resolves #1157.

No breaking current functionality.

Not at the moment, only a draft of feature.

A draft of README as been added.

### Notes
The format of JSON data can be modified. As the moment, only data of SenML are send by a POST request to URL `<MF_HTTP_FORWARDER_URL>/channels/<channel_id>/subtopic`.


jonathandreyer:
It is a mistake of mine.

drasko:
@jonathandreyer nice contribution

@mteodor please check this PR and tell us if it makes sense adding HTTP support in Export. It should be generic, meaning implemening `Publisher` interface - something similar that @dusanb94 did with `mainflux/messaging`. Then implementation can be MQTT or HTTP.

cmd/http-forwarder/main.go
```diff
@@ -0,0 +1,126 @@
...
+		subjectsCfgPath: mainflux.Env(envSubjectsCfgPath, defSubjectsCfgPath),
+		contentType:     mainflux.Env(envContentType, defContentType),
+	}
+
+	return cfg
+}
+
+func makeMetrics() (*kitprometheus.Counter, *kitprometheus.Summary) {
+	counter := kitprometheus.NewCounterFrom(stdprometheus.CounterOpts{
+		Namespace: "httpforwarder",
```
manuio: Use `http_forwarder`

cmd/http-forwarder/main.go
```diff
@@ -0,0 +1,126 @@
...
+		Subsystem: "message_writer",
```
manuio, line 107: Why `message_writer` here?

docker/addons/http-forwarder/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+  docker_mainflux-base-net:
+    external: true
+
+services:
+  http-forwarder:
+    image: mainflux/http-forwarder:latest
+    container_name: mainflux-http-forwarder
+    restart: on-failure
+    environment:
+      MF_HTTP_FORWARDER_LOG_LEVEL: debug
```
manuio: use `${MF_HTTP_FORWARDER_LOG_LEVEL}` here

writers/httpforwarder/messages.go
```diff
@@ -0,0 +1,101 @@
...
+	return &httpforwarderRepo{
+		url: url,
+	}
+}
+
+func (repo *httpforwarderRepo) Save(messages ...senml.Message) error {
+	msgs := make(map[string][]*fields)
+	for _, msg := range messages {
+		_t := repo.fullTopic(&msg)
+		_fields := repo.fieldsOf(&msg)
```
manuio: Use `t` and `fields` instead of `_t` and `_fields` please

writers/httpforwarder/messages.go
```diff
@@ -0,0 +1,101 @@
...
+		if resp.StatusCode != http.StatusCreated && resp.StatusCode != http.StatusOK {
+			return errors.Wrap(errSaveMessage, errors.New(resp.Status))
+		}
+	}
+
+	return nil
+}
+
+func (repo *httpforwarderRepo) fullTopic(msg *senml.Message) string {
+	_t := fmt.Sprintf("channels.%s.%s", msg.Channel, msg.Subtopic)
```
manuio: `t` instead of `_t` please

writers/httpforwarder/messages.go
```diff
@@ -0,0 +1,101 @@
...
+
+		req.Header.Add("Content-Type", "application/json")
+		client := &http.Client{}
+		resp, err := client.Do(req)
+		if err != nil {
+			return errors.Wrap(errSaveMessage, err)
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode != http.StatusCreated && resp.StatusCode != http.StatusOK {
```
manuio: Do we need this check: `resp.StatusCode != http.StatusCreated` ?

manuio (changes requested):

writers/httpforwarder/messages.go
jonathandreyer: You're right, it's not necessary.

cmd/http-forwarder/main.go
jonathandreyer: Good question. Maybe `message_forwarder` it is better?

jonathandreyer:
From my side, I have some interrogations about the output format of data. For me JSON is OK but is there any requirement about which data needed to be present? And the format of URL, is it OK?
Do you have any recommendations?

drasko:
Most of the work got delayed because of the `0.11` release.

@jonathandreyer coming back to this - this looks like something that should be a part of Export service as I mentioned before, and I understand that you need simple solution. However, I am still not convinced that this should be a part of the core, and I see it as an external application.

Would it be possible for you to maintain this application somewhere in your private GitHub repo, and we can leave a links to these community repos in our docs?

drasko:
@jonathandreyer can you please tell us about this idea - that you keep this as a community Writer - not to put it in the Core. If will be easier for us if you can maintain this.

jonathandreyer:
@drasko I will try to be the most complete as possible :-)

I fully agree about the point of mainflux-core. I have added this feature inside this repository for some reasons. In first, this feature is quite similar to DB-writer and easier to code inside this repository. The second point, it is because the structure of NATS message is inside this repository. I can find other explanations but those two are the main reasons. Probably (but I ignore how to do that), it is possible to do that inside another repository but I have not the knowledge in Go language (importation of some part of mainflux-core inside another repository, etc).

The goal of this PR is to show the idea of HTTP-forwarder. A short example of the code is usually more explicit than many sentences. (A picture's worth a thousand words.)

I have also developed (before writing this PR in Go) a draft of this idea in python (because I use more easily Python) but the NATS format of messages is quite problematic (not because of the structure of data but because it is with protobuf). When a modification is done in the structure of message, the package with the structure of the message needs to be compiled again, etc.

My point of view about this feature is that she is really necessary to transfer message to another application. But I don’t know in which context do you use (or planned to use) Mainflux and what is the future roadmap about this platform. So, it is only some speculation. For me, the HTTP protocol has some gain regarding the MQTT protocol to send data to another application. It is possible to use a reverse proxy to filter and redirect data behind the application. It is also possible to add a load balancer, etc.

I would prefer that this feature is reviewed for your team as well because I think you have a better idea of where you want to go with the Mainflux project and you also have a better knowledge of Go. About my knowledge in Go, I never used this language in other projects.
From my side, I think we can move (recode) this idea inside the export "module", maybe after a refactor of this module (or some help from your side to implement this feature in this module). What is your point of view about that?

drasko:
@jonathandreyer as I said, I would prefer if you could keep this extension in your private repos.

We are trying to remove all the unnecessary code from the platform, there is already too much. This also removes responsibility from Mainflux maintainers, and offloads it to community.

What we can do is review your implementation in your repo. It can be Python or any other kind of service, it is a micro-service based architecture, so please use your language of choice.

We will also keep a list of community repos and extensions in the documentation, so please tell us once your repo is ready.

I will close this PR for now. Please paste the link to the repo once it is ready.

jonathandreyer:
@drasko OK, copy that. I will implement this feature in another repository and be back when it is implemented.

jonathandreyer:
@drasko I have migrated the code of this PR into a new [repository](https://github.com/jonathandreyer/mainflux-http-forwarder). Currently, it is for v0.11.0 but I will update the code to be synced with the master release. I will add a message in the related issue that this extension resolve it.
As discussed below, if you can take a look, it is with pleasure.

PR #1164, opened 2020-05-05
MF-1079 - Add MQTT forwarder

### LINKED ISSUE #1079: Add MQTT forwarder
**BUG REPORT**

I am trying to publish messages using HTTP or Websocket so a device using MQTT subscribe can receive messages.

When publishing messages using HTTP or Websocket over a channel, a device subscribe to that channel using MQTT should receive messages.

MQTT sub are not receiving messages published by HTTP or Websocket.

- Create two things and a channel and connect both things to that channel.
- Subscribe to the channel using thing-1-id and thing-1-key. 
- Publish a message using thing-2-id and thing-2-key over the channel using HTTP or Websocket

- Latest Docker images deployed using docker-compose.
- Tried everything on one Linux machine, or separate pub and sub into two Linux machines.

- MQTT sub receives MQTT pub without any issues.
- Websocket receives HTTP, Websocket and MQTT published messages without any issues.
- Subscribe using mosquitto
```mosquitto_sub -u 7287b133-942c-45f1-a19f-ff77167db47b -P ebd001f4-a6ec-4c03-b0fb-6f78bd16608c -t channels/434b9b66-94be-4f99-913d-6de7bb1c2008/messages/ct/application_senml-json -h localhost```
- Subscribe using Paho
Created a Java program using Paho library to subscribe to the same channel.
- Publish using CURL 
```curl -s -S -X POST -H "Content-Type: application/senml+json" -H "Authorization: 0250e814-deca-4cca-8201-9ab535ed32a9" http://localhost/http/channels/434b9b66-94be-4f99-913d-6de7bb1c2008/messages -d '[{"bn":"some-base-name:","bt":1.276020076001e+09, "bu":"A","bver":5, "n":"voltage","u":"V","v":120.1}, {"n":"current","t":-5,"v":1.2}, {"n":"current","t":-4,"v":1.3}]'```
- Publish using CLI
```docker run -it --rm mainflux/cli -m http://<host ip> messages send 434b9b66-94be-4f99-913d-6de7bb1c2008 '[{"bn":"some-base-name:","bt":1.276020076001e+09, "bu":"A","bver":5, "n":"voltage","u":"V","v":120.1}, {"n":"current","t":-5,"v":1.2}, {"n":"current","t":-4,"v":1.3}]' 0250e814-deca-4cca-8201-9ab535ed32a9```
- Publish using Websocket
Created a JS page and published same message using Websocket.

### PR DESCRIPTION
**Pull request by dborovcanin**
### What does this do?
This pull request introduces MQTT forwarder for multi-protocol support explained in #1079.

This pull request fixes #1079.

Messages published from other protocols will be received on MQTT subs.

No.

This fix will make the MQTT adapter work as intended.

### DIFF
--- a/cmd/mqtt/main.go
+++ b/cmd/mqtt/main.go
@@ -147,7 +147,8 @@ func main() {
 		logger.Error(fmt.Sprintf("Failed to create MQTT publisher: %s", err))
 		os.Exit(1)
 	}
-	if err := mqtt.Forward(nats.SubjectAllChannels, nps, mp); err != nil {
+	fwd := mqtt.NewForwarder(nats.SubjectAllChannels)
+	if err := fwd.Forward(nps, mp); err != nil {
 		logger.Error(fmt.Sprintf("Failed to forward NATS messages: %s", err))
 		os.Exit(1)
 	}
--- a/cmd/mqtt/main.go
+++ b/cmd/mqtt/main.go
@@ -34,14 +34,17 @@ const (
 	defLogLevel = "error"
 	envLogLevel = "MF_MQTT_ADAPTER_LOG_LEVEL"
 	// MQTT
-	defMQTTHost       = "0.0.0.0"
-	defMQTTPort       = "1883"
-	defMQTTTargetHost = "0.0.0.0"
-	defMQTTTargetPort = "1884"
-	envMQTTHost       = "MF_MQTT_ADAPTER_MQTT_HOST"
-	envMQTTPort       = "MF_MQTT_ADAPTER_MQTT_PORT"
-	envMQTTTargetHost = "MF_MQTT_ADAPTER_MQTT_TARGET_HOST"
-	envMQTTTargetPort = "MF_MQTT_ADAPTER_MQTT_TARGET_PORT"
+	defMQTTHost             = "0.0.0.0"
+	defMQTTPort             = "1883"
+	defMQTTTargetHost       = "0.0.0.0"
+	defMQTTTargetPort       = "1883"
+	defMQTTForwarderTimeout = "30" // in seconds
+
+	envMQTTHost             = "MF_MQTT_ADAPTER_MQTT_HOST"
+	envMQTTPort             = "MF_MQTT_ADAPTER_MQTT_PORT"
+	envMQTTTargetHost       = "MF_MQTT_ADAPTER_MQTT_TARGET_HOST"
+	envMQTTTargetPort       = "MF_MQTT_ADAPTER_MQTT_TARGET_PORT"
+	envMQTTForwarderTimeout = "MF_MQTT_ADAPTER_FORWARDER_TIMEOUT"
 	// HTTP
 	defHTTPHost       = "0.0.0.0"
 	defHTTPPort       = "8080"
@@ -84,28 +87,29 @@ const (
 )
 
 type config struct {
-	mqttHost          string
-	mqttPort          string
-	mqttTargetHost    string
-	mqttTargetPort    string
-	httpHost          string
-	httpPort          string
-	httpScheme        string
-	httpTargetHost    string
-	httpTargetPort    string
-	httpTargetPath    string
-	jaegerURL         string
-	logLevel          string
-	thingsURL         string
-	thingsAuthURL     string
-	thingsAuthTimeout time.Duration
-	natsURL           string
-	clientTLS         bool
-	caCerts           string
-	instance          string
-	esURL             string
-	esPass            string
-	esDB              string
+	mqttHost             string
+	mqttPort             string
+	mqttTargetHost       string
+	mqttTargetPort       string
+	mqttForwarderTimeout time.Duration
+	httpHost             string
+	httpPort             string
+	httpScheme           string
+	httpTargetHost       string
+	httpTargetPort       string
+	httpTargetPath       string
+	jaegerURL            string
+	logLevel             string
+	thingsURL            string
+	thingsAuthURL        string
+	thingsAuthTimeout    time.Duration
+	natsURL              string
+	clientTLS            bool
+	caCerts              string
+	instance             string
+	esURL                string
+	esPass               string
+	esDB                 string
 }
 
 func main() {
@@ -130,20 +134,13 @@ func main() {
 
 	cc := thingsapi.NewClient(conn, thingsTracer, cfg.thingsAuthTimeout)
 
-	pub, err := nats.NewPublisher(cfg.natsURL)
-	if err != nil {
-		logger.Error(fmt.Sprintf("Failed to connect to NATS: %s", err))
-		os.Exit(1)
-	}
-	defer pub.Close()
-
 	ps, err := nats.NewPubSub(cfg.natsURL, "mqtt", logger)
 	if err != nil {
 		logger.Error(fmt.Sprintf("Failed to connect to NATS: %s", err))
 		os.Exit(1)
 	}
 	defer ps.Close()
-	f, err := mqtt.NewForwarder(fmt.Sprintf("%s:%s", cfg.mqttHost, cfg.mqttPort))
+	f, err := mqtt.NewForwarder(fmt.Sprintf("%s:%s", cfg.mqttTargetHost, cfg.mqttTargetPort), cfg.mqttForwarderTimeout)
 	if err != nil {
 		logger.Error(fmt.Sprintf("Failed to create MQTT forwarder: %s", err))
 		os.Exit(1)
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: messaging/mqtt/subscriber.go [hunk 1]
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package mqtt
+
+import (
+	"errors"
```
blokovi: why not `"github.com/mainflux/mainflux/errors"` ?

cmd/mqtt/main.go
```diff
@@ -130,6 +136,22 @@ func main() {
...
+		logger.Error(fmt.Sprintf("Failed to connect to NATS: %s", err))
+		os.Exit(1)
+	}
+	defer ps.Close()
+	mqttPub, err := mqttpub.NewPublisher(fmt.Sprintf("%s:%s", cfg.mqttTargetHost, cfg.mqttTargetPort), cfg.mqttForwarderTimeout)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Failed to create MQTT publisher: %s", err))
+		os.Exit(1)
+	}
+	if err := mqtt.Forward(nats.SubjectAllChannels, ps, mqttPub); err != nil {
```
manuio, line 150: `nats.SubjectAllChannels` should be used as arg here `nats.NewPubSub(cfg.natsURL, "mqtt", logger)`

manuio: should we configure `nats.SubjectAllChannels` from pubsub?

mqtt/forwarder.go
```diff
@@ -0,0 +1,33 @@
...
+func Forward(topic string, sub messaging.Subscriber, pub messaging.Publisher) error {
+	return sub.Subscribe(topic, handle(pub))
+}
+
+func handle(pub messaging.Publisher) messaging.MessageHandler {
+	return func(msg messaging.Message) error {
+		if msg.Protocol == protocol {
+			return nil
+		}
+		topic := fmt.Sprintf("%s.%s.%s.%s", channels, msg.Channel, messages, msg.Subtopic)
```
manuio: Is `topic` ok if `Subtopic` is empty?

manuio (changes requested):

cmd/mqtt/main.go
```diff
@@ -130,6 +136,22 @@ func main() {
 
 	cc := thingsapi.NewClient(conn, thingsTracer, cfg.thingsAuthTimeout)
 
+	ps, err := nats.NewPubSub(cfg.natsURL, "mqtt", logger)
```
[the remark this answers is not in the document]
drasko: Maybe `nps` can be a good compromise? Also `mqttPub` can become `mp` and `pub` below can become `np`.

In any case, using long or short names - we should be consistent - which is not currently the case.

messaging/mqtt/pubsub.go
```diff
@@ -0,0 +1,39 @@
...
+
+func newClient(address string, timeout time.Duration) (mqtt.Client, error) {
+	opts := mqtt.NewClientOptions().
+		AddBroker(address).
+		SetUsername(id).
+		SetPassword(id).
+		SetClientID(id).
+		SetCleanSession(false)
+	client := mqtt.NewClient(opts)
+	tkn := client.Connect()
```
drasko: What does `tkn` stand for?

messaging/mqtt/subscriber.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+
```
drasko: Kill this white line

drasko (changes requested):

messaging/mqtt/pubsub.go
dborovcanin: It stands for `token`. That's the construct Paho internally returns on the async operations.

manuio: Why not `token`?

drasko: It is bit confusing IMHO, maybe comment could help?

manuio: @drasko why a comment?

mteodor: why not `token` than?

drasko (dismissed):
LGTM

messaging/mqtt/publisher.go
```diff
@@ -0,0 +1,48 @@
...
+		client:  client,
+		timeout: timeout,
+	}
+	return ret, nil
+}
+
+func (pub publisher) Publish(topic string, msg messaging.Message) error {
+	token := pub.client.Publish(topic, qos, false, msg.Payload)
+	ok := token.WaitTimeout(pub.timeout)
+	if ok && token.Error() != nil {
```
manuio, line 43: What if `token.Error() != nil` in the previous line?

manuio (changes requested):

mqtt/forwarder.go
```diff
@@ -0,0 +1,51 @@
...
+		if msg.Protocol == protocol {
+			return nil
+		}
+		// Use concatenation instead of mft.Sprintf for the
+		// sake of simplicity and performance.
+		topic := channels + "." + msg.Channel + "." + messages
+		if msg.Subtopic != "" {
+			topic += "." + msg.Subtopic
+		}
+		topic = strings.ReplaceAll(topic, ".", "/")
```
manuio: why not simply use `/`? 
Also, use fmt.Sprint() to concatenate strings

manuio (changes requested):

mqtt/forwarder.go
dborovcanin: Changed to `/`. String buffer would be an overkill here, and the comment explains why I don't want to use `Sprintf`.

messaging/mqtt/publisher.go
mteodor: dont you like this `if token.Wait() && token.Error() != nil {` better ?

messaging/mqtt/subscriber.go
```diff
@@ -0,0 +1,80 @@
...
+		logger:  logger,
+	}
+	return ret, nil
+}
+
+func (sub subscriber) Subscribe(topic string, handler messaging.MessageHandler) error {
+	token := sub.client.Subscribe(topic, qos, sub.mqttHandler(handler))
+	ok := token.WaitTimeout(sub.timeout)
+	if ok && token.Error() != nil {
+		return token.Error()
```
mteodor, line 52: if we return `token.Error()` on multiple places it would be good to wrap with another error so that exact place can be more easily detected

mqtt/forwarder.go
drasko: Makes sense, and I think we should consider avoiding `sprintf` in general where `+` string concatenation can bring simplicity. We need to sync internally and understand when to use one and when another.

messaging/mqtt/subscriber.go
dborovcanin: I would leave error wrapping for the layer above MQTT Publisher or Subscriber. We only use it as a wrapper around MQTT client, so returning the real error is sufficient, IMHO. However, I'll improve the message of the errors we return if token is expired to be more specific ("failed to subscribe/unsubscribe due to timeout reached" is better than just "failed to subscribe/unsubscribe").

cmd/mqtt/main.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/mainflux/mainflux"
 	"github.com/mainflux/mainflux/logger"
 	"github.com/mainflux/mainflux/messaging"
+	mqttpub "github.com/mainflux/mainflux/messaging/mqtt"
 	"github.com/mainflux/mainflux/messaging/nats"
 	mqtt "github.com/mainflux/mainflux/mqtt"
 	mr "github.com/mainflux/mainflux/mqtt/redis"
```
manuio: I would use `mqttRedis` here.

messaging/mqtt/pubsub.go
```diff
@@ -0,0 +1,39 @@
...
+
+func newClient(address string, timeout time.Duration) (mqtt.Client, error) {
+	opts := mqtt.NewClientOptions().
+		AddBroker(address).
+		SetUsername(id).
+		SetPassword(id).
+		SetClientID(id).
+		SetCleanSession(false)
+	client := mqtt.NewClient(opts)
+	token := client.Connect()
```
manuio, line 29: Should we add a check here?

manuio (changes requested):

messaging/mqtt/pubsub.go
dborovcanin: I added the check.

manuio (approved):
LGTM

cmd/mqtt/main.go
drasko: Why Redis? How is this MQTT forwarding related to Redis?

mteodor (approved):

PR #1165, opened 2020-05-05
NOISSUE - Add subtopic to twin state payload key

darkodraskovic:

Currently, state payload keys are identical to definition attribute names. This does not take into account subtopic contribution to the state payload and attribute semantics.

### DIFF
--- a/twins/service.go
+++ b/twins/service.go
@@ -374,7 +374,7 @@ func prepareState(st *State, tw *Twin, rec senml.Record, msg *messaging.Message)
 				}
 			}
 			val := findValue(rec)
-			st.Payload[attr.Name] = val
+			st.Payload[getKeyName(attr)] = val
 			break
 		}
 	}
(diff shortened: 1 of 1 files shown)


twins/service.go
```diff
@@ -401,9 +401,17 @@ func findValue(rec senml.Record) interface{} {
 	return nil
 }
 
-func findAttribute(name string, attrs []Attribute) (idx int) {
+func getKeyName(attr Attribute) string {
+	keyName := attr.Name
+	if len(attr.Subtopic) > 0 {
+		keyName = attr.Name + "/" + attr.Subtopic
```
manuio, line 407: Use `fmt.Sprintf()` please

twins/service.go
```diff
@@ -401,9 +401,17 @@ func findValue(rec senml.Record) interface{} {
 	return nil
 }
 
-func findAttribute(name string, attrs []Attribute) (idx int) {
+func getKeyName(attr Attribute) string {
+	keyName := attr.Name
+	if len(attr.Subtopic) > 0 {
```
manuio, line 406: If `attr.Subtopic` is a string I would use `!= ""`

manuio (changes requested):

darkodraskovic:
Sorry, I realized that this is already covered by the possibility to give arbitrary names to attributes. Closing this pull request.

PR #1169, opened 2020-05-07
NOISSUE - Remove thing related code from twins service

darkodraskovic:
Formerly, twin service was dependent on Mainflux thing. Every twin was connected to one and only one thing. Now twin is decoupled from thing and is thing agnostic, since its definition relies only on channels and subtopics.

Also, since notifications were published via mqtt, we needed a MF thing in order to publish them. Now they are published directly on NATS, so there is no need to specify thing via env var.

The code related to thing in the upper two senses is now removed.

### DIFF
--- a/twins/README.md
+++ b/twins/README.md
@@ -1,10 +1,10 @@
 # Twins
 
 Service twins is used for CRUD and update of digital twins. Twin is a semantic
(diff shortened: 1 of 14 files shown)


drasko (approved):
LGTM

dborovcanin (approved):

PR #1172, opened 2020-05-12
MF-1154 - Move UUID provider to pkg

### LINKED ISSUE #1154: Move identity providers to project root
Mutualize Things, Authn, and Users identity providers

manuio:

Resolves #1154


mteodor (dismissed):
LGTM

blokovi (dismissed):

authn/service_test.go
```diff
@@ -23,7 +23,7 @@ const (
 
 func newService() authn.Service {
 	repo := mocks.NewKeyRepository()
-	idp := mocks.NewIdentityProvider()
+	idp := mocks.NewUUIDProvider()
```
drasko: I do not like the idea to call this var `idp` any more

drasko (changes requested):

mteodor (dismissed):

nmarcetic:
@mainflux/maintainers Why not moving this to `/pkg` it's a great candidate as a shared package.

manuio:
@nmarcetic I'm not against. I also wanted to know where do you think that we should put uuid mock. In `pkg/uuid/mocks` directory or in a specific `pkg/mocks` directory?

nmarcetic:
> @nmarcetic I'm not against. I also wanted to know where do you think that we should put uuid mock. In `pkg/uuid/mocks` directory or in a specific `pkg/mocks` directory?

If those mocks are not sharable (e.g not used by there packages, but only by `/pkg/uuid`) then I would keep them under `/pkg/uuid`

drasko (approved):
LGTM

nmarcetic (approved):

PR #1173, opened 2020-05-12
NOISSUE - Add MF_DOCKER_IMAGE_NAME_PREFIX to Makefile

sprql:
Allow custom docker image names

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 # Copyright (c) Mainflux
 # SPDX-License-Identifier: Apache-2.0
 
+MF_DOCKER_IMAGE_NAME_PREFIX ?= mainflux
 BUILD_DIR = build
 SERVICES = users things http coap lora influxdb-writer influxdb-reader mongodb-writer \
 	mongodb-reader cassandra-writer cassandra-reader postgres-writer postgres-reader cli \
@@ -22,7 +23,7 @@ define make_docker
 		--build-arg SVC=$(svc) \
 		--build-arg GOARCH=$(GOARCH) \
 		--build-arg GOARM=$(GOARM) \
-		--tag=mainflux/$(svc) \
+		--tag=$(MF_DOCKER_IMAGE_NAME_PREFIX)/$(svc) \
 		-f docker/Dockerfile .
 endef
 
@@ -32,7 +33,7 @@ define make_docker_dev
 	docker build \
 		--no-cache \
 		--build-arg SVC=$(svc) \
-		--tag=mainflux/$(svc) \
+		--tag=$(MF_DOCKER_IMAGE_NAME_PREFIX)/$(svc) \
 		-f docker/Dockerfile.dev ./build
 endef
 
@@ -47,20 +48,20 @@ cleandocker:
 	# Stop all containers (if running)
 	docker-compose -f docker/docker-compose.yml stop
 	# Remove mainflux containers
-	docker ps -f name=mainflux -aq | xargs -r docker rm
+	docker ps -f name=$(MF_DOCKER_IMAGE_NAME_PREFIX) -aq | xargs -r docker rm
 
 	# Remove exited containers
-	docker ps -f name=mainflux -f status=dead -f status=exited -aq | xargs -r docker rm -v
+	docker ps -f name=$(MF_DOCKER_IMAGE_NAME_PREFIX) -f status=dead -f status=exited -aq | xargs -r docker rm -v
 
 	# Remove unused images
-	docker images "mainflux\/*" -f dangling=true -q | xargs -r docker rmi
+	docker images "$(MF_DOCKER_IMAGE_NAME_PREFIX)\/*" -f dangling=true -q | xargs -r docker rmi
 
 	# Remove old mainflux images
-	docker images -q mainflux\/* | xargs -r docker rmi
+	docker images -q "$(MF_DOCKER_IMAGE_NAME_PREFIX)\/*" | xargs -r docker rmi
 
 ifdef pv
 	# Remove unused volumes
-	docker volume ls -f name=mainflux -f dangling=true -q | xargs -r docker volume rm
+	docker volume ls -f name=$(MF_DOCKER_IMAGE_NAME_PREFIX) -f dangling=true -q | xargs -r docker volume rm
 endif
 
 install:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -7,6 +7,7 @@ SERVICES = users things http coap lora influxdb-writer influxdb-reader mongodb-w
 	bootstrap opcua authn twins mqtt provision
 DOCKERS = $(addprefix docker_,$(SERVICES))
 DOCKERS_DEV = $(addprefix docker_dev_,$(SERVICES))
+DOCKER_IMAGE_NAME_PREFIX ?= ""
```
manuio: I would use `mainflux` as prefix here and I would remove it where it's hard-coded.

manuio (changes requested):

Makefile
sprql: I've thought about it but I'm confused about what to do with `/`: 
should be `/` part of `DOCKER_IMAGE_NAME_PREFIX` or not.

manuio: I prefer to keep it outside but let's see what others think about

drasko:
I do not see a great value of this PR in the core (mostly because of additional code complexity for no big gains).

@sprql  what is the particular use-case that you have to name images differently?

I think that this might be some particular use-case, which can probably be solved with a simple custom script (or even shell one-liner) that renames Mainflux images, but I would like to first hear from you the requirement.

sprql:
@drasko It's very handy for custom builds, for example when using different docker registry.
Yes, it is possible to create custom shell scripts, but isn't it easier to solve this in the first place, in Makefile?

I think it's better to introduce MF_DOCKER_IMAGE_NAME_PREFIX env var, so it will be easy to export it in a shell.

drasko:
> ...for example when using different docker registry.

Can you please elaborate on this, because I think that even in the private registries (non-DockerHub) you can have `mainflux/<service_name>`. Or am I wrong?

sprql:
@drasko, as I know, to push an image to non-DockerHub registry I should specify host and port in the Docker image name. So to push `mainflux/service:v1` to `registry.example.com` I should tag the image with `registry.example.com/mainflux/service:v1` and then push it as `registry.example.com/mainflux/service:v1`

https://docs.docker.com/engine/reference/commandline/push/

Maybe there is another approach which I don't know…

drasko:
I see... I thought that you can give a param to `docker push` in some other way, but apparently not.

OK, this is good enough reason to add this prefix to the core. Please update the branch and I will do the review.

Makefile
drasko: @manuio I think `mainflux` should be default, and then `<docker_registry_url>/mainflux` can be used for other cases.

I preferred previous implementation, looks cleaner and more configurable.

drasko (changes requested):

Makefile
manuio: @drasko I'm no sure to understand. I proposed `DOCKER_IMAGE_NAME_PREFIX ?= "mainflux"`
and `$(DOCKER_IMAGE_NAME_PREFIX)/$(svc)` to simplify an eventual renaming and because it's hard-coded for many commands.

drasko: Ah, cool - then we are on the same page.

Makefile
```diff
@@ -87,7 +88,7 @@ dockers_dev: $(DOCKERS_DEV)
 
 define docker_push
 	for svc in $(SERVICES); do \
-		docker push mainflux/$$svc:$(1); \
+		docker push $(DOCKER_IMAGE_NAME_PREFIX)mainflux/$$svc:$(1); \
```
manuio: I think that we can use `$(svc)` here

sprql: Without `$(MF_DOCKER_IMAGE_NAME_PREFIX)` there will be just service name, so for docker it would be different images

manuio: I'm maybe wrong but my question was about `$$svc`

sprql: I assume that is needed because there is used `shell for loop`, and to access shell variables we should use `$$` instead of `$`

manuio: right

Makefile
sprql: Good, I made `MF_DOCKER_IMAGE_NAME_PREFIX ?= mainflux/`

Makefile
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) Mainflux
 # SPDX-License-Identifier: Apache-2.0
 
+MF_DOCKER_IMAGE_NAME_PREFIX ?= mainflux/
```
drasko: Why not just `mainflux`? I would prefer that service is always prefixed with some name, then followed by `/`, so you can have `$(MF_DOCKER_IMAGE_NAME_PREFIX)/$(svc)`

It is more readable then concatenated like this.

drasko (changes requested):

Makefile
drasko: We proposed `mainflux`, not `mainflux/` ;)

Makefile
sprql: I thought it would be more flexible

Makefile
sprql: ok, updated:

`MF_DOCKER_IMAGE_NAME_PREFIX ?= mainflux`

drasko (dismissed):
LGTM

manuio (dismissed):
LGTM

manuio:
@sprql update your branch please

manuio:
@sprql Please, can you update again your branch? We had an issue with the CI...

Makefile
```diff
@@ -47,20 +48,20 @@ cleandocker:
...
-	docker ps -f name=mainflux -aq | xargs -r docker rm
+	docker ps -f name=$(MF_DOCKER_IMAGE_NAME_PREFIX) -aq | xargs -r docker rm
 
 	# Remove exited containers
-	docker ps -f name=mainflux -f status=dead -f status=exited -aq | xargs -r docker rm -v
+	docker ps -f name=$(MF_DOCKER_IMAGE_NAME_PREFIX) -f status=dead -f status=exited -aq | xargs -r docker rm -v
 
 	# Remove unused images
-	docker images "mainflux\/*" -f dangling=true -q | xargs -r docker rmi
+	docker images "$(MF_DOCKER_IMAGE_NAME_PREFIX)\/*" -f dangling=true -q | xargs -r docker rmi
```
drasko, line 57: Are the quotes really needed in `"$(MF_DOCKER_IMAGE_NAME_PREFIX)\/*"` or can be removed?

sprql: I think it depends on that is `sh` on your system

manuio (approved):
LGTM

drasko (approved):
LGTM

PR #1174, opened 2020-05-13
MF-995 - Add Twins tests for endpoint list twins and list states

### LINKED ISSUE #995: Write service, database and endpoint tests for state saving in Twins service

darkodraskovic:
Partially resolves #995


twins/api/http/endpoint_states_test.go
```diff
@@ -0,0 +1,256 @@
...
+	}{
+		{
+			desc:   "get a list of states",
+			auth:   token,
+			status: http.StatusOK,
+			url:    baseURL,
+			res:    data[0:10],
+		},
+		{
+			desc:   "get a list of states with with valid offset and limit",
```
manuio: `with with`

twins/api/http/endpoint_states_test.go
```diff
@@ -0,0 +1,256 @@
...
+		},
+		{
+			desc:   "get a list of states with empty token",
+			auth:   "",
+			status: http.StatusForbidden,
+			url:    fmt.Sprintf(queryFmt, baseURL, 0, 5),
+			res:    nil,
+		},
+		{
+			desc:   "get a list of states with with  + limit > total",
```
manuio: `with with`

twins/api/http/endpoint_states_test.go
```diff
@@ -0,0 +1,256 @@
...
+		},
+		{
+			desc:   "get a list of states without limit",
+			auth:   token,
+			status: http.StatusOK,
+			url:    fmt.Sprintf("%s?offset=%d", baseURL, 14),
+			res:    data[14:24],
+		},
+		{
+			desc:   "get a list of states with invalid number of params",
```
manuio: I would use `parameters`

twins/api/http/endpoint_states_test.go
```diff
@@ -0,0 +1,256 @@
...
+			auth:   token,
+			status: http.StatusBadRequest,
+			url:    fmt.Sprintf("%s%s", baseURL, "?offset=4&limit=4&limit=5&offset=5"),
+			res:    nil,
+		},
+		{
+			desc:   "get a list of states with redundant query params",
```
manuio: same here

twins/mocks/service.go
```diff
@@ -0,0 +1,15 @@
+package mocks
+
+import (
+	"github.com/mainflux/mainflux/twins"
+)
+
+func NewService(tokens map[string]string) twins.Service {
```
manuio, line 18: We are trying to use `New()` as much as posible

manuio (changes requested):

twins/api/http/endpoint_states_test.go
```diff
@@ -0,0 +1,256 @@
...
+		Name:         name,
+		Channel:      id,
+		Subtopic:     subtopic,
+		PersistState: true,
+	}
+}
+
+func createSenML(n int, bn string) []senml.Record {
+	var recs []senml.Record
+	bt := time.Now().Unix()
```
manuio: Use `bt := time.Now().UnixNano() / float64(1e9)`

twins/api/http/endpoint_states_test.go
```diff
@@ -0,0 +1,256 @@
...
+	for i := 0; i < n; i++ {
+		rec := senml.Record{
+			BaseName: bn,
+			BaseTime: float64(bt),
+			Time:     float64(i),
```
manuio: Why `Time` is `float64(i)`?  Probably you can use the same timestamp for all messages.

manuio (changes requested):

manuio (changes requested):

twins/api/http/endpoint_states_test.go
mteodor: and extra space

twins/api/http/endpoint_states_test.go
```diff
@@ -0,0 +1,256 @@
...
+			Value:    nil,
+		}
+		recs = append(recs, rec)
+	}
+	return recs
+}
+
+func createMessage(attr twins.Attribute, recs []senml.Record) *messaging.Message {
+	mRecs, _ := json.Marshal(recs)
```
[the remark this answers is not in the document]
darkodraskovic: Yes, but it's a helper function - not a test func with `t *testing.T` and the code guarantees correct values, since the values are generated in this package for this purpose and in this program flow - i.e. they don't come from outside.

twins/api/http/endpoint_states_test.go
manuio: I meant why `i` but forget it it's a test. Resolving this one.

twins/api/http/endpoint_states_test.go
```diff
@@ -0,0 +1,255 @@
...
+		},
+		{
+			desc:   "get a list of states with empty token",
+			auth:   "",
+			status: http.StatusForbidden,
+			url:    fmt.Sprintf(queryFmt, baseURL, 0, 5),
+			res:    nil,
+		},
+		{
+			desc:   "get a list of states with  + limit > total",
```
manuio: `with  +` there are 2 white spaces here

manuio (changes requested):

twins/api/http/endpoint_states_test.go
```diff
@@ -0,0 +1,255 @@
...
+	return recs
+}
+
+func createMessage(attr twins.Attribute, recs []senml.Record) *messaging.Message {
+	mRecs, _ := json.Marshal(recs)
+	return &messaging.Message{
+		Channel:   attr.Channel,
+		Subtopic:  attr.Subtopic,
+		Payload:   mRecs,
+		Publisher: "twins",
```
manuio: Create a const for `twins`.  btw, it could be  `twins_publisher`

manuio (changes requested):

manuio (dismissed):
LGTM

mteodor (dismissed):

twins/api/http/endpoint_states_test.go
```diff
@@ -0,0 +1,266 @@
...
+	attrSubtopic2 = "chassis"
+
+	publisher = "twins"
+)
+
+type stateRes struct {
+	TwinID     string `json:"twin_id"`
+	ID         int64  `json:"id"`
+	Definition int    `json:"definition"`
+	// Created    time.Time              `json:"created"`
```
drasko: Leave commented code, or find a way to use it

drasko (changes requested):

twins/api/http/endpoint_states_test.go
```diff
@@ -0,0 +1,259 @@
...
+		Publisher: publisher,
+	}, nil
+}
+
+func createStateResponse(id int, tw twins.Twin, rec senml.Record) stateRes {
+	return stateRes{
+		TwinID:     tw.ID,
+		ID:         int64(id),
+		Definition: tw.Definitions[len(tw.Definitions)-1].ID,
+		Payload:    map[string]interface{}{rec.BaseName: nil},
```
mteodor, line 211: I don't think that you need to set this to nil, zero value of pointer is `nil`

twins/api/http/endpoint_states_test.go
```diff
@@ -0,0 +1,259 @@
...
+			status: http.StatusOK,
+			url:    baseURL,
+			res:    data[0:10],
+		},
+		{
+			desc:   "get a list of states with valid offset and limit",
+			auth:   token,
+			status: http.StatusOK,
+			url:    fmt.Sprintf(queryFmt, baseURL, 20, 15),
+			res:    data[20:35],
```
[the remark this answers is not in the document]
darkodraskovic: We fetch 15 elements with an offset of 20. Therefore, we need elements from 20 up to (not including) 35.

twins/api/http/endpoint_states_test.go
darkodraskovic: I have to initialize this map to `{rec.BaseName: nil}`, therefore, I need to put explicit values.

twins/service_test.go
```diff
@@ -23,18 +23,8 @@ const (
...
-	statesRepo := mocks.NewStateRepository()
-	idp := mocks.NewIdentityProvider()
-	subs := map[string]string{"chanID": "chanID"}
-	broker := mocks.New(subs)
-	return twins.New(broker, auth, twinsRepo, statesRepo, idp, "chanID", nil)
-}
-
 func TestAddTwin(t *testing.T) {
-	svc := newService(map[string]string{token: email})
+	svc := mocks.New(map[string]string{token: email})
```
dborovcanin: I didn't dig into the code logic but `svc := mocks.New(...)` sounds like you're using the mock service in `service_test.go`, that is, you're mocking the very service you want to test. If that's the case, please fix this to use real service, and mock only dependencies.

darkodraskovic: This is the endpoints mock. I'm not mocking the service, I'm using the real service (look at the last line of the code):

```
func New(tokens map[string]string) twins.Service {
	auth := NewAuthNServiceClient(tokens)
	twinsRepo := NewTwinRepository()
	statesRepo := NewStateRepository()
	idp := NewIdentityProvider()
	subs := map[string]string{"chanID": "chanID"}
	broker := NewBroker(subs)
	return twins.New(broker, auth, twinsRepo, statesRepo, idp, "chanID", nil)
}
```

Formerly, the name of this function was `NewService`, but @manuio said that we should simply use `New` instead.

dborovcanin: In that case, can you, please, rename `svc` to `pub`. @manuio We do prefer `New` over `New<Suffix>`, but only in case `New` is unambiguous. In `mocks` package with multiple mocks inside it makes perfect sense to have `NewRepository` or `NewPublisher`, and in the code we'll have a nice `pub := mocks.NewPublisher(..)`.

darkodraskovic: @dusanb94 yes, that was exactly my argument: it will be ambiguous in the code. I'm fixing it bask to `NewService`.

manuio: @dusanb94 I agree.  As I said last time to @darkodraskovic I didn't noticed that it was a mock and I was fine to let it as it was.

twins/api/http/endpoint_states_test.go
mteodor: ok, its map init, my bad

mteodor (dismissed):

drasko (approved):
LGTM

blokovi (approved):

PR #1178, opened 2020-05-18
NOISSUE - Add VerneMQ docker image build from source

drasko:
Signed-off-by: Drasko Draskovic <drasko.draskovic@gmail.com>

### DIFF
--- a/docker/vernemq/bin/vernemq.sh
+++ b/docker/vernemq/bin/vernemq.sh
@@ -1,0 +1,155 @@
+#!/usr/bin/env bash
+
+IP_ADDRESS=$(ip -4 addr show ${DOCKER_NET_INTERFACE:-eth0} | grep -oE '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}' | sed -e "s/^[[:space:]]*//" | head -n 1)
+IP_ADDRESS=${DOCKER_IP_ADDRESS:-${IP_ADDRESS}}
+
+# Ensure the Erlang node name is set correctly
+if env | grep "DOCKER_VERNEMQ_NODENAME" -q; then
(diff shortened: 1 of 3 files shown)


nmarcetic (approved):

dborovcanin (approved):

docker/vernemq/bin/vernemq.sh
```diff
@@ -0,0 +1,155 @@
...
+$password
+EOF
+    done
+
+    echo "erlang.distribution.port_range.minimum = 9100" >> /vernemq/etc/vernemq.conf
+    echo "erlang.distribution.port_range.maximum = 9109" >> /vernemq/etc/vernemq.conf
+    echo "listener.tcp.default = ${IP_ADDRESS}:1883" >> /vernemq/etc/vernemq.conf
+    echo "listener.ws.default = ${IP_ADDRESS}:8080" >> /vernemq/etc/vernemq.conf
+    echo "listener.vmq.clustering = ${IP_ADDRESS}:44053" >> /vernemq/etc/vernemq.conf
+    echo "listener.http.metrics = ${IP_ADDRESS}:8888" >> /vernemq/etc/vernemq.conf
```
blokovi, line 109: You can turn off plugins.vmq_passwd here, it's not used
`echo "plugins.vmq_passwd = off" >> /vernemq/etc/vernemq.conf`

drasko: This is true, but my goal here is to produce identical image as VerneMQ official image. Changing their scripts can be confusing, especially when we start investigating through a doc and we can not find what is changed where.

I propose that we stay aligned to their scripts and change explicitly in our k8s scripts via ENV vars - it is more documenting. Also, it permits us easier updates if their scripts change.

PR #1184, opened 2020-05-22
MF-1180 - Add redis based twins and states cache

### LINKED ISSUE #1180: Add twins cache and states cache

darkodraskovic:
Resolves #1180


drasko (changes requested):

cmd/twins/main.go
```diff
@@ -232,16 +248,32 @@ func connectToAuth(cfg config, logger logger.Logger) *grpc.ClientConn {
 	return conn
 }
 
-func newService(ps messaging.PubSub, chanID string, users mainflux.AuthNServiceClient, dbTracer opentracing.Tracer, db *mongo.Database, logger logger.Logger) twins.Service {
+func connectToRedis(cacheURL, cachePass, cacheDB string, logger logger.Logger) *redis.Client {
+	db, err := strconv.Atoi(cacheDB)
+	if err != nil {
```
mteodor, line 253: why dont  you return an error and handle it in the caller and thus no need for logger here?

twins/mongodb/states.go
```diff
@@ -56,7 +56,7 @@ func (sr *stateRepository) Update(ctx context.Context, st twins.State) error {
 func (sr *stateRepository) Count(ctx context.Context, tw twins.Twin) (int64, error) {
 	coll := sr.db.Collection(statesCollection)
 
-	filter := bson.D{{"twinid", tw.ID}}
+	filter := bson.M{"twinid": tw.ID}
```
mteodor: shouldn't `"twinid"` be constant?

darkodraskovic: `twinid` is the string key name of the field in the bson. Therefore, i give it here as is.

cmd/twins/main.go
darkodraskovic: This is a copy-pasted func that we use elsewhere also. We use the same pass logger signature also in `connectToAuth`. Maybe it's better to change it everywhere. @dusanb94 ?

twins/redis/twins.go
```diff
@@ -0,0 +1,97 @@
...
+
+// NewTwinCache returns redis twin cache implementation.
+func NewTwinCache(client *redis.Client) twins.TwinCache {
+	return &twinCache{
+		client: client,
+	}
+}
+
+func (tc *twinCache) Save(_ context.Context, twin twins.Twin) error {
+	def := twin.Definitions[len(twin.Definitions)-1]
```
mteodor: lets check the len before access

darkodraskovic: By design, twin always have at least an empty definition because of this line https://github.com/mainflux/mainflux/blob/master/twins/service.go#L147 In that sense, that would be a superfluous check. However, might make sense if we let twins creation happen outside Save mtethod. Not sure.

twins/service.go
```diff
@@ -154,6 +156,10 @@ func (ts *twinsService) AddTwin(ctx context.Context, token string, twin Twin, de
 	id = twin.ID
 	b, err = json.Marshal(twin)
 
+	if err := ts.twinCache.Save(ctx, twin); err != nil {
```
drasko: I think you can directly `return ts.twinCache.Save(ctx, twin)` here

twins/service.go
```diff
@@ -205,6 +211,13 @@ func (ts *twinsService) UpdateTwin(ctx context.Context, token string, twin Twin,
 	id = twin.ID
 	b, err = json.Marshal(tw)
 
+	if err := ts.twinCache.Remove(ctx, id); err != nil {
+		return err
+	}
+	if err := ts.twinCache.Save(ctx, tw); err != nil {
```
drasko: You can directly `return ts.twinCache.Save(ctx, tw)` here

twins/service.go
```diff
@@ -240,6 +253,10 @@ func (ts *twinsService) RemoveTwin(ctx context.Context, token, id string) (err e
 		return err
 	}
 
+	if err := ts.twinCache.Remove(ctx, id); err != nil {
```
drasko: You can directly `return ts.twinCache.Remove(ctx, id)` here

drasko (changes requested):

twins/mocks/service.go
```diff
@@ -42,11 +40,10 @@ func CreateDefinition(names []string, subtopics []string) twins.Definition {
 }
 
 // CreateSenML creates SenML record array
-func CreateSenML(n int, bn string) []senml.Record {
+func CreateSenML(n int) []senml.Record {
```
mteodor: I think that that it is better to precreate array
like var recs[n]senml.Record and then initialize each element

twins/service.go
darkodraskovic: I wouldn't be returning an empty Twin then, which is probably not important.

twins/mongodb/twins.go
```diff
@@ -90,7 +90,7 @@ func (tr *twinRepository) RetrieveByAttribute(ctx context.Context, channel, subt
 			"_id": 0,
 		},
 	}
-	match := bson.M{
+	match1 := bson.M{
```
manuio: Why `match1`?

twins/service.go
drasko: You are right, this does not return 2 values, I did not see.

Sorry for this remark, please ignore it.

cmd/twins/main.go
dborovcanin: The purpose of these functions inside `main.go` is to simplify the main method itself, so I would vote to keep it like it is.

twins/redis/twins.go
```diff
@@ -0,0 +1,101 @@
...
+// NewTwinCache returns redis twin cache implementation.
+func NewTwinCache(client *redis.Client) twins.TwinCache {
+	return &twinCache{
+		client: client,
+	}
+}
+
+func (tc *twinCache) Save(_ context.Context, twin twins.Twin) error {
+	if len(twin.Definitions) < 1 {
+		return nil
```
mteodor: isnt this an error?

twins/mocks/service.go
```diff
@@ -18,21 +17,20 @@ const (
...
+func CreateDefinition(channels []string, subtopics []string) twins.Definition {
 	var def twins.Definition
-	for i, v := range names {
-		id, _ := uuid.New().ID()
+	for i := range channels {
 		attr := twins.Attribute{
-			Name:         v,
-			Channel:      id,
+			Channel:      channels[i],
 			Subtopic:     subtopics[i],
```
manuio, line 33: Channels and subtopics slices have always the same size?

darkodraskovic: Yes, the idea is, you send to arrays of the same length with names of channels and subtopics. Maybe it's not the best solution... This is only a mock.

twins/redis/twins.go
darkodraskovic: No. If twin does not have a definition, we don't need to save anything.

darkodraskovic: It's only a precaution.

twins/mocks/service.go
darkodraskovic: Yes, the idea is, you send to arrays of the same length with names of channels and subtopics. Maybe it's not the best solution... This is only a mock.

twins/twins.go
```diff
@@ -76,3 +76,18 @@ type TwinRepository interface {
...
 	Remove(ctx context.Context, id string) error
 }
+
+// TwinCache contains twin caching interface.
+type TwinCache interface {
+	// Save stores twin ID as element of channel-subtopic keyed set and vice versa.
+	Save(context.Context, Twin) error
+
+	// SaveIDs stores twin IDs as elements of channel-subtopic keyed set and vice versa.
+	SaveIDs(_ context.Context, channel, subtopic string, ids []string) error
```
manuio: why not `ctx context.Context` here?

twins/tracing/twins.go
```diff
@@ -85,6 +86,52 @@ func (trm twinRepositoryMiddleware) Remove(ctx context.Context, id string) error
 	return trm.repo.Remove(ctx, id)
 }
 
+type twinCacheMiddleware struct {
+	tracer opentracing.Tracer
+	cache  twins.TwinCache
+}
+
+// TwinCacheMiddleware tracks request and their latency, and adds spans
+// to context.
```
manuio: I would keep a single line for this comment

twins/service.go
```diff
@@ -262,10 +268,22 @@ func (ts *twinsService) ListStates(ctx context.Context, token string, offset uin
 }
 
 func (ts *twinsService) SaveStates(msg *messaging.Message) error {
-	ids, err := ts.twins.RetrieveByAttribute(context.TODO(), msg.Channel, msg.Subtopic)
+	var ids []string
+
+	ctx := context.TODO()
+	channel, subtopic := msg.Channel, msg.Subtopic
+	ids, err := ts.twinCache.IDs(ctx, channel, subtopic)
+
```
manuio: You can remove this blank line

drasko (dismissed):
LGTM

twins/redis/twins.go
```diff
@@ -0,0 +1,121 @@
...
+var ErrRedisTwinSave = errors.New("saving twin in redis cache error")
+
+// ErrRedisTwinUpdate indicates error while saving Twin in redis cache
+var ErrRedisTwinUpdate = errors.New("updating twin in redis cache error")
+
+// ErrRedisTwinIDs indicates error while geting Twin IDs from redis cache
+var ErrRedisTwinIDs = errors.New("get twin id from redis cache error")
+
+// ErrRedisTwinRemove indicates error while removing Twin from redis cache
+var ErrRedisTwinRemove = errors.New("remove twin from redis cache error")
```
manuio: Better use `failed to` in error messages.

twins/redis/twins_test.go
```diff
@@ -0,0 +1,298 @@
...
+	}{
+		{
+			desc:     "Save ids to cache",
+			channel:  channels[0],
+			subtopic: subtopics[0],
+			ids:      twinIDs,
+			err:      nil,
+		},
+		{
+			desc:     "Save emtpy ids array to cache",
```
manuio: fix `emtpy`

manuio (changes requested):

twins/service.go
```diff
@@ -205,7 +207,10 @@ func (ts *twinsService) UpdateTwin(ctx context.Context, token string, twin Twin,
 	id = twin.ID
 	b, err = json.Marshal(tw)
 
-	return nil
+	if err := ts.twinCache.Remove(ctx, id); err != nil {
```
mteodor: if remove succeeds and save doesn't we have lost the data, right?

darkodraskovic: Forgot to fix that :)

cmd/twins/main.go
```diff
@@ -65,6 +70,9 @@ const (
 	envDB              = "MF_TWINS_DB"
 	envDBHost          = "MF_TWINS_DB_HOST"
 	envDBPort          = "MF_TWINS_DB_PORT"
+	envCacheURL        = "MF_TWINS_CACHE_URL"
+	envCachePass       = "MF_TWINS_CACHE_PASS"
+	envCacheDB         = "MF_TWINS_CACHE_DB"
```
manuio, line 75: Add this envars where in .env and docker-compose

manuio (changes requested):

twins/mocks/service.go
```diff
@@ -42,18 +40,12 @@ func CreateDefinition(names []string, subtopics []string) twins.Definition {
...
-			BaseTime: float64(time.Now().Unix()),
-			Time:     float64(i),
-			Value:    nil,
-		}
-		recs = append(recs, rec)
+func CreateSenML(n int, recs []senml.Record) {
+	for i, rec := range recs {
+		rec.BaseTime = float64(time.Now().Unix())
+		rec.Time = float64(i)
+		rec.Value = nil
```
manuio, line 46: Maybe you can also use `float64(i)` here.

twins/twins.go
```diff
@@ -76,3 +76,21 @@ type TwinRepository interface {
...
+	SaveIDs(ctx context.Context, channel, subtopic string, ids []string) error
+
+	// Update updates update twin id and channel-subtopic attributes mapping
+	Update(_ context.Context, twin Twin) error
+
+	// ID returns twin IDs for given attribute.
+	IDs(ctx context.Context, channel, subtopic string) ([]string, error)
+
+	// Removes twin from cache.
+	Remove(context.Context, string) error
```
manuio: Use named args. For example `Remove(ctx context.Context, id string) error`

manuio (changes requested):

twins/mocks/service.go
darkodraskovic: I've tried many things. Problem is, rec.Value is a pointer. It's value gets somehow lost in "transition". Anyway, we don't really care for the values in these tests - we're not testing for the same values.

dborovcanin (dismissed):

twins/twins.go
```diff
@@ -71,8 +71,26 @@ type TwinRepository interface {
 	RetrieveByAttribute(ctx context.Context, channel, subtopic string) ([]string, error)
 
 	// RetrieveAll retrieves the subset of twins owned by the specified user.
-	RetrieveAll(context.Context, string, uint64, uint64, string, Metadata) (Page, error)
+	RetrieveAll(ctx context.Context, owner string, offset, limit uint64, name string, metadata Metadata) (Page, error)
 
 	// Remove removes the twin having the provided identifier.
 	Remove(ctx context.Context, id string) error
```
manuio: Definition is `func (tcm twinCacheMiddleware) Remove(ctx context.Context, twinID string)`. You could use `twinID` here

manuio (dismissed):
LGTM

manuio (dismissed):
LGTM

drasko (dismissed):
LGTM

manuio (dismissed):
LGTM

twins/redis/twins.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/mainflux/mainflux/pkg/errors"
+	"github.com/mainflux/mainflux/twins"
+)
+
+const (
+	prefix = "twin"
+)
+
+// ErrRedisTwinSave indicates error while saving Twin in redis cache
+var ErrRedisTwinSave = errors.New("failed to save twin in redis cache")
```
dborovcanin: You can wrap errors in one `var` block.

twins/redis/twins_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+	"github.com/mainflux/mainflux/pkg/uuid"
+	"github.com/mainflux/mainflux/twins"
+	"github.com/mainflux/mainflux/twins/mocks"
+	"github.com/mainflux/mainflux/twins/redis"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var subtopics = []string{"engine", "chassis", "wheel_2"}
```
dborovcanin: Same here, one `var` block.

twins/redis/twins_test.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	for _, tc := range cases {
+		err := twinCache.Update(ctx, tc.twin)
+		assert.Nil(t, err, fmt.Sprintf("%s: expected %s got %s", tc.desc, tc.err, err))
+
+		attr := tc.twin.Definitions[0].Attributes[0]
+		ids, err := twinCache.IDs(ctx, attr.Channel, attr.Subtopic)
+		require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+
```
dborovcanin: You can remove this empty line to be consistent with other tests.

twins/service.go
```diff
@@ -252,20 +254,31 @@ func (ts *twinsService) ListTwins(ctx context.Context, token string, offset uint
...
-	ids, err := ts.twins.RetrieveByAttribute(context.TODO(), msg.Channel, msg.Subtopic)
+	var ids []string
+
+	ctx := context.TODO()
+	channel, subtopic := msg.Channel, msg.Subtopic
+	ids, err := ts.twinCache.IDs(ctx, channel, subtopic)
 	if err != nil {
 		return err
 	}
+	if (len(ids)) < 1 {
```
dborovcanin:
```go
if len(ids) < 1 {...}
```

twins/service.go
```diff
@@ -252,20 +254,31 @@ func (ts *twinsService) ListTwins(ctx context.Context, token string, offset uint
...
+		ids, err = ts.twins.RetrieveByAttribute(ctx, channel, subtopic)
+		if err != nil {
+			return err
+		}
+		ts.twinCache.SaveIDs(ctx, channel, subtopic, ids)
```
dborovcanin: If `len(ids) < 1`, you can do `return ts.twinCache.SaveIDs(ctx, channel, subtopic, ids)`, so that you don't ignore possible error here; there is no need to proceed to iteration over an empty array, as well.

dborovcanin (changes requested):

dborovcanin (approved):

manuio (approved):
LGTM

PR #1187, opened 2020-06-01
MF-1177 - Implement caching in MQTT adapter

### LINKED ISSUE #1177: Implement caching in MQTT adapter
**FEATURE REQUEST**

Currently Mainflux uses Redis chache connected to Things service: a gRPC request is send to Things service upon every CONNECT, PUBLISH or SUBSCRIBE and is processed by Things service in a manner that data is first looksed for in Redis and then upon cache miss in PostgresDB.

Because MQTT is the work-horse of Mainflux, and because once established connection will be probably re-established by the same device (who loose connectivity), and publish/subscribe will come from the same device many times in the row from the same `thing` on the same `channel`, it makes sense that cache is moved directly to an adapter in a form of map or some of the fast embedded DBs - like [Bbolt](https://github.com/etcd-io/bbolt) or [BadgerDB](https://github.com/dgraph-io/badger) (as a side-note Bolt seems to have much lower memory footprint).

It would be good that this caching be implemented in a form of a library, so that it can be reused in other adapters (CoAP comes in mind).

If than that way, we should consider removing Redis completely, for several reasons:
- C code, harder compilation (especially cross-compilation), no Windows port
- NATS 2.0 can replace RedisStreams for ES functionality

This would significantly simplify our stack. 

Must-have

dborovcanin:
### What does this do?
This pull request adds caching Auth requests to the Mainflux MQTT adapter.

This pull request resolves #1177.

There are no such changes. The basic idea is described in [this comment](https://github.com/mainflux/mainflux/issues/1177#issuecomment-632223386)

No.

Yes.


cmd/mqtt/main.go
```diff
@@ -85,6 +86,13 @@ const (
 	defESURL  = "localhost:6379"
 	defESPass = ""
 	defESDB   = "0"
+	// AUTH
+	envAuthURL  = "MF_MQTT_ADAPTER_AUTH_URL"
+	envAuthPass = "MF_MQTT_ADAPTER_AUTH_PASS"
+	envAuthDB   = "MF_MQTT_ADAPTER_AUTH_DB"
+	defAuthURL  = "localhost:6379"
```
drasko: This is actually Cache, and should be called like this IMHO - to avoid confusion with Auth service

cmd/mqtt/main.go
```diff
@@ -85,6 +86,13 @@ const (
 	defESURL  = "localhost:6379"
 	defESPass = ""
 	defESDB   = "0"
+	// AUTH
+	envAuthURL  = "MF_MQTT_ADAPTER_AUTH_URL"
```
drasko: Looks like this is missing in the README

mqtt/auth/client.go
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) Mainflux
```
drasko, line 1: This should be an independent lib, shared between all adapters (in the future). It should wrap Things gRPC client, preceded by Redis cache query.

drasko (changes requested):

cmd/mqtt/main.go
```diff
@@ -35,35 +36,30 @@ const (
...
-	defHTTPHost       = "0.0.0.0"
 	defHTTPPort       = "8080"
 	defHTTPScheme     = "ws"
 	defHTTPTargetHost = "localhost"
 	defHTTPTargetPort = "8080"
 	defHTTPTargetPath = "/mqtt"
-	envHTTPHost       = "MF_MQTT_ADAPTER_WS_HOST"
 	envHTTPPort       = "MF_MQTT_ADAPTER_WS_PORT"
-	envHTTPScheme     = "MF_MQTT_ADAPTER_WS_SCHEMA"
+	envHTTPScheme     = "MF_MQTT_ADAPTER_WS_SCHEME"
```
drasko: I think that this is unnecessary option in general and should always be hardcoded to `ws` in mProxy

cmd/mqtt/main.go
```diff
@@ -35,35 +36,30 @@ const (
...
 	envHTTPTargetHost = "MF_MQTT_ADAPTER_WS_TARGET_HOST"
 	envHTTPTargetPort = "MF_MQTT_ADAPTER_WS_TARGET_PORT"
 	envHTTPTargetPath = "MF_MQTT_ADAPTER_WS_TARGET_PATH"
 	// Things
 	defThingsAuthURL     = "localhost:8181"
-	defThingsAuthTimeout = "1" // in seconds
+	defThingsAuthTimeout = "1s" // 1 second
```
drasko, line 58: Why is this extension `s` needed?

mqtt/README.md
```diff
@@ -0,0 +1,111 @@
...
+default values.
+
+| Variable                          | Description                                            | Default               |
+| --------------------------------- | ------------------------------------------------------ | --------------------- |
+| MF_MQTT_ADAPTER_LOG_LEVEL         | mProxy Log level                                       | error                 |
+| MF_MQTT_ADAPTER_MQTT_PORT         | mProxy port                                            | 1883                  |
+| MF_MQTT_ADAPTER_MQTT_TARGET_HOST  | MQTT broker host                                       | 0.0.0.0               |
+| MF_MQTT_ADAPTER_MQTT_TARGET_PORT  | MQTT broker port                                       | 1883                  |
+| MF_MQTT_ADAPTER_WS_PORT           | mProxy MQTT iver WS port                               | 8080                  |
+| MF_MQTT_ADAPTER_WS_SCHEME         | MQTT over WS protocol scheme                           | ws                    |
```
drasko: Not really needed. Should be removed even from mProxy.

drasko (changes requested):

cmd/mqtt/main.go
dborovcanin: Because it specifies the time unit. It's clear from the value that this is 1 second. On the opposite, having 1 is not self-documenting.

cmd/mqtt/main.go
dborovcanin: I agree. If we add TLS termination on the MQTT adapter side, we can update. At the moment, this is a useless option that we'll never change form its default value. I'm removing it.

cmd/mqtt/main.go
drasko: Very interesting, I did not know about this `ParseDuration()`. Nicely solved.

mqtt/auth/client.go
drasko: We can merge this for a start, then move several libs in `pkg``

drasko (approved):
LGTM

PR #1207, opened 2020-06-11
MF-1198 - Add errors package tests

### LINKED ISSUE #1198: Add errors package tests
**FEATURE REQUEST**

Errors package from `/pkg` is missing unit tests.

This is a must-have.

darkodraskovic:
Add unit tests to `errors` package from `pkg/`.

Resolves #1198

### DIFF
--- a/pkg/errors/errors_test.go
+++ b/pkg/errors/errors_test.go
@@ -5,6 +5,7 @@ package errors_test
 
 import (
 	"fmt"
+	"math/rand"
 	"strconv"
 	"testing"
 
@@ -34,7 +35,7 @@ func (ce *customError) Error() string {
 	return ce.msg
 }
 
-func TestWrap(t *testing.T) {
+func TestError(t *testing.T) {
 	cases := []struct {
 		desc string
 		err  error
@@ -63,7 +64,8 @@ func TestWrap(t *testing.T) {
 	}
 
 	for _, tc := range cases {
-		assert.Equal(t, tc.msg, tc.err.Error(), fmt.Sprintf("%s: expected %s got %s\n", tc.desc, tc.err, tc.err.Error()))
+		errMsg := tc.err.Error()
+		assert.Equal(t, tc.msg, errMsg, fmt.Sprintf("%s: expected %s got %s\n", tc.desc, tc.msg, errMsg))
 	}
 }
 
@@ -96,10 +98,39 @@ func TestContains(t *testing.T) {
 	}
 	for _, tc := range cases {
 		contains := errors.Contains(tc.container, tc.contained)
-		cntStr := strconv.FormatBool(contains)
-		assert.Equal(t, true, contains, fmt.Sprintf("%s: expected %s got %s\n", tc.desc, "true", cntStr))
+		assert.Equal(t, true, contains, fmt.Sprintf("%s: expected %v to contain %v\n", tc.desc, tc.container, tc.contained))
 	}
 
+}
+
+func TestWrap(t *testing.T) {
+	cases := []struct {
+		desc  string
+		level int
+	}{
+		{
+			desc:  "level 1 wrap",
+			level: 1,
+		},
+		{
+			desc:  "level 5 wrap",
+			level: 5,
+		},
+		{
+			desc:  "level 10 wrap",
+			level: 10,
+		},
+	}
+
+	for _, tc := range cases {
+		err := wrap(tc.level)
+		msg := message(tc.level)
+		errMsg := err.Error()
+		assert.Equal(t, msg, errMsg, fmt.Sprintf("%s: expected %s got %s\n", tc.desc, msg, errMsg))
+		contained := errors.New(strconv.Itoa(rand.Intn(tc.level)))
+		contains := errors.Contains(err, contained)
+		assert.Equal(t, true, contains, fmt.Sprintf("%s: expected %v to contain %v\n", tc.desc, err, contained))
+	}
 }
 
 func wrap(n int) error {
--- a/pkg/errors/errors_test.go
+++ b/pkg/errors/errors_test.go
@@ -1,0 +1,114 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package errors_test
+
+import (
+	"fmt"
+	"strconv"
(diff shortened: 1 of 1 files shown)


blokovi:
Did you consider to include test cases when error is `nil` (for any of these tests)?

pkg/errors/errors_test.go
```diff
@@ -0,0 +1,149 @@
...
+func wrap(n int) error {
+	if n < 1 {
+		return errors.New(strconv.Itoa(n))
+	}
+	return errors.Wrap(errors.New(strconv.Itoa(n)), wrap(n-1))
+}
+
+func message(level int) string {
+	msg := strconv.Itoa(level)
+	for i := level - 1; i >= 0; i-- {
```
manuio: I would prefer `for i := 0; i < level; i++ {`

pkg/errors/errors_test.go
```diff
@@ -0,0 +1,149 @@
...
+		errMsg := err.Error()
+		assert.Equal(t, msg, errMsg, fmt.Sprintf("%s: expected %s got %s\n", tc.desc, msg, errMsg))
+		contained := errors.New(strconv.Itoa(rand.Intn(tc.level)))
+		contains := errors.Contains(err, contained)
+		assert.Equal(t, true, contains, fmt.Sprintf("%s: expected %v to contain %v\n", tc.desc, err, contained))
+	}
+}
+
+func wrap(n int) error {
+	if n < 1 {
```
manuio: I would use n == 0

pkg/errors/errors_test.go
darkodraskovic: The order is not relevant. However, semantically, we expect that higher number errors wrap lower number errors. This function corresponds to the `wrap` function, i.e. produces an error message of a n level wrapped error.

pkg/errors/errors_test.go
```diff
@@ -0,0 +1,149 @@
...
+		},
+		{
+			desc: "level 2 wrapped error",
+			err:  errors.Wrap(err1, errors.Wrap(err2, err3)),
+			msg:  fmt.Sprintf("%s : %s : %s", msg1, msg2, msg3),
+		},
+		{
+			desc: fmt.Sprintf("level %d wrapped error", level),
+			err:  wrap(level),
+			msg:  message(level),
```
manuio, line 48: Why you are not using error messages here?

manuio (changes requested):

pkg/errors/errors_test.go
darkodraskovic: Function `message` produces error message of n level wrapped message. We use this error message to compare it with the error of an actually wrapped error with func `err`.

pkg/errors/errors_test.go
manuio: This is more readable: 
```
for i := 0; i < level; i++ {
    msg = fmt.Sprintf("%s : %s", msg, strconv.Itoa(level - i))
}
```

pkg/errors/errors_test.go
darkodraskovic: Although in this case it is not possible, it is recommended to test for less then if the exact number is not important. For example, if you decrement twice, you'll be at the `n == -1` and that'll cause stack overflow. However, in your solution intention is more clear.

darkodraskovic:
> Did you consider to include test cases when error is `nil` (for any of these tests)?
Actually, a very good idea :D and an omission from my side...

pkg/errors/errors_test.go
darkodraskovic: This part is tricky for me `strconv.Itoa(level - i)`. Solution is nice, however... However, it does not work out of the  box.

darkodraskovic: You need this in order to work `msg = fmt.Sprintf("%s : %s", msg, strconv.Itoa(level - i - 1))`. It's kind of ugly.

pkg/errors/errors_test.go
manuio: What I mean is that for other tests you have an error message. You could have a funcrtion `createWrap(msg, level)` that you could use to generate wrapped error in all your tests. For example, you could replace `fmt.Sprintf("%s : %s : %s", msg1, msg2, msg3)`  by `createWrap(msg, 3)`

pkg/errors/errors_test.go
manuio: Start by 1 then:
```
for i := 1; i < level; i++ {
    msg = fmt.Sprintf("%s : %s", msg, strconv.Itoa(level - i))
}
```

darkodraskovic: It does not work. You need to do it like this: `msg = fmt.Sprintf("%s : %s", msg, strconv.Itoa(level - i - 1))`. However, maybe a sign of a poor design :)

pkg/errors/errors_test.go
darkodraskovic: Yeah, maybe a unified func would be better. For now, I've replaced `fmt.Sprintf("%s : %s : %s", msg1, msg2, msg3)` with `message` func.

pkg/errors/errors_test.go
manuio: I'm not sure then. But the inversed for loop can be confusing

pkg/errors/errors_test.go
```diff
@@ -67,31 +59,66 @@ func TestContains(t *testing.T) {
 		desc      string
 		container error
 		contained error
+		contains  bool
 	}{
+		{
+			desc:      "nil contains nil",
+			container: nil,
+			contained: nil,
+			contains:  true,
```
manuio, line 69: I would use `err`, `contained` and `res` or something easier to differentiate

darkodraskovic: IMO, this is a nice triplet. You know instantly which error contains, which error is contained and what should be result.

darkodraskovic: @manuio I also use this "terminology" in `TestWrap` and also a similar double term like wrapper and wrapped. IMO, it's very clear what this all means, while `err` and `res` are too generic.

blokovi (approved):

manuio (approved):
LGTM

PR #1208, opened 2020-06-16
NOISSUE - Use `pgcrypto` instead `uuid-ossp` for UUIDs generation 

sprql:
According to 
https://www.postgresql.org/docs/current/uuid-ossp.html

```
Note
If you only need randomly-generated (version 4) UUIDs, consider using the gen_random_uuid() function from the pgcrypto module instead.
```

### DIFF
--- a/users/postgres/init.go
+++ b/users/postgres/init.go
@@ -63,9 +63,9 @@ func migrateDB(db *sqlx.DB) error {
 			{
 				Id: "users_3",
 				Up: []string{
-					`CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+					`CREATE EXTENSION IF NOT EXISTS "pgcrypto";
 					ALTER TABLE IF EXISTS users ADD COLUMN IF NOT EXISTS
-					id UUID DEFAULT uuid_generate_v4()`,
+					id UUID NOT NULL DEFAULT gen_random_uuid()`,
 				},
 			},
 		},


drasko:
@manuio @dusanb94 @mteodor please check

manuio:
@sprql This looks good. I didn't understand what's the problem with `uuid_generate_v4` but the doc is clear about the recommendation.  Just give me some time to test before to approve.

nmarcetic (dismissed):

users/postgres/init.go
```diff
@@ -63,9 +63,9 @@ func migrateDB(db *sqlx.DB) error {
 			{
 				Id: "users_3",
 				Up: []string{
-					`CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+					`CREATE EXTENSION IF NOT EXISTS "pgcrypto";
 					ALTER TABLE IF EXISTS users ADD COLUMN IF NOT EXISTS
-					id UUID DEFAULT uuid_generate_v4()`,
+					id UUID DEFAULT gen_random_uuid()`,
```
dborovcanin: Before I approve, can you just set `id` column to NOT NULL (`id UUID NOT NULL DEFAULT uuid_generate_v4()`), as well? We didn't include this change to any release, so it's safe to change this migration instead of adding another one.

dborovcanin (changes requested):

users/postgres/init.go
sprql: Good point!

dborovcanin (approved):

drasko (approved):
LGTM

PR #1211, opened 2020-06-29
NOISSUE - Remove unused `MF_THINGS_SECRET` env var

sprql:


dborovcanin (approved):

drasko:
@manuio @dusanb94 @mteodor @nmarcetic where is the key for signing JWTs kept currently? Is it hard-coded? Shouldn't we make this configurable?

dborovcanin:
> @manuio @dusanb94 @mteodor @nmarcetic where is the key for signing JWTs kept currently? Is it hard-coded? Shouldn't we make this configurable?
@drasko The key is [here](https://github.com/mainflux/mainflux/blob/ecff0664218356f5ddbad32a7b8cacafbcaa8be2/cmd/authn/main.go#L63). It's not hardcoded, we can configure it using env vars.

manuio (approved):
LGTM

PR #1214, opened 2020-07-06
NOISSUE - Add subtopic wildcard for twin attribute's definition

darkodraskovic:
### What does this do?
Twin stores semantic definition of a data system. Such definition consists of attributes which are uniquely defined by the combination of channel and subtopic. Currently, we can only define attribute to match the combination of an empty or non-empty subtopic and channel. However, we don't have means to define attribute to match all the combinations of a certain channel and subtopics, i.e. we don't have means to store to states all message of a channel irrespective of the subtopic used to send messages. Adition of a wildcard subtopic enables attribute to pick up all messages sent via it's channel.

N/A

Yes

Yes

### DIFF
--- a/twins/README.md
+++ b/twins/README.md
@@ -27,7 +27,6 @@ default values.
 | MF_TWINS_CLIENT_TLS        | Flag that indicates if TLS should be turned on                       | false                 |
 | MF_TWINS_CA_CERTS          | Path to trusted CAs in PEM format                                    |                       |
 | MF_TWINS_CHANNEL_ID        | NATS notifications channel id                                        |                       |
-| MF_TWINS_SUBTOPIC_WILDCARD | Subtopic wildcard for attribute's definition                         | #                     |
 | MF_NATS_URL                | Mainflux NATS broker URL                                             | nats://localhost:4222 |
 | MF_AUTHN_GRPC_URL          | AuthN service gRPC URL                                               | localhost:8181        |
 | MF_AUTHN_GRPC_TIMEOUT      | AuthN service gRPC request timeout in seconds                        | 1s                    |
@@ -64,7 +63,6 @@ services:
       MF_TWINS_CLIENT_TLS: [Flag that indicates if TLS should be turned on]
       MF_TWINS_CA_CERTS: [Path to trusted CAs in PEM format]
       MF_TWINS_CHANNEL_ID: [NATS notifications channel id]
-      MF_TWINS_SUBTOPIC_WILDCARD: [Subtopic wildcard for attribute's definition]
       MF_NATS_URL: [Mainflux NATS broker URL]
       MF_AUTHN_GRPC_URL: [AuthN service gRPC URL]
       MF_AUTHN_GRPC_TIMEOUT: [AuthN service gRPC request timeout in seconds]
@@ -101,7 +99,6 @@ MF_TWINS_SINGLE_USER_TOKEN: [User token for single user mode] \
 MF_TWINS_CLIENT_TLS: [Flag that indicates if TLS should be turned on] \
 MF_TWINS_CA_CERTS: [Path to trusted CAs in PEM format] \
 MF_TWINS_CHANNEL_ID: [NATS notifications channel id] \
-MF_TWINS_SUBTOPIC_WILDCARD: [Subtopic wildcard for attribute's definition] \
 MF_NATS_URL: [Mainflux NATS broker URL] \
 MF_AUTHN_GRPC_URL: [AuthN service gRPC URL] \
 MF_AUTHN_GRPC_TIMEOUT: [AuthN service gRPC request timeout in seconds] \
--- a/twins/README.md
+++ b/twins/README.md
@@ -26,7 +26,7 @@ default values.
 | MF_TWINS_SINGLE_USER_TOKEN | User token for single user mode that should be passed in auth header |                       |
 | MF_TWINS_CLIENT_TLS        | Flag that indicates if TLS should be turned on                       | false                 |
 | MF_TWINS_CA_CERTS          | Path to trusted CAs in PEM format                                    |                       |
-| MF_TWINS_CHANNEL_ID        | NATS notifications channel id                                        |                       |
+| MF_TWINS_CHANNEL_ID        | NATS notifications channel ID                                        |                       |
 | MF_NATS_URL                | Mainflux NATS broker URL                                             | nats://localhost:4222 |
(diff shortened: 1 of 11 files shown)


twins/README.md
```diff
@@ -26,11 +26,11 @@ default values.
 | MF_TWINS_SINGLE_USER_TOKEN | User token for single user mode that should be passed in auth header |                       |
 | MF_TWINS_CLIENT_TLS        | Flag that indicates if TLS should be turned on                       | false                 |
 | MF_TWINS_CA_CERTS          | Path to trusted CAs in PEM format                                    |                       |
-| MF_TWINS_MQTT_URL          | Mqtt broker URL for twin CRUD and states update notifications        | tcp://localhost:1883  |
-| MF_TWINS_CHANNEL_ID        | Mqtt notifications topic                                             |                       |
+| MF_TWINS_CHANNEL_ID        | NATS notifications channel id                                        |                       |
+| MF_TWINS_SUBTOPIC_WILDCARD | Subtopic wildcard for attribute's definition                         | #                     |
```
drasko: Why this is configurable? Are you using MQTT to get these attributes? If yes, then "catch all joker can not be anything else (it can be `+` for a singe-level topic, though).

twins/redis/twins.go
```diff
@@ -74,6 +74,11 @@ func (tc *twinCache) IDs(_ context.Context, channel, subtopic string) ([]string,
 	if err != nil {
 		return nil, errors.Wrap(ErrRedisTwinIDs, err)
 	}
+	idsWildcard, err := tc.client.SMembers(attrKey(channel, "#")).Result()
```
drasko: Here `#` looks hard-coded, although `README` says it is configurable.

drasko (changes requested):

twins/README.md
```diff
@@ -26,11 +26,10 @@ default values.
 | MF_TWINS_SINGLE_USER_TOKEN | User token for single user mode that should be passed in auth header |                       |
 | MF_TWINS_CLIENT_TLS        | Flag that indicates if TLS should be turned on                       | false                 |
 | MF_TWINS_CA_CERTS          | Path to trusted CAs in PEM format                                    |                       |
-| MF_TWINS_MQTT_URL          | Mqtt broker URL for twin CRUD and states update notifications        | tcp://localhost:1883  |
-| MF_TWINS_CHANNEL_ID        | Mqtt notifications topic                                             |                       |
+| MF_TWINS_CHANNEL_ID        | NATS notifications channel id                                        |                       |
```
manuio: `channel ID`

twins/README.md
```diff
@@ -63,8 +62,7 @@ services:
       MF_TWINS_SINGLE_USER_TOKEN: [User token for single user mode]
       MF_TWINS_CLIENT_TLS: [Flag that indicates if TLS should be turned on]
       MF_TWINS_CA_CERTS: [Path to trusted CAs in PEM format]
-      MF_TWINS_MQTT_URL: [Mqtt broker URL for twin CRUD and states]
-      MF_TWINS_CHANNEL_ID: [Mqtt notifications topic]
+      MF_TWINS_CHANNEL_ID: [NATS notifications channel id]
```
manuio: `channel ID`

twins/mocks/states.go
```diff
@@ -55,12 +55,11 @@ func (srm *stateRepositoryMock) RetrieveAll(ctx context.Context, offset uint64,
 	srm.mu.Lock()
 	defer srm.mu.Unlock()
 
-	items := make([]twins.State, 0)
-
 	if limit <= 0 {
 		return twins.StatesPage{}, nil
 	}
 
+	items := make([]twins.State, 0)
```
manuio: `var items []twins.State`

twins/redis/twins.go
```diff
@@ -74,6 +74,11 @@ func (tc *twinCache) IDs(_ context.Context, channel, subtopic string) ([]string,
 	if err != nil {
 		return nil, errors.Wrap(ErrRedisTwinIDs, err)
 	}
+	idsWildcard, err := tc.client.SMembers(attrKey(channel, twins.SubtopicWildcard)).Result()
```
manuio, line 77: `SMembers` could have better naming

manuio (changes requested):

twins/redis/twins.go
darkodraskovic: Please elaborate on this. `SMembers` is redis function.

dborovcanin (approved):

manuio (approved):
LGTM

twins/mocks/service.go
```diff
@@ -38,6 +41,15 @@ func CreateDefinition(channels []string, subtopics []string) twins.Definition {
 	return def
 }
 
+// CreateTwin creates twin
+func CreateTwin(channels []string, subtopics []string) twins.Twin {
+	id++
```
drasko, line 46: I do not know if having a global var would be a good idea here. I do not go in deeper analysis, but there is some "code smell". I would prefer some internal structure, created in the `main()` via `New()` fnc, then passed back here... Or something like this.

drasko (changes requested):

twins/mocks/service.go
darkodraskovic: This `id` is not a global var, it's a package level var, not exported, used for the sole purpose of generating unique id's for the mock entitites that we use to test service. It's rather so called closure variable. The function users can't know anything about it.

manuio: You can probably use this UUID mock here: https://github.com/mainflux/mainflux/blob/master/pkg/uuid/mock.go

manuio: Looks like you already have it here https://github.com/mainflux/mainflux/blob/e8852a16ec984045094612ae16512d5c8f082885/twins/mocks/service.go#L24

drasko: @darkodraskovic sorry I did not see this was for the mock. I will approve then. I let you take a look at this Manu's remark optionally, but this is good to me.

drasko (approved):
LGTM

twins/mocks/service.go
darkodraskovic: @manuio i purposefully didn't use the mock uuid generator, cause i don't want to handle uuid error creation. I really need only mock id's, can be really anything, it's just for the purpose of differentiating the twins, so I just give them unique serial number like this - no possibility of error, so no need for error handling.

PR #1215, opened 2020-07-07
NOISSUE: fix envs for nginx

mteodor:

### What does this do?

This fixes setting ports for nginx with `.env` file

### DIFF
--- a/docker/nginx/entrypoint.sh
+++ b/docker/nginx/entrypoint.sh
@@ -13,6 +13,8 @@ envsubst '
     ${MF_USERS_HTTP_PORT}
     ${MF_THINGS_HTTP_PORT}
     ${MF_HTTP_ADAPTER_PORT}
+    ${MF_NGINX_MQTT_PORT}
+    ${MF_NGINX_MQTTS_PORT}
     ${MF_WS_ADAPTER_PORT}' < /etc/nginx/nginx.conf.template > /etc/nginx/nginx.conf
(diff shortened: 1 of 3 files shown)


dborovcanin (approved):

docker/nginx/entrypoint.sh
```diff
@@ -13,6 +13,8 @@ envsubst '
     ${MF_USERS_HTTP_PORT}
     ${MF_THINGS_HTTP_PORT}
     ${MF_HTTP_ADAPTER_PORT}
+    ${MF_NGINX_MQTT_PORT}
```
manuio, line 16: Why `MF_NGINX_MQTT_PORT`?

drasko: @manuio what would you suggest?

manuio: @drasko something without `NGINX`. Maybe `MF_MQTT_BROKER_PORT`

mteodor: ok, whatever you find suitable, but this is MQTT port on NGINX, and there is MQTT port on mqtt-adapter and MQTT port on mqtt broker, so I think it is not bad to denote this in naming

manuio: ok sorry, my bad

manuio (approved):
LGTM

PR #1217, opened 2020-07-10
MF-983 - Add HTTP query param to connections list endpoints to fetch disconnected Things or Channels

### LINKED ISSUE #983: Fetch not connected Things or Channels
**FEATURE REQUEST**

Fetch Channels or Things IDs not connected to a given Thing or Channel.
nice-to-have

manuio:

Resolves #983


mteodor (changes requested):
swagger needs update

dborovcanin:
@manuio Please update PR, CI is failing and the branch is out of date.

nmarcetic:
@manuio IMHO `connected` bool is much better than `disconnected`. If we have this query param, by default it should return all (booth conn/disconn) and if you pass query connected bool then filter and return filtered result.

manuio:
> @manuio IMHO `connected` bool is much better than `disconnected`. If we have this query param, by default it should return all (booth conn/disconn) and if you pass query connected bool then filter and return filtered result.

@nmarcetic do you mean to return connected by default? And disconnected if connected = false? If we return connected and disconnected we are returning all things :) I wanted a flag where the default value is false to simplify the code. Maybe we can find a better naming.

nmarcetic:
> > @manuio IMHO `connected` bool is much better than `disconnected`. If we have this query param, by default it should return all (booth conn/disconn) and if you pass query connected bool then filter and return filtered result.
> 
> @nmarcetic do you mean to return connected by default? And disconnected if connected = false? If we return connected and disconnected we are returning all things :) I wanted a flag where the default value is false to simplify the code. Maybe we can find a better naming.

@manuio Yes, return all things owned by user by default. then filter it with a query like &connected=true/false if you want. We have pagination. This would be the cleanest way.

manuio:
> > > @manuio IMHO `connected` bool is much better than `disconnected`. If we have this query param, by default it should return all (booth conn/disconn) and if you pass query connected bool then filter and return filtered result.
> > 
> > 
> > @nmarcetic do you mean to return connected by default? And disconnected if connected = false? If we return connected and disconnected we are returning all things :) I wanted a flag where the default value is false to simplify the code. Maybe we can find a better naming.
> 
> @manuio Yes, return all things owned by user by default. then filter it with a query like &connected=true/false if you want. We have pagination. This would be the cleanest way.

@nmarcetic Why would I specify a channel or thing ID (`things/<thing_id>/channels` or `channels/<channel_id>/things`) to fetch all existent things or channels? It's not really the same feature. I would keep it as it is but find a better naming. Maybe `off` instead of `disconnected`?

nmarcetic:
> `things/<thing_id>/channels` or `channels/<channel_id>/things`

@manuio Why would you specify ID (`things/<thing_id>/channels` or `channels/<channel_id>/things`) to fetch something that does not belong here :) like disconnected things/channels? I think something is definitely wrong with the design here.

Ok, keep it as is, I think `connected` bool is best (because of our internal terminology of connecting things to channel ). 
We can discuss about the design later with all @mainflux/maintainers

things/postgres/things.go
```diff
@@ -252,20 +252,33 @@ func (tr thingRepository) RetrieveAll(ctx context.Context, owner string, offset,
...
 	q := `SELECT id, name, key, metadata
-	      FROM things th
-	      INNER JOIN connections co
-		  ON th.id = co.thing_id
-		  WHERE th.owner = :owner AND co.channel_id = :channel
-		  ORDER BY th.id
-		  LIMIT :limit
-		  OFFSET :offset;`
+  FROM things th
+  INNER JOIN connections co
```
dborovcanin: Using `conn` instead of `co` is less ambiguous.

things/postgres/things.go
```diff
@@ -252,20 +252,33 @@ func (tr thingRepository) RetrieveAll(ctx context.Context, owner string, offset,
...
 
 	q := `SELECT id, name, key, metadata
-	      FROM things th
-	      INNER JOIN connections co
-		  ON th.id = co.thing_id
-		  WHERE th.owner = :owner AND co.channel_id = :channel
-		  ORDER BY th.id
-		  LIMIT :limit
-		  OFFSET :offset;`
+  FROM things th
```
dborovcanin: You can use `switch` here, it's cleaner than using `if`.

dborovcanin (changes requested):
I agree with @nmarcetic that using `connected` param is better than `disconnected`.

things/postgres/channels.go
```diff
@@ -193,20 +193,31 @@ func (cr channelRepository) RetrieveAll(ctx context.Context, owner string, offse
...
-	q := `SELECT id, name, metadata
-	      FROM channels ch
-	      INNER JOIN connections co
-		  ON ch.id = co.channel_id
-		  WHERE ch.owner = :owner AND co.thing_id = :thing
-		  ORDER BY ch.id
-		  LIMIT :limit
-		  OFFSET :offset`
+	q := `SELECT id, name, metadata FROM channels ch
+    INNER JOIN connections co ON ch.id = co.channel_id
```
mteodor: this doesnt look properly aligned

things/postgres/things.go
```diff
@@ -252,20 +252,35 @@ func (tr thingRepository) RetrieveAll(ctx context.Context, owner string, offset,
...
 
-	q := `SELECT id, name, key, metadata
-	      FROM things th
-	      INNER JOIN connections co
-		  ON th.id = co.thing_id
-		  WHERE th.owner = :owner AND co.channel_id = :channel
-		  ORDER BY th.id
-		  LIMIT :limit
-		  OFFSET :offset;`
+	var q string
```
dborovcanin: Same here.

dborovcanin (changes requested):

things/postgres/channels.go
```diff
@@ -193,20 +193,48 @@ func (cr channelRepository) RetrieveAll(ctx context.Context, owner string, offse
...
 
-	q := `SELECT id, name, metadata
-	      FROM channels ch
-	      INNER JOIN connections co
-		  ON ch.id = co.channel_id
-		  WHERE ch.owner = :owner AND co.thing_id = :thing
-		  ORDER BY ch.id
-		  LIMIT :limit
-		  OFFSET :offset`
+	var q string
```
dborovcanin: You can use `var q, qc string`.

things/postgres/channels.go
```diff
@@ -193,20 +193,33 @@ func (cr channelRepository) RetrieveAll(ctx context.Context, owner string, offse
...
+	switch connected {
+	case true:
```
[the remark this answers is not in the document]
dborovcanin: Can you indent relative to `SELECT`, like:
```go
case true:
        q = `SELECT id, name, key, metadata
                FROM things th
                INNER JOIN connections conn
                ON th.id = conn.thing_id
                WHERE th.owner = :owner AND conn.channel_id = :channel
                ORDER BY th.id
                LIMIT :limit
                OFFSET :offset;`
        qc = `SELECT COUNT(*)
                 FROM things th
                 INNER JOIN connections conn
                 ON th.id = conn.thing_id
                 WHERE th.owner = $1 AND conn.channel_id = $2;`
```
It's just more readable and it's aligned to `init.go`.

mteodor (dismissed):

things/postgres/things.go
dborovcanin: Same remark for indentation.

dborovcanin (changes requested):

dborovcanin (approved):

mteodor (approved):

drasko (commented):
LGTM

PR #1221, opened 2020-07-17
NOISSUE - Fix certs update in bootstrap config and make content handling in config.toml user friendly

mteodor:
### What does this do?
fix certs update in bootstrap config
content in `config.toml` that is used for content in bootstrap configuration currently is represented as string.
representing content as map will be more readable for user.


certs/pki/vault.go
```diff
@@ -0,0 +1,162 @@
...
+// Package pki wraps vault client
+package pki
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/hashicorp/vault/api"
```
dborovcanin: Does it make sense to completely remove the Vault API client? I had a pretty bad experience with using Vault API and the client they provide is quite useless. I see that you also have to do a lot of parsing manually. I'm asking this because Vault is a giant dependency, and the only reason to have it is this client, which is not particularly useful.

certs/pki/vault.go
```diff
@@ -0,0 +1,162 @@
...
+
+	s, _ := api.ParseSecret(resp.Body)
+	cert := Cert{}
+
+	if err = mapstructure.Decode(s.Data, &cert); err != nil {
+		return Cert{}, errors.Wrap(errFailedCertDecoding, err)
+	}
+
+	// Expire time calc must be revised
+	// value doesnt look correct
```
manuio: I would use one line for this comment

certs/postgres/certs.go
```diff
@@ -0,0 +1,196 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+
```
manuio: Remove this blank line

certs/service.go
```diff
@@ -0,0 +1,303 @@
...
+	if err != nil {
+		return c, errors.Wrap(ErrUnauthorizedAccess, err)
+	}
+
+	thing, err := cs.sdk.Thing(thingID, token)
+	if err != nil {
+		return c, errors.Wrap(errFailedCertCreation, err)
+	}
+
+	// If PKIClient == nil we don't use 3rd party PKI service.
```
manuio: Please don't use code annotations for comments. btw, why you check `PKIHost` but the comment says `PKIClient`?

manuio (changes requested):

cmd/provision/main.go
```diff
@@ -169,6 +171,15 @@ func loadConfig() (provision.Config, error) {
 		return provision.Config{}, errFailSettingKeyBits
 	}
 
+	var content map[string]interface{}
+	c := mainflux.Env(envBSContent, defBSContent)
+	if c != "" {
+		err = json.Unmarshal([]byte(c), content)
+		if err != nil {
```
drasko: Inline the error check.

cmd/provision/main.go
```diff
@@ -169,6 +171,15 @@ func loadConfig() (provision.Config, error) {
 		return provision.Config{}, errFailSettingKeyBits
 	}
 
+	var content map[string]interface{}
+	c := mainflux.Env(envBSContent, defBSContent)
+	if c != "" {
```
drasko: Is this `c` used below? If not, please inline the check.

provision/service.go
```diff
@@ -152,16 +153,21 @@ func (ps *provisionService) Provision(token, name, externalID, externalKey strin
...
+		bootstrap := needsBootstrap(thing)
+
 		for _, ch := range channels {
 			chanIDs = append(chanIDs, ch.ID)
 		}
+		content, err := json.Marshal(ps.conf.Bootstrap.Content)
+		if err != nil {
+			return Result{}, errors.Wrap(ErrFailedBootstrap, err)
+		}
+		fmt.Println("content:" + string(content))
```
drasko: Remove console prints.

provision/service.go
```diff
@@ -152,16 +153,21 @@ func (ps *provisionService) Provision(token, name, externalID, externalKey strin
...
-	var bs SDK.BootstrapConfig
+	var bsConfig SDK.BootstrapConfig
 	for _, thing := range things {
-		bootstrap := false
-		if _, ok := thing.Metadata[externalIDKey]; ok {
-			bootstrap = true
-		}
 		var chanIDs []string
+
+		bootstrap := needsBootstrap(thing)
```
drasko: No need to declare this var, use `needsBootstrap(thing)` directly blow.

provision/service.go
```diff
@@ -196,6 +202,12 @@ func (ps *provisionService) Provision(token, name, externalID, externalKey strin
 			res.ClientCert[thing.ID] = cert.ClientCert
 			res.ClientKey[thing.ID] = cert.ClientKey
 			res.CACert = cert.CACert
+
+			if bootstrap {
```
drasko: Use `needsBootstrap(thing)` directly

drasko (changes requested):

drasko (dismissed):
LGTM

drasko:
@dusanb94 please review and merge if OK

certs/pki/vault.go
```diff
@@ -106,8 +106,7 @@ func (p *pkiAgent) IssueCert(cn string, ttl, keyType string, keyBits int) (Cert,
 		return Cert{}, errors.Wrap(errFailedCertDecoding, err)
 	}
 
-	// Expire time calc must be revised
-	// value doesnt look correct
+	// Expire time calc must be revised value doesnt look correct
```
dborovcanin, line 109: Can you explain this? Can we remove this comment?

docker/addons/provision/configs/config.toml
```diff
@@ -1,5 +1,56 @@
 [bootstrap]
-  content = "{\"log_level\":\"info\",\"http_port\":\"9003\",\"mqtt_url\":\"localhost:1883\",\"edgex_url\":\"http://localhost:48090/api/v1/\",\"nats_url\":\"localhost:4222\"}"
+  [bootstrap.content]
```
dborovcanin, line 2: This is way better than a single string value.

nmarcetic:
@mteodor Pls update branch.

certs/pki/vault.go
mteodor: @dusanb94  I had a problem here with converting response from vault API client to a structure, I used `mapstructure` lib since the response from `s, _ := api.ParseSecret(resp.Body)` in `s.Data` is `map[string]interface{}` , all fields were good but expiration. 
Now since we are going to remove vault api client with #1225 this piece should be refactored and since the response to be handled will be JSON i dont expect problem with converting value to appropriate data type
I will update issue #1225 not to forget to tackle this problem.

dborovcanin (dismissed):

dborovcanin (approved):

blokovi (approved):

PR #1223, opened 2020-07-20
MF-435 - Add support for env file loading

### LINKED ISSUE #435: Enable service configuration by using configuration files

hoanga:
Signed-off-by: Al Hoang <hoanga@gmail.com>

### What does this do?

Adds support for loading environment variables from a file path

This pull request resolves #435.

n/a

yes


dborovcanin (commented):
While I like the idea, have you considered using a different lib (such as https://github.com/joho/godotenv) or writing your simple parser? I don't like adding a new dependency, especially since this one does not look well maintained.

manuio:
@hoanga Let's use this issue as reference for this PR: https://github.com/mainflux/mainflux/issues/435

hoanga:
@dusanb94 one of the underlying config file parsing libraries references this library in this pr which is the basis for choosing the library in question. I understand the concern regarding keeping the number of introduced dependencies low, but with the availability of libraries that fall right into this domain, if the footprint is small enough, I would suggest considering using a more battle-tested library for this use case instead of re-implementing a (possible) buggy rewrite.

Both this env parsing library and godotenv are part of the dependency graph for this project from what I can discern from inspection. I've setup an alternate version using godotenv (see #1224). The API should remain the same, the underlying implementation is not.

dborovcanin:
> @dusanb94 one of the underlying config file parsing libraries references this library in this pr which is the basis for choosing the library in question. I understand the concern regarding keeping the number of introduced dependencies low, but with the availability of libraries that fall right into this domain, if the footprint is small enough, I would suggest considering using a more battle-tested library for this use case instead of re-implementing a (possible) buggy rewrite.
> 
> Both this env parsing library and godotenv are part of the dependency graph for this project from what I can discern from inspection. I've setup an alternate version using godotenv (see #1224). The API should remain the same, the underlying implementation is not.

@hoanga OK, makes sense. I'll close the other PR, let's keep only this. It's easier to follow the conversation. Can you add an example of using this function? Also, please check dependencies, CI is failing.

drasko:
@hoanga can you please proceed on thi PR as @dusanb94 explained? Otherwise I will close it due to inactivity.

hoanga:
@dusanb94 @drasko I have put in updates to update the branch to track latest. Below is a sample calling this function as the first steps in initialization of a program that uses environment settings to:

```

package main

import (
)

const defaultEnvPath = ".env"

func main() {
        if err := loadEnvironment(); err != nil {
                os.Exit(1)
        }
        // .. Further initializations here

        // .. Add your business logic here
}

func loadEnvironment() error {
        // Populate environment vars from file
        err := mainflux.LoadEnvFile(defaultEnvPath)
        if err != nil {
                return error
        }
}
```

dborovcanin:
> @dusanb94 @drasko I have put in updates to update the branch to track latest. Below is a sample calling this function as the first steps in initialization of a program that uses environment settings to:
> 
> ```
> 
> package main
> 
> import (
>         "github.com/mainflux/mainflux"
> )
> 
> const defaultEnvPath = ".env"
> 
> func main() {
>         if err := loadEnvironment(); err != nil {
>                 os.Exit(1)
>         }
>         // .. Further initializations here
> 
>         // .. Add your business logic here
> }
> 
> func loadEnvironment() error {
>         // Populate environment vars from file
>         err := mainflux.LoadEnvFile(defaultEnvPath)
>         if err != nil {
>                 return error
>         }
> }
> ```

@hoanga When I asked for a code example, I meant that you actually use this function in the main methods of Mainflux services. You can paste this code sample to `main.go` files in `cmd`. Also, you can move `const defaultEnvFilePath = ".env"` to `env.go` and modify `LoadEnv` to:
```go
// LoadEnvFile loads environment variables defined in an .env formatted file.
func LoadEnvFile(envFilePath string) error {
	if envFilePath == "" {
            envFilePath = defEnvFilePath
        }
        return gotenv.Load(envfilepath)
}
```

drasko:
@hoanga can you please take a look at these remarks from @dusanb94

hoanga:
@dusanb94 I'm still unclear on what you mean when you are asking for a code sample. If you are asking me to modify all of the existing mainflux services main() method to include this change, that sounds like it would be better handled in the scope of another PR (perhaps with tests). The intent for this PR is to put in the facilities to allow loading from an env file and enable services that could use env file loading as needed.

It would make more sense to show how this could be done for a brand new service that could illustrate how to use this but what new service that would be I feel is also outside of the scope of this current PR. If the mainflux authors happen to have a suggestion/direction for such a brand new service it might make sense to comment in that future PR on how to use this.

Originally, I thought having this method call would be beneficial for contexts I was working with and the mainflux maintainers as well. I will leave it up to the project maintainers how to handle this PR at this point.

manuio:
@hoanga @dusanb94 We can add here: https://github.com/mainflux/mainflux/blob/master/env.go#L12  something similar to`os.Getenv(key)`  but to read the key from a `.env` file. It would avoid to modify any main.go.

drasko:
@dusanb94 please take a look at this urgently.

dborovcanin (approved):

manuio (approved):
LGTM

PR #1229, opened 2020-09-08
NOISSUE - Fix malformed Swagger API specs

darkodraskovic:


certs/swagger.yml
```diff
@@ -19,21 +19,21 @@ paths:
...
-             - rsa_bits
-           properties:
-              thing_id:
-                type: string
-              days_valid:
-                type: string
-              rsa_bits:
-                type: integer
+        name: cert
+        description: TODO
```
manuio: TODO :)

http/swagger.yaml
```diff
@@ -51,3 +52,65 @@ paths:
...
+        description: Value Sum
+      t:
+        type: number
+        format: double
+        description: Time
+      ut:
+        type: number
+        format: double
+        description: Update Time
+  array:
```
manuio: I would call this one SenML

http/swagger.yaml
```diff
@@ -51,3 +52,65 @@ paths:
           description: Message discarded due to invalid or missing content type.
         500:
           description: Unexpected server-side error occurred.
+definitions:
+  SenML:
```
manuio: I would call this one SenML_record

readers/swagger.yml
```diff
@@ -27,7 +27,7 @@ paths:
         200:
           description: Data retrieved.
           schema:
-            $ref: "#/definitions/MessagesPage"
+            $ref: "#/definitions/MessagePage"
```
manuio: Why singular?

certs/swagger.yml
```diff
@@ -109,7 +109,7 @@ definitions:
         description: Key for the client_cert.
       issuing_ca:
         type: string
-        description: 
+        description: TODO
```
manuio: TODO

manuio (changes requested):

certs/swagger.yml
drasko: It's OK to leave `TODO`, especially if the validator demands something to be written as a value to a key.

But if we can add description, that would be better.

http/swagger.yaml
darkodraskovic: There is SenML declaration just above this declaration.

http/swagger.yaml
darkodraskovic: Ok. Got the point. Do we use snake_case or CamelCase?

readers/swagger.yml
darkodraskovic: Don't really know. I was just fixing so this thing is error-free. Didn't touch the logic or semantics.

certs/swagger.yml
darkodraskovic: @mteodor has to fix this, since I don't know what the service does exactly. As said above, "Don't really know. I was just fixing so this thing is error-free. Didn't touch the logic or semantics."

certs/swagger.yml
darkodraskovic: @mteodor has to fix this, since I don't know what the service does exactly.

http/swagger.yaml
manuio: Looks like you should use camelCase, `senml` and `senmlRecord` . Maybe even `senmlMessage` and `senmlRecord`

readers/swagger.yml
manuio: I would rename the structure and keep plural as `thingsPage` for example.

certs/swagger.yml
manuio: I would kindly ask him to help you in this PR. It can't be so long :)

manuio:
@darkodraskovic Just a detail but we are using `NOISSUE` for the title

http/swagger.yaml
```diff
@@ -51,3 +52,65 @@ paths:
...
+        format: double
+        description: Value Sum
+      t:
+        type: number
+        format: double
+        description: Time
+      ut:
+        type: number
+        format: double
+        description: Update Time
```
manuio, line 112: You can add here the parameters that are required. Something like 
```
    required:
      - n
      - t
```

manuio (changes requested):

certs/swagger.yml
mteodor: ``` description: CA Certificate that is used to issue client certs, usually intermediate.```

certs/swagger.yml
mteodor: @darkodraskovic  can you add 
 ```description: |
        Issues a certificate that is required for mTLS. To create a certificate for a thing
        provide a thing id, data identifying particular thing will be embedded into the Certificate.
        x509 and ECC certificates are supported when using when Vault is used as PKI.```

http/swagger.yaml
darkodraskovic: I'm in dilemma here, because the spec says "Value of the entry.  Optional if a Sum value is present; otherwise, it's required." So there is no singular field that is by itself required. Please, cf. https://tools.ietf.org/html/rfc8428

http/swagger.yaml
darkodraskovic: I opted here for SenMLRecord and SenMLArray, because we are using capitals for the first letter of definitions.

http/swagger.yaml
manuio: Right, you can omit `t` but I think that `v` and `n` are always required

manuio: Yes, hard to explain with a swagger

manuio (approved):
LGTM

mteodor (approved):

PR #1245, opened 2020-09-29
MF-1244 - Return UserID alongside with user Email in Identify response

### LINKED ISSUE #1244: Return UserID alongside with user Email in Identify response
**FEATURE REQUEST**

NO

We will need user id in authenticated token when introducing policies, also there is a need to change reference to user in thing db from email to user id so this would be on of intermediary steps.

must-have

dborovcanin:

### What does this do?
This pull request adds UserID to the AuthN Identify response.

This pull request resolves #1244.

Both AuthN gRPC API and the Service interface are changed. Since AuthN gRPC API is used by other services, this introduces some breaking changes to the inter-service communication.

Yes, tests are fixed according to the introduced changes.

No - only an internal API is changed.


authn.proto
```diff
@@ -39,11 +39,13 @@ message Token {
     string value = 1;
 }
 
-message UserID {
-    string value = 1;
+message UserIdentity {
```
drasko, line 42: Maybe just `User`? Like - do we know what else in the future might be sent.

Btw. do we still need e-mail for AuthN at all?

authn/api/logging.go
```diff
@@ -53,10 +53,10 @@ func (lm *loggingMiddleware) Revoke(ctx context.Context, owner, id string) (err
...
 	}(time.Now())
 
-	return lm.svc.Revoke(ctx, owner, id)
+	return lm.svc.Revoke(ctx, token, id)
 }
 
-func (lm *loggingMiddleware) Retrieve(ctx context.Context, owner, id string) (key authn.Key, err error) {
+func (lm *loggingMiddleware) Retrieve(ctx context.Context, token, id string) (key authn.Key, err error) {
 	defer func(begin time.Time) {
 		message := fmt.Sprintf("Method retrieve for key %s took %s to complete", id, time.Since(begin))
```
mteodor, line 61: token or id?

authn/postgres/init.go
```diff
@@ -47,16 +47,31 @@ func migrateDB(db *sqlx.DB) error {
...
 				Up: []string{
 					`CREATE TABLE IF NOT EXISTS keys (
-						id          UUID NOT NULL,
-						type        SMALLINT,
-						issuer      VARCHAR(254) NOT NULL,
-						issued_at   TIMESTAMP NOT NULL,
-						expires_at  TIMESTAMP,
-						PRIMARY KEY (id, issuer)
-					)`,
+                        id          UUID NOT NULL,
```
mteodor: alignment

authn/service.go
```diff
@@ -40,22 +39,24 @@ var (
 
 // Service specifies an API that must be fullfiled by the domain service
 // implementation, and all of its decorators (e.g. logging & metrics).
+// Token is a string value of the actual Key and is used to authenticate
+// an AuthN service request.
 type Service interface {
 	// Issue issues a new Key.
```
mteodor: and secret?

authn.proto
dborovcanin: I prefer to keep `UserIdentity` because it does not represent a User, but a response from the `Identify` method. I think that we can remove an email, but that requires some modifications in the Users service and it's probably better to do it in a separate PR.

authn/api/grpc/endpoint_test.go
```diff
@@ -123,43 +125,52 @@ func TestIdentify(t *testing.T) {
...
 		token string
-		id    string
+		idt   mainflux.UserIdentity
 		err   error
 		code  codes.Code
 	}{
+		{
+			desc:  "identify user with user token",
+			token: loginSecret,
+			idt:   mainflux.UserIdentity{Email: email, Id: email},
```
manuio: `Id: email` ?

drasko (commented):
@dusanb94 this looks OK to me - please take into account @manuio 's remarks and I will approve

authn.proto
dborovcanin: I created an issue for this: #1273.

drasko (approved):
LGTM

mteodor (approved):

PR #1247, opened 2020-10-01
NOISSUE - Fix group retrieval when parent id is not specified

mteodor:

When parent id is specified we retrieve group for specified id and all the group ancestors filtered by metadata - this worked properly.
When parent id retrieve groups should retrieve all groups filtered by metadata - this was broken.

### DIFF
--- a/users/postgres/groups.go
+++ b/users/postgres/groups.go
@@ -153,12 +153,16 @@ func (gr groupRepository) RetrieveAllWithAncestors(ctx context.Context, groupID 
 	if err != nil {
 		return users.GroupPage{}, errors.Wrap(errRetrieveDB, err)
 	}
+	if mq != "" {
+		mq = fmt.Sprintf("WHERE %s", mq)
+	}
 
-	subQuery := fmt.Sprintf("SELECT id, owner_id, parent_id, name, description, metadata from groups WHERE 1=1 %s", mq)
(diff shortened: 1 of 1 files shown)


users/postgres/groups.go
```diff
@@ -154,15 +154,23 @@ func (gr groupRepository) RetrieveAllWithAncestors(ctx context.Context, groupID
...
-						FROM groups
-						WHERE id = :id 
-						UNION
-							SELECT groups.id, groups.owner_id, groups.parent_id, groups.name, groups.description, groups.metadata
-							FROM groups 
-							INNER JOIN subordinates s ON s.id = groups.parent_id %s
-					) SELECT * FROM subordinates ORDER BY id LIMIT :limit OFFSET :offset`, mq)
+	subQuery := fmt.Sprintf("SELECT id, owner_id, parent_id, name, description, metadata from groups WHERE 1=1 %s", mq)
+	q := fmt.Sprintf("%s ORDER BY id LIMIT :limit OFFSET :offset", subQuery)
+	cq := fmt.Sprintf("SELECT COUNT(*) FROM groups WHERE 1=1 %s", mq)
```
dborovcanin: I understand the "hack", but I'm not sure that I like it. This can be solved with a simple `if` which is easier to understand, right?

drasko (commented):
Please edit the PR description and explain the problem it solves and why.

dborovcanin (approved):

manuio (approved):
LGTM

PR #1259, opened 2020-10-16
MF-397 - Introduce Thing Groups

mteodor:

Adding group of things


things/api/logging.go
```diff
@@ -264,3 +264,120 @@ func (lm *loggingMiddleware) Identify(ctx context.Context, key string) (id strin
 
 	return lm.svc.Identify(ctx, key)
 }
+
+func (lm *loggingMiddleware) CreateGroup(ctx context.Context, token string, group things.Group) (g things.Group, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method create_group for key %s and name %s took %s to complete", token, g.Name(), time.Since(begin))
```
manuio: Here and for all methods: `key` or `token`?

things/api/things/http/transport.go
```diff
@@ -172,6 +172,76 @@ func MakeHandler(tracer opentracing.Tracer, svc things.Service) http.Handler {
 		opts...,
 	))
 
+	r.Get("/things/:memberID/groups", kithttp.NewServer(
```
manuio, line 177: `thingID`

things/api/things/http/transport.go
```diff
@@ -172,6 +172,76 @@ func MakeHandler(tracer opentracing.Tracer, svc things.Service) http.Handler {
...
+	))
+
+	r.Delete("/groups/:groupID", kithttp.NewServer(
+		kitot.TraceServer(tracer, "delete_group")(things.DeleteGroupEndpoint(svc)),
+		things.DecodeGroupRequest,
+		encodeResponse,
+		opts...,
+	))
+
+	r.Put("/groups/:groupID/things/:memberID", kithttp.NewServer(
```
manuio, line 205: `thingID`

things/api/things/http/transport.go
```diff
@@ -172,6 +172,76 @@ func MakeHandler(tracer opentracing.Tracer, svc things.Service) http.Handler {
...
+		kitot.TraceServer(tracer, "assign_user_to_group")(things.AssignMemberToGroup(svc)),
+		things.DecodeMemberGroupRequest,
+		encodeResponse,
+		opts...,
+	))
+
+	r.Delete("/groups/:groupID/things/:memberID", kithttp.NewServer(
```
manuio, line 212: `thingID`

manuio (changes requested):

things/api/things/http/transport.go
mteodor: reason why it is `memberID` is because the methods that server this endpoint could be used for other entities that we want to group

things/api/metrics.go
```diff
@@ -192,3 +192,99 @@ func (ms *metricsMiddleware) Identify(ctx context.Context, key string) (string,
...
+		ms.latency.With("method", "members").Observe(time.Since(begin).Seconds())
+	}(time.Now())
+
+	return ms.svc.Members(ctx, token, groupID, offset, limit, meta)
+}
+
+func (ms *metricsMiddleware) Memberships(ctx context.Context, token, groupID string, offset, limit uint64, meta things.Metadata) (gp things.GroupPage, err error) {
+	defer func(begin time.Time) {
+		ms.counter.With("method", "memberships").Add(1)
+		ms.latency.With("method", "membeships").Observe(time.Since(begin).Seconds())
```
drasko: Fix typo

things/groups.go
```diff
@@ -0,0 +1,168 @@
...
+
+type Group interface {
+	ID() string
+	OwnerID() string
+	ParentID() string
+	Name() string
+	Description() string
+	Metadata() Metadata
+
+	SetID(id string)
```
drasko: Do we really need all these setters? Why not passing group struct pointer and changing the value directly.

This looks like some OO encapsulation, maybe it is needed but it Go it gives me "code smell".

drasko (changes requested):

things/groups.go
mteodor: Idea here and in `things/http.go` is to extract the common code that can be used for other entities that we wish to group
so `Group` is an interface that would have its implementations in each entity package itself
Of course if this solution is ok, http.go and `groups.go` would be moved to appropriate package

things/api/things/http/transport.go
manuio: Until we decide let's have something consistent

mteodor: we need to decide first

manuio: What I meant is that if we finally decide to don't use generic groups you will have to remember and modify this. If we use generic groups this will moved from here in any case.

things/api/logging.go
```diff
@@ -264,3 +264,146 @@ func (lm *loggingMiddleware) Identify(ctx context.Context, key string) (id strin
 
 	return lm.svc.Identify(ctx, key)
 }
+
+func (lm *loggingMiddleware) CreateGroup(ctx context.Context, token string, group things.Group) (g things.Group, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method create_group for key %s and name %s took %s to complete", token, group.Name(), time.Since(begin))
```
manuio: key or token?

manuio (changes requested):

things/groups.go
drasko: This is OK, but I do not like these setters. You can create an var struct without the setters, and then change internal struct directly through pointer receiver.

things/groups.go
```diff
@@ -0,0 +1,168 @@
...
+func (g *group) SetMetadata(meta Metadata) {
+	g.metadata = meta
+}
+
+func NewGroup() Group {
+	return &group{}
+}
+
+type GroupRepository interface {
+
```
dborovcanin: Please remove an empty line.

things/groups.go
```diff
@@ -0,0 +1,168 @@
+package things
+
+import (
+	"context"
+)
+
+type Group interface {
```
dborovcanin: I'm also not sure about this idea. Why can't we use Group struct here? This interface resembles struct anyways (explains fields, rather than behavior). It is a limitation compared to the interface, but it also provides better type safety and this looks sufficient for Mainflux needs anyways.

things/postgres/groups.go
```diff
@@ -0,0 +1,601 @@
...
+	}
+
+	return group, nil
+}
+
+func (gr groupRepository) Delete(ctx context.Context, groupID string) error {
+	qd := `DELETE FROM thing_groups WHERE id = :id`
+	group := things.NewGroup()
+	group.SetID(groupID)
+	dbg, err := toDBGroup(group)
```
dborovcanin: Can you simply create `dbGroup{ID: groupID}`?

dborovcanin (changes requested):

things/groups.go
mteodor: Because Thing groups and User groups could differ in implementation and this interface should be common denominator.
Idea was to have at least some parts modular, that I put this into a package and reuse it in Users, Things and Channels groups, no need to rewrite again each endpoint for example.
This is something that was asked for by @drasko, at least one part. I can easily copy paste and refactor this into the usual place

drasko: @mteodor this is OK, but remove getters and setters, use strucs and change the fields directly via pointer receiver.

dborovcanin: Group entity itself, however, is the same in both cases, right?

things/service.go
```diff
@@ -382,3 +396,123 @@ func (ts *thingsService) hasThing(ctx context.Context, chanID, thingKey string)
...
+		return groups.Group{}, errors.Wrap(ErrUnauthorizedAccess, err)
+	}
+
+	uid, err := uuidProvider.New().ID()
+	if err != nil {
+		return groups.Group{}, errors.Wrap(ErrCreateGroup, err)
+	}
+	group.ID = uid
+
+	// Temporary solution, until we have ID in response
```
drasko: Make this comment single line.

things/tracing/groups.go
```diff
@@ -0,0 +1,137 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+// Package tracing contains middlewares that will add spans
+// to existing traces.
```
drasko: No need to break line here

drasko (changes requested):

internal/groups/groups.go
```diff
@@ -0,0 +1,109 @@
...
+	CreateGroup(ctx context.Context, token string, group Group) (Group, error)
+
+	// UpdateGroup updates the group identified by the provided ID.
+	UpdateGroup(ctx context.Context, token string, group Group) (Group, error)
+
+	// ViewGroup retrieves data about the group identified by ID.
+	ViewGroup(ctx context.Context, token, id string) (Group, error)
+
+	// ListGroups retrieves groups.
+	ListGroups(ctx context.Context, token string, offset, limit uint64, meta Metadata) (GroupPage, error)
```
manuio: Use `m Metadata` please

internal/groups/groups.go
```diff
@@ -0,0 +1,109 @@
...
+	Delete(ctx context.Context, groupID string) error
+
+	// RetrieveByID retrieves group by its id
+	RetrieveByID(ctx context.Context, id string) (Group, error)
+
+	// RetrieveByName retrieves group by its name
+	RetrieveByName(ctx context.Context, name string) (Group, error)
+
+	// RetrieveAll retrieves all groups.
+	RetrieveAll(ctx context.Context, offset, limit uint64, gm Metadata) (GroupPage, error)
```
manuio: Same here `m Metadata`

things/api/logging.go
```diff
@@ -264,3 +265,146 @@ func (lm *loggingMiddleware) Identify(ctx context.Context, key string) (id strin
 
 	return lm.svc.Identify(ctx, key)
 }
+
+func (lm *loggingMiddleware) CreateGroup(ctx context.Context, token string, group groups.Group) (g groups.Group, err error) {
```
manuio: Use `g groups.Group` please

things/api/logging.go
```diff
@@ -264,3 +265,146 @@ func (lm *loggingMiddleware) Identify(ctx context.Context, key string) (id strin
...
+			lm.logger.Warn(fmt.Sprintf("%s with error: %s.", message, err))
+			return
+		}
+		lm.logger.Info(fmt.Sprintf("%s without errors.", message))
+	}(time.Now())
+
+	return lm.svc.CreateGroup(ctx, token, group)
+}
+
+func (lm *loggingMiddleware) UpdateGroup(ctx context.Context, token string, group groups.Group) (g groups.Group, err error) {
```
manuio: same here

things/postgres/things.go
```diff
@@ -22,6 +22,14 @@ const (
 	errTruncation = "string_data_right_truncation"
 )
 
+var (
+	errSaveDB     = errors.New("Save to DB failed")
+	errUpdateDB   = errors.New("Update to DB failed")
+	errRetrieveDB = errors.New("Retreiving from DB failed")
+	errMarshal    = errors.New("Failed to marshal metadata")
+	errUnmarshal  = errors.New("Failed to unmarshal metadata")
```
manuio: Use lower case for error messages. You can also improve the consistency starting messages by `failed to ...`

things/service.go
```diff
@@ -36,6 +40,12 @@ var (
 
 	// ErrDisconnect indicates error in removing connection
 	ErrDisconnect = errors.New("remove connection failed")
+
+	// ErrCreateGroup indicates error in creating group.
+	ErrCreateGroup = errors.New("failed to create group")
+
+	// ErrFailedToRetrieveThings failed to retrieve things.
+	ErrFailedToRetrieveThings = errors.New("failed to retrieve things for group")
```
manuio: Maybe better `failed to retrieve group things` or `failed to retrieve group members`

manuio (changes requested):

internal/groups/groups.go
```diff
@@ -0,0 +1,111 @@
...
+	ViewGroup(ctx context.Context, token, id string) (Group, error)
+
+	// ListGroups retrieves groups.
+	ListGroups(ctx context.Context, token string, offset, limit uint64, m Metadata) (GroupPage, error)
+
+	// ListChildren retrieves groups that are children to group identified by parentID
+	ListChildren(ctx context.Context, token, parentID string, offset, limit uint64, m Metadata) (GroupPage, error)
+
+	// ListParents retrieves groups that are parent to group identified by childID.
+	ListParents(ctx context.Context, token, childID string, offset, limit uint64, m Metadata) (GroupPage, error)
```
dborovcanin: While I understand that using `parents` and `children` terminology is natural, I'm afraid it's also ambiguous because `parents` imply that one Group has multiple parents, and `children` implies only direct descendants.

dborovcanin: Also, did you consider adding a feature to search genealogy only up to some depth, something like: give me only my children? Or: give me my only my children and children of my children. Same for parents. There are many possibilities for searching group members, but let's not get into that in this PR, this is just a remark for future reference.

internal/groups/http.go
```diff
@@ -0,0 +1,685 @@
...
+			return nil, err
+		}
+		if err := svc.RemoveGroup(ctx, req.token, req.groupID); err != nil {
+			return nil, err
+		}
+		return groupDeleteRes{}, nil
+	}
+}
+
+func AssignMemberToGroup(svc Service) endpoint.Endpoint {
```
dborovcanin: Please use consistent naming. We usually use the `Endpoint` suffix. If the method name is too long, you can shorten it like `AssignEndpoint` or `AssignMemberEndpoint`, where `member` implies that we're talking about Groups.

internal/groups/http.go
```diff
@@ -0,0 +1,685 @@
...
+			return nil, err
+		}
+		if err := svc.Assign(ctx, req.token, req.memberID, req.groupID); err != nil {
+			return nil, err
+		}
+		return assignMemberToGroupRes{}, nil
+	}
+}
+
+func RemoveMemberFromGroup(svc Service) endpoint.Endpoint {
```
dborovcanin: Same here: `UnassignEndpoint` or `UnassignMemberEndpoint`.

internal/groups/http.go
```diff
@@ -0,0 +1,685 @@
...
+			return nil, err
+		}
+		if err := svc.Unassign(ctx, req.token, req.memberID, req.groupID); err != nil {
+			return nil, err
+		}
+		return removeMemberFromGroupRes{}, nil
+	}
+}
+
+func ListMembersForGroupEndpoint(svc Service) endpoint.Endpoint {
```
dborovcanin: Same here: Member implies Group, so `ListMembersEndpoint` is sufficient.

internal/groups/http.go
```diff
@@ -0,0 +1,685 @@
+package groups
```
dborovcanin: Consider moving this to a separate `http` or `api` package (`/internal/groups/http` or `/internal/groups/api`) and breaking this file to the usual Mainflux API package structure (endpoints.go, requests.go, responses.go...).

internal/groups/http.go
```diff
@@ -0,0 +1,685 @@
...
+			Name:        g.Name,
+			Description: g.Description,
+			Metadata:    g.Metadata,
+			ParentID:    g.ParentID,
+		}
+		return res, nil
+	}
+}
+
+func DecodeListGroupsRequest(_ context.Context, r *http.Request) (interface{}, error) {
```
dborovcanin: Consider separating decode methods from endpoint methods (break in two files `endpoint.go` and `transport.go`).

internal/groups/http.go
```diff
@@ -0,0 +1,685 @@
...
+		},
+		Members: []interface{}{},
+	}
+	for _, m := range mp.Members {
+		res.Members = append(res.Members, m)
+	}
+	return res
+}
+
+// responses.go
```
dborovcanin: Exactly, please move this to a new file.

internal/groups/http.go
```diff
@@ -0,0 +1,685 @@
...
+
+func (res removeMemberFromGroupRes) Headers() map[string]string {
+	return map[string]string{}
+}
+
+func (res removeMemberFromGroupRes) Empty() bool {
+	return true
+}
+
+// requests.go
```
dborovcanin: Same.

pkg/sdk/go/things_test.go
```diff
@@ -49,7 +49,7 @@ func newThingsService(tokens map[string]string) things.Service {
 	thingCache := mocks.NewThingCache()
 	uuidProvider := uuid.NewMock()
 
-	return things.New(auth, thingsRepo, channelsRepo, chanCache, thingCache, uuidProvider)
+	return things.New(auth, thingsRepo, channelsRepo, nil, chanCache, thingCache, uuidProvider)
```
dborovcanin, line 52: Please open an issue for SDK support for Things groups.

internal/groups/groups.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type GroupPage struct {
+	PageMetadata
+	Groups []Group
+}
+
+type MemberPage struct {
+	PageMetadata
+	Members []Member
```
dborovcanin, line 43: Do we need Member type definition or `interface{}` is sufficient? I have no preferences, but we probably don't gain anything with this since we need to typecast anyways.

things/api/logging.go
```diff
@@ -264,3 +265,146 @@ func (lm *loggingMiddleware) Identify(ctx context.Context, key string) (id strin
 
 	return lm.svc.Identify(ctx, key)
 }
+
+func (lm *loggingMiddleware) CreateGroup(ctx context.Context, token string, g groups.Group) (gr groups.Group, err error) {
```
dborovcanin: Please be consistent when using `token` and `key`. I'm OK with using either option. It's a JSON Web **Token** which is Mainflux **Key**, but, using it as `key` here and as `token` there might be interpreted as if it's not the same thing.

things/postgres/init.go
```diff
@@ -93,6 +93,29 @@ func migrateDB(db *sqlx.DB) error {
...
+					`ALTER TABLE IF EXISTS things ADD CONSTRAINT things_id_key UNIQUE (id)`,
+					`CREATE TABLE IF NOT EXISTS thing_groups ( 
+						id          UUID NOT NULL,
+						parent_id   UUID, 
+						owner_id    UUID,
+						name        VARCHAR(254) UNIQUE NOT NULL,
+						description VARCHAR(1024),
+						metadata    JSONB,
+						PRIMARY KEY (id),
+						FOREIGN KEY (parent_id) REFERENCES thing_groups (id)  ON DELETE CASCADE ON UPDATE CASCADE
```
dborovcanin: Please check the formatting, it looks like there are multiple whitespaces in some lines that start with `FOREIGN_KEY`.

things/service.go
```diff
@@ -382,3 +396,122 @@ func (ts *thingsService) hasThing(ctx context.Context, chanID, thingKey string)
...
+	}
+
+	uid, err := uuidProvider.New().ID()
+	if err != nil {
+		return groups.Group{}, errors.Wrap(ErrCreateGroup, err)
+	}
+	group.ID = uid
+
+	// Temporary solution, until we have ID in response from Identify gRPC call.
+	owner, err := uuidProvider.New().ID()
```
dborovcanin: We need to prioritize #1245 and merge it before this PR.

dborovcanin (changes requested):

internal/groups/api/endpoint.go
```diff
@@ -0,0 +1,252 @@
...
+		if err != nil {
+			return createGroupRes{}, errors.Wrap(groups.ErrCreateGroup, err)
+		}
+		return createGroupRes{
+			created:     true,
+			ID:          gp.ID,
+			ParentID:    gp.ParentID,
+			Description: gp.Description,
+			Metadata:    gp.Metadata,
+			Name:        gp.Name,
```
manuio: Only `ID` is needed for the "location header". Remove other fields please

manuio (changes requested):

internal/groups/api/endpoint.go
```diff
@@ -0,0 +1,241 @@
...
+			Description: g.Description,
+			Metadata:    g.Metadata,
+			ParentID:    g.ParentID,
+			OwnerID:     g.OwnerID,
+		}
+		return res, nil
+	}
+}
+
+// endpoint.go
```
manuio: What is this comment?

manuio (changes requested):

internal/groups/api/requests.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (req updateGroupReq) validate() error {
+	if req.token == "" {
+		return groups.ErrUnauthorizedAccess
+	}
+	if req.ID == "" {
+		return groups.ErrMalformedEntity
+	}
+
```
manuio, line 56: check name size here

manuio (changes requested):

internal/groups/groups.go
```diff
@@ -0,0 +1,111 @@
...
+
+type Group struct {
+	ID          string
+	OwnerID     string
+	ParentID    string
+	Name        string
+	Description string
+	Metadata    Metadata
+	Level       int
+	Path        string
```
dborovcanin: I'm not sure that these belong to the Group struct. If we need info about the position of the Group in the hierarchy, we can expose an API for that. I doubt that we'll need it. Also, please consider limiting the possible depth of this hierarchy as well as the number of direct children to some sane number (10 or 20 should be enough) due to the performance constraints.

bootstrap/mocks/things.go
```diff
@@ -230,3 +231,47 @@ func findIndex(list []string, val string) int {
 
 	return -1
 }
+
+func (svc *mainfluxThings) CreateGroup(ctx context.Context, token string, group groups.Group) (groups.Group, error) {
+	panic("not implemented")
```
drasko, line 236: No need to panic and break the whole service

drasko: Also, please open an issue for this and paste the issue reference as a comment to this PR.

internal/groups/api/responses.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+type viewGroupRes struct {
+	ID          string                 `json:"id,omitempty"`
+	Name        string                 `json:"name,omitempty"`
+	ParentID    string                 `json:"parent_id,omitempty"`
+	OwnerID     string                 `json:"owner_id,omitempty"`
+	Description string                 `json:"description,omitempty"`
+	Metadata    map[string]interface{} `json:"metadata,omitempty"`
+	Level       int                    `json:"level,omitempty"`
```
drasko: What is this level? It is unclear from the code reader's perspective and probably demands a comment.

drasko (changes requested):

bootstrap/mocks/things.go
mteodor: This is a mock, and while it isn't used it will not break anything similar to other places

internal/groups/groups.go
dborovcanin: Regarding https://github.com/mainflux/mainflux/pull/1259/files#r511910630, do Service methods return direct ascendents and descendants or the entire tree? This should be better explained in the documentation comments so that the one who implements knows the desirable behavior of the `Service` interface.

things/postgres/groups_test.go
```diff
@@ -0,0 +1,490 @@
...
+		},
+	}
+
+	for _, tc := range cases {
+		err := repo.Delete(context.Background(), tc.groupID)
+		assert.True(t, errors.Contains(err, tc.err), fmt.Sprintf("%s: expected %s got %s\n", tc.desc, tc.err, err))
+	}
+}
+
+// func TestAssignUser(t *testing.T) {
```
dborovcanin: Please remove the commented code.

things/postgres/groups_test.go
```diff
@@ -0,0 +1,490 @@
...
+	"github.com/mainflux/mainflux/internal/groups"
+	"github.com/mainflux/mainflux/pkg/errors"
+	"github.com/mainflux/mainflux/pkg/uuid"
+	uuidProvider "github.com/mainflux/mainflux/pkg/uuid"
+	"github.com/mainflux/mainflux/things/postgres"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+/*
```
dborovcanin: Please remove the commented code.

things/postgres/groups.go
```diff
@@ -0,0 +1,631 @@
...
+	ID       uuid.NullUUID `db:"id"`
+	OwnerID  uuid.NullUUID `db:"owner_id"`
+	ParentID uuid.NullUUID `db:"parent_id"`
+	Metadata dbMetadata    `db:"metadata"`
+	Limit    uint64
+	Offset   uint64
+	Size     uint64
+}
+
+func toUUID(id string) (uuid.NullUUID, error) {
```
dborovcanin, line 473: This looks way too complicated. Can you use `uuid.FromStringOrNil(id)` or at least simplify this method?

things/postgres/groups.go
```diff
@@ -0,0 +1,631 @@
...
+	}
+	_, mq, err := getGroupsMetadataQuery("subordinates", gm)
+	if err != nil {
+		return groups.GroupPage{}, errors.Wrap(errRetrieveDB, err)
+	}
+	if mq != "" {
+		mq = fmt.Sprintf("WHERE %s", mq)
+	}
+
+	sq := `WITH RECURSIVE subordinates (id, owner_id, parent_id, name, description, metadata,  level, path) AS (
```
dborovcanin: This query is complex and it hard to read and understand. A large part of complexity comes from having `path` and `level`. If we remove it, do we still have all the necessary data to reconstruct Group genealogy? I mean, can we construct a tree or whichever data structure we want by programmatically tracking IDs of the retrieved groups?

dborovcanin (changes requested):

internal/groups/api/requests.go
```diff
@@ -0,0 +1,117 @@
...
+	token    string
+	groupID  uuid.NullUUID
+	memberID uuid.NullUUID
+}
+
+func (req memberGroupReq) validate() error {
+	if req.token == "" {
+		return groups.ErrUnauthorizedAccess
+	}
+	// if req.groupID == "" && req.memberID == "" {
```
dborovcanin: Please remove the commented code.

internal/groups/api/requests.go
```diff
@@ -0,0 +1,117 @@
...
+	token   string
+	groupID uuid.NullUUID
+	name    string
+}
+
+func (req groupReq) validate() error {
+	if req.token == "" {
+		return groups.ErrUnauthorizedAccess
+	}
+	// // if req.groupID == "" && req.name == "" {
```
dborovcanin: Please remove the commented code.

internal/groups/api/requests.go
```diff
@@ -0,0 +1,117 @@
...
+	name     string
+	groupID  uuid.NullUUID
+	memberID uuid.NullUUID
+}
+
+func (req listMemberGroupReq) validate() error {
+	if req.token == "" {
+		return groups.ErrUnauthorizedAccess
+	}
+	// if req.groupID == "" && req.memberID == "" {
```
dborovcanin: Please remove the commented code.

things/postgres/groups.go
```diff
@@ -0,0 +1,576 @@
...
+	if err != nil {
+		return groups.Group{}, errors.Wrap(errUpdateDB, err)
+	}
+	if group.Name == "" {
+		group.Name = old.Name
+	}
+	if group.Description == "" {
+		group.Description = old.Description
+	}
+	// if group.ParentID == "" {
```
dborovcanin: Please remove the commented code.

dborovcanin (changes requested):

internal/groups/groups.go
```diff
@@ -0,0 +1,122 @@
...
+	Unassign(ctx context.Context, token string, memberID, groupID uuid.NullUUID) error
+}
+
+type Repository interface {
+
+	// Save group
+	Save(ctx context.Context, group Group) (Group, error)
+
+	// Update a group
+	Update(ctx context.Context, group Group) (Group, error)
```
manuio: `g Group`

internal/groups/groups.go
```diff
@@ -0,0 +1,122 @@
...
+	RetrieveAll(ctx context.Context, offset, limit uint64, gm Metadata) (GroupPage, error)
+
+	// RetrieveAllParents retrieves all groups that are ancestors to the group with given groupID.
+	RetrieveAllParents(ctx context.Context, groupID uuid.NullUUID, offset, limit uint64, gm Metadata) (GroupPage, error)
+
+	// RetrieveAllChildren retrieves all children from group with given groupID.
+	RetrieveAllChildren(ctx context.Context, groupID uuid.NullUUID, offset, limit uint64, gm Metadata) (GroupPage, error)
+
+	// Retrieves list of groups that member belongs to
+	Memberships(ctx context.Context, memberID uuid.NullUUID, offset, limit uint64, gm Metadata) (GroupPage, error)
```
manuio: `m Metadata`

things/api/things/http/transport.go
```diff
@@ -172,6 +174,83 @@ func MakeHandler(tracer opentracing.Tracer, svc things.Service) http.Handler {
...
+
+	r.Post("/groups", kithttp.NewServer(
+		kitot.TraceServer(tracer, "add_group")(groupsAPI.CreateGroupEndpoint(svc)),
+		groupsAPI.DecodeGroupCreate,
+		encodeResponse,
+		opts...,
+	))
+
+	r.Get("/groups", kithttp.NewServer(
+		kitot.TraceServer(tracer, "groups")(groupsAPI.ListGroupsEndpoint(svc)),
```
manuio: `list_groups`

things/api/things/http/transport.go
```diff
@@ -172,6 +174,83 @@ func MakeHandler(tracer opentracing.Tracer, svc things.Service) http.Handler {
...
+
+	r.Delete("/groups/:groupID", kithttp.NewServer(
+		kitot.TraceServer(tracer, "delete_group")(groupsAPI.DeleteGroupEndpoint(svc)),
+		groupsAPI.DecodeGroupRequest,
+		encodeResponse,
+		opts...,
+	))
+
+	r.Put("/groups/:groupID/things/:memberID", kithttp.NewServer(
+		kitot.TraceServer(tracer, "assign_user_to_group")(groupsAPI.AssignEndpoint(svc)),
```
manuio: `assign`

things/api/things/http/transport.go
```diff
@@ -172,6 +174,83 @@ func MakeHandler(tracer opentracing.Tracer, svc things.Service) http.Handler {
...
+		groupsAPI.DecodeMemberGroupRequest,
+		encodeResponse,
+		opts...,
+	))
+
+	r.Delete("/groups/:groupID/things/:memberID", kithttp.NewServer(
+		kitot.TraceServer(tracer, "remove_thing_from_group")(groupsAPI.UnassignEndpoint(svc)),
```
manuio: unassign

things/api/things/http/transport.go
```diff
@@ -172,6 +174,83 @@ func MakeHandler(tracer opentracing.Tracer, svc things.Service) http.Handler {
...
+		groupsAPI.DecodeMemberGroupRequest,
+		encodeResponse,
+		opts...,
+	))
+
+	r.Get("/groups/:groupID/things", kithttp.NewServer(
+		kitot.TraceServer(tracer, "members")(groupsAPI.ListMembersEndpoint(svc)),
```
manuio: list_members

things/api/things/http/transport.go
```diff
@@ -172,6 +174,83 @@ func MakeHandler(tracer opentracing.Tracer, svc things.Service) http.Handler {
...
+
+	r.Get("/groups/:groupID/parents", kithttp.NewServer(
+		kitot.TraceServer(tracer, "list_parent_groups")(groupsAPI.ListGroupParentsEndpoint(svc)),
+		groupsAPI.DecodeListGroupsRequest,
+		encodeResponse,
+		opts...,
+	))
+
+	r.Get("/groups/:groupID", kithttp.NewServer(
+		kitot.TraceServer(tracer, "group")(groupsAPI.ViewGroupEndpoint(svc)),
```
manuio: `view_group`

things/service.go
```diff
@@ -382,3 +397,118 @@ func (ts *thingsService) hasThing(ctx context.Context, chanID, thingKey string)
 	}
 	return thingID, nil
 }
+
+func (ts *thingsService) CreateGroup(ctx context.Context, token string, group groups.Group) (groups.Group, error) {
```
manuio: `g groups.Group`

things/service.go
```diff
@@ -382,3 +397,118 @@ func (ts *thingsService) hasThing(ctx context.Context, chanID, thingKey string)
...
+	if err != nil {
+		return groups.Group{}, errors.Wrap(ErrCreateGroup, err)
+	}
+
+	group.ID = groups.UUIDFromString(uid)
+	group.OwnerID = groups.UUIDFromString(user.GetId())
+	return ts.groups.Save(ctx, group)
+}
+
+func (ts *thingsService) ListGroups(ctx context.Context, token string, offset, limit uint64, meta groups.Metadata) (groups.GroupPage, error) {
```
manuio: `gm groups.Metadata`

things/service.go
```diff
@@ -382,3 +397,118 @@ func (ts *thingsService) hasThing(ctx context.Context, chanID, thingKey string)
...
+	_, err := ts.auth.Identify(ctx, &mainflux.Token{Value: token})
+	if err != nil {
+		return groups.GroupPage{}, errors.Wrap(ErrUnauthorizedAccess, err)
+	}
+	return ts.groups.RetrieveAll(ctx, offset, limit, meta)
+
+}
+
+func (ts *thingsService) ListParents(ctx context.Context, token string, childID uuid.NullUUID, offset, limit uint64, meta groups.Metadata) (groups.GroupPage, error) {
```
manuio: `gm groups.Metadata`

things/service.go
```diff
@@ -382,3 +397,118 @@ func (ts *thingsService) hasThing(ctx context.Context, chanID, thingKey string)
...
+	_, err := ts.auth.Identify(ctx, &mainflux.Token{Value: token})
+	if err != nil {
+		return groups.GroupPage{}, errors.Wrap(ErrUnauthorizedAccess, err)
+	}
+	return ts.groups.RetrieveAllParents(ctx, childID, offset, limit, meta)
+}
+
+func (ts *thingsService) ListChildren(ctx context.Context, token string, parentID uuid.NullUUID, offset, limit uint64, meta groups.Metadata) (groups.GroupPage, error) {
```
manuio: `gm groups.Metadata`

things/tracing/groups.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/mainflux/mainflux/internal/groups"
+	opentracing "github.com/opentracing/opentracing-go"
+)
+
+const (
+	assignMember         = "assign_member"
+	saveGroup            = "save_group"
+	deleteGroup          = "delete_group"
+	updateGroup          = "update_group"
+	retrieveGroupByID    = "retrieve_group_by_id"
```
manuio: `	retrieveByID    = "retrieve_by_id"`

things/tracing/groups.go
```diff
@@ -0,0 +1,137 @@
...
+
+var _ groups.Repository = (*groupRepositoryMiddleware)(nil)
+
+type groupRepositoryMiddleware struct {
+	tracer opentracing.Tracer
+	repo   groups.Repository
+}
+
+// GroupRepositoryMiddleware tracks request and their latency, and adds spans to context.
+func GroupRepositoryMiddleware(tracer opentracing.Tracer, repo groups.Repository) groups.Repository {
```
manuio: `gr groups.Repository`

things/tracing/groups.go
```diff
@@ -0,0 +1,137 @@
...
+	return groupRepositoryMiddleware{
+		tracer: tracer,
+		repo:   repo,
+	}
+}
+func (grm groupRepositoryMiddleware) Save(ctx context.Context, group groups.Group) (groups.Group, error) {
```
manuio: `g groups.Group`

things/tracing/groups.go
```diff
@@ -0,0 +1,137 @@
...
+	span := createSpan(ctx, grm.tracer, saveGroup)
+	defer span.Finish()
+	ctx = opentracing.ContextWithSpan(ctx, span)
+
+	return grm.repo.Save(ctx, group)
+}
+func (grm groupRepositoryMiddleware) Update(ctx context.Context, group groups.Group) (groups.Group, error) {
```
manuio: `g groups.Group`

manuio (changes requested):

internal/groups/api/requests.go
```diff
@@ -0,0 +1,126 @@
...
+	if len(req.Name) > maxNameSize || req.Name == "" || !groupRegexp.MatchString(req.Name) {
+		return errors.Wrap(groups.ErrMalformedEntity, groups.ErrBadGroupName)
+	}
+
+	return nil
+}
+
+type updateGroupReq struct {
+	token       string
+	ID          string                 `json:"id,omitempty"`
```
manuio: This field don't need to be exported. It comes as path parameter not from the body of the request.

internal/groups/api/transport.go
```diff
@@ -0,0 +1,201 @@
...
+
+func DecodeGroupUpdate(_ context.Context, r *http.Request) (interface{}, error) {
+	if !strings.Contains(r.Header.Get("Content-Type"), contentType) {
+		return nil, groups.ErrUnsupportedContentType
+	}
+	var req updateGroupReq
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return nil, errors.Wrap(groups.ErrFailedDecode, err)
+	}
+	req.ID = bone.GetValue(r, "groupID")
```
manuio: Here it is

manuio (changes requested):

users/service.go
```diff
@@ -330,23 +330,18 @@ func (svc usersService) CreateGroup(ctx context.Context, token string, group Gro
 		return Group{}, ErrMalformedEntity
 	}
 
-	email, err := svc.identify(ctx, token)
+	id, err := svc.identify(ctx, token)
```
manuio: AFAIK `svc.identify` returns an `email`

things/tracing/groups.go
```diff
@@ -0,0 +1,138 @@
...
+
+import (
+	"context"
+
+	"github.com/mainflux/mainflux/internal/groups"
+	opentracing "github.com/opentracing/opentracing-go"
+)
+
+const (
+	assignMember         = "assign_member"
```
manuio: This could be `assign`

things/tracing/groups.go
```diff
@@ -0,0 +1,138 @@
...
+	deleteGroup          = "delete_group"
+	updateGroup          = "update_group"
+	retrieveByID         = "retrieve_by_id"
+	retrieveAllAncestors = "retrieve_all_ancestors"
+	retrieveAllChildren  = "retrieve_all_children"
+	retrieveAll          = "retrieve_all_groups"
+	retrieveByName       = "retrieve_by_name"
+	memberships          = "memberships"
+	members              = "members"
+	unassignMember       = "unassign_member"
```
manuio: This could be `unassign`

things/service.go
```diff
@@ -382,3 +395,115 @@ func (ts *thingsService) hasThing(ctx context.Context, chanID, thingKey string)
...
+	p, err := ts.groups.Members(ctx, groupID, offset, limit, gm)
+	if err != nil {
+		return groups.MemberPage{}, errors.Wrap(ErrFailedToRetrieveThings, err)
+	}
+	mp := groups.MemberPage{
+		PageMetadata: groups.PageMetadata{
+			Total:  p.Total,
+			Offset: p.Offset,
+			Limit:  p.Limit,
+			Name:   "things",
```
manuio: Use a const for `"things"`

things/service.go
```diff
@@ -6,12 +6,15 @@ package things
...
 
+	"github.com/mainflux/mainflux/internal/groups"
 	"github.com/mainflux/mainflux/pkg/errors"
 
 	"github.com/mainflux/mainflux"
+	uuidProvider "github.com/mainflux/mainflux/pkg/uuid"
 )
 
 var (
+
```
manuio, line 17: Remove this blank line

things/postgres/things.go
```diff
@@ -22,6 +22,11 @@ const (
 	errTruncation = "string_data_right_truncation"
 )
 
+var (
+	errUpdateDB   = errors.New("failed to update DB")
+	errRetrieveDB = errors.New("failed retrieving from DB")
```
manuio: not sure about capital letters in logs

things/postgres/init.go
```diff
@@ -93,6 +93,31 @@ func migrateDB(db *sqlx.DB) error {
...
+					`ALTER TABLE IF EXISTS things ADD CONSTRAINT things_id_key UNIQUE (id)`,
+					`CREATE extension LTREE`,
+					`CREATE TABLE IF NOT EXISTS thing_groups ( 
+						id          UUID NOT NULL,
+						parent_id   UUID, 
+						owner_id    UUID,
+						name        VARCHAR(254) UNIQUE NOT NULL,
+						description VARCHAR(1024),
+						metadata    JSONB,
+						path 		LTREE,
```
manuio: Looks like tabulation is not correct here

things/postgres/groups_test.go
```diff
@@ -0,0 +1,309 @@
...
+		require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+
+		group := groups.Group{
+			Description: "description",
+			ID:          groupID,
+			OwnerID:     ownerID,
+			Name:        fmt.Sprintf("mainflux%d", i),
+		}
+
+		// Create Things with metadata.
```
manuio: `Things` or `Groups`?

things/postgres/groups_test.go
```diff
@@ -0,0 +1,309 @@
...
+		groupID string
+		err     error
+	}{
+		{
+			desc:    "retrieve a group",
+			groupID: groupID,
+			err:     nil,
+		},
+		{
+			desc:    "retrieve a group with incorrect id",
```
manuio: Maybe `invalid` instead of `incorrect`

things/postgres/groups_test.go
```diff
@@ -0,0 +1,309 @@
...
+	incorrectID, err := uuidProvider.New().ID()
+	require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+
+	cases := []struct {
+		desc    string
+		groupID string
+		err     error
+	}{
+		{
+			desc:    "retrieve a group",
```
manuio: `retrieve group with valid id`

things/postgres/groups_test.go
```diff
@@ -0,0 +1,309 @@
...
+			err:   groups.ErrGroupConflict,
+		},
+
+		{
+			desc:  "create group with parent id",
+			group: group,
+			err:   nil,
+		},
+		{
+			desc:  "create group with invalid parentID",
```
manuio: `invalid parent id`

manuio (changes requested):

internal/groups/api/transport.go
```diff
@@ -0,0 +1,221 @@
...
+		groupID:  bone.GetValue(r, "groupID"),
+	}
+	return req, nil
+}
+
+func DecodeListMemberGroupRequest(_ context.Context, r *http.Request) (interface{}, error) {
+	if !strings.Contains(r.Header.Get("Content-Type"), contentType) {
+		return nil, groups.ErrUnsupportedContentType
+	}
+	o, err := readUintQuery(r, offsetKey, defOffset)
```
drasko, line 73: Add empty line here, if you added them below.

internal/groups/api/transport.go
```diff
@@ -0,0 +1,221 @@
...
+		metadata: m,
+		tree:     t,
+	}
+	return req, nil
+}
+
+func DecodeGroupCreate(_ context.Context, r *http.Request) (interface{}, error) {
+	if !strings.Contains(r.Header.Get("Content-Type"), contentType) {
+		return nil, groups.ErrUnsupportedContentType
+	}
```
drasko, line 114: Add empty line below.

internal/groups/api/transport.go
```diff
@@ -0,0 +1,221 @@
...
+	var req createGroupReq
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return nil, errors.Wrap(groups.ErrFailedDecode, err)
+	}
```
drasko, line 119: Add empty line below.

internal/groups/api/transport.go
```diff
@@ -0,0 +1,221 @@
...
+	req.token = r.Header.Get("Authorization")
+	return req, nil
+}
+
+func DecodeGroupUpdate(_ context.Context, r *http.Request) (interface{}, error) {
+	if !strings.Contains(r.Header.Get("Content-Type"), contentType) {
+		return nil, groups.ErrUnsupportedContentType
+	}
+	var req updateGroupReq
```
drasko, line 130: Same here

things/service.go
```diff
@@ -382,3 +396,115 @@ func (ts *thingsService) hasThing(ctx context.Context, chanID, thingKey string)
...
+		return groups.Group{}, errors.Wrap(ErrCreateGroup, err)
+	}
+
+	g.ID = uid
+	g.OwnerID = user.GetId()
+	return ts.groups.Save(ctx, g)
+}
+
+func (ts *thingsService) ListGroups(ctx context.Context, token string, level uint64, gm groups.Metadata) (groups.GroupPage, error) {
+	_, err := ts.auth.Identify(ctx, &mainflux.Token{Value: token})
```
drasko: If value is ignored, embed the check.

things/service.go
```diff
@@ -382,3 +396,115 @@ func (ts *thingsService) hasThing(ctx context.Context, chanID, thingKey string)
...
+	if err != nil {
+		return groups.GroupPage{}, errors.Wrap(ErrUnauthorizedAccess, err)
+	}
+	return ts.groups.RetrieveAll(ctx, level, gm)
+
+}
+
+func (ts *thingsService) ListParents(ctx context.Context, token string, childID string, level uint64, gm groups.Metadata) (groups.GroupPage, error) {
+	_, err := ts.auth.Identify(ctx, &mainflux.Token{Value: token})
```
drasko: Same here

drasko (changes requested):

internal/groups/api/endpoint.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+}
+
+func ListMembersEndpoint(svc groups.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(listMemberGroupReq)
+		if err := req.validate(); err != nil {
+			return memberPageRes{}, errors.Wrap(groups.ErrMalformedEntity, err)
+		}
+		mp, err := svc.ListMembers(ctx, req.token, req.groupID, req.offset, req.limit, req.metadata)
```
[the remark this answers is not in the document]
mteodor: Member page

internal/groups/api/endpoint.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+
+		group := groups.Group{
+			Name:        req.Name,
+			Description: req.Description,
+			ParentID:    req.ParentID,
+			Metadata:    req.Metadata,
+		}
+
+		g, err := svc.CreateGroup(ctx, req.token, group)
```
drasko: It is unclear from the code (on reading) what is the difference between `group` and `g` - they both suggest to be groups. It is also expected for `CreateGroup()` to return group, but on the other hand you created it manually above.

internal/groups/api/endpoint.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+}
+
+func ListGroupParentsEndpoint(svc groups.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(listGroupsReq)
+		if err := req.validate(); err != nil {
+			return groupPageRes{}, errors.Wrap(groups.ErrMalformedEntity, err)
+		}
+		page, err := svc.ListParents(ctx, req.token, req.groupID, req.level, req.metadata)
```
drasko, line 166: Please stay consistent in adding empty lines comparing to other function.

internal/groups/api/endpoint.go
```diff
@@ -0,0 +1,316 @@
...
+		return buildUsersResponse(page), nil
+	}
+}
+
+func buildGroupsResponseTree(page groups.GroupPage) groupPageRes {
+	groupsMap := map[string]*groups.Group{}
+	// Parents map keeps its array of children.
+	parentsMap := map[string][]*groups.Group{}
+	for _, g := range page.Groups {
+		group := g
```
drasko: Why is this `group` var needed? Can't just `g` be used here directly?

internal/groups/api/endpoint.go
```diff
@@ -0,0 +1,316 @@
...
+		Level:       g.Level,
+		Path:        g.Path,
+		Children:    make([]*viewGroupRes, 0),
+	}
+
+	for _, ch := range g.Children {
+		child := toViewGroupRes(*ch)
+		view.Children = append(view.Children, &child)
+	}
+	return view
```
drasko, line 289: If empty line was added above, add one here as well

internal/groups/api/requests.go
```diff
@@ -0,0 +1,126 @@
...
+	Description string                 `json:"description,omitempty"`
+	ParentID    string                 `json:"parent_id,omitempty"`
+	Metadata    map[string]interface{} `json:"metadata,omitempty"`
+}
+
+func (req updateGroupReq) validate() error {
+	if req.token == "" {
+		return groups.ErrUnauthorizedAccess
+	}
+
```
drasko, line 44: If function below do not have white lines, please suppers these as well.

drasko (changes requested):

internal/groups/api/endpoint.go
mteodor: no it cant, it creates new object as a copy of `g` so that its address can be used for pointer in next lines

internal/groups/api/endpoint.go
mteodor: same as `CreateThings` works

internal/groups/api/requests.go
manuio: To be consistent I'd add it in the function where it's missing.

internal/groups/api/endpoint.go
manuio: Probably this would solve your issue:
```
	for i := range page.Groups {
		if _, ok := groupsMap[group.ID]; !ok {
			groupsMap[group.ID] = &page.Groups[i]
			parentsMap[group.ID] = make([]*groups.Group, 0)
		}
	}
```

internal/groups/api/endpoint.go
```diff
@@ -0,0 +1,336 @@
...
+			groupsMap[group.ID] = &group
+			parentsMap[group.ID] = make([]*groups.Group, 0)
+		}
+	}
+
+	for _, g := range groupsMap {
+		if ch, ok := parentsMap[g.ParentID]; ok {
+			ch = append(ch, g)
+			parentsMap[g.ParentID] = ch
+			fmt.Printf("%s, array> %d\n", g.ParentID, len(ch))
```
manuio: Remove this log

manuio (changes requested):

internal/groups/api/endpoint.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+
+		group := groups.Group{
+			Name:        req.Name,
+			Description: req.Description,
+			ParentID:    req.ParentID,
+			Metadata:    req.Metadata,
+		}
+
+		created, err := svc.CreateGroup(ctx, req.token, group)
```
manuio: `created` should be called `group` but in fact `svc.CreateGroup` should return `groupID` instead of whole Group.

manuio (changes requested):

authn/postgres/setup_test.go
```diff
@@ -32,7 +32,7 @@ func TestMain(m *testing.M) {
 		"POSTGRES_PASSWORD=test",
 		"POSTGRES_DB=test",
 	}
-	container, err := pool.Run("postgres", "10.2-alpine", cfg)
+	container, err := pool.Run("postgres", "10.8-alpine", cfg)
```
dborovcanin, line 35: Consider updating version in correpsonding `docker-compose.ylm` files.

internal/groups/api/responses.go
```diff
@@ -0,0 +1,151 @@
...
+	ID          string                 `json:"id,omitempty"`
+	Name        string                 `json:"name,omitempty"`
+	ParentID    string                 `json:"parent_id,omitempty"`
+	OwnerID     string                 `json:"owner_id,omitempty"`
+	Description string                 `json:"description,omitempty"`
+	Metadata    map[string]interface{} `json:"metadata,omitempty"`
+	// Level if retrieved with Parents or Children method
+	// indicates a level in hierarchy from first group node.
+	Level    int             `json:"level,omitempty"`
+	Path     string          `json:"path"`
```
dborovcanin: A short explanation similar to the one for `level` field would be nice. Also, is level empty (thus, omitted) if it's not Parents/Children method?

internal/groups/groups.go
```diff
@@ -0,0 +1,112 @@
...
+
+type Group struct {
+	ID          string
+	OwnerID     string
+	ParentID    string
+	Name        string
+	Description string
+	Metadata    Metadata
+	// Level if retrieved with Parents or Children method
+	// indicates a level in hierarchy from first group node.
```
dborovcanin: Is level 0 when the Group is not fetched using Parents/Children? This can be confusing and misleading if not documented. Also, consider using a pointer (or at least setting it to a negative value) so that there is a difference between 0 level and level not fetched at all.

internal/groups/groups.go
```diff
@@ -0,0 +1,112 @@
...
+	Update(ctx context.Context, g Group) (Group, error)
+
+	// Delete a group
+	Delete(ctx context.Context, groupID string) error
+
+	// RetrieveByID retrieves group by its id
+	RetrieveByID(ctx context.Context, id string) (Group, error)
+
+	// RetrieveByName retrieves group by its name
+	RetrieveByName(ctx context.Context, name string) (Group, error)
```
dborovcanin: This implies that `Name` is unique. Instead, please consider making (owner, name) pair unique. If two users, for example, group Things by the same criteria (say, spacial), it's very likely that there will be a name collision.

things/postgres/groups.go
```diff
@@ -0,0 +1,647 @@
...
+	if err != nil {
+		return groups.Group{}, errors.Wrap(errUpdateDB, err)
+	}
+	if g.Name == "" {
+		g.Name = old.Name
+	}
+	if g.Description == "" {
+		g.Description = old.Description
+	}
+	if g.ParentID == "" {
```
dborovcanin: I'm not sure that we should allow parent change. While it's an extremely powerful feature, I find it rather dangerous (upcoming policies are going to be heavily affected). @mainflux/maintainers Opinions?

things/postgres/groups.go
```diff
@@ -0,0 +1,647 @@
...
+	old, err := gr.RetrieveByID(ctx, g.ID)
+	if err != nil {
+		return groups.Group{}, errors.Wrap(errUpdateDB, err)
+	}
+	if g.Name == "" {
+		g.Name = old.Name
+	}
+	if g.Description == "" {
+		g.Description = old.Description
+	}
```
dborovcanin: Consider allowing `name` and `description` to be empty. While it's pointless to have the `name` field empty, it will eliminate this entire fetch from the DB, as well as potential transaction problem we have here (fetching and updating are executed in separate transactions) due to the lack of optimistic lock.

things/postgres/groups.go
dborovcanin: This whole Null UUID handling looks complicated. Can you try to simplify it?

dborovcanin (changes requested):

internal/groups/api/responses.go
mteodor: path and level are always there

internal/groups/groups.go
```diff
@@ -0,0 +1,122 @@
...
+	Name        string
+	Description string
+	Metadata    Metadata
+}
+
+type PageMetadata struct {
+	Total  uint64
+	Offset uint64
+	Limit  uint64
+	Name   string
```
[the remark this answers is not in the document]
mteodor: Yes I think, it is useful for example when returning members, since members can be either things or users, so name is additional info

internal/groups/groups.go
mteodor: Name is unique

things/postgres/groups.go
```diff
@@ -0,0 +1,647 @@
...
+	_, mq, err := getGroupsMetadataQuery("thing_groups", gm)
+	if err != nil {
+		return groups.GroupPage{}, errors.Wrap(errRetrieveDB, err)
+	}
+	if mq != "" {
+		mq = fmt.Sprintf("AND %s", mq)
+	}
+
+	q := fmt.Sprintf("SELECT id, name, owner_id, parent_id, description, metadata, path, nlevel(path) as level FROM thing_groups WHERE path @> :path AND nlevel(:path) - nlevel(path) >= 0 AND nlevel(:path) - nlevel(path) <= :level %s", mq)
+	cq := fmt.Sprintf("SELECT COUNT(*) FROM thing_groups WHERE path @> :path AND nlevel(:path) - nlevel(path) >= 0 AND nlevel(:path) - nlevel(path) <= :level %s", mq)
```
manuio: Split this in multiple lines

things/postgres/groups.go
```diff
@@ -0,0 +1,647 @@
...
+	_, mq, err := getGroupsMetadataQuery("thing_groups", gm)
+	if err != nil {
+		return groups.GroupPage{}, errors.Wrap(errRetrieveDB, err)
+	}
+	if mq != "" {
+		mq = fmt.Sprintf("AND %s", mq)
+	}
+
+	q := fmt.Sprintf("SELECT id, name, owner_id, parent_id, description, metadata, path, nlevel(path) as level FROM thing_groups WHERE path <@ :path AND nlevel(path) - nlevel(:path) <= :level %s", mq)
+	cq := fmt.Sprintf("SELECT COUNT(*) FROM thing_groups WHERE path <@ :path AND nlevel(path) - nlevel(:path) <= :level %s", mq)
```
manuio: same here and for all other queries:

things/postgres/groups.go
```diff
@@ -0,0 +1,647 @@
...
+
+func (gr groupRepository) Members(ctx context.Context, groupID string, offset, limit uint64, gm groups.Metadata) (groups.MemberPage, error) {
+	m, mq, err := getGroupsMetadataQuery("things_group", gm)
+	if err != nil {
+		return groups.MemberPage{}, errors.Wrap(errRetrieveDB, err)
+	}
+
+	q := fmt.Sprintf(`SELECT th.id, th.name, th.key, th.metadata FROM things th, thing_group_relations g
+                      WHERE th.id = g.thing_id AND g.group_id = :group 
+                      %s ORDER BY id LIMIT :limit OFFSET :offset;`, mq)
```
manuio: The tabulation seems wrong here

things/postgres/groups.go
```diff
@@ -0,0 +1,647 @@
...
+	}
+
+	if mq != "" {
+		mq = fmt.Sprintf("AND %s", mq)
+	}
+	q := fmt.Sprintf(`SELECT g.id, g.owner_id, g.parent_id, g.name, g.description, g.metadata 
+					  FROM thing_group_relations gr, thing_groups g
+					  WHERE gr.group_id = g.id and gr.thing_id = :userID 
+		  			  %s ORDER BY id LIMIT :limit OFFSET :offset;`, mq)
+
```
manuio, line 365: same here

things/postgres/groups.go
```diff
@@ -0,0 +1,647 @@
...
+		gr, err := toGroup(dbgr)
+		if err != nil {
+			return groups.GroupPage{}, err
+		}
+		items = append(items, gr)
+	}
+
+	cq := fmt.Sprintf(`SELECT COUNT(*) 
+					   FROM thing_group_relations gr, thing_groups g
+					   WHERE gr.group_id = g.id and gr.thing_id = :userID %s;`, mq)
```
manuio: and here

things/postgres/groups.go
```diff
@@ -0,0 +1,647 @@
...
+		Group: grID,
+		Thing: thID,
+	}, nil
+}
+
+func getGroupsMetadataQuery(db string, m groups.Metadata) ([]byte, string, error) {
+	mq := ""
+	mb := []byte("{}")
+	if len(m) > 0 {
+		mq = db + `.metadata @> :metadata`
```
manuio, line 600: If `db` is empty is `.metadata @> :metadata` going to work?

manuio (changes requested):

things/service.go
```diff
@@ -382,3 +396,117 @@ func (ts *thingsService) hasThing(ctx context.Context, chanID, thingKey string)
...
+func (ts *thingsService) ListParents(ctx context.Context, token string, childID string, level uint64, gm groups.Metadata) (groups.GroupPage, error) {
+	if _, err := ts.auth.Identify(ctx, &mainflux.Token{Value: token}); err != nil {
+		return groups.GroupPage{}, errors.Wrap(ErrUnauthorizedAccess, err)
+	}
+	return ts.groups.RetrieveAllParents(ctx, childID, level, gm)
+}
+
+func (ts *thingsService) ListChildren(ctx context.Context, token string, parentID string, level uint64, gm groups.Metadata) (groups.GroupPage, error) {
+	_, err := ts.auth.Identify(ctx, &mainflux.Token{Value: token})
+	if err != nil {
```
manuio: Please, inline this check here and for all other methods

manuio (changes requested):

things/service.go
```diff
@@ -36,6 +40,12 @@ var (
 
 	// ErrDisconnect indicates error in removing connection
 	ErrDisconnect = errors.New("remove connection failed")
+
+	// ErrCreateGroup indicates error in creating group.
+	ErrCreateGroup = errors.New("failed to create group")
+
+	// ErrFailedToRetrieveThings failed to retrieve things.
+	ErrFailedToRetrieveThings = errors.New("failed to retrieve group members")
```
manuio, line 48: `ErrFailedToRetrieveThings` ?

manuio (changes requested):

things/postgres/groups_test.go
```diff
@@ -0,0 +1,420 @@
...
+	metadata  = groups.Metadata{"department": "IoT"}
+)
+
+func TestGroupsSave(t *testing.T) {
+	dbMiddleware := postgres.NewDatabase(db)
+	groupRepo := postgres.NewGroupRepo(dbMiddleware)
+	cleanAll(t, dbMiddleware)
+
+	ownerID, err := uuidProvider.New().ID()
+	require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
```
manuio: Can you please find/replace `got unexpected error` by `unexpected error`

manuio (changes requested):

things/postgres/groups.go
mteodor: I already did simplification and this best I can do

internal/groups/errors.go
```diff
@@ -0,0 +1,49 @@
...
+	// ErrUnsupportedContentType indicates unacceptable or lack of Content-Type
+	ErrUnsupportedContentType = errors.New("unsupported content type")
+
+	// ErrFailedDecode indicates failed to decode request body
+	ErrFailedDecode = errors.New("failed to decode request body")
+
+	// ErrMissingParent indicates that parent cant be found
+	ErrMissingParent = errors.New("failed to retrieve parent")
+
+
```
manuio: Remove these blank lines

internal/groups/errors.go
```diff
@@ -0,0 +1,49 @@
...
+	// ErrUnassignFromGroup indicates failure to unassign member from a group.
+	ErrUnassignFromGroup = errors.New("failed to unassign member from a group")
+
+	// ErrUnsupportedContentType indicates unacceptable or lack of Content-Type
+	ErrUnsupportedContentType = errors.New("unsupported content type")
+
+	// ErrFailedDecode indicates failed to decode request body
+	ErrFailedDecode = errors.New("failed to decode request body")
+
+	// ErrMissingParent indicates that parent cant be found
```
manuio: `can't`

internal/groups/groups.go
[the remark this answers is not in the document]
drasko: Can you give an example please, I am not convinced with this explanation.

internal/groups/groups.go
drasko: @mteodor how do you guarantee name uniqueness in a distributed system?

internal/groups/api/responses.go
```diff
@@ -0,0 +1,152 @@
...
+	_ mainflux.Response = (*assignMemberToGroupRes)(nil)
+	_ mainflux.Response = (*removeMemberFromGroupRes)(nil)
+)
+
+type pageRes struct {
+	Total  uint64 `json:"total"`
+	Offset uint64 `json:"offset"`
+	Limit  uint64 `json:"limit"`
+	Name   string `json:"name"`
+}
```
drasko, line 24: Add one empty line after the struct definition

drasko (changes requested):

bootstrap/mocks/things.go
dborovcanin: Comment resolved by issue #1271.

internal/groups/groups.go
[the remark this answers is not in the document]
dborovcanin: I'm also not convinced. You can't confuse Things and Users Groups because the Members will be of different types.

things/postgres/init.go
```diff
@@ -93,6 +93,32 @@ func migrateDB(db *sqlx.DB) error {
...
+					`CREATE extension LTREE`,
+					`CREATE TABLE IF NOT EXISTS thing_groups ( 
+						id          UUID UNIQUE NOT NULL,
+						parent_id   UUID, 
+						owner_id    UUID,
+						name        VARCHAR(254) NOT NULL,
+						description VARCHAR(1024),
+						metadata    JSONB,
+						path        LTREE, 
+						PRIMARY KEY (name, path),
```
dborovcanin: Does it make sense to have an ID that's not a primary key?

internal/groups/groups.go
dborovcanin: Based on the other conversations and on the code in the implementation (`init.go`), the name doesn't seem to be unique. Should this method signature be updated?

things/postgres/groups.go
```diff
@@ -0,0 +1,636 @@
...
+	row.Next()
+	if err := row.Scan(&id); err != nil {
+		return groups.Group{}, err
+	}
+	g.ID = id
+	return g, nil
+}
+
+func (gr groupRepository) Update(ctx context.Context, g groups.Group) (groups.Group, error) {
+	q := `UPDATE thing_groups SET name = :name, description = :description, parent_id = :parent_id, metadata = :metadata  WHERE id = :id`
```
dborovcanin: Is it safe to enable Parent update? I'm afraid that it might be dangerous, especially once we introduce policies attached to the groups.

things/postgres/groups.go
```diff
@@ -0,0 +1,636 @@
...
+		return groups.GroupPage{}, errors.Wrap(errRetrieveDB, err)
+	}
+	if mq != "" {
+		mq = fmt.Sprintf("AND %s", mq)
+	}
+
+	q := fmt.Sprintf(`SELECT id, name, owner_id, parent_id, description, metadata, path, nlevel(path) as level FROM thing_groups 
+					  WHERE path <@ :path AND nlevel(path) - nlevel(:path) <= :level %s`, mq)
+	cq := fmt.Sprintf("SELECT COUNT(*) FROM thing_groups WHERE path <@ :path AND nlevel(path) - nlevel(:path) <= :level %s", mq)
+	g, err := gr.RetrieveByID(ctx, groupID)
```
dborovcanin: Looks like you are fetching the group only to fetch a path for the response. Can this call be avoided (it's a separate transaction - one more DB driver call; also, because not everything is done in the transactions and can change parent in the current implementation - I'm afraid that we need an optimistic lock here)?

dborovcanin (changes requested):

things/postgres/init.go
mteodor: https://github.com/mainflux/mainflux/pull/1259#discussion_r525561057 Im talking about json response

internal/groups/groups.go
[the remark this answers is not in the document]
mteodor:
```
curl -s -S -X GET http://localhost:9001/groups/894f95a2-14b9-4f4b-9164-ff39a5ee7114/children\?level\=5\&tree\=true  -H "Authorization: $TOK" -H 'Content-Type: application/json' | jq                                                                    
{                                                                                                                                                                                                                                                                                         
  "total": 3,                                                                                                                                                                                                                                                                             
  "offset": 0,                                                                                                                                                                                                                                                                            
  "limit": 0,                                                                                                                                                                                                                                                                             
  "name": "things",                                                                                                                                                                                                                                                                             
  "groups": [                                                                                                                                                                                                                                                                             
    {                                                                                                                                                                                                                                                                                     
      "id": "894f95a2-14b9-4f4b-9164-ff39a5ee7114",                                                                                                                                                                                                                                       
      "name": "group6",                                                                                                                                                                                                                                                                   
      "owner_id": "d3861ca6-a950-4a51-91ff-50cfb5559a11",                                                                                                                                                                                                                                 
      "metadata": {                                                                                                                                                                                                                                                                       
        "admin": "true"                                                                                                                                                                                                                                                                   
      },                                                                                                                                                                                                                                                                                  
      "level": 1,                                                                                                                                                                                                                                                                         
      "path": "group6",                                                                                                                                                                                                                                                                   
      "children": [               
```


and url would be similar for users just different port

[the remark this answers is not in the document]
dborovcanin: Thanks for the explanation. I'm OK with this.

drasko (dismissed):
LGTM

manuio (dismissed):
LGTM

manuio (dismissed):
LGTM

things/postgres/groups.go
```diff
@@ -0,0 +1,626 @@
...
+		return groups.GroupPage{}, errors.Wrap(errRetrieveDB, err)
+	}
+	if mq != "" {
+		mq = fmt.Sprintf("AND %s", mq)
+	}
+
+	q := fmt.Sprintf(`SELECT g.id, g.name, g.owner_id, g.parent_id, g.description, g.metadata, g.path, nlevel(g.path) as level FROM 
+					 (SELECT thing_groups.path FROM thing_groups where id = :parent_id) parent LEFT JOIN
+					 (SELECT id, name, owner_id, parent_id, description,metadata, path FROM thing_groups) g 
+					  ON g.path @> parent.path AND nlevel(parent.path) - nlevel(g.path) >= 0 AND nlevel(parent.path) - nlevel(g.path) <= :level %s`, mq)
```
dborovcanin: Is `g.path @> parent.path AND nlevel(parent.path) - nlevel(g.path) >= 0` necessary, or we can use only `g.path @> parent.path`?

things/postgres/groups.go
```diff
@@ -0,0 +1,626 @@
...
+
+	_, mq, err := getGroupsMetadataQuery("thing_groups", gm)
+	if err != nil {
+		return groups.GroupPage{}, errors.Wrap(errRetrieveDB, err)
+	}
+	if mq != "" {
+		mq = fmt.Sprintf("AND %s", mq)
+	}
+
+	q := fmt.Sprintf(`SELECT g.id, g.name, g.owner_id, g.parent_id, g.description, g.metadata, g.path, nlevel(g.path) as level FROM
```
dborovcanin: Consider using simpler `self join`. You can try this query:
```sql
SELECT g.id, g.name, g.owner_id, g.parent_id, g.description, g.metadata, g.path, nlevel(g.path) as level
FROM thing_groups parent, thing_groups g
WHERE parent.id = :id AND g.path @> parent.path AND nlevel(parent.path) - nlevel(g.path) <= :level;
```
It's a little simpler (and shorter) and possibly more performant.

dborovcanin (changes requested):

manuio (dismissed):
LGTM

blokovi (dismissed):

internal/groups/api/responses.go
```diff
@@ -0,0 +1,153 @@
...
+	return map[string]string{}
+}
+
+func (res viewGroupRes) Empty() bool {
+	return false
+}
+
+type groupRes struct {
+	id      string
+	created bool
```
drasko, line 74: Missing `updated`?

drasko, line 74: `created` is a time, not bool. Time of creation.

drasko (changes requested):

internal/groups/groups.go
```diff
@@ -0,0 +1,114 @@
...
+	Metadata    Metadata
+	// Indicates a level in hierarchy from first group node.
+	// For a root node level is 1.
+	Level int
+	// Path is a path in a tree, consisted of group names
+	// parentName.childrenName1.childrenName2 .
+	Path      string
+	Children  []*Group
+	CreatedAt time.Time
+	UpdateAt  time.Time
```
dborovcanin: `UpdatedAt`

dborovcanin (dismissed):

dborovcanin (approved):

drasko (approved):
LGTM

PR #1260, opened 2020-10-20
MF-1254 - Create universal JSON writer

### LINKED ISSUE #1254: Create universal JSON writer
**FEATURE REQUEST**

Mainflux writers understand [SenML](https://tools.ietf.org/html/rfc8428), and this is the only format that they can parse and write in the DB.

While this makes a great telemetry choice for a payload - as it is an IEEE standard, and comes in both JSON and CBOR (binary) well defined flavors - inn many cases it is needed to send custom JSON.

Mainflux provides [transformers](https://github.com/mainflux/mainflux/tree/master/pkg/transformers) package to handle transforming of the internal [`messaging.Message{}`](https://github.com/mainflux/mainflux/blob/master/pkg/messaging/message.proto) into something that will be output to writers and written in the DB. It is even used internally to transform internal mainflux message struct into SenML message.

While using `transformers` package is handy, this concept can be pushed forward, and universal JSON writer can be made - similar to what [Telegraf JSON plugin](https://github.com/influxdata/telegraf/tree/master/plugins/parsers/json) is doing.

This way a user could provide only `config.toml` and eventually filter out some keys to produce desired result.

This universal JSON transformer should produce internal Mainflux message format, in JSON, and a universal writers should pick this up and write into DB.

This way it will be easy to send arbitrary JSON from various devices and all messages will end-up in the DB, in a unified Mainflux format.

Must-have

dborovcanin:
### What does this do?
This pull request adds a universal JSON Normalizer and support for it in Mainflux Writers. Universal JSON writer enables storing arbitrary JSON data as a Mainflux message payload. This means that the user will be able to store any JSON object as a Mainflux message payload. Mainflux JSON Transformer parses Mainflux message and transforms its payload to the generic `map[string]interface{}`. How the message is stored depends on the Writer's implementation and the underlying technology. For example, for PostgreSQL the message is stored like the following:
```sql
      id            UUID,
      channel       UUID,
      subtopic      VARCHAR(254),
      publisher     UUID,
      protocol      TEXT,
      payload       JSONB,
      PRIMARY KEY   (id)
```
An example row for PostgreSQL looks like:
| time                | channel                              | protocol | publisher                            | subtopic | payload            |
| ------------------- | ------------------------------------ | -------- | ------------------------------------ | -------- | ------------------ |
| 1603274654745903159 | d5c0dd88-0f8c-4e66-aec6-795ced81504b | http     | 909a84e9-135b-4f60-b934-1c74165831ab | sub.topic | {"a":1, "a.b":"c"} |

For InfluxDB, an example measurement would look like:

| time                | channel                              | protocol | publisher                            | subtopic | a   | a.b |
| ------------------- | ------------------------------------ | -------- | ------------------------------------ | -------- | --- | --- |
| 1603274654745903159 | d5c0dd88-0f8c-4e66-aec6-795ced81504b | http     | 909a84e9-135b-4f60-b934-1c74165831ab | sub.topic | 1   | "c" |

For MnogoDB, an example document looks like:
```json
{
"_id":{"$oid":"5f8f7117b26e1ce5692ce47b"},
"channel":"d5c0dd88-0f8c-4e66-aec6-795ced81504b",
"subtopic":"sub.topic",
"publisher":"909a84e9-135b-4f60-b934-1c74165831ab",
"protocol":"HTTP",
"payload":{
    "a":1,
    "a.b": "c"
    }
}
```

There are a few drawbacks to this implementation:

1. Only an object can be payload - no arrays allowed.
2. Storing many different formats of payload in some time-series DB (such as an InfluxDB) can cause a lot of empty values.
3. Potentially limits querying data and performance.

This pull request resolves #1254.

InfluxDB measurement naming is changed due to added separate measurements for SenML and generic JSON messages.

Yes.

No.


pkg/transformers/json/transformer.go
```diff
@@ -5,88 +5,63 @@ package json
...
-			msgs = append(msgs, kvmgs...)
+func flatten(prefix string, m, m1 map[string]interface{}) (map[string]interface{}, error) {
+	for k, v := range m1 {
+		if k == "publisher" || k == "protocol" || k == "channel" || k == "subtopic" || strings.Contains(k, ".") {
+			return nil, errInvalidKey
+		}
+		switch val := v.(type) {
+		case map[string]interface{}:
+			var err error
+			m, err = flatten(prefix+k+".", m, val)
```
mteodor: I would use `_` instead of `.` cause `.` can be used in query and it may be confusing

dborovcanin: Since `_` is common in keys, I'll use `/` instead.

writers/writer.go
```diff
@@ -66,33 +55,28 @@ func (c *consumer) handler(msg messaging.Message) error {
...
-		return errMessageConversion
-	}
 
-	return c.repo.Save(msgs...)
+	return c.repo.Save(t)
 }
 
 type filterConfig struct {
-	Filter []string `toml:"filter"`
+	List []string `toml:"filter"`
```
manuio: I would follow ``toml:"filter"`` naming

writers/postgres/messages.go
```diff
@@ -82,12 +97,43 @@ func (pr postgresRepo) Save(messages ...transformers.Message) (err error) {
 	return err
 }
 
+func (pr postgresRepo) saveJSON(messages interface{}) error {
+	msg, ok := messages.(mfjson.Message)
+	if !ok {
+		return errSaveMessage
+	}
+	q := `INSERT INTO json (id, channel, subtopic, publisher, protocol, payload)
+    VALUES (:id, :channel, :subtopic, :publisher, :protocol, :payload);`
```
manuio: align tab

manuio (changes requested):

manuio (approved):
LGTM

writers/cassandra/README.md
```diff
@@ -18,7 +18,7 @@ default values.
 | MF_CASSANDRA_WRITER_DB_USER         | Cassandra DB username                                     |                        |
 | MF_CASSANDRA_WRITER_DB_PASS         | Cassandra DB password                                     |                        |
 | MF_CASSANDRA_WRITER_DB_PORT         | Cassandra DB port                                         | 9042                   |
-| MF_CASSANDRA_WRITER_CONFIG_PATH     | Configuration file path with filters list                 | /configs.toml          |
+| MF_CASSANDRA_WRITER_SUBJECTS_CONFIG | Configuration file path with subjects list                | /config/subjects.toml  |
```
drasko: I would only add here maybe `NATS subjects list`, or the people who land in the project and do not know NATS terminology.

writers/mongodb/README.md
```diff
@@ -16,7 +16,7 @@ default values.
 | MF_MONGO_WRITER_DB              | Default MongoDB database name              | messages               |
 | MF_MONGO_WRITER_DB_HOST         | Default MongoDB database host              | localhost              |
 | MF_MONGO_WRITER_DB_PORT         | Default MongoDB database port              | 27017                  |
-| MF_MONGO_WRITER_CONFIG_PATH     | Configuration file path with filters list  | /configs.toml          |
+| MF_MONGO_WRITER_SUBJECTS_CONFIG | Configuration file path with subjects list | /config/subjects.toml  |
```
drasko: Same here - NATS subjects

drasko (changes requested):

writers/cassandra/README.md
manuio: This could be `config.toml` and contain 
```
[nats]
subjects = ["channels.>"]

[transformer]
filter=["temperature", "humidity"]
```

cmd/influxdb-writer/main.go
```diff
@@ -26,40 +26,40 @@ import (
...
-	envContentType = "MF_INFLUX_WRITER_CONTENT_TYPE"
+	defNatsURL         = "nats://localhost:4222"
+	defLogLevel        = "error"
+	defPort            = "8180"
+	defDB              = "mainflux"
+	defDBHost          = "localhost"
+	defDBPort          = "8086"
+	defDBUser          = "mainflux"
+	defDBPass          = "mainflux"
+	defSubjectsCfgPath = "/config.toml"
```
manuio: `defConfigPath`

cmd/influxdb-writer/main.go
```diff
@@ -26,40 +26,40 @@ import (
...
+
+	envNatsURL         = "MF_NATS_URL"
+	envLogLevel        = "MF_INFLUX_WRITER_LOG_LEVEL"
+	envPort            = "MF_INFLUX_WRITER_PORT"
+	envDB              = "MF_INFLUX_WRITER_DB"
+	envDBHost          = "MF_INFLUX_WRITER_DB_HOST"
+	envDBPort          = "MF_INFLUX_WRITER_DB_PORT"
+	envDBUser          = "MF_INFLUX_WRITER_DB_USER"
+	envDBPass          = "MF_INFLUX_WRITER_DB_PASS"
+	envSubjectsCfgPath = "MF_INFLUX_WRITER_CONFIG_PATH"
```
manuio: `envConfigPath`

cmd/influxdb-writer/main.go
```diff
@@ -26,40 +26,40 @@ import (
...
-	contentType string
+	natsURL         string
+	logLevel        string
+	port            string
+	dbName          string
+	dbHost          string
+	dbPort          string
+	dbUser          string
+	dbPass          string
+	subjectsCfgPath string
```
manuio: `configPath`

writers/cassandra/README.md
```diff
@@ -18,7 +18,7 @@ default values.
 | MF_CASSANDRA_WRITER_DB_USER         | Cassandra DB username                                     |                        |
 | MF_CASSANDRA_WRITER_DB_PASS         | Cassandra DB password                                     |                        |
 | MF_CASSANDRA_WRITER_DB_PORT         | Cassandra DB port                                         | 9042                   |
-| MF_CASSANDRA_WRITER_CONFIG_PATH     | Configuration file path with filters list                 | /configs.toml          |
+| MF_CASSANDRA_WRITER_SUBJECTS_CONFIG | Configuration file path with subjects list                | /config.toml           |
```
manuio: `MF_CASSANDRA_WRITER_CONFIG_PATH`

manuio (changes requested):

writers/cassandra/README.md
dborovcanin: I kept `MF_<NAME>_WRITER_CONFIG_PATH` and updated code, docs, and compose files to match it. In README files, we mention the NATS subject, so it should be clear. I removed the `keys` filter, but we'll possibly add something like that later - it's a nice option, but I'm not sure if it belongs to the Writers or Transformers. That's also the reason I didn't want to use NATS in config file path naming - it will probably be used for more complex configuration in the future and some of that config won't be related to NATS. This will probably need an update in k8s scripts (but nothing is broken due to the sensible defaults).

manuio (approved):
LGTM

drasko (approved):
LGTM

drasko:
Great contribution. Thanks a lot @dusanb94.

PR #1262, opened 2020-10-20
NOISSUE - Add ListUsers, ViewUser and ViewProfile methods 

manuio:

### DIFF
--- a/users/api/endpoint.go
+++ b/users/api/endpoint.go
@@ -81,7 +81,7 @@ func viewUserEndpoint(svc users.Service) endpoint.Endpoint {
 			return nil, err
 		}
 
-		u, err := svc.User(ctx, req.token, req.userID)
+		u, err := svc.ViewUser(ctx, req.token, req.userID)
 		if err != nil {
 			return nil, err
 		}
@@ -100,7 +100,7 @@ func viewProfileEndpoint(svc users.Service) endpoint.Endpoint {
 			return nil, err
 		}
 
-		u, err := svc.Profile(ctx, req.token)
+		u, err := svc.ViewProfile(ctx, req.token)
 		if err != nil {
 			return nil, err
 		}
@@ -118,7 +118,7 @@ func listUsersEndpoint(svc users.Service) endpoint.Endpoint {
 		if err := req.validate(); err != nil {
 			return users.UserPage{}, err
 		}
-		up, err := svc.Users(ctx, req.token, req.offset, req.limit, req.email, req.metadata)
+		up, err := svc.ListUsers(ctx, req.token, req.offset, req.limit, req.email, req.metadata)
 		if err != nil {
 			return users.UserPage{}, err
 		}
@@ -227,13 +227,13 @@ func removeUserFromGroup(svc users.Service) endpoint.Endpoint {
 	}
 }
 
-func listUsersForGroupEndpoint(svc users.Service) endpoint.Endpoint {
+func listMembersEndpoint(svc users.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(listUserGroupReq)
 		if err := req.validate(); err != nil {
 			return users.UserPage{}, err
 		}
-		up, err := svc.Members(ctx, req.token, req.groupID, req.offset, req.limit, req.metadata)
+		up, err := svc.ListMembers(ctx, req.token, req.groupID, req.offset, req.limit, req.metadata)
 		if err != nil {
 			return users.UserPage{}, err
 		}
@@ -241,13 +241,13 @@ func listUsersForGroupEndpoint(svc users.Service) endpoint.Endpoint {
 	}
 }
 
-func listUserGroupsEndpoint(svc users.Service) endpoint.Endpoint {
+func listMembershipsEndpoint(svc users.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(listUserGroupReq)
 		if err := req.validate(); err != nil {
 			return users.UserPage{}, err
 		}
-		gp, err := svc.Memberships(ctx, req.token, req.userID, req.offset, req.limit, req.metadata)
+		gp, err := svc.ListMemberships(ctx, req.token, req.userID, req.offset, req.limit, req.metadata)
 		if err != nil {
 			return groupPageRes{}, err
 		}
@@ -285,7 +285,7 @@ func viewGroupEndpoint(svc users.Service) endpoint.Endpoint {
 		if err := req.validate(); err != nil {
 			return viewGroupRes{}, err
 		}
-		group, err := svc.Group(ctx, req.token, req.groupID)
+		group, err := svc.ViewGroup(ctx, req.token, req.groupID)
 		if err != nil {
 			return viewGroupRes{}, err
 		}
@@ -304,7 +304,7 @@ func listGroupsEndpoint(svc users.Service) endpoint.Endpoint {
 		if err := req.validate(); err != nil {
 			return groupPageRes{}, err
 		}
-		gp, err := svc.Groups(ctx, req.token, req.groupID, req.offset, req.limit, req.metadata)
+		gp, err := svc.ListGroups(ctx, req.token, req.groupID, req.offset, req.limit, req.metadata)
 		if err != nil {
 			return groupPageRes{}, err
 		}
--- a/users/api/endpoint.go
+++ b/users/api/endpoint.go
@@ -93,6 +93,25 @@ func viewUserEndpoint(svc users.Service) endpoint.Endpoint {
 	}
 }
 
+func viewProfileEndpoint(svc users.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(viewUserReq)
+		if err := req.validate(); err != nil {
+			return nil, err
+		}
+
+		u, err := svc.Profile(ctx, req.token)
+		if err != nil {
+			return nil, err
+		}
+		return viewUserRes{
+			ID:       u.ID,
+			Email:    u.Email,
+			Metadata: u.Metadata,
+		}, nil
+	}
+}
+
 func listUsersEndpoint(svc users.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(listUsersReq)
--- a/users/api/endpoint.go
+++ b/users/api/endpoint.go
@@ -81,7 +81,7 @@ func viewUserEndpoint(svc users.Service) endpoint.Endpoint {
 			return nil, err
 		}
 
-		u, err := svc.User(ctx, req.token)
+		u, err := svc.User(ctx, req.token, req.userID)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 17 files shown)


users/postgres/users.go
```diff
@@ -141,6 +141,54 @@ func (ur userRepository) RetrieveByID(ctx context.Context, id string) (users.Use
 	return toUser(dbu)
 }
 
+func (ur userRepository) Users(ctx context.Context, offset, limit uint64) (users.UserPage, error) {
+	q := `SELECT id, email, metadata FROM users ORDER BY id LIMIT :limit OFFSET :offset;`
```
mteodor: what about filtering by metadata?

users/service.go
```diff
@@ -198,15 +201,21 @@ func (svc usersService) Login(ctx context.Context, user User) (string, error) {
 	return svc.issue(ctx, dbUser.Email, authn.UserKey)
 }
 
-func (svc usersService) User(ctx context.Context, token string) (User, error) {
+func (svc usersService) User(ctx context.Context, token, id string) (User, error) {
```
mteodor: this actually doesnt get user for ID only for token?

mteodor (changes requested):

users/postgres/users.go
manuio: I'm adding this now

users/api/endpoint.go
```diff
@@ -93,6 +93,20 @@ func viewUserEndpoint(svc users.Service) endpoint.Endpoint {
 	}
 }
 
+func listUsersEndpoint(svc users.Service) endpoint.Endpoint {
```
drasko, line 115: I am not quite sure, but I am leaning more towards `getUsers()`, because we have `update` below. Again - not quite sure. @dusanb94 @nmarcetic opinions?

drasko (changes requested):

users/api/endpoint.go
drasko: Looks actually that this terminology is used in other services, so for consistency it makes sense.

manuio: This naming is used for any endpoint that returns a list. For example `listThingsEndpoint`

dborovcanin: @drasko Makes sense, but we use `list` in the other services. I think that the only reason is that `viewEntity` and `viewEntitites` are too much alike, so it's a bit more readable with `list`. Other than that, I don't see any reasons for using `list` instead of `get`.

users/service.go
```diff
@@ -215,12 +221,17 @@ func (svc usersService) User(ctx context.Context, token string) (User, error) {
 	}, nil
 }
 
-func (svc usersService) ListUsers(ctx context.Context, token string, groupID string, offset, limit uint64, um Metadata) (UserPage, error) {
-	_, err := svc.identify(ctx, token)
+func (svc usersService) Users(ctx context.Context, token string, offset, limit uint64, email string, m Metadata) (UserPage, error) {
+	aemail, err := svc.identify(ctx, token)
```
drasko: Please add a comment that this is an admin email, because `aemail` looks confusing in the code.

drasko (changes requested):

users/service.go
manuio: This is not really an admin email but simply an user email (or identity). I renamed by `user`

users/service.go
```diff
@@ -198,15 +203,32 @@ func (svc usersService) Login(ctx context.Context, user User) (string, error) {
...
 }
 
-func (svc usersService) User(ctx context.Context, token string) (User, error) {
+func (svc usersService) User(ctx context.Context, token, id string) (User, error) {
 	email, err := svc.identify(ctx, token)
 	if err != nil {
 		return User{}, err
 	}
+
+	// Retrieve User infos by ID if User is Admin and UUID is valid
```
drasko: `User info`, not `User infos`. "Information" is uncountable noun, it does not nave plural.

drasko (changes requested):

users/service.go
manuio: For `id` if it's a valid UUID. If empty or not valid it uses Email

users/postgres/users.go
```diff
@@ -141,6 +141,95 @@ func (ur userRepository) RetrieveByID(ctx context.Context, id string) (users.Use
...
+		return q, qMeta, nil
+	}
+
+	if len(m) > 0 {
+		b, err := json.Marshal(m)
+		if err != nil {
+			return "", []byte{}, err
+		}
+		qMeta = b
+	}
```
dborovcanin: Can you generalize this, [getUsersMetadataQuery](https://github.com/mainflux/mainflux/blob/420b598ac711e56e6d4d86987f8dd0e1caeb9532/users/postgres/users.go#L293), and [getGroupsMetadataQuery](https://github.com/mainflux/mainflux/blob/420b598ac711e56e6d4d86987f8dd0e1caeb9532/users/postgres/groups.go#L427) to a single parameterized method?

users/postgres/users.go
```diff
@@ -141,6 +141,95 @@ func (ur userRepository) RetrieveByID(ctx context.Context, id string) (users.Use
...
+	q := ""
+	qMeta := []byte("{}")
+
+	if email == "" && len(m) == 0 {
+		return q, qMeta, nil
+	}
+
+	if email != "" && len(m) == 0 {
+		return q, qMeta, nil
+	}
```
dborovcanin: Duplicated lines.

dborovcanin (changes requested):

users/postgres/users.go
manuio: I accidentally removed a query. No it's ok. Thanks!

users/postgres/users.go
```diff
@@ -290,17 +366,30 @@ func toUser(dbu dbUser) (users.User, error) {
...
+		return query, param, nil
+	}
+
+	return query, param, nil
+}
 
-		b, err := json.Marshal(m)
+func createMetadataQuery(entity string, meta users.Metadata) (string, []byte, error) {
+	query := ""
+	param := []byte("{}")
```
dborovcanin: Is param allowed to be `nil` if len(meta) == 0? If so, this can be simplified:
```go

func createMetadataQuery(entity string, meta users.Metadata) (string, []byte, error) {
	if len(meta) == 0 {
		return "", nil, nil
	}
	param, err := json.Marshal(meta)
	if err != nil {
		return "", nil, err
	}
	query := fmt.Sprintf("%smetadata @> :metadata", entity)
	return query, param, nil
}
```

users/postgres/users.go
```diff
@@ -290,17 +366,30 @@ func toUser(dbu dbUser) (users.User, error) {
 	}, nil
 }
 
-func getUsersMetadataQuery(m users.Metadata) ([]byte, string, error) {
-	mq := ""
-	mb := []byte("{}")
-	if len(m) > 0 {
-		mq = ` AND users.metadata @> :metadata`
+func createEmailQuery(entity string, email string) (string, string, error) {
```
dborovcanin, line 369: Maybe you can simplify a bit:
```go

func createEmailQuery(entity, email string) (string, string, error) {
	if email == "" {
		return "", "", nil
	}
	// Create LIKE operator to search Users with an email containing a given string.
	param := fmt.Sprintf(`%%%s%%`, email)
	query := fmt.Sprintf("%semail LIKE :email", entity)
	return query, param, nil
}
```

users/postgres/users.go
```diff
@@ -290,17 +366,30 @@ func toUser(dbu dbUser) (users.User, error) {
...
+
+	return query, param, nil
+}
 
-		b, err := json.Marshal(m)
+func createMetadataQuery(entity string, meta users.Metadata) (string, []byte, error) {
+	query := ""
+	param := []byte("{}")
+	if len(meta) > 0 {
+		query = fmt.Sprintf("%smetadata @> :metadata", entity)
```
dborovcanin: Should this be:
```go
query = fmt.Sprintf("%s.metadata @> :metadata", entity)
```

dborovcanin: Maybe we should add better tests, I suspect that it would detect this kind of issues.

dborovcanin (changes requested):

users/postgres/users.go
manuio: No, because if entity is empty you would search `.metadata` instead of `metadata` . There are test covering this part (I had many errors building this queries) but I can add more

users/service_test.go
```diff
@@ -116,6 +116,49 @@ func TestLogin(t *testing.T) {
...
+			user:   u,
+			token:  token,
+			userID: id,
+			err:    nil,
+		},
+		"view user with unauthorized token": {
+			user:   users.User{},
+			token:  "",
+			userID: id,
+			err:    users.ErrUnauthorizedAccess,
```
mteodor, line 145: why dont we put errors like this in separate package, `errors` maybe so that we can use it in each service not repeating it?

mteodor (dismissed):

users/api/endpoint.go
```diff
@@ -194,27 +227,27 @@ func removeUserFromGroup(svc users.Service) endpoint.Endpoint {
 	}
 }
 
-func listUsersForGroupEndpoint(svc users.Service) endpoint.Endpoint {
+func listMembersEndpoint(svc users.Service) endpoint.Endpoint {
```
dborovcanin, line 230: I like that you renamed this. I like `memberships` and `members` terminology, it's unambiguous and short.

users/mocks/users.go
```diff
@@ -89,7 +89,14 @@ func (urm *userRepositoryMock) RetrieveByID(ctx context.Context, id string) (use
 	return val, nil
 }
 
-func (urm *userRepositoryMock) Members(ctx context.Context, groupID string, offset, limit uint64, gm users.Metadata) (users.UserPage, error) {
+func (urm *userRepositoryMock) RetrieveAll(ctx context.Context, offset, limit uint64, email string, um users.Metadata) (users.UserPage, error) {
```
dborovcanin, line 92: Why don't we mock this method? Do we miss some tests?

users/service.go
```diff
@@ -198,15 +206,37 @@ func (svc usersService) Login(ctx context.Context, user User) (string, error) {
 	return svc.issue(ctx, dbUser.Email, authn.UserKey)
 }
 
-func (svc usersService) User(ctx context.Context, token string) (User, error) {
+func (svc usersService) ViewUser(ctx context.Context, token, id string) (User, error) {
+	// TODO: Check Policy
```
dborovcanin: Please remove `TODO` and open an issue for this.

users/service.go
```diff
@@ -215,12 +245,14 @@ func (svc usersService) User(ctx context.Context, token string) (User, error) {
 	}, nil
 }
 
-func (svc usersService) ListUsers(ctx context.Context, token string, groupID string, offset, limit uint64, um Metadata) (UserPage, error) {
+func (svc usersService) ListUsers(ctx context.Context, token string, offset, limit uint64, email string, m Metadata) (UserPage, error) {
+	// TODO: Check Policy
```
dborovcanin: Same here. No need to open two issues, just reference the methods that need to be updated in one issue.

manuio: Here it is: https://github.com/mainflux/mainflux/issues/1267

users/mocks/users.go
manuio: Yes, some service tests are missing. I added a mock.

users/api/endpoint.go
manuio: We agreed with @mteodor to use this naming. I'm just polishing

users/mocks/users.go
drasko: Can we add an issue for the tests?

manuio: Here: https://github.com/mainflux/mainflux/pull/1262/files#diff-28ebac99dae6f7fda6ef1772531588fc3e201a5bac3a41c3ea0c760989edb96aR152

drasko (approved):
LGTM

dborovcanin (approved):

Issue #935, opened 2019-11-05
Add Vault docs


dborovcanin:
**FEATURE REQUEST**

1. Is there an open issue addressing this request? If it does, please add a "+1" reaction to the
   existing issue, otherwise proceed to step 2.
No.

2. Describe the feature you are requesting, as well as the possible use case(s) for it.
[Vault](https://www.vaultproject.io) is the official recommendation for a certificate management tool, so we should add steps for setting it up and using it for client certificate management to our documentation.

3. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).
This is a should-have feature.

arvindh123:
New feature and documentation are added in following PR 
https://github.com/absmach/magistrala/pull/2084
https://github.com/absmach/magistrala/pull/2091

dborovcanin:
We are moving from Vault to OpenBao. Closing this one.

PR #1266, opened 2020-10-23
NOISSUE - Vault integration as an addon

tritao:
### What does this do?

Integrates Vault PKI service as a service addon.

Also adds some helper scripts to help setup the CA in Vault, as well as
some docs to explain how to use them.

Originally based from https://github.com/mteodor/vault.

Signed-off-by: Joao Matos <joao@tritao.eu>

Could be related to https://github.com/mainflux/mainflux/issues/935.

Yes.


dborovcanin (commented):
@tritao Thanks for the contribution. @mainflux/maintainers will review and comment during this week. Did you test integration with the Provision service?

nmarcetic:
> @tritao Thanks for the contribution. @mainflux/maintainers will review and comment during this week. Did you test integration with the Provision service?

@tritao [Provision service](https://mainflux.readthedocs.io/en/latest/provision/#provision-service). Doc doest mention Vault, but you can see it in [Certs service](https://mainflux.readthedocs.io/en/latest/certs/#certs-service)

tritao:
I didn't use the Provision service yet, however I did test Vault with the Certs service and I can successfully issue new certificates.

There are some improvements to Certs service around Vault service that I want to send as another PR, however I was waiting until this one was merged, because it depends on the new Vault environment variables in `.env`.

Should I send Certs PR now or wait for this one?

docker/addons/vault/Dockerfile
```diff
@@ -0,0 +1,37 @@
+# base image
```
nmarcetic: Can we use the official Vaulet docker image and avoid custom or any Dockerfiles?

dborovcanin:
@tritao I think that this PR will be merged, but it's better to wait for the proper review so that you don't have to redo things.

docker/addons/vault/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+
+volumes:
+  mainflux-vault-volume:
+
+services:
+  vault:
+    image: mainflux/vault:latest
+    container_name: mainflux-vault
+    ports:
+      - 8200:8200
```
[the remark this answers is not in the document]
tritao: Updated with variable, but used the default Vault port in container to simplify the config.

tritao:
Updated with all feedback.

Now using the official Vault image, unfortunately I could not get rid of the custom Docker image.

I spent some time trying but there is a bug with docker-compose handling of relative paths when being called with multiple compose files, like: `docker-compose -f docker/docker-compose.yml f docker/addons/vault/docker-compose.yml`

It will assume paths are relative to the root file so it breaks relative path behavior when using volumes to pass files (I tested and absolute paths worked OK).

mteodor:
> Updated with all feedback.
> 
> Now using the official Vault image, unfortunately I could not get rid of the custom Docker image.
> 
> I spent some time trying but there is a bug with docker-compose handling of relative paths when being called with multiple compose files, like: `docker-compose -f docker/docker-compose.yml f docker/addons/vault/docker-compose.yml`
> 
> It will assume paths are relative to the root file so it breaks relative path behavior when using volumes to pass files (I tested and absolute paths worked OK).

than you for your contribution
I think that it is better to use official vault image without custom Dockerfile, check the other addons how they handle volumes and file paths, so same principle can be applied here
It is better to use two commands and have official vault image.
Tako care that if official vault is used I'm not sure  which shell it uses if it has any ( regarding the execution of entrypoint.sh)

docker/addons/vault/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+docker build . -t mainflux/vault:latest
```
mteodor: this should be in main Makefile

docker/addons/vault/.gitignore
```diff
@@ -0,0 +1 @@
+data
```
mteodor, line 1: remove this file from PR

.env
```diff
@@ -153,8 +153,22 @@ MF_CERTS_SIGN_RSA_BITS=2048
 MF_CERTS_VAULT_HOST=
 MF_CERTS_VAULT_PKI_PATH=pki_int
```
mteodor, line 154: default values here are `pki_int_mainflux` and role `mainflux` right? we should put them here than

.env
```diff
@@ -153,8 +153,22 @@ MF_CERTS_SIGN_RSA_BITS=2048
...
+MF_CERTS_VAULT_TOKEN=
 
+### Vault
+MF_VAULT_HOST=vault
+MF_VAULT_PORT=8200
+MF_VAULT_UNSEAL_KEY_1=
+MF_VAULT_UNSEAL_KEY_2=
+MF_VAULT_UNSEAL_KEY_3=
+MF_VAULT_TOKEN=
+MF_VAULT_CA_NAME=mainflux5
```
mteodor: are these default values right?

.env
tritao: I am going to change these in the next PR which is going to be related to certs service.

.env
tritao: I am using the same defaults you were using in https://github.com/mteodor/vault/blob/master/set-pki.sh#L11

Do you want me to change them? And if so, what to?

docker/addons/vault/.gitignore
tritao: Why remove this file? 

The `init.sh` script creates a `data/secrets` file, along with the certificates, so this is a perfectly valid usage of Git ignore.

Not having this will lead to Git always having files for commit that should not be under version control.

tritao:
> I think that it is better to use official vault image without custom Dockerfile, check the other addons how they handle volumes and file paths, so same principle can be applied here

The other addons have the same bug.

Docker-compose will always compute relative paths relative to the the directory of the first compose file that is passed.

So if you launch like this:

`docker-compose -f docker/docker-compose.yml up`

`docker/docker-compose.yml` paths will be relative to `docker` and everything works.

But if you launch like this:

`docker-compose -f docker/docker-compose.yml -f docker/addons/addon/docker-compose.yml up`

Then `addon/docker-compose.yml` paths will still be relative to `docker`, so a line like ` - ./subjects.toml:/config/subjects.toml` will fail to find `subjects.toml`. Docker will mount an empty directory instead with the name of the file.

You should be able to easily confirm this.

Given this docker-compose behavior (bug?), I think a solution should be found, because right now, depending on how user launches the addons in the system, it leads to broken platform (and hard to debug problem).

Potential solutions:
1. Build docker images with files (like I did here)
2. Move all addons folders to the `docker` folder (and do not use nested folders with compose files)
3. Build some kind of tool or make commands that call docker-compose with working directory from `docker-compose.yml`

I think solution 1. is easiest, and good idea to have Mainflux containers for all services with all necessary files to make them self contained when deploying.

If you want to me change this as you asked to still use relative paths, I will for the sake of merging. But it's a broken design given this docker-compose behavior.

dborovcanin:
> Updated with all feedback.
> 
> Now using the official Vault image, unfortunately I could not get rid of the custom Docker image.
> 
> I spent some time trying but there is a bug with docker-compose handling of relative paths when being called with multiple compose files, like: `docker-compose -f docker/docker-compose.yml f docker/addons/vault/docker-compose.yml`
> 
> It will assume paths are relative to the root file so it breaks relative path behavior when using volumes to pass files (I tested and absolute paths worked OK).

If the only reason for the custom Docker image is multiple compose files support - I'd favor the official image over the multiple compose file support. Just mention it somewhere in the docs, and that's it.

tritao:
> If the only reason for the custom Docker image is multiple compose files support - I'd favor the official image over the multiple compose file support. Just mention it somewhere in the docs, and that's it.

I think an even more important reason is simplification of production deployments.

Docker containers should be, if possible self-contained, so that an user can push them to an organization Docker registry, and from a production machine, either using an orchestrator like Nomad or Kubernetes, or just using docker-compose, pull them from the registry and `up` them.

If the containers are not self-contained, then a separate system for uploading (and rollback, if necessary) the necessary support files will be necessary. In this specific case, the shell file `entrypoint.sh` (and config file) are crucial to the startup of the Vault instance. Without them, the system will not work. They belong inside the container, where they can be distributed and versioned in a self-contained way.

They can still be overridden from a `docker-compose.yml` file if necessary by the user.

If you think that is a good idea, I can add support for building the containers and pushing to registry via the Makefile system, and change other containers to work like that.

If not, then I will revert this back to just using a relative file like requested.

Let me know what I should do.

docker/addons/vault/.gitignore
mteodor: ok, my bad

.env
mteodor: ok, this like my 5th iteration, if it is not a problem just put `mainflux`

drasko:
@tritao avoid custom images. We will not be ones maintaining Vault image. Also - it is extremely beneficial to have configurable deployments, so all config files must be outside the image and mapped via `volume`.

Sometimes mapping config files is enough, and yes - if special scripts need to be executed, than `entrypoint.sh` is needed. This is not a problem at all, as all our deployments are orchestrated - either via docker-compose or via K8s.

tritao:
> @tritao avoid custom images. We will not be ones maintaining Vault image. Also - it is extremely beneficial to have configurable deployments, so all config files must be outside the image and mapped via `volume`.
> 
> Sometimes mapping config files is enough, and yes - if special scripts need to be executed, than `entrypoint.sh` is needed. This is not a problem at all, as all our deployments are orchestrated - either via docker-compose or via K8s.

@drasko Just to make it clear, there is not any custom Vault image, the PR was updated to be based from the official image based upon the initial feedback, it uses `FROM vault:latest`. So I think the maintainability concerns do not apply.

It just creates another Docker image based off the official one with the necessary files for it to work. The shell script (`entrypoint.sh`) is *always necessary* to unseal the Vault, this is why I think it makes sense that it is inside the container image.

Anyway, I've updated the PR to work like requested.

docker/addons/vault/README.md
```diff
@@ -0,0 +1,104 @@
+This is Vault service deployment to be used with Mainflux.
+
+When the Vault service is started, some initialization steps need to be done to set things up.
+
+## Build
+
+To build the Vault service container image, run the `build.sh` script.
+
```
dborovcanin: This can be removed now.

dborovcanin (commented):
Please update the branch, remove the unused section in the README and it's ready.

tritao:
Updated.

dborovcanin (approved):

drasko (approved):
LGTM

drasko:
Merged!

Thanks @tritao for this great contribution!

PR #1269, opened 2020-10-26
NOISSUE - Fix ViewGroup and UpdateGroup

manuio:

### DIFF
--- a/users/api/endpoint.go
+++ b/users/api/endpoint.go
@@ -263,6 +263,7 @@ func updateGroupEndpoint(svc users.Service) endpoint.Endpoint {
 		}
 
 		group := users.Group{
+			ID:          req.id,
 			Name:        req.Name,
 			Description: req.Description,
 			Metadata:    req.Metadata,
--- a/users/api/endpoint.go
+++ b/users/api/endpoint.go
@@ -259,23 +259,20 @@ func updateGroupEndpoint(svc users.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(updateGroupReq)
 		if err := req.validate(); err != nil {
-			return createGroupRes{}, err
+			return updateGroupRes{}, err
 		}
+
 		group := users.Group{
 			Name:        req.Name,
 			Description: req.Description,
 			Metadata:    req.Metadata,
 		}
+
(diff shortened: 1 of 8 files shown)


users/api/requests.go
```diff
@@ -129,6 +129,7 @@ func (req createGroupReq) validate() error {
 
 type updateGroupReq struct {
 	token       string
+	id          string
 	Name        string                 `json:"name,omitempty"`
 	Description string                 `json:"description,omitempty"`
```
mteodor, line 135: we need parent id here also

users/api/requests.go
```diff
@@ -138,12 +140,16 @@ func (req updateGroupReq) validate() error {
...
 	}
+	if req.id == "" {
+		return users.ErrMalformedEntity
+	}
 	if req.Name == "" {
 		return users.ErrMalformedEntity
 	}
 	if len(req.Name) > maxNameSize {
 		return users.ErrMalformedEntity
 	}
```
dborovcanin: You can `or` these conditions.

users/api/endpoint.go
```diff
@@ -259,23 +259,20 @@ func updateGroupEndpoint(svc users.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(updateGroupReq)
 		if err := req.validate(); err != nil {
-			return createGroupRes{}, err
+			return updateGroupRes{}, err
 		}
+
 		group := users.Group{
 			Name:        req.Name,
```
mteodor, line 267: mising ID

mteodor (dismissed):

drasko (commented):
OK for me, just take into account @dusanb94 's remark and I'll approve

drasko (approved):
LGTM

users/api/requests.go
dborovcanin: I meant that you can `or` all of these checks: `if id == "" || name == "" || len(name) > maxNameLen`.

manuio: I agreed to group this checks by type but for consistency with other services and readability I would keep it like that.
You can check here: https://github.com/mainflux/mainflux/blob/master/things/api/things/http/requests.go

dborovcanin (approved):

PR #1274, opened 2020-10-27
MF-1268 - CLI improvements

### LINKED ISSUE #1268: CLI improvements
**FEATURE REQUEST**

As a recent user of Mainflux, I have spent some time using the CLI tool for creating users, things, channels,
as well as certificates and found some problems with user experience, even had to go through the source code to understand how to use some commands.

As such, I would like to suggest some changes that I think could help the new user experience a bit as well as make the CLI tool a bit more self-documented and easier to use.

## Introduce flags to commands instead of just using positional parameters

Lets take an example of the `certs issue` command:

Right now, it is like this: `"issue <thing_id> <keybits> <keytype> <hoursvalid> <user_auth_token>"`

The user has no way of knowing what each parameter expects, and also there is no way to tell the CLI about default good values. I suggest using flags instead for the optional parameters, which can have default values set in Cobra.

`usage: issue <thing_id> [--keysize=2048] [--keytype=rsa] [--ttl 128000]`

This provides the user with some information about the default values, as well as makes them optional if the default values are good enough (which most often are).

The long help for each command can be then used to document each flag, for instance, `--keytype` can take `rsa` or `ec`.

## Introduce a CLI config file as well as a new `--config <path>` flag.

The purpose of the config file would be to be able to contain settings for the CLI tool.
The config file would be read at launch and the CLI tool would automatically know default parameters for some flags.

Things like the Mainflux URL instance, and default user auth token could be setup, as well as whatever comes up in the future.

I was thinking maybe something like https://github.com/spf13/viper could be used, what do you think?

## Remove explicit user token positional parameters from all commands and use a global flag instead

It gets tiring to always have to copy paste the user auth token, and since it is used by a lot of commands,
I would suggest to introduce a global flag `--auth-token` or such, that would be used by default for the commands
that need to pass it to Mainflux API calls.

The default value would be read from config file. Each command that needs it would check to see if there is the
user auth token set, if not it would give an error message to the user telling how to set it.

Nice-to-have.

Let me know your thoughts on these suggestions, I can send a PR for this if you think it is a good idea.

tritao:
### What does this do?

Introduces an initial set of improvements to the CLI as suggested in the issue.

There is a new `~/.config/mainflux/cli-config.toml` file that can have an `auth_token` set.
This config file is read by default put a custom config can be passed with `--config` CLI flag.

Some commands have been refactored to take advantage of this config token.
Other positional arguments have also been converted to flags.

This is an initial pass, and other commands can be refactored in another PR to make this easier to review.

Resolves #1268 

I renamed the `cert` command to `certs` to be consistent with all the other commands.


dborovcanin (commented):
Please update the branch and fix CI (it's linter, nothing big). Thanks!

cmd/cli/main.go
```diff
@@ -32,6 +32,11 @@ func main() {
 	var rootCmd = &cobra.Command{
 		Use: "mainflux-cli",
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			err := cli.ParseConfig()
```
drasko: If error is not needed below, and it looks like it is not, you can embed this check and function call in one line.

cli/users.go
```diff
@@ -117,23 +123,22 @@ var cmdUsers = []cobra.Command{
...
 		Run: func(cmd *cobra.Command, args []string) {
-			logUsage("Usage: users [create | get | update | token | password]")
+			logUsage("users [create | get | update | token | password]")
 		},
 	}
 
-	for i := range cmdUsers {
-		cmd.AddCommand(&cmdUsers[i])
-	}
+	cmd.AddCommand(&createCmd)
```
drasko: Adding command by cmmand is maybe good for documentation purposes, but I kinda liked this geeky loop.

@dusanb94 @manuio @nmarcetic opnions?

drasko (changes requested):

cli/users.go
dborovcanin: I'm OK with either solution, I really don't have preferences.

tritao: It's not possible to add flags easily when the commands are in the array, this is why I changed it.

manuio: If we do this here we will have to do it for all commands and this should be done in a separate PR and discussed before. I also prefer to geeky loop :)

tritao:
Fixed the linting issues, by the way is there any way to run the linter locally?

cli/config.go
```diff
@@ -0,0 +1,82 @@
...
+
+type Config struct {
+	AuthToken string `toml:"auth_token"`
+}
+
+// Save - store config in a file
+func Save(c Config, file string) error {
+	b, err := toml.Marshal(c)
+	if err != nil {
+		return errors.New(fmt.Sprintf("Error reading config file: %s", err))
```
manuio: Please, avoid using `error` in error messages.  Maybe you can use `failed to read config file`

manuio (dismissed):

cmd/cli/main.go
tritao: Sure, should I do the same in `getUserAuthToken`?

manuio:
> Fixed the linting issues, by the way is there any way to run the linter locally?

@tritao If you have golint installed you can do it with: `golint ./...`

tritao:
Updated PR with all feedback.

* Added back array-based command registration
* Improved error handling by moving checking to parent functions
* Improved error messages

cli/config.go
```diff
@@ -0,0 +1,83 @@
...
+		return errors.New(fmt.Sprintf("failed to read config file: %s", err))
+	}
+	if err := ioutil.WriteFile(file, b, 0644); err != nil {
+		return errors.New(fmt.Sprintf("failed to write config TOML: %s", err))
+	}
+	return nil
+}
+
+// Read - retrieve config from a file
+func Read(file string) (Config, error) {
```
dborovcanin: Are `Read` and `GetConfigPath` used by any other package? If not, there is no need to export them.

dborovcanin (commented):
Only one simple remark/question. Other than that, it's ready to be merged.

cli/config.go
dborovcanin: If you decide to make them not exported, please move the unexported code after the exported methods (`Save` and `Parse`).

tritao: Makes sense, updated.

Save is currently not used, but will use it in a later PR.

tritao:
Updated with latest feedback.

cmd/cli/main.go
```diff
@@ -32,6 +32,8 @@ func main() {
 	var rootCmd = &cobra.Command{
 		Use: "mainflux-cli",
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			cli.ParseConfig()
```
drasko, line 35: Function returns an error, so please check it and treat it.

drasko (changes requested):

cmd/cli/main.go
drasko: Not sure that you can do the same here, because you use value below.

cli/certs.go
```diff
@@ -1,54 +1,57 @@
...
+// NewCertsCmd returns certificate command.
+func NewCertsCmd() *cobra.Command {
+	var keySize uint16
+	var keyType string
+	var ttl uint32
+
+	issueCmd := cobra.Command{
 		Use:   "issue",
-		Short: "issue <thing_id> <keybits> <keytype> <hoursvalid> <user_auth_token>",
+		Short: "issue <thing_id> [--keysize=2048] [--keytype=rsa] [--ttl=<hoursvalid>]",
```
drasko: Why are `keysize` and `keytype` values hardcoded in this example, and `ttl` is not? For the consistency sake, either represent all with their default values, or none.

drasko (changes requested):

drasko:
@tritao just to double-check, using config file stays optional, does not become mandatory - right?

tritao:
> @tritao just to double-check, using config file stays optional, does not become mandatory - right?

Correct, configs are optional.

Good catch on the error, it was some leftover return type, updated the PR.

drasko (approved):
LGTM

dborovcanin (approved):

drasko:
Merged! Thanks @tritao for the contribution.

PR #1279, opened 2020-11-09
NOISSUE - Improve storing arbitrary JSON data

dborovcanin:
### What does this do?
This pull request improves storing arbitrary JSON data. JSON arrays in the root are now supported, as well as storing different formats in different collections/tables/measurements.

There is no such issue(s).

JSON data will no more be stored in a single collection/table/measurement.

Yes.

No.

### DIFF
--- a/pkg/transformers/json/README.md
+++ b/pkg/transformers/json/README.md
@@ -2,3 +2,53 @@
 
 JSON Transformer provides Message Transformer for JSON messages.
 To transform Mainflux Message successfully, the payload must be a JSON object.
+
+For the messages that contain _JSON array as the root element_, JSON Transformer does normalization of the data: it creates a separate JSON message for each JSON object in the root. In order to be processed and stored properly, JSON messages need to contain message format information. For the sake of the simpler storing of the messages, nested JSON objects are flatten to a single JSON object, using composite keys with the default separator `/`. This implies that the separator character (`/`) _is not allowed in the JSON object key_. For example, the following JSON object:
+```json
+{
+    "name": "name",
+    "id":8659456789564231564,
+    "in": 3.145,
+    "alarm": true,
+    "ts": 1571259850000,
+    "d": {
+        "tmp": 2.564,
+        "hmd": 87,
+        "loc": {
+            "x": 1,
+            "y": 2
+        }
+    }
+}
+```
+
+will be transformed to:
+
+```json
+
+{
+    "name": "name",
+    "id":8659456789564231564,
+    "in": 3.145,
+    "alarm": true,
+    "ts": 1571259850000,
+    "d/tmp": 2.564,
+    "d/hmd": 87,
+    "d/loc/x": 1,
+    "d/loc/y": 2
+}
+```
+
+The message format is stored in *the subtopic*. It's the last part of the subtopic. In the example:
+
+```
+http://localhost:8185/channels/<channelID>/messages/home/temperature/myFormat
+```
+
+the message format is `myFormat`. It can be any valid subtopic name, JSON transformer is format-agnostic. The format is used by the JSON message consumers so that they can process the message properly. If the format is not present (i.e. message subtopic is empty), JSON Transformer will report an error. Since the Transformer is agnostic to the format, having format in the subtopic does not prevent the publisher to send the content of different formats to the same subtopic. It's up to the consumer to handle this kind of issue. Message writers, for example, will store the message(s) in the table/collection/measurement (depending on the underlying database) with the name of the format (which in the example is `myFormat`). Mainflux writers will try to save any format received (whether it will be successful depends on the writer implementation and the underlying database), but it's recommended that the publisher takes care not to send different formats to the same subtopic.
+
+Having a message format in the subtopic means that the subscriber has an option to subscribe to only one message format. This is a nice feature because message subscribers know what's the expected format of the message so that they can process it. If the message format is not important, wildcard subtopic can always be used to subscribe to any message format:
+
+```
+http://localhost:8185/channels/<channelID>/messages/home/temperature/*
+```
--- a/pkg/transformers/json/transformer.go
+++ b/pkg/transformers/json/transformer.go
@@ -40,38 +40,40 @@ func transformer(msg messaging.Message) (interface{}, error) {
 		Channel:   msg.Channel,
 		Subtopic:  msg.Subtopic,
 	}
-	if err := json.Unmarshal(msg.Payload, &ret.Payload); err != nil {
+	var payload interface{}
+	if err := json.Unmarshal(msg.Payload, &payload); err != nil {
 		return nil, err
 	}
-	switch payload := ret.Payload.(type) {
+	switch p := payload.(type) {
 	case map[string]interface{}:
 		pld := make(map[string]interface{})
-		p, err := flatten("", pld, payload)
+		flat, err := flatten("", pld, p)
 		if err != nil {
 			return nil, err
 		}
-		ret.Payload = p
+		ret.Payload = flat
+		return []Message{ret}, nil
 	case []interface{}:
-		// Convert an array to the flat map array.
-		resPld := []map[string]interface{}{}
-		for _, val := range payload {
+		res := []Message{}
+		// Make an array of messages from the root array.
+		for _, val := range p {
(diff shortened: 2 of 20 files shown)


drasko:
Probably `content-type` (`ct`) subtopic info needs to be re-introduced so that we can pass this additional info that is needed.

writers/postgres/init.go
```diff
@@ -67,15 +67,16 @@ func migrateDB(db *sqlx.DB) error {
...
 				},
 			},
 			{
 				Id: "messages_2",
 				Up: []string{
 					`CREATE TABLE IF NOT EXISTS json (
                         id            UUID,
-                        channel       UUID,
+						channel       UUID,
+						created       BIGINT,
```
manuio: Is this ok?

pkg/transformers/json/transformer.go
```diff
@@ -14,8 +14,14 @@ import (
...
 
-// ErrInvalidKey represents an invalid JSON key format.
-var ErrInvalidKey = errors.New("invalid object key")
+var (
+	// ErrInvalidKey represents an invalid JSON key format.
+	ErrInvalidKey = errors.New("invalid object key")
+
+	errUnknownFormat     = errors.New("unknown format of JSON message")
+	errInvalidFormat     = errors.New("invalid JSON object")
+	errInvalidNestedJSON = errors.New("invalid nested JSON object")
```
manuio, line 25: capital letters?

manuio (changes requested):

writers/cassandra/messages.go
```diff
@@ -45,3 +60,43 @@ func (cr *cassandraRepository) Save(messages interface{}) error {
...
+}
+
+func (cr *cassandraRepository) insertJSON(msgs mfjson.Messages) error {
+	for _, msg := range msgs.Messages {
+		pld, err := json.Marshal(msg.Payload)
+		if err != nil {
+			return err
+		}
+		cql := `INSERT INTO %s (id, channel, created, subtopic, publisher, payload)
+            VALUES (?, ?, ?, ?, ?, ?)`
```
manuio: tabulation?

manuio (changes requested):

pkg/transformers/json/transformer.go
dborovcanin: Fixed. All the specific errors are wrapped in the single `ErrTransform`.

writers/cassandra/messages.go
```diff
@@ -45,3 +60,43 @@ func (cr *cassandraRepository) Save(messages interface{}) error {
 
 	return nil
 }
+
+func (cr *cassandraRepository) saveJSON(msgs mfjson.Messages) error {
+	if err := cr.insertJSON(msgs); err != nil {
+		if err == errNoTable {
+			if err := cr.creteTable(msgs.Format); err != nil {
```
mteodor: createTable

writers/postgres/messages.go
```diff
@@ -97,57 +104,94 @@ func (pr postgresRepo) saveSenml(messages interface{}) error {
...
 }
 
-func (pr postgresRepo) saveJSON(messages interface{}) error {
-	msg, ok := messages.(mfjson.Message)
-	if !ok {
-		return errSaveMessage
+func (pr postgresRepo) saveJSON(msgs mfjson.Messages) error {
+	if err := pr.insertJSON(msgs); err != nil {
+		if err == errNoTable {
+			if err := pr.createTable(msgs.Format); err != nil {
```
mteodor, line 110: why do we createTable here and we dont in saveSenml

writers/cassandra/messages.go
```diff
@@ -45,3 +60,43 @@ func (cr *cassandraRepository) Save(messages interface{}) error {
...
+	return nil
+}
+
+func (cr *cassandraRepository) insertJSON(msgs mfjson.Messages) error {
+	for _, msg := range msgs.Messages {
+		pld, err := json.Marshal(msg.Payload)
+		if err != nil {
+			return err
+		}
+		cql := `INSERT INTO %s (id, channel, created, subtopic, publisher, payload)
```
mteodor: this query could be created in one line

writers/cassandra/messages.go
```diff
@@ -45,3 +60,43 @@ func (cr *cassandraRepository) Save(messages interface{}) error {
...
+			if err.Error() == fmt.Sprintf("unconfigured table %s", msgs.Format) {
+				return errNoTable
+			}
+			return errors.Wrap(errSaveMessage, err)
+		}
+	}
+	return nil
+}
+
+func (cr *cassandraRepository) creteTable(name string) error {
```
mteodor: createTable

writers/postgres/messages.go
mteodor, line 108: why do we createTable here and in `saveSenml` we dont

writers/postgres/messages.go
```diff
@@ -97,57 +104,94 @@ func (pr postgresRepo) saveSenml(messages interface{}) error {
...
 		}
 
-		return errors.Wrap(errSaveMessage, err)
+		if err = tx.Commit(); err != nil {
+			err = errors.Wrap(errSaveMessage, err)
+		}
+		return
+	}()
+
+	q := `INSERT INTO %s (id, channel, created, subtopic, publisher, protocol, payload)
```
mteodor, line 137: can this be in one line

writers/postgres/messages.go
dborovcanin: Because all SenML messages are stored in a single table (created in `init.go`), while JSON messages are stored in separate tables depending on the format. While this does not make too much sense for some writer implementations (such as Postgres - we store payload as a JSON column anyway, so we'll end up with multiple tables with the same schema, but different content of `payload` JSON column), it helps logically organizing messages and easier consuming messages from the DB later.

writers/postgres/messages.go
dborovcanin: Unlike [in this case](https://github.com/mainflux/mainflux/pull/1279#discussion_r526319716), we're using `NamedExec` here, so the rest of the query is much larger and can't fit in a single line.

mteodor: actually I meant together with this `q = fmt.Sprintf(q, msgs.Format)`

dborovcanin: I find it a bit easier to read this way:
```go
q := `INSERT INTO %s (id, channel, created, subtopic, publisher, protocol, payload)
      VALUES (:id, :channel, :created, :subtopic, :publisher, :protocol, :payload);`
q = fmt.Sprintf(q, msgs.Format)
```
than this:
```go
q = fmt.Sprintf(`INSERT INTO %s (id, channel, created, subtopic, publisher, protocol, payload) 
                 VALUES (:id, :channel, :created, :subtopic, :publisher, :protocol, :payload);`, msgs.Format)
```
If you insist, I'll use the second option.

dborovcanin:
> Probably `content-type` (`ct`) subtopic info needs to be re-introduced so that we can pass this additional info that is needed.

As we agreed, content-type (format) is handled as [the last subtopic](https://github.com/mainflux/mainflux/pull/1279/files#diff-6b97c5458728b7166c8a9d837208c49996bf949d558c66f5c32864f7905d0d52R48) This was the simplest solution, and we can change it in the future if it turns that it's not viable.

pkg/transformers/json/README.md
```diff
@@ -2,3 +2,53 @@
...
+
+```json
+
+{
+    "name": "name",
+    "id":8659456789564231564,
+    "in": 3.145,
+    "alarm": true,
+    "ts": 1571259850000,
+    "d/tmp": 2.564,
```
drasko, line 35: Why do we use `/` instead of `.` for flattening?

dborovcanin: The only reason is that it seems to be more likely that `.` is already a part of the key. Basically, we can choose any character to be the delimiter, and that's the only character that _can't be in any key in the received JSON object_. It can be some of the usual delimiters: `,`, `/`, `\` `.`, `|`, or special chars such as `#`, `?`, `!`, `>`...

drasko (approved):
LGTM

mteodor (approved):

PR #1282, opened 2020-11-13
NOISSUE - Improve AuthN service docs

dborovcanin:
### What does this do?
This pull request improves AuthN service docs and refactors the service code.

There is no such issue.

Instead of not returning an error if the API key is expired, the special error for expired API keys is created.

Yes, tests are updated.

Yes, docs are updated.


drasko (dismissed):
LGTM

PR #1283, opened 2020-11-16
NOISSUE - Fix hardcoded env var values

darkodraskovic:


docker/ssl/Makefile
```diff
@@ -11,29 +11,29 @@ all: clean_certs ca server_crt
 # CA name and key is "ca".
 ca:
 	openssl req -newkey rsa:2048 -x509 -nodes -sha512 -days 1095 \
-				-keyout $(CRT_LOCATION)/ca.key -out $(CRT_LOCATION)/ca.crt -subj "/CN=localhost/O=Mainflux/OU=IoT/emailAddress=info@mainflux.com"
+				-keyout $(CRT_LOCATION)/ca.key -out $(CRT_LOCATION)/ca.crt -subj "/CN=$(CN)/O=$(O)/OU=$(OU)/emailAddress=$(EA)"
```
mteodor, line 14: Docs that describe the use of this script should reflect this change
so we should put something like
`make ca CN=localhost O=Mainflux OU=IoT`

manuio (dismissed):
LGTM

docker/ssl/Makefile
```diff
@@ -11,29 +11,29 @@ all: clean_certs ca server_crt
...
-	# Remove CSR.
+
+# Remove CSR.
 	rm $(CRT_LOCATION)/mainflux-server.csr
 
 thing_cert:
-	# Create mainflux server key and CSR.
+# Create mainflux server key and CSR.
 	openssl req -new -sha256 -newkey rsa:4096 -nodes -keyout $(CRT_LOCATION)/$(CRT_FILE_NAME).key \
 				-out $(CRT_LOCATION)/$(CRT_FILE_NAME).csr -subj "/CN=$(KEY)/O=$(O)/OU=$(OU)/emailAddress=$(EA)"
```
blokovi, line 31: Can you also change default value for `KEY`, now it is `default`. I would use some value in format of Mainflux Thing Key.

docker/ssl/Makefile
```diff
@@ -11,29 +11,29 @@ all: clean_certs ca server_crt
 # CA name and key is "ca".
 ca:
 	openssl req -newkey rsa:2048 -x509 -nodes -sha512 -days 1095 \
-				-keyout $(CRT_LOCATION)/ca.key -out $(CRT_LOCATION)/ca.crt -subj "/CN=localhost/O=Mainflux/OU=IoT/emailAddress=info@mainflux.com"
+				-keyout $(CRT_LOCATION)/ca.key -out $(CRT_LOCATION)/ca.crt -subj "/CN=$(CN)/O=$(O)/OU=$(OU)/emailAddress=$(EA)"
 
 # Server cert and key name is "mainflux-server".
 server_cert:
-	# Create mainflux server key and CSR.
+# Create mainflux server key and CSR.
```
drasko: Donot move these comments to the beginning of the line - they belong to indented code.

drasko (changes requested):

docker/ssl/Makefile
blokovi: Can you also change default value for KEY, now it is default. I would use some value in format of Mainflux Thing Key.

blokovi (changes requested):

docker/ssl/Makefile
darkodraskovic: Besides, it will be treated as a command by shell and it will be echoed as a comment.

drasko: `Blank lines and lines of just comments may appear among the recipe lines; they are ignored.` - they should be completely ignored.

Did you try and verify if the shell echoes these in the case they are indented and not in the case they are not?

dborovcanin (approved):

docker/ssl/Makefile
darkodraskovic: Did a PR.

drasko (approved):
LGTM

PR #1292, opened 2020-11-25
NOISSUE - Add different CNs for CA and certs

darkodraskovic:

### DIFF
--- a/docker/ssl/Makefile
+++ b/docker/ssl/Makefile
@@ -1,9 +1,11 @@
 CRT_LOCATION = certs
-THING_KEY = d7cc2964-a48b-4a6e-871a-08da28e7883d # example value
 O = Mainflux
-OU = mainflux
+OU_CA = mainflux_ca
+OU_CRT = mainflux_crt
 EA = info@mainflux.com
-CN = localhost
+CN_CA = Mainflux_Self_Signed_CA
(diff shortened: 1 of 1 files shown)


blokovi:
This Makefile is referenced in mainflux/docs (in two chapters, I think).
Can you please check later if there should be some changes also

docker/ssl/Makefile
```diff
@@ -1,23 +1,25 @@
 CRT_LOCATION = certs
-THING_KEY = d7cc2964-a48b-4a6e-871a-08da28e7883d # example value
 O = Mainflux
-OU = mainflux
+OU_CA = mainflux_ca
+OU_CRT = mainflux_crt
```
mteodor, line 4: maybe there is no need to add additional variables, since `make ca` and `make server_cert` are consecutive commands we can just update docs and specify that commands should be ran with different values

mteodor (changes requested):

docker/ssl/Makefile
darkodraskovic: Yes, but that adds a further step to get the thigns going. Plus, if you just try to run the script, you'll be screwed and you won't know why.

drasko (approved):
LGTM

PR #1300, opened 2020-12-09
NOISSUE - Fix dates not being init properly on save, change path construction, replace UUID with ULID for group ID

mteodor:
### What does this do?

* Fix `created_at` and `updated_at` not being saved properly
* Change the way that path is constructed, instead of the name of the group being used group id will be used
```
things=# select name, id, path from thing_groups;
  name   |             id             |                         path                          
---------+----------------------------+-------------------------------------------------------
 group13 | 01ESM3GW5VD6VSQ2YYDPCDQHXH | 01ESM3GW5VD6VSQ2YYDPCDQHXH
 group13 | 01ESM3H75DCYQPWHYZS8BAT733 | 01ESM3GW5VD6VSQ2YYDPCDQHXH.01ESM3H75DCYQPWHYZS8BAT733
```
* Name of the group can be changed
* Name of the group is not unique


drasko (changes requested):
This looks a bot complex... let's think if it is really needed.

Also - does name still stays unique? We should not demand that it is unique if we use IDs.

things/postgres/groups.go
```diff
@@ -574,21 +551,17 @@ func toGroup(dbu dbGroup) (groups.Group, error) {
 }
 
 type dbGroupRelation struct {
-	Group uuid.UUID `db:"group_id"`
+	Group string    `db:"group_id"`
 	Thing uuid.UUID `db:"thing_id"`
```
manuio: This should be `GroupID` and `ThingID`

pkg/ulid/ulid.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/mainflux/mainflux/pkg/errors"
+	"github.com/oklog/ulid/v2"
+
+	mathrand "math/rand"
+)
+
+// ErrGeneratingID indicates error in generating ULID
+var ErrGeneratingID = errors.New("generating id failed")
+
+var _ mainflux.UUIDProvider = (*ulidProvider)(nil)
```
manuio: Should we rename this  to `mainflux.IDProvider` ?

things/service.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/mainflux/mainflux/pkg/errors"
 
 	"github.com/mainflux/mainflux"
-	uuidProvider "github.com/mainflux/mainflux/pkg/uuid"
+	mfxulid "github.com/mainflux/mainflux/pkg/ulid"
```
drasko: Why `mfxulid` and not `ulidProvider` or `IDProvider`? Maybe we can also remove everything and use just `ulid`?

drasko (changes requested):

drasko (dismissed):
LGTM

manuio (dismissed):
LGTM

internal/groups/api/requests.go
```diff
@@ -50,10 +50,6 @@ func (req updateGroupReq) validate() error {
 		return groups.ErrParentInvariant
 	}
 
-	if req.Name != "" {
-		return groups.ErrNameInvariant
```
dborovcanin, line 54: We should remove this error if it's not used elsewhere.

dborovcanin (approved):

manuio (approved):
LGTM

drasko (approved):
LGTM

PR #1303, opened 2020-12-15
NOISSUE - remove owner id from user table and object

mteodor:
### What does this do?
Removing owner id from user object as no needed


manuio (approved):
LGTM

drasko (approved):
LGTM

PR #1305, opened 2020-12-15
NOISSUE - Add health check for MQTT broker

blokovi:

### What does this do?
If health-check path is set, mqtt adapter will wait that path is up and returning `200` status code. For VerneMQ health-check path is `/health` so it's default in docker composition where VerneMQ is used.

Related to  [issue](https://github.com/mainflux/mproxy/issues/27) in mProxy repo


cmd/mqtt/main.go
```diff
@@ -135,6 +138,24 @@ func main() {
...
+			if err != nil {
+				logger.Info(fmt.Sprintf("Broker not ready: %s ", err.Error()))
+				time.Sleep(1 * time.Second)
+				continue
+			}
+			if res.StatusCode != 200 {
+				logger.Info(fmt.Sprintf("Broker not ready, status code: %d ", res.StatusCode))
+				time.Sleep(1 * time.Second)
+				continue
+			}
```
manuio: Looks like you are logging twice

manuio (changes requested):

cmd/mqtt/main.go
blokovi: First check is if broker is up at all, because adapter will be up before broker (most of the time), and you will get `err` because http request will not be succesfull.
Second check is when broker is up, but is it ready to accept connections and joined cluster? Then you will get status `200` (when broker can accept connections) or `503` when broker is up but not yet ready to accept connections. That is how VerneMQ health check function: 
https://docs.vernemq.com/monitoring/health-check

manuio: Resolving this one if `err` is null even if status is different than `200`

cmd/mqtt/main.go
```diff
@@ -135,6 +138,24 @@ func main() {
...
 
+	if cfg.mqttTargetHealthCheck != "" {
+		for {
+			res, err := http.Head(cfg.mqttTargetHealthCheck)
+			if err != nil {
+				logger.Info(fmt.Sprintf("Broker not ready: %s ", err.Error()))
+				time.Sleep(1 * time.Second)
+				continue
+			}
+			if res.StatusCode != 200 {
```
drasko: Do not use `200`, there are [status code constants](https://golang.org/src/net/http/status.go)

.env
```diff
@@ -99,6 +99,7 @@ MF_HTTP_ADAPTER_PORT=8185
 MF_MQTT_ADAPTER_LOG_LEVEL=debug
 MF_MQTT_ADAPTER_MQTT_PORT=1883
 MF_MQTT_BROKER_PORT=1883
+MF_MQTT_BROKER_HEALTH_CHECK=http://0.0.0.0:8888/health
```
drasko: Needs to be documented in README

cmd/mqtt/main.go
```diff
@@ -135,6 +138,24 @@ func main() {
...
+				logger.Info(fmt.Sprintf("Broker not ready, status code: %d ", res.StatusCode))
+				time.Sleep(1 * time.Second)
```
drasko: Move constants to the top of the file.

drasko (changes requested):

cmd/mqtt/main.go
```diff
@@ -135,6 +138,24 @@ func main() {
 	}
 	defer nps.Close()
 
+	if cfg.mqttTargetHealthCheck != "" {
+		for {
+			res, err := http.Head(cfg.mqttTargetHealthCheck)
```
dborovcanin: Should you handle the response body close? Not closing it is usually considered to be a resource leak and prevents internal reuse of the underlying TCP connection.

blokovi: This is `HEAD` method http request, not `GET`. And there is no response body with HEAD. 
All I care is status code, so that's why decided to use it.

dborovcanin: Since `http.Head()` is just the shortcut for calling `DefaultClient.Do()`, there should be no difference between this and any other HTTP request. Since the [official docs (second paragraph)](https://golang.org/pkg/net/http/) say that the client is responsible for closing the response body, I would simply add after the error check:
```go
if err := res.Body.Close(); err != nil {
logger.Warn(....)
}
```
I doubt that there will be any problems with the current implementation, but I'd also like to respect the original design and the official docs.

mqtt/README.md
```diff
@@ -10,30 +10,31 @@ The service is configured using the environment variables presented in the
...
-| MF_AUTH_CACHE_PASS                | Auth cache password                                    | ""                    |
-| MF_AUTH_CACHE_DB                  | Auth cache database                                    | "0"                   |
+| Variable                                 | Description                                            | Default               |
+|------------------------------------------|--------------------------------------------------------|-----------------------|
+| MF_MQTT_ADAPTER_LOG_LEVEL                | mProxy Log level                                       | error                 |
+| MF_MQTT_ADAPTER_MQTT_PORT                | mProxy port                                            | 1883                  |
+| MF_MQTT_ADAPTER_MQTT_TARGET_HOST         | MQTT broker host                                       | 0.0.0.0               |
+| MF_MQTT_ADAPTER_MQTT_TARGET_PORT         | MQTT broker port                                       | 1883                  |
+| MF_MQTT_ADAPTER_MQTT_TARGET_HEALTH_CHECK | URL of broker health check                             | ""                    |
```
drasko, line 19: Is this default value - empty string? I thought it was `http://0.0.0.0:8888/health`. BTW, maybe you should consider `localhost` for consistency?

drasko (changes requested):

mqtt/README.md
blokovi: Default is empty.
In adapter there is `MF_MQTT_ADAPTER_MQTT_TARGET_HEALTH_CHECK` that is empty by default, because when you start mainflux natively (`make rundev`) you don't have broker started, so you don't know if it's going to be verne or mosquitto or some other broker.
In docker-compose `MF_MQTT_ADAPTER_MQTT_TARGET_HEALTH_CHECK` gets value of `MF_MQTT_BROKER_HEALTH_CHECK` . That other envvar has default value `http://0.0.0.0:8888/health`
So, only when you start docker composition (`make run`) you run mainflux with VerneMQ  and also check its health

dborovcanin: Since we have all the `MQTT_TARGET` env vars default values set to our default MQTT broker, I think that health check should not be different.

blokovi: That example is actually very similiar with my solution.
Target host by default is hardcoded to `0.0.0.0` in [main.go](https://github.com/mainflux/mainflux/blob/master/cmd/mqtt/main.go#L40) and then it's overwritten in [docker-compose](https://github.com/mainflux/mainflux/blob/master/docker/docker-compose.yml#L275) to `vernemq`
I just didn't hardocode values, I think it's better to use env vars, but same principle: in main it's empty string, in docker-compose (where you use `vernemq` as target host) health-check get value of VerneMQ health check endpoint.

blokovi: On the other hand, it's not a good example because whatever broker you choose it will have default mqtt endpoint on localhost:1883. But health-check endpoint on `:8888/health` is something that is specific just for VerneMQ, this is why I think it's better to be initialized only in places where we use VerneMQ, like docker-compose and helm charts

drasko (dismissed):
LGTM

drasko (approved):
LGTM, very nice

cmd/mqtt/main.go
```diff
@@ -122,6 +126,30 @@ func main() {
 		log.Fatalf(err.Error())
 	}
 
+	if cfg.mqttTargetHealthCheck != "" {
+		for {
```
mteodor: should this be in the go routine?

drasko: Makes sense, good candidate for goroutine.

drasko (changes requested):

cmd/mqtt/main.go
drasko: But wait - this is done only on init, and mProxy is blocked here until broker is up. However, I agree that we can continue to check health constantly in the background and restart mProxy if VerneMQ is down.

drasko: I took a quick look [this](https://medium.com/appsflyer/health-checks-like-a-pro-3fa362942883#id_token=eyJhbGciOiJSUzI1NiIsImtpZCI6ImUxOTdiZjJlODdiZDE5MDU1NzVmOWI2ZTVlYjQyNmVkYTVkNTc0ZTMiLCJ0eXAiOiJKV1QifQ.eyJpc3MiOiJodHRwczovL2FjY291bnRzLmdvb2dsZS5jb20iLCJuYmYiOjE2MDgxNDk1NjQsImF1ZCI6IjIxNjI5NjAzNTgzNC1rMWs2cWUwNjBzMnRwMmEyamFtNGxqZGNtczAwc3R0Zy5hcHBzLmdvb2dsZXVzZXJjb250ZW50LmNvbSIsInN1YiI6IjEwNzQ4NTcwMTQzNTA2MDQ4NDk5NCIsImVtYWlsIjoiZHJhc2tvLmRyYXNrb3ZpY0BnbWFpbC5jb20iLCJlbWFpbF92ZXJpZmllZCI6dHJ1ZSwiYXpwIjoiMjE2Mjk2MDM1ODM0LWsxazZxZTA2MHMydHAyYTJqYW00bGpkY21zMDBzdHRnLmFwcHMuZ29vZ2xldXNlcmNvbnRlbnQuY29tIiwibmFtZSI6IkRyYXNrbyBEUkFTS09WSUMiLCJwaWN0dXJlIjoiaHR0cHM6Ly9saDMuZ29vZ2xldXNlcmNvbnRlbnQuY29tL2EtL0FPaDE0R2dYWXBVZ2JLWDRyaXhKQXBwWTBNZGVwSTRROVZoOGJlS1JPdDMxeGc9czk2LWMiLCJnaXZlbl9uYW1lIjoiRHJhc2tvIiwiZmFtaWx5X25hbWUiOiJEUkFTS09WSUMiLCJpYXQiOjE2MDgxNDk4NjQsImV4cCI6MTYwODE1MzQ2NCwianRpIjoiMGJkZGVhZGQyMzVkZmY0MWYyNTk1MzgwOTZhY2RiMTIxODY3MjRhNyJ9.kcO5gj10CVXWR9WKvfAGAVVv-koh_dtEirN08t0t-zeLd6PXn9BGXUaPm6tNPxOBR6QMU_axfi06kTY-8q4utRUE1t6-wfM_TUQge5b6dGv-zC8NfJ2HHXLEVrYI7XQ8hYjvvsk-FQxalogrbvDqFVsPiVVeX8B9wPcZUIz8fadUfFi9H0hFnolul_sOZKG-SqfnMsVeBA3bJ6s7cUZVzdYKW7e1IyYXMGKT9yo7ehSZNIhDMrCoS8oEN9Z-Zgez9qJZJsp0zt3o5OKTGGHDUhVqSzjwZK60tdy_HcelpEnYcr9vMTkTXcESqbiq_oMkoyWbpbM7iWCDPU587wOqpQ) approach seems interesting.

The mentioned [go-sundheit](https://github.com/AppsFlyer/go-sundheit) library allows you to [schedule](https://github.com/AppsFlyer/go-sundheit/blob/c78d43b4328a771cf4c94d8c66e5d16e6c46ad85/health.go#L142) an async check in a [go routine](https://github.com/AppsFlyer/go-sundheit/blob/c78d43b4328a771cf4c94d8c66e5d16e6c46ad85/health.go#L191).

dborovcanin (dismissed):

cmd/mqtt/main.go
blokovi: moved

cmd/mqtt/main.go
blokovi: I actually used http.Get so I can log body response. And closing body of course.

mqtt/README.md
blokovi: Actualy, this was error, in docker composition address of VerneMQ container is not `0.0.0.0` but `vernemq`. Fixed.

cmd/mqtt/main.go
blokovi: Restarting mProxy when VerneMQ is down will be done by kubernetes. There is [PR](https://github.com/mainflux/devops/pull/74) in devops repo that defines liveness probe for mProxy container with checking health checker on VerneMQ. If it's not healthy, both containers (VerneMQ and mProxy) will be restarted by kubernetes. 
It would be redundant and probably conflicting to have same same job done here in code and by orchestration tool.

mteodor (dismissed):

mteodor (dismissed):

docker/vernemq/Dockerfile
```diff
@@ -47,7 +47,7 @@ EXPOSE 1883 8883 8080 44053 4369 8888 \
 
 VOLUME ["/vernemq/log", "/vernemq/data", "/vernemq/etc"]
 
-HEALTHCHECK CMD vernemq ping | grep -q pong
+HEALTHCHECK CMD wget --no-verbose --tries=1 --spider http://localhost:8888/health || exit 1
```
drasko: Can we use `curl` here instead of `wget`?

drasko (changes requested):

docker/vernemq/Dockerfile
blokovi: Alternative with `curl` could be:
```HEALTHCHECK CMD  [[ "$(curl -o /dev/null -s -w "%{http_code}\n" http://localhost:8888/health)" == "200" ]]```

The command’s exit status indicates the health status of the container. The possible values are:
0: success - the container is healthy and ready for use
1: unhealthy - the container is not working correctly

That's why this command compare  if return http code with `curl` is `"200"`

Comment (drasko) [reply docker/vernemq/Dockerfile]: I prefer `curl` command. It looks more complicated on a first look, but it explains better what it does.

But I would prefer even more following @jpetazzo's [instructions](https://github.com/jpetazzo/container.training/blob/master/slides/swarm/healthchecks.md#health-checks-and-auto-rollbacks) with `curl -f`.

Comment (blokovi) [reply docker/vernemq/Dockerfile]: changed

File: cmd/mqtt/main.go [hunk 9]
```diff
@@ -36,14 +36,17 @@ const (
...
-	defMQTTForwarderTimeout = "30s" // 30 seconds
-	envMQTTPort             = "MF_MQTT_ADAPTER_MQTT_PORT"
-	envMQTTTargetHost       = "MF_MQTT_ADAPTER_MQTT_TARGET_HOST"
-	envMQTTTargetPort       = "MF_MQTT_ADAPTER_MQTT_TARGET_PORT"
-	envMQTTForwarderTimeout = "MF_MQTT_ADAPTER_FORWARDER_TIMEOUT"
+	defMQTTPort              = "1883"
+	defMQTTTargetHost        = "0.0.0.0"
+	defMQTTTargetPort        = "1883"
+	defMQTTForwarderTimeout  = "30s" // 30 seconds
+	healthCheckSleep         = 1 * time.Second
```
dborovcanin: It's OK not to have this configurable, but I'd extract it to a separate constant (`const mqttHealthCheckSleep = time.Second)`.

cmd/mqtt/main.go
```diff
@@ -122,6 +126,30 @@ func main() {
...
+				if err := res.Body.Close(); err != nil {
+					logger.Warn(fmt.Sprintf("Error closing response body: %s", err.Error()))
+				}
+				if res.StatusCode == http.StatusOK {
+					logger.Info("MQTT Broker health check successful")
+					break
+				}
+				logger.Info(fmt.Sprintf("Broker not ready, status code: %d, body: %s", res.StatusCode, body))
+			}
+			time.Sleep(healthCheckSleep)
```
dborovcanin: I would add a backoff to prevent this loop from running indefinitely:
```go
if i == maxHealthCheckRetries {
    logger.Error("MQTT healthcheck limit exceeded, exiting.")
    os.Exit(1)
}
time.Sleep(i * mqttHealthcheckSleep)
```

This can come in handy in the case of running service natively.

dborovcanin (changes requested):

cmd/mqtt/main.go
```diff
@@ -122,6 +130,34 @@ func main() {
 		log.Fatalf(err.Error())
 	}
 
+	if cfg.mqttTargetHealthCheck != "" {
+		for i := 1; i <= maxHealthCheckRetries; i++ {
+			res, err := http.Get(cfg.mqttTargetHealthCheck)
+			if err != nil {
+				logger.Info(fmt.Sprintf("Broker not ready: %s ", err.Error()))
```
drasko: Shouldn't we do `continue` here?

cmd/mqtt/main.go
drasko: I do not like this whole construct, it looks clunky.

Can you please use https://github.com/cenkalti/backoff and solve this with [`Retry()`](https://github.com/cenkalti/backoff/blob/c2975ffa541a1caeca5f76c396cb8c3e7b3bb5f8/example_test.go#L14) function.

Then if `Retry()` succeeds you can try reading the body, and if this fails just restart (that means broker is ready, but there were some more serious error. And this usually should not happen).

drasko (changes requested):

cmd/mqtt/main.go
dborovcanin: Simply extracting to a separate function and calling it here will help.

cmd/mqtt/main.go
blokovi: I think not, without `continue` rest of the loop body is skipped anyway, except `sleep` on the end. 
With `continue` next iteration would be right away.

drasko: Yes, this is exactly what I meant - that the rest of the operations in this loop body should be skipped. Because if you can not do HTTP GET, do you really need to check the result? And do you really need to read HTTP body (remember that GET failed)? And then do you need to close the `Body` (remember, GET failed, and you did not read anything)?

cmd/mqtt/main.go
```diff
@@ -122,6 +130,34 @@ func main() {
...
+				if err := res.Body.Close(); err != nil {
+					logger.Warn(fmt.Sprintf("Error closing response body: %s", err.Error()))
+				}
+				if res.StatusCode == http.StatusOK {
+					logger.Info("MQTT Broker health check successful")
+					break
+				}
+				logger.Info(fmt.Sprintf("Broker not ready, status code: %d, body: %s", res.StatusCode, body))
+			}
+			if i == maxHealthCheckRetries {
```
drasko: Should be probably moved to the top of the loop.

cmd/mqtt/main.go
drasko: I would like that we try with https://github.com/cenkalti/backoff in order to prepare terrain for the eventual changing of other services.

drasko (changes requested):

cmd/mqtt/main.go
blokovi: > Yes, this is exactly what I meant - that the rest of the operations in this loop body should be skipped. Because if you can not do HTTP GET, do you really need to check the result? And do you really need to read HTTP body (remember that GET failed)? And then do you need to close the Body (remember, GET failed, and you did not read anything)?

Rest of code is already skipped - except sleep.
All operation you mention here is in `if res != nil` where `res` is result of `http.Get`. If that fails (and that is normal because adapter will be up before broker) then sleep and retry again.

blokovi: > Should be probably moved to the top of the loop.

Moving it to the top would do unnecessary sleep in situation when success is in firts iteration.

drasko: My question was - why are you even checking `res` if you already had an `err` in `http.Get(cfg.mqttTargetHealthCheck)`? It's  like hoping that `res` will somehow be valid, although there was an error in the GET function itself.

cmd/mqtt/main.go
```diff
@@ -122,6 +130,35 @@ func main() {
 		log.Fatalf(err.Error())
 	}
 
+	if cfg.mqttTargetHealthCheck != "" {
+		for i := 1; i <= maxHealthCheckRetries; i++ {
```
dborovcanin: Since you are already checking the condition below, you can ignore it here `for i:=1; ;i++ {...}`. Also, please consider increasing the sleep period in each iteration by multiplying it by `i`. In that case, `maxHealthCheckRetries` should probably be less than 10 (5?) to prevent long waits. This block can be extracted to a separate function so that we don't pollute the `main`.

cmd/mqtt/main.go
```diff
@@ -122,6 +130,35 @@ func main() {
...
+			res, err := http.Get(cfg.mqttTargetHealthCheck)
```
drasko: New `res` object is created on every iteration, so use `defer` to close it's body. Defers will be intelligent enough to know to which object they apply at the end of the function.

drasko (changes requested):

cmd/mqtt/main.go
```diff
@@ -122,6 +127,37 @@ func main() {
 		log.Fatalf(err.Error())
 	}
 
+	if cfg.mqttTargetHealthCheck != "" {
+
```
dborovcanin: Please remove the empty line.

cmd/mqtt/main.go
```diff
@@ -122,6 +127,37 @@ func main() {
...
+			return nil
+		}
+
+		notify := func(e error, next time.Duration) {
+			logger.Info(fmt.Sprintf("Broker not ready: %s, next try in %s", e.Error(), next))
+		}
+
+		err := backoff.RetryNotify(operation, backoff.NewExponentialBackOff(), notify)
+		if err != nil {
+			// Handle error.
```
dborovcanin: No need for this comment.

cmd/mqtt/main.go
```diff
@@ -122,6 +127,37 @@ func main() {
 		log.Fatalf(err.Error())
 	}
 
+	if cfg.mqttTargetHealthCheck != "" {
+
+		// An operation that may fail.
+		operation := func() error {
+			res, err := http.Get(cfg.mqttTargetHealthCheck)
```
dborovcanin: Consider extracting this function outside of main and just use it in `backoff.RetryNotify` call.

cmd/mqtt/main.go
blokovi: switched to cenkalti/backoff

cmd/mqtt/main.go
blokovi: removed

blokovi:
bunch of files are removed from `vendor` directory after
`go mod vendor`

dborovcanin (dismissed):

dborovcanin (approved):

docker/vernemq/Dockerfile
```diff
@@ -47,7 +47,7 @@ EXPOSE 1883 8883 8080 44053 4369 8888 \
 
 VOLUME ["/vernemq/log", "/vernemq/data", "/vernemq/etc"]
 
-HEALTHCHECK CMD vernemq ping | grep -q pong
+HEALTHCHECK CMD curl -s -f http://localhost:8888/health || false
```
drasko, line 50: Just a comment here: if we changed VerneMQ Dockerfile, we'll need to build and upload new image. I am not sure of the current CI covers the VerneMQ - if not, please add it.

drasko (commented):
LGTM

drasko (approved):

PR #1307, opened 2020-12-16
NOISSUE - Remove authz from docker comp

mteodor:
### What does this do?
Remove `authz` service from docker composition as `authz` functionality will be merged with `authn` service

### DIFF
--- a/users/postgres/init.go
+++ b/users/postgres/init.go
@@ -96,13 +96,6 @@ func migrateDB(db *sqlx.DB) error {
 				)`,
 				},
 			},
-			{
-				Id: "users_5",
-				Up: []string{
-					`ALTER TABLE IF EXISTS users DROP CONSTRAINT IF EXISTS users_owner_id_fkey`,
-					`ALTER TABLE IF EXISTS users DROP COLUMN IF EXISTS owner_id`,
-				},
-			},
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


users/postgres/init.go
```diff
@@ -96,6 +96,13 @@ func migrateDB(db *sqlx.DB) error {
 				)`,
 				},
 			},
+			{
+				Id: "users_5",
+				Up: []string{
+					`ALTER TABLE IF EXISTS users DROP CONSTRAINT IF EXISTS users_owner_id_fkey`,
+					`ALTER TABLE IF EXISTS users DROP COLUMN IF EXISTS owner_id`,
+				},
+			},
```
manuio: Dead migration?

manuio (changes requested):

manuio (approved):
LGTM

nmarcetic (changes requested):
@mteodor  You still have 3 commits. Is there is a special reason why you are removing docker-compose stuff before `authz` is merged in `authn`? Does this will break functionality?

mteodor:
> @mteodor You still have 3 commits. Is there is a special reason why you are removing docker-compose stuff before `authz` is merged in `authn`? Does this will break functionality?

authz is not used, and there is a problem with starting the service in docker composition so that is why I removed it
authz was not complete since it was the idea to break it in few PR's
in a meanwhile we decided to have authn and authz in one service

nmarcetic:
@mteodor I see. Thanks for the clarification. 
Maybe is a good idea to remove it from all builds like Makefile (build, push etc...) and leave just source code for now.

nmarcetic (approved):

PR #1309, opened 2020-12-17
NOISSUE - Fix MQTT Forwarder client id

dborovcanin:
Don't set client ID for MQTT client in MQTT forwarder. It results in
error in case of multiple instances of the MQTT adapter because they all
share MQTT client ID, which makes broker disconnect all the other
adapters.

### What does this do?
This pull request fixes shared MQTT client id between MQTT adapter instances.

There is no such issue.

MQTT adapter wasn't scaling properly because the internal MQTT client used for forwarding NATS messages used to have a fixed client ID, which was shared between instances. This results in only one MQTT forwarder being subscribed at the time.

No.

No.


blokovi (dismissed):

blokovi (dismissed):

nmarcetic:
@dusanb94 It's not really important to know ID while you debugging messaging on the MQTT side (so you know it's a forwarder, not an actual client/thing)?  If so, maybe it's better to stick rand Int on `mqtt-publisher-%rand`?

manuio (dismissed):
LGTM

dborovcanin:
> @dusanb94 It's not really important to know ID while you debugging messaging on the MQTT side (so you know it's a forwarder, not an actual client/thing)? If so, maybe it's better to stick rand Int on `mqtt-publisher-%rand`?

@nmarcetic Good catch. I'd rather let the broker generate an id, but  I added a username that can be used to differentiate MQTT publishers and clients.

nmarcetic:
@dusanb94 Make sense.

nmarcetic (approved):

blokovi (approved):

drasko (approved):
LGTM

PR #1310, opened 2020-12-17
NOISSUE - Correct readers openapi.yml

manuio:


drasko (approved):
LGTM

readers/openapi.yml
```diff
@@ -123,6 +125,21 @@ components:
...
+      schema:
+        type: string
+        format: uuid
+      required: false
+    Name:
+      name: name
+      description: SenML message name.
+      in: query
+      schema:
+        type: string
```
darkodraskovic, line 141: "OpenAPI 3 introduces support for objects as query parameters." Please see here https://www.baeldung.com/openapi-json-query-parameters, specifically here https://www.baeldung.com/openapi-json-query-parameters#OAP3

darkodraskovic (changes requested):

readers/openapi.yml
manuio: @darkodraskovic Looks like this explains more how to generate an API that accept JSON objects. AFAIK we don't support this in our services. As you previously did I would follow: https://swagger.io/docs/specification/describing-parameters/

darkodraskovic: This also allows for automatic Postman collections etc. I'm not sure, but it seams like a standard way to specify json string in query. Just add
```
        schema:
          type: object
```
And see if the editor complains.

darkodraskovic (approved):

Issue #1061, opened 2020-03-05
Improve readers message retrival querying


mteodor:
When getting messages from reader it would be nice to have possibility to have more flexible query of the results. For example `give me all messages that name contains "appartment"`
we can have messages where names are `appartment1,appartment2,appartment3` 
and it would be good to retrieve all these messages in a single http request like this
`curl -S -H "Content-Type: application/json" -H "Authorization:8bf59204-aa6a-4efd-a646-e3ca0b049c3e" -X GET https://clover.mainflux.io/pgreader/channels/cb4b3cbf-ee82-44d9-9723-9dda7b86469d/messages\?offset\=0\&publisher\=3ad0f341-42d3-49e8-8dbb-6b08dda4de2c\&limit\=100\&name\=appartment`
so name would be checked against data in databse with something similare to `select * from messages where name like 'appartment%'`

manuio:
It would be also very useful to be able to fetch messages for a certain period in order to implement date range pickers from applications.

manuio:
@mteodor btw, isn't name filter already available here: https://github.com/mainflux/mainflux/blob/master/readers/influxdb/messages.go#L130 ?

manuio:
@mteodor the problem is that these filters are generated in these functions:
 https://github.com/mainflux/mainflux/blob/master/readers/mongodb/messages.go#L113
 https://github.com/mainflux/mainflux/blob/master/readers/cassandra/messages.go#L99
 https://github.com/mainflux/mainflux/blob/master/readers/postgres/messages.go#L83
 
 We are generating a `map[string]string` using this fields: https://github.com/mainflux/mainflux/blob/master/readers/api/transport.go#L33 and it's hard to understand how this is propagated in the code. Also,  if we support`"value", "v", "vs", "vb", "vd"` this should be documented. Maybe it would be better to be more explicit  and use
 a structure. Something like  we did for things service: 
 ```
     pageMetadata: readers.PageMetadata{
	    Offset:   o,
	    Limit:    l,
	    Publisher: p,
	    Value: v,
	    Name: n,
	    From: f,
	    To: t,
    }
```

PR #1312, opened 2020-12-22
MF-1061 - Implement InfluxDB filters value, v, vb, vs, vd, from, to

manuio:

Related to #1061

### DIFF
--- a/readers/api/endpoint_test.go
+++ b/readers/api/endpoint_test.go
@@ -27,11 +27,11 @@ const (
 )
 
 var (
-	val       float64 = 5
-	stringVal         = "value"
-	boolVal           = true
-	dataVal           = "dataValue"
-	sum       float64 = 42
+	v   float64 = 5
+	vs          = "value"
+	vb          = true
+	vd          = "dataValue"
+	sum float64 = 42
 )
 
 func newService() readers.MessageRepository {
@@ -47,13 +47,13 @@ func newService() readers.MessageRepository {
 
 		switch count {
 		case 0:
-			msg.Value = &val
+			msg.Value = &v
 		case 1:
-			msg.BoolValue = &boolVal
+			msg.BoolValue = &vb
 		case 2:
-			msg.StringValue = &stringVal
+			msg.StringValue = &vs
 		case 3:
-			msg.DataValue = &dataVal
+			msg.DataValue = &vd
 		case 4:
 			msg.Sum = &sum
 		}
@@ -167,22 +167,22 @@ func TestReadAll(t *testing.T) {
 			status: http.StatusOK,
 		},
 		"read page with value": {
-			url:    fmt.Sprintf("%s/channels/%s/messages?v=%f", ts.URL, chanID, val),
+			url:    fmt.Sprintf("%s/channels/%s/messages?v=%f", ts.URL, chanID, v),
 			token:  token,
 			status: http.StatusOK,
 		},
 		"read page with boolean value": {
-			url:    fmt.Sprintf("%s/channels/%s/messages?vb=%t", ts.URL, chanID, boolVal),
+			url:    fmt.Sprintf("%s/channels/%s/messages?vb=%t", ts.URL, chanID, vb),
 			token:  token,
 			status: http.StatusOK,
 		},
 		"read page with string value": {
-			url:    fmt.Sprintf("%s/channels/%s/messages?vs=%s", ts.URL, chanID, dataVal),
+			url:    fmt.Sprintf("%s/channels/%s/messages?vs=%s", ts.URL, chanID, vd),
 			token:  token,
 			status: http.StatusOK,
 		},
 		"read page with data value": {
-			url:    fmt.Sprintf("%s/channels/%s/messages?vd=%s", ts.URL, chanID, dataVal),
+			url:    fmt.Sprintf("%s/channels/%s/messages?vd=%s", ts.URL, chanID, vd),
 			token:  token,
 			status: http.StatusOK,
 		},
--- a/readers/api/endpoint_test.go
+++ b/readers/api/endpoint_test.go
@@ -27,11 +27,11 @@ const (
 )
 
 var (
-	v       float64 = 5
-	stringV         = "value"
-	boolV           = true
-	dataV           = "base64"
-	sum     float64 = 42
+	val       float64 = 5
+	stringVal         = "value"
+	boolVal           = true
+	dataVal           = "dataValue"
+	sum       float64 = 42
 )
 
 func newService() readers.MessageRepository {
@@ -47,13 +47,13 @@ func newService() readers.MessageRepository {
 
 		switch count {
 		case 0:
-			msg.Value = &v
+			msg.Value = &val
 		case 1:
-			msg.BoolValue = &boolV
+			msg.BoolValue = &boolVal
 		case 2:
-			msg.StringValue = &stringV
+			msg.StringValue = &stringVal
 		case 3:
-			msg.DataValue = &dataV
+			msg.DataValue = &dataVal
 		case 4:
 			msg.Sum = &sum
 		}
@@ -163,6 +163,41 @@ func TestReadAll(t *testing.T) {
 		},
 		"read page with default limit": {
 			url:    fmt.Sprintf("%s/channels/%s/messages?offset=0", ts.URL, chanID),
+			token:  token,
+			status: http.StatusOK,
+		},
+		"read page with from": {
+			url:    fmt.Sprintf("%s/channels/%s/messages?from=1577836800", ts.URL, chanID),
+			token:  token,
+			status: http.StatusOK,
+		},
+		"read page with to": {
+			url:    fmt.Sprintf("%s/channels/%s/messages?to=1608630754", ts.URL, chanID),
+			token:  token,
+			status: http.StatusOK,
+		},
+		"read page with value": {
+			url:    fmt.Sprintf("%s/channels/%s/messages?value=%f", ts.URL, chanID, val),
+			token:  token,
+			status: http.StatusOK,
+		},
+		"read page with v": {
+			url:    fmt.Sprintf("%s/channels/%s/messages?v=%f", ts.URL, chanID, val),
+			token:  token,
+			status: http.StatusOK,
+		},
+		"read page with vb": {
+			url:    fmt.Sprintf("%s/channels/%s/messages?vb=%t", ts.URL, chanID, boolVal),
+			token:  token,
+			status: http.StatusOK,
+		},
+		"read page with vs": {
+			url:    fmt.Sprintf("%s/channels/%s/messages?vs=%s", ts.URL, chanID, dataVal),
+			token:  token,
+			status: http.StatusOK,
+		},
+		"read page with vd": {
+			url:    fmt.Sprintf("%s/channels/%s/messages?vd=%s", ts.URL, chanID, dataVal),
 			token:  token,
 			status: http.StatusOK,
 		},
--- a/readers/api/endpoint_test.go
+++ b/readers/api/endpoint_test.go
@@ -166,16 +166,6 @@ func TestReadAll(t *testing.T) {
 			token:  token,
 			status: http.StatusOK,
 		},
-		"read page with from": {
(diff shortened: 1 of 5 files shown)


readers/openapi.yml
```diff
@@ -140,6 +147,55 @@ components:
...
+    Vd:
+      name: vd
+      description: SenML message data value.
+      in: query
+      schema:
+        type: string
+      required: false
+    From:
+      name: from
+      description: SenML message time in nanoseconds (int64).
```
drasko: Is this Unix time?

drasko (changes requested):

readers/openapi.yml
manuio: This is `time.Now().UnixNano()`. We should maybe use float64 here and convert it before the DB query.
Internally, InfluxDB use int64 in nanoseconds. But our readers return float64.

readers/influxdb/messages_test.go
```diff
@@ -58,31 +60,37 @@ func TestReadAll(t *testing.T) {
...
+	boolMsgs := []senml.Message{}
+	stringMsgs := []senml.Message{}
+	dataMsgs := []senml.Message{}
+
 	now := time.Now().UnixNano()
 	for i := 0; i < msgsNum; i++ {
 		// Mix possible values as well as value sum.
-		count := i % valueFields
 		msg := m
+		msg.Time = float64(now)/float64(1e9) - float64(i)
```
darkodraskovic: `msg.Time = float64(now)/1e9 - i` should be enough for the same reasons specified above.

readers/influxdb/messages_test.go
```diff
@@ -149,10 +157,85 @@ func TestReadAll(t *testing.T) {
...
+			query:  map[string]string{"vd": dataVal},
+			page: readers.MessagesPage{
+				Total:    msgsValNum,
 				Offset:   0,
 				Limit:    10,
-				Messages: subtopicMsgs[0:10],
+				Messages: dataMsgs[0:10],
 			},
 		},
 	}
```
darkodraskovic, line 230: You can replace all those `0` and `10` constants with constants or vars.

readers/openapi.yml
```diff
@@ -140,6 +147,55 @@ components:
       schema:
         type: string
       required: false
+    Value:
+      name: value
+      description: SenML message value.
+      in: query
+      schema:
+        type: string
+      required: false
```
darkodraskovic: This seems redundant. You have `V` param.

readers/openapi.yml
```diff
@@ -140,6 +147,55 @@ components:
...
+    Vd:
+      name: vd
+      description: SenML message data value.
+      in: query
+      schema:
+        type: string
+      required: false
+    From:
+      name: from
+      description: SenML message time in nanoseconds as float64 (integer part represent seconds).
```
darkodraskovic: "represents"

readers/openapi.yml
```diff
@@ -140,6 +147,55 @@ components:
...
+      in: query
+      schema:
+        type: integer
```
darkodraskovic: Shouldn't this be `type: float`?

readers/openapi.yml
```diff
@@ -140,6 +147,55 @@ components:
...
+      required: false
+    To:
+      name: to
+      description: SenML message time in nanoseconds as float64 (integer part represent seconds).
+      in: query
+      schema:
+        type: integer
```
darkodraskovic: The same here, `type: float`.

darkodraskovic (changes requested):

readers/api/endpoint_test.go
```diff
@@ -166,6 +166,41 @@ func TestReadAll(t *testing.T) {
...
+			url:    fmt.Sprintf("%s/channels/%s/messages?vb=%t", ts.URL, chanID, boolVal),
+			token:  token,
+			status: http.StatusOK,
+		},
+		"read page with vs": {
+			url:    fmt.Sprintf("%s/channels/%s/messages?vs=%s", ts.URL, chanID, dataVal),
+			token:  token,
+			status: http.StatusOK,
+		},
+		"read page with vd": {
```
mteodor: do we need `>` and `<`

readers/api/endpoint_test.go
```diff
@@ -166,6 +166,41 @@ func TestReadAll(t *testing.T) {
 			token:  token,
 			status: http.StatusOK,
 		},
+		"read page with from": {
```
mteodor: maybe `from timestamp`, also can we fetch interval?

readers/influxdb/messages_test.go
manuio: Without to  cast this returns `mismatched types float64 and int`.  Let's keep this one

readers/openapi.yml
manuio: It is, but it means that we support `v` or `value`. I did it following this filters: https://github.com/mainflux/mainflux/blob/master/readers/api/transport.go#L33

readers/openapi.yml
manuio: Right, good catch!

readers/api/endpoint_test.go
manuio: There are 2 test `from` and `to`

readers/influxdb/messages_test.go
manuio: This wouldn't be consistent with all other tests and I find it simpler to understand when you see the value of limit, offset for each test.

darkodraskovic: In things tests we use `n` var for non zero limit values https://github.com/mainflux/mainflux/blob/master/things/postgres/things_test.go Anyway, this is just a recommendation.

readers/influxdb/messages_test.go
manuio: I removed the unnecessary cast of `1e9` the one from `i` is needed because it's an `int`.  In your example is probably done automatically but if you use the index of the for loop, cast is needed.

darkodraskovic: This `i` will be auto-converted to float64 because of the precision of the left operand `float64(now)/1e9`. So the whole expression will be float64 type. If you need an expression to be a float64 type, then this should work and it seems to me that you need it.

readers/influxdb/messages_test.go
manuio: ok I can add this `n`

readers/influxdb/messages_test.go
darkodraskovic: However, you're right about `i` because you implictely declared it as an int and i used a literal value.

darkodraskovic (dismissed):
LGTM

readers/influxdb/messages_test.go
manuio: I fact it's not correct. In Go there is no implicit casting with variables. Here probably `i`is forcing to cast all values.

darkodraskovic: In fact, there is no casting in go. Only converting. And there is implicit converting. See here https://golang.org/ref/spec#Conversions

Also from the same page,

```An untyped constant has a default type which is the type to which the constant is implicitly converted in contexts where a typed value is required, for instance, in a short variable declaration such as i := 0 where there is no explicit type. The default type of an untyped constant is bool, rune, int, float64, complex128 or string respectively, depending on whether it is a boolean, rune, integer, floating-point, complex, or string constant.```

Comment (darkodraskovic) [reply readers/influxdb/messages_test.go]: As far as I understand Go can implicitly convert only literal values that happen to be in expressions, so that's why this works:

```
nowF := float64(now)/1e9 - 100
```

while this does not work:

```
i := 100
nowF := float64(now)/1e9 - i
```

because Go does not implicitly converts `i` to float64, but converts litteral value of 100 to float64 (please see here https://yourbasic.org/golang/conversions/#implicit-conversions also).

Comment (manuio) [reply readers/influxdb/messages_test.go]: Yes,that's it.I reverted too previous version. I forgot but I already had this issue. I changed both at the same time... At least the difference is clear now.

Comment (darkodraskovic) [reply readers/influxdb/messages_test.go]: IMO, you should still use `iVal := int64(fVal * 1e9)`, because you fVal is float64 and 1e9 is implicitly converted to float64 per explanations above. 

The same goes for `float64(now)/1e9`. Only the `i` cannot be converted implicitly, because it's type is already determined as int. Implicit conversation works only on literal value.

File: readers/openapi.yml (hunk 4 above)
Comment (dborovcanin) [reply readers/openapi.yml]: Can you use `$ref` to reference Value instead of copy-pasting definition?

Comment (manuio) [reply readers/openapi.yml]: The name is different. It's not really copy paste. Should I maybe remove `value` parameter and keep `v`, `vs`,  `vd` and `vb`?

Comment (dborovcanin) [reply readers/openapi.yml]: I agree. The input JSON is using `v`, so it makes sense to use it in the filter.

**Review by dborovcanin — COMMENTED** [review]
The other readers need to support this filtering, as well.

**Review by mteodor — DISMISSED** [review]

File: readers/influxdb/messages_test.go (hunk 2 above)
Comment (darkodraskovic) [inline readers/influxdb/messages_test.go]: This one should be `msg.Time = float64(now)/1e9 - float64(i)`. Otherwise LGTM.

**Review by darkodraskovic — DISMISSED** [review]

File: readers/api/transport.go [hunk 10]
```diff
@@ -30,7 +30,7 @@ var (
 	errInvalidRequest     = errors.New("received invalid request")
 	errUnauthorizedAccess = errors.New("missing or invalid credentials provided")
 	auth                  mainflux.ThingsServiceClient
-	queryFields           = []string{"subtopic", "publisher", "protocol", "name", "value", "v", "vs", "vb", "vd"}
+	queryFields           = []string{"subtopic", "publisher", "protocol", "name", "v", "vs", "vb", "vd", "from", "to"}
```
dborovcanin, line 33: Consider extracting these fields to a separate struct and sending them as a JSON.

manuio: As explained here https://github.com/mainflux/mainflux/issues/1061 I'm considering to do it but I would to it in another PR.

mteodor (dismissed):

readers/api/endpoint_test.go
```diff
@@ -27,11 +27,11 @@ const (
 )
 
 var (
-	v       float64 = 5
-	stringV         = "value"
-	boolV           = true
-	dataV           = "base64"
-	sum     float64 = 42
+	val       float64 = 5
```
drasko: Why not `v`, `sv`, and so on?

readers/api/transport.go
drasko: Is it a big work that demands an additional PR? If not - it is probably better to do it in this one.

drasko (changes requested):

readers/openapi.yml
drasko: Then we should mention this in the description:
```
description: SenML message time in nanoseconds in Unix time format (int64).
```

readers/api/transport.go
manuio: @drasko It have to be done for all readers. Similar to what we recently did for Things and Channels `PageMetadata`. IMO it's better to separate this in another PR.

readers/influxdb/messages_test.go
manuio: You are right. I got confused with docker tests. There is a strange behavior if you run tests several times. Sometimes it fails and if you repeat then it pass... it could be some docker tests cache? 
In any case I removed the cast, it should be ok now.

drasko (approved):
LGTM

readers/api/transport.go
drasko: OK then, I approved. Please do not forget to do this in a subsequent PR. Joyeux Noël.

dborovcanin (approved):

drasko:
@manuio it would probably be interesting to add a few more filters around `value`:
- Less than (`<some_value>`)
- Higher than (`<some_value>`)
- In range (less than `<value_max>` but higher than `<value_min>`)
- Not in range (less than `<value_min>` but higher than `<value_max>`)

I'll open a dedicated issue as a reminder and then we can discuss if this needs to be implemented.

PR #1313, opened 2020-12-22
NOISSUE - Merge authz and authn into new service auth

mteodor:
### What does this do?
New auth service will perform functions of authn service 
* issue token
* identify
Additionally it will support groups features
* assign member to a group ( member can be user or thing )
* get group members


auth/README.md
```diff
@@ -0,0 +1,118 @@
...
+- revoke (API keys only)
+
+## Configuration
+
+The service is configured using the environment variables presented in the
+following table. Note that any unset variables will be replaced with their
+default values.
+
+| Variable                  | Description                                                              | Default       |
+|---------------------------|--------------------------------------------------------------------------|---------------|
```
drasko, line 42: Align

auth/README.md
```diff
@@ -0,0 +1,118 @@
...
+| MF_AUTH_DB_SSL_CERT      | Path to the PEM encoded certificate file                                |                |
+| MF_AUTH_DB_SSL_KEY       | Path to the PEM encoded key file                                        |                |
+| MF_AUTH_DB_SSL_ROOT_CERT | Path to the PEM encoded root certificate file                           |                |
+| MF_AUTH_HTTP_PORT        | Authn service HTTP port                                                 | 8180           |
+| MF_AUTH_GRPC_PORT        | Authn service gRPC port                                                 | 8181           |
+| MF_AUTH_SERVER_CERT      | Path to server certificate in pem format                                |                |
+| MF_AUTH_SERVER_KEY       | Path to server key in pem format                                        |                |
+| MF_AUTH_SECRET           | String used for signing tokens                                          | auth           |
+| MF_JAEGER_URL             | Jaeger server URL                                                       | localhost:6831 |
```
drasko: Align

drasko (changes requested):

auth/api/http/transport.go
```diff
@@ -46,6 +47,83 @@ func MakeHandler(svc authn.Service, tracer opentracing.Tracer) http.Handler {
 		opts...,
 	))
 
+	mux.Get("/members/:memberID/groups", kithttp.NewServer(
```
dborovcanin: Please reorder methods in CRUD order (where read = get one, list all, list members, list member of...).

auth/postgres/init.go
```diff
@@ -0,0 +1,122 @@
...
+						issuer_id   UUID NOT NULL,
+						issued_at   TIMESTAMP NOT NULL,
+						expires_at  TIMESTAMP,
+						PRIMARY KEY (id, issuer_id)
+					)`,
+					`CREATE extension LTREE`,
+					`CREATE TABLE IF NOT EXISTS group_type (
+						id INTEGER UNIQUE NOT NULL,
+						type VARCHAR(254) UNIQUE NOT NULL,
+						PRIMARY KEY (id)
```
dborovcanin, line 61: Since `type` is actually a string representation of group type (i.e. it's name), consider renaming to `name`. We can also add a short description of the group type, but it's probably unnecessary at the moment.

auth/postgres/tracing.go
```diff
@@ -21,6 +21,12 @@ type database struct {
 type Database interface {
 	NamedExecContext(context.Context, string, interface{}) (sql.Result, error)
 	QueryRowxContext(context.Context, string, ...interface{}) *sqlx.Row
+	QueryxContext(context.Context, string, ...interface{}) (*sqlx.Rows, error)
+	// NamedExecContext(context.Context, string, interface{}) (sql.Result, error)
```
dborovcanin: Please remove unused methods, both from the interface and from the implementation.

authn.proto
```diff
@@ -13,9 +13,12 @@ service ThingsService {
     rpc Identify(Token) returns (ThingID) {}
```
dborovcanin, line 13: This should be renamed to `auth.proto`.

dborovcanin (changes requested):

drasko (changes requested):

auth/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Deployment
+
+The service itself is distributed as Docker container. The following snippet
+provides a compose file template that can be used to deploy the service container
+locally:
+
+```yaml
+version: "2"
```
drasko: We use version `3.8` or something AFAIK. Please verify.

auth/api/grpc/client.go
```diff
@@ -0,0 +1,222 @@
...
+	return identityRes{id: res.GetId(), email: res.GetEmail(), err: nil}, nil
+}
+
+func (client grpcClient) Authorize(ctx context.Context, req *mainflux.AuthorizeReq, _ ...grpc.CallOption) (r *mainflux.AuthorizeRes, err error) {
+	ctx, close := context.WithTimeout(ctx, client.timeout)
+	defer close()
+
+	res, err := client.authorize(ctx, AuthZReq{Act: req.Act, Obj: req.Obj, Sub: req.Sub})
+	if err != nil {
+		return &mainflux.AuthorizeRes{Authorized: false, Err: err.Error()}, err
```
drasko: Please align naming of `AuthZReq{}` and `AuthorizeRes{}`

auth/api/grpc/requests.go
```diff
@@ -0,0 +1,119 @@
...
+		return auth.ErrMalformedEntity
+	}
+	return nil
+}
+
+// AuthZReq represents authorization request. It contains:
+// 1. subject - an action invoker
+// 2. object - an entity over which action will be executed
+// 3. action - type of action that will be executed (read/write)
+type AuthZReq struct {
```
drasko: Consider aligning this name (you are were using something like `AuthorizeRequest / Response`). I am not against `AuthZReq` also, but then align other entities.

auth/service.go
```diff
@@ -179,3 +193,109 @@ func (svc service) login(token string) (string, string, error) {
...
+	p, err := svc.groups.Members(ctx, groupID, offset, limit, gm)
+	if err != nil {
+		return groups.MemberPage{}, errors.Wrap(ErrFailedToRetrieveMembers, err)
+	}
+	mp := groups.MemberPage{
+		PageMetadata: groups.PageMetadata{
+			Total:  p.Total,
+			Offset: p.Offset,
+			Limit:  p.Limit,
+			// Name:   things,
```
drasko: Remove dead code

drasko (changes requested):

auth.proto
```diff
@@ -49,3 +52,36 @@ message IssueReq {
...
+message MembersReq {
+    string token    = 1;
+    string groupID  = 2;
+    uint64 offset   = 3;
+    uint64 limit    = 4;
+    string type     = 5;
+}
+
+message MembersRes {
+    uint64 total            = 1;
```
dborovcanin, line 81: Please allign.

auth/README.md
```diff
@@ -0,0 +1,118 @@
+# Auth - Authentication and Authorization service
+
+Auth service provides authentication features as an API for managing authentication keys. User service is using AuthN service gRPC API to obtain login token or password reset token. Authentication key consists of the following fields:
```
dborovcanin, line 3: `...provides authentication and authorization... Auth service gRPC API...`
I would split this into 3 sections: Authentication (where we talk about Keys), Groups, and Authorization (where we talk about policies).

auth/api/grpc/responses.go
```diff
@@ -0,0 +1,30 @@
...
+type issueRes struct {
+	value string
+	err   error
+}
+
+type authorizeRes struct {
+	authorized bool
+	err        string
+}
+type membersRes struct {
```
dborovcanin, line 18: Should this contain an err field, as well? Please put an empty line between type definitions (`membersRes` and `emptyRes`).

auth.proto
```diff
@@ -13,9 +13,12 @@ service ThingsService {
     rpc Identify(Token) returns (ThingID) {}
 }
 
-service AuthNService {
+service AuthService {
     rpc Issue(IssueReq) returns (Token) {}
```
dborovcanin, line 17: Since Auth service now works with three entity types (Keys, Groups, Policies), we should probably be more specific here: `IssueKey` or `IssueToken`.

auth/service.go
```diff
@@ -57,22 +65,28 @@ type Service interface {
 	// is returned. If token is invalid, or invocation failed for some
 	// other reason, non-nil error value is returned in response.
 	Identify(ctx context.Context, token string) (Identity, error)
+
```
dborovcanin, line 84: Please update the Service comment, it's a leftover from AuthN service (```Token is a string value of the actual Key and is used to authenticate an AuthN service request.```)

auth.proto
```diff
@@ -49,3 +52,36 @@ message IssueReq {
...
+    string token    = 1;
+    string groupID  = 2;
+    uint64 offset   = 3;
+    uint64 limit    = 4;
+    string type     = 5;
+}
+
+message MembersRes {
+    uint64 total            = 1;
+	uint64 offset           = 2;
```
blokovi: check indentation here, looks like it's messed up

auth/service.go
```diff
@@ -1,14 +1,16 @@
...
 
-package authn
+package auth
 
 import (
 	"context"
 	"time"
 
 	"github.com/mainflux/mainflux"
+	"github.com/mainflux/mainflux/internal/groups"
```
dborovcanin, line 11: Since Groups are now in the Auth service (we'll have generic groups), there is no need to put this in `internal`, since it won't be shared with other services. We can simply move it to `auth` package.

dborovcanin (changes requested):

twins/README.md
```diff
@@ -28,8 +28,8 @@ default values.
 | MF_TWINS_CA_CERTS          | Path to trusted CAs in PEM format                                    |                       |
 | MF_TWINS_CHANNEL_ID        | NATS notifications channel ID                                        |                       |
 | MF_NATS_URL                | Mainflux NATS broker URL                                             | nats://localhost:4222 |
-| MF_AUTHN_GRPC_URL          | AuthN service gRPC URL                                               | localhost:8181        |
-| MF_AUTHN_GRPC_TIMEOUT      | AuthN service gRPC request timeout in seconds                        | 1s                    |
+| MF_AUTH_GRPC_URL          | AuthN service gRPC URL                                               | localhost:8181        |
+| MF_AUTH_GRPC_TIMEOUT      | AuthN service gRPC request timeout in seconds                        | 1s                    |
```
blokovi, line 32: Alignment of table is broken with new envars, just add one space

things/README.md
```diff
@@ -44,8 +44,8 @@ default values.
 | MF_THINGS_SINGLE_USER_EMAIL | User email for single user mode (no gRPC communication with users)     |                |
 | MF_THINGS_SINGLE_USER_TOKEN | User token for single user mode that should be passed in auth header   |                |
 | MF_JAEGER_URL               | Jaeger server URL                                                      | localhost:6831 |
-| MF_AUTHN_GRPC_URL           | AuthN service gRPC URL                                                 | localhost:8181 |
-| MF_AUTHN_GRPC_TIMEOUT       | AuthN service gRPC request timeout in seconds                          | 1s              |
+| MF_AUTH_GRPC_URL           | AuthN service gRPC URL                                                 | localhost:8181 |
```
blokovi: add one space in renamed envars because of table alingment

auth/README.md
mteodor: yes, I would leave that for later, when everything is done

auth/service.go
mteodor: not sure about that since some endpoints may be in other services

dborovcanin: Yes, some endpoints probably will be in other services, but we won't share them; we won't use the `groups` package as an internally shared package, but as a dependency from the `auth` service. That is, we won't have multiple implementations of the Groups.

auth/postgres/init.go
```diff
@@ -0,0 +1,124 @@
...
+	Name        string
+	SSLMode     string
+	SSLCert     string
+	SSLKey      string
+	SSLRootCert string
+}
+
+// Connect creates a connection to the PostgreSQL instance and applies any
+// unapplied database migrations. A non-nil error is returned to indicate
+// failure.
```
drasko: Move this `failure` word to upper line for easier reading.

drasko (changes requested):

auth/service.go
```diff
@@ -0,0 +1,309 @@
...
+// implementation, and all of its decorators (e.g. logging & metrics).
+// Token is a string value of the actual Key and is used to authenticate
+// an Auth service request.
+type Service interface {
+	// Issue issues a new Key, returning its token value alongside.
+	IssueKey(ctx context.Context, token string, key Key) (Key, string, error)
+
+	// Revoke removes the Key with the provided id that is
+	// issued by the user identified by the provided key.
+	RevokeKey(ctx context.Context, token, id string) error
```
nmarcetic: Same question, is `Key` is necessary ?

auth/service.go
```diff
@@ -0,0 +1,309 @@
...
+	switch key.Type {
+	case APIKey, RecoveryKey, UserKey:
+		return Identity{ID: key.IssuerID, Email: key.Subject}, nil
+	default:
+		return Identity{}, ErrUnauthorizedAccess
+	}
+}
+
+func (svc service) Authorize(ctx context.Context, token, sub, obj, act string) (bool, error) {
+	return true, nil
```
nmarcetic, line 163: Is this is a placeholder for impl ?

auth/service.go
```diff
@@ -0,0 +1,309 @@
...
+	errRevoke    = errors.New("failed to remove key")
+	errRetrieve  = errors.New("failed to retrieve key data")
+	errIdentify  = errors.New("failed to validate token")
+)
+
+// Service specifies an API that must be fullfiled by the domain service
+// implementation, and all of its decorators (e.g. logging & metrics).
+// Token is a string value of the actual Key and is used to authenticate
+// an Auth service request.
+type Service interface {
```
nmarcetic: Would embeding interfaces makes more sense ? So i can see the [union of multiple interfaces](https://golang.org/doc/effective_go.html#embedding) (Groups and Auth) in Auth spec. IMHO it would be much more obvious.

internal/groups/groups.go
```diff
@@ -6,6 +6,7 @@ import (
 )
```
nmarcetic, line 7: Why we didn't put groups in `/pkg` ? Its not a topic of this PR , but I am curious. Looks like a good candidate to me.

auth/jwt/token_test.go
```diff
@@ -57,7 +57,7 @@ func TestParse(t *testing.T) {
 	require.Nil(t, err, fmt.Sprintf("issuing key expected to succeed: %s", err))
 
 	apiKey := key()
-	apiKey.Type = authn.APIKey
+	apiKey.Type = auth.APIKey
 	apiKey.ExpiresAt = time.Now().UTC().Add(-1 * time.Minute).Round(time.Second)
 	apiToken, err := tokenizer.Issue(apiKey)
```
nmarcetic, line 62: if you will follow naming, then token makes more sense than tokenizer (interface name).

auth/service.go
```diff
@@ -0,0 +1,309 @@
...
+	// Issue issues a new Key, returning its token value alongside.
+	IssueKey(ctx context.Context, token string, key Key) (Key, string, error)
```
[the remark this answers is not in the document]
mteodor: I did it because https://github.com/mainflux/mainflux/pull/1313#discussion_r548873164

auth/service.go
mteodor: yes

internal/groups/groups.go
mteodor: initialy I've put it there, but then I was asked to move it to `internal`, `pkg` holds packages that usually get imported into other  projects I think that was the reasoning.

auth/service.go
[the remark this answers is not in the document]
nmarcetic: I see. It makes sense. However, I do not see `IssueToken()` or IssueAnythig then IssueKey. Anyway, keep it like if you will have more then one thing to issue :)

internal/groups/groups.go
nmarcetic: IMHO `/pkg` is the right place for this. It's sharable and exported :)  True, but `internal` is for things that you explicitly don't someone import. Nvm, it's out of the scope of this PR. Thanks for the info

auth/postgres/tracing.go
```diff
@@ -30,14 +32,24 @@ func NewDatabase(db *sqlx.DB) Database {
 	}
 }
 
-func (d database) NamedExecContext(ctx context.Context, query string, args interface{}) (sql.Result, error) {
+func (dm database) NamedQueryContext(ctx context.Context, query string, args interface{}) (*sqlx.Rows, error) {
```
drasko: Why was this changed to `dm`, and what `m` really represents here (a bit confusing)?

twins/mocks/auth.go
```diff
@@ -0,0 +1,47 @@
...
+type authNServiceClient struct {
+	users map[string]string
+}
+
+// NewAuthServiceClient creates mock of auth service.
+func NewAuthServiceClient(users map[string]string) mainflux.AuthServiceClient {
+	return &authNServiceClient{users}
+}
+
+func (svc authNServiceClient) Identify(ctx context.Context, in *mainflux.Token, opts ...grpc.CallOption) (*mainflux.UserIdentity, error) {
```
drasko: `authN` or `auth`?

drasko (changes requested):

writers/influxdb/messages_test.go
```diff
@@ -84,7 +84,7 @@ func TestSave(t *testing.T) {
 	}
 
 	for _, tc := range cases {
-		// Clean previously saved messages.
+		// // Clean previously saved messages.
```
drasko: No need for double `//`. Please remove.

drasko (changes requested):

auth/api/grpc/endpoint.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+func issueEndpoint(svc auth.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(issueReq)
+		if err := req.validate(); err != nil {
+			return issueRes{}, err
+		}
+
+		now := time.Now().UTC()
```
drasko: I do not think that we need this var - just call the fnc directly in the struct declaration below.

auth/api/metrics.go
```diff
@@ -0,0 +1,172 @@
...
+var _ auth.Service = (*metricsMiddleware)(nil)
+
+type metricsMiddleware struct {
+	counter metrics.Counter
+	latency metrics.Histogram
+	svc     auth.Service
+}
+
+// MetricsMiddleware instruments core service by tracking request count and
+// latency.
```
drasko: Move this word to upper line to simplify comment reading.

auth/service.go
drasko: Please remove dead code.

drasko (changes requested):

dborovcanin (dismissed):

drasko (approved):
LGTM

PR #1318, opened 2021-01-08
MF-1034 - Wrapping MQTT client

### LINKED ISSUE #1034: Wrapping MQTT client
**FEATURE REQUEST**

Maybe we should consider wrapping MQTT client with few methods like Connect(srv.opts) Pub() Sub() Close() etc...and hide paho impl. I saw the same code at least 10 times (and it's not short, we will need it in Twins, e2e testing, etc...all places where we use MQTT).

Same for NATS, extend our MessagePublisher interface and hide nats impl. Consider adding multiple publishers' as an option, so I can pass the engine I use (kafka or NATS). The idea is to use our API in Core and hide the fact that we use nats or Kafka or even MQTT, make abstraction

nice-to-have

blokovi:

Resolves #1034


cmd/lora/main.go
```diff
@@ -115,7 +119,24 @@ func main() {
...
+		m := lora.Message{}
+		if err := json.Unmarshal(msg.Payload, &m); err != nil {
+			logger.Warn(fmt.Sprintf("Failed to Unmarshal message: %s", err.Error()))
+			return err
+		}
+		if err := svc.Publish(context.Background(), "", m); err != nil {
+			return err
+		}
+		return nil
+	})
```
manuio: I would keep this in a separate method like it was previously done with `subscribeToLoRaBroker`

cmd/lora/main.go
```diff
@@ -115,7 +119,24 @@ func main() {
...
 
-	go subscribeToLoRaBroker(svc, mqttConn, logger)
+	msub, err := mqttsub.NewSubscriber(cfg.loraMsgURL, cfg.subTimeout, logger)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Failed to create MQTT subscriber: %s", err))
+		os.Exit(1)
+	}
+
+	go msub.Subscribe(loraServerTopic, func(msg messaging.Message) error {
+		m := lora.Message{}
```
drasko: I think we have an internal convention to use something like `var m lora.Message` for the declaration.

cmd/lora/main.go
drasko: @manuio that might demand passing subscriber as a param, and I do not think it will be more elegant.

@blokovi please check how it is done in other services and align.

drasko (changes requested):

cmd/lora/main.go
blokovi: @drasko subscribers and publishers are passing as a params, so we can do that also here

cmd/lora/main.go
```diff
@@ -115,7 +119,14 @@ func main() {
 		}, []string{"method"}),
 	)
 
-	go subscribeToLoRaBroker(svc, mqttConn, logger)
+	msub, err := mqttsub.NewSubscriber(cfg.loraMsgURL, cfg.subTimeout, logger)
```
manuio: I would use `mqtt` without package alias

manuio (dismissed):
LGTM

cmd/lora/main.go
drasko: I agree - no need for this alias as it it is not overlapping with anything.

Only argument might be that it is somehow documenting wight from the start the intention of what we want to do with it in this module, and somehow prevent some other programmer to use it for PUB (or change the alias at this moment). If that was the intention, then I can accept it.

@dusanb94 @nmarcetic what is your opinion on this?

drasko (changes requested):

cmd/lora/main.go
nmarcetic: Agree, if there is no overlapping, no need for an alias.

cmd/lora/main.go
```diff
@@ -115,7 +119,14 @@ func main() {
...
+	if err != nil {
+		logger.Error(fmt.Sprintf("Failed to create MQTT subscriber: %s", err))
```
nmarcetic, line 124: I thought we switched to `errors.Wrap()`? I see `fmt.Sprintf` on a lot of places, not just here.

cmd/lora/main.go
dborovcanin: I really don't mind, both look fine with me.

dborovcanin (dismissed):

nmarcetic (dismissed):

cmd/lora/main.go
blokovi: I was following how it is used in mqtt:
https://github.com/mainflux/mainflux/blob/master/cmd/mqtt/main.go#L24
but there is overlapping so it had to be with alias.

Since here there is no other mqtt package, I removed alias.

manuio (approved):
LGTM

drasko (approved):
LGTM

PR #1322, opened 2021-01-11
MF-1061 - Implement protocol, name, v, vb, vs, vd and from/to postgre…

manuio:

Related to #1061

### DIFF
--- a/readers/postgres/messages.go
+++ b/readers/postgres/messages.go
@@ -51,13 +51,19 @@ func (tr postgresRepository) ReadAll(chanID string, offset, limit uint64, query 
 	LIMIT :limit OFFSET :offset;`, table, fmtCondition(chanID, query), order)
 
 	params := map[string]interface{}{
-		"channel":   chanID,
-		"limit":     limit,
-		"offset":    offset,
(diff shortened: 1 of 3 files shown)


blokovi (dismissed):

readers/postgres/messages.go
```diff
@@ -122,11 +123,54 @@ func fmtCondition(chanID string, query map[string]string) string {
...
+		case "from":
+			condition = fmt.Sprintf(`%s AND time >= :from`, condition)
+		case "to":
+			condition = fmt.Sprintf(`%s AND time < :to`, condition)
 		}
 	}
 	return condition
 }
 
+func fmtCount(chanID string, query map[string]string) (string, []interface{}) {
```
dborovcanin: Can named variables be used here (similar to the previous case)? I think that `sqlx` does not allow it, but please double-check. If it can be done, these two methods can be merged, because they will mostly be the same.

dborovcanin (changes requested):

dborovcanin (approved):

mteodor (approved):

drasko (approved):
LGTM

PR #1323, opened 2021-01-11
NOISSUE - Rename package aliases uuidProvider into uuid

manuio:


drasko (changes requested):
If I understand well, this is to remove dependency against Mainflux UUID provider?

However, I am afraid that whoever starts removing Mainflux UUID provider and replacing it with the ULID will completely forget and miss this file, as it will be no longer using it.
his is why I prefer keeping dependency, and reminding that the more general task should be undertaken - overall switch to ULID.

mteodor (dismissed):

readers/postgres/messages_test.go
```diff
@@ -38,13 +38,13 @@ var (
 func TestReadSenml(t *testing.T) {
 	writer := pwriter.New(db)
 
-	chanID, err := uuidProvider.New().ID()
+	chanID, err := uuid.New().ID()
```
dborovcanin: There is no need to instantiate a new ID provider for each ID. You can create one provider and use it for all IDs in this test. Check all the other places in tests where you have multiple instances.

dborovcanin (changes requested):

auth/postgres/key_test.go
```diff
@@ -12,18 +12,28 @@ import (
...
 	"github.com/mainflux/mainflux/auth/postgres"
 	"github.com/mainflux/mainflux/pkg/errors"
-	uuidProvider "github.com/mainflux/mainflux/pkg/uuid"
+	"github.com/mainflux/mainflux/pkg/uuid"
 	"github.com/opentracing/opentracing-go"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
```
dborovcanin:
```go
const email = "user-save@example.com"
```

readers/cassandra/messages_test.go
```diff
@@ -36,6 +36,8 @@ var (
 	vb          = true
 	vd          = "base64"
 	sum float64 = 42
+
```
dborovcanin, line 39: This `var` block looks like it's actually `const`. The only var is `uuidProvider`.

dborovcanin (changes requested):

readers/cassandra/messages_test.go
manuio: Not really, we need `var` to use the address (with `const`you get error `cannot take the address of v`). `addr` is also reassigned in `setup_test`

mteodor (dismissed):

users/service.go
```diff
@@ -159,21 +158,23 @@ type UserPage struct {
...
-	groups GroupRepository
-	hasher Hasher
-	email  Emailer
-	auth   mainflux.AuthServiceClient
+	users        UserRepository
+	groups       GroupRepository
+	hasher       Hasher
+	email        Emailer
+	auth         mainflux.AuthServiceClient
+	uuidProvider mainflux.IDProvider
```
dborovcanin: This is not necessarily a UUID provider (we'll switch to ULID soon) and the underlying format should be transparent to the Users service (that's one of the reasons we use the interface), so `idp` or `idProvider` sounds like a better naming.

users/service.go
```diff
@@ -186,7 +187,7 @@ func (svc usersService) Register(ctx context.Context, user User) (string, error)
 		return "", errors.Wrap(ErrMalformedEntity, err)
 	}
 	user.Password = hash
-	uid, err := uuidProvider.New().ID()
+	uid, err := svc.uuidProvider.ID()
```
dborovcanin: I'm glad you fixed this, it was a horrible mistake.

cmd/users/main.go
```diff
@@ -284,7 +285,9 @@ func newService(db *sqlx.DB, tracer opentracing.Tracer, auth mainflux.AuthServic
 		logger.Error(fmt.Sprintf("Failed to configure e-mailing util: %s", err.Error()))
 	}
 
-	svc := users.New(userRepo, groupRepo, hasher, auth, emailer)
+	uuidProvider := uuid.New()
```
dborovcanin: Since we are using `hasher` as an instance of the hasher (`bcrypt`), we should do the same for the ID provider. We don't really care what's the underlying implementation.

dborovcanin (changes requested):

users/service.go
manuio: Right, let's follow interface naming.

blokovi (dismissed):

dborovcanin (dismissed):

dborovcanin (approved):

mteodor (approved):

drasko (approved):
LGTM

PR #1324, opened 2021-01-11
MF-1311 - Add Notifications service

### LINKED ISSUE #1311: Add Notifications service
**FEATURE REQUEST**

No.

I want to create a Mainflux Writer that can send me an email when a message to some topic (channel + subtopic) is received. In the ideal case, these messages are processed using some sort of rules engine so that I can specify rules when receiving emails to avoid spam messages and unnecessary traffic, but creating an SMTP writer is the first step in that direction. Potential,  use cases are limitless, but most of them are related to notifications and alarms.

This is a should-have.

dborovcanin:
### What does this do?
This pull request introduces Mainflux notifier services and adds an implementation of SMTP Notifier (capability of sending email notifications).

This pull request resolves #1311.

There is no such functionality.

No.

Yes, but more documentation is needed, as well as a follow-up pull request for [docs](https://www.github.com/mainflux/docs).


consumers/notifiers/api/logging.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		lm.logger.Info(fmt.Sprintf("%s without errors.", message))
+	}(time.Now())
+
+	return lm.svc.ViewSubscription(ctx, token, topic)
+}
+
+func (lm *loggingMiddleware) ListSubscriptions(ctx context.Context, token, topic string) (res []notifiers.Subscription, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method list_subscription for topic %s and token %s took %s to complete", topic, token, time.Since(begin))
```
manuio: `list_subscriptions`

consumers/notifiers/api/responses.go
```diff
@@ -0,0 +1,90 @@
...
+func (res viewSubRes) Headers() map[string]string {
+	return map[string]string{}
+}
+
+func (res viewSubRes) Empty() bool {
+	return false
+}
+
+type listSubsRes struct {
+	Data []viewSubRes `json:"subscriptions"`
```
manuio: `Data` is not consistent with `json:"subscriptions"`

manuio (changes requested):

cmd/smtp-notify/main.go
```diff
@@ -0,0 +1,320 @@
...
+func connectToDB(dbConfig postgres.Config, logger logger.Logger) *sqlx.DB {
+	db, err := postgres.Connect(dbConfig)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Failed to connect to postgres: %s", err))
+		os.Exit(1)
+	}
+	return db
+}
+
+func connectToAuthn(cfg config, tracer opentracing.Tracer, logger logger.Logger) (mainflux.AuthServiceClient, func() error) {
```
drasko: You probably mean `Auth`

drasko (changes requested):

cmd/smtp-notify/main.go
```diff
@@ -0,0 +1,320 @@
...
+			opts = append(opts, grpc.WithTransportCredentials(tpc))
+		}
+	} else {
+		opts = append(opts, grpc.WithInsecure())
+		logger.Info("gRPC communication is not encrypted")
+	}
+
+	conn, err := grpc.Dial(cfg.authURL, opts...)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Failed to connect to authn service: %s", err))
```
manuio: `auth`

cmd/smtp-notify/main.go
```diff
@@ -0,0 +1,320 @@
...
+		signal.Notify(c, syscall.SIGINT)
+		errs <- fmt.Errorf("%s", <-c)
+	}()
+
+	err = <-errs
+	logger.Error(fmt.Sprintf("Users service terminated: %s", err))
+}
+
+func loadConfig() config {
+	authnTimeout, err := time.ParseDuration(mainflux.Env(envAuthTimeout, defAuthTimeout))
```
manuio: `authTimeout`

manuio (changes requested):

drasko (dismissed):
LGTM

nmarcetic (dismissed):

cmd/smtp-notify/main.go
```diff
@@ -0,0 +1,320 @@
...
+
+	authTracer, closer := initJaeger("auth", cfg.jaegerURL, logger)
+	defer closer.Close()
+
+	auth, close := connectToAuth(cfg, authTracer, logger)
+	if close != nil {
+		defer close()
+	}
+
+	tracer, closer := initJaeger("notify-smtp", cfg.jaegerURL, logger)
```
mteodor: here https://github.com/mainflux/mainflux/pull/1324/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R8 it is `smtp-notify`

consumers/notify/api/transport.go
```diff
@@ -0,0 +1,195 @@
...
+
+// MakeHandler returns a HTTP handler for API endpoints.
+func MakeHandler(svc notify.Service, tracer opentracing.Tracer) http.Handler {
+	opts := []kithttp.ServerOption{
+		kithttp.ServerErrorEncoder(encodeError),
+	}
+
+	mux := bone.New()
+
+	mux.Post("/subscriptions", kithttp.NewServer(
```
mteodor: singular?

manuio: We are using plural for other services

consumers/notify/service.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	return ns.subs.Retrieve(ctx, id)
+}
+
+func (ns *notifierService) ListSubscriptions(ctx context.Context, token string, pm PageMetadata) (Page, error) {
+	_, err := ns.auth.Identify(ctx, &mainflux.Token{Value: token})
+	if err != nil {
+		return Page{}, errors.Wrap(ErrUnauthorizedAccess, err)
+	}
```
manuio: You can inline here and for others

consumers/notify/postgres/subscriptions.go
```diff
@@ -0,0 +1,170 @@
...
+		args["limit"] = pm.Limit
+	}
+
+	rows, err := repo.db.NamedQueryContext(ctx, q, args)
+	if err != nil {
+		return notify.Page{}, errors.Wrap(notify.ErrSelectEntity, err)
+	}
+	defer rows.Close()
+
+	subs := []notify.Subscription{}
```
manuio: You can use `var subs []notify.Subscription`

consumers/notify/postgres/init.go
```diff
@@ -0,0 +1,68 @@
...
+		Migrations: []*migrate.Migration{
+			{
+				Id: "subscriptions_1",
+				Up: []string{
+					`CREATE TABLE IF NOT EXISTS subscriptions (
+                        id          VARCHAR(254) PRIMARY KEY,
+                        owner_id    VARCHAR(254) NOT NULL,
+                        contact     VARCHAR(254),
+                        topic       TEXT,
+                        UNIQUE(topic, contact)
```
manuio: Is this properly aligned?

consumers/notify/postgres/subscriptions_test.go
```diff
@@ -0,0 +1,244 @@
...
+	id1, err := idProvider.ID()
+	require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+
+	id2, err := idProvider.ID()
+	require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+
+	sub1 := notify.Subscription{
+		OwnerID: "owner",
+		ID:      id1,
+		Contact: "ownersave@example.com",
```
manuio: You can create a const for `Contact`

consumers/notify/postgres/subscriptions_test.go
```diff
@@ -0,0 +1,244 @@
...
+	repo := postgres.New(dbMiddleware)
+
+	id1, err := idProvider.ID()
+	require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+
+	id2, err := idProvider.ID()
+	require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+
+	sub1 := notify.Subscription{
+		OwnerID: "owner",
```
manuio: I would use a real ID

consumers/notify/postgres/subscriptions_test.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func TestRetrieveAll(t *testing.T) {
+	_, err := db.Exec("DELETE FROM subscriptions")
+	require.Nil(t, err, fmt.Sprintf("cleanup must not fail: %s", err))
+
+	dbMiddleware := postgres.NewDatabase(db)
+	repo := postgres.New(dbMiddleware)
+
+	const numSubs = 100
```
manuio: Can we maybe define this as global? I saw other constants with value `100`

manuio (changes requested):

nmarcetic (commented):
Good work @dusanb94 
I have just one informal question.
[Here](https://github.com/mainflux/mainflux/blob/136e3e60a79fa66056a35b01e9363433feb71ff6/consumers/writers/postgres/consumer.go#L69)
Is this is an error from BeginTxx? Not from query execution?
If so, what we are rolling back? It should fail transactions (when message q exec fails).
One from naming,
It would be better to call it `notifiers` to keep consistency with writers and readers

consumers/notify/api/requests.go
```diff
@@ -0,0 +1,64 @@
...
+
+var (
+	errInvalidTopic   = errors.New("invalid Subscritpion topic")
+	errInvalidContact = errors.New("invalid Subscritpion contact")
+	errNotFound       = errors.New("invalid or empty Subscription id")
+)
+
+type createSubReq struct {
+	token   string
+	Topic   string
```
darkodraskovic, line 19: Is there a reason to omit unmarshal string?

darkodraskovic (changes requested):

consumers/notify/api/requests.go
darkodraskovic: Is there a reason to omit unmarshal string?

darkodraskovic (changes requested):

consumers/notify/api/requests.go
dborovcanin: Yes, `topic` is extracted from the URL path, rather than the JSON body.

consumers/notify/postgres/init.go
dborovcanin: Yes, it's aligned with the row above.

dborovcanin:
@nmarcetic 
> Good work @dusanb94
> I have just one informal question.
> [Here](https://github.com/mainflux/mainflux/blob/136e3e60a79fa66056a35b01e9363433feb71ff6/consumers/writers/postgres/consumer.go#L69)
> Is this is an error from BeginTxx? Not from query execution?
> If so, what we are rolling back? It should fail transactions (when message q exec fails).

Oh, it looks like that's the case, which means it's a bug. I'll open an issue. Great catch.
> One from naming,
> It would be better to call it `notifiers` to keep consistency with writers and readers

I was initially calling it `notifiers` but I renamed it to `notify` and named an interface that's in charge of sending notifications `Notifier`. I can rename if other @mainflux/maintainers agree.

nmarcetic:
> @nmarcetic
> 
> > Good work @dusanb94
> > I have just one informal question.
> > [Here](https://github.com/mainflux/mainflux/blob/136e3e60a79fa66056a35b01e9363433feb71ff6/consumers/writers/postgres/consumer.go#L69)
> > Is this is an error from BeginTxx? Not from query execution?
> > If so, what we are rolling back? It should fail transactions (when message q exec fails).
> 
> Oh, it looks like that's the case, which means it's a bug. I'll open an issue. Great catch.
> 
> > One from naming,
> > It would be better to call it `notifiers` to keep consistency with writers and readers
> 
> I was initially calling it `notifiers` but I renamed it to `notify` and named an interface that's in charge of sending notifications `Notifier`. I can rename if other @mainflux/maintainers agree.

I don't have preferences really, it's just consistent naming with other `consumers` like `writers`. 
Let's hear from @mainflux/maintainers , feel free to mark it as resolved.

consumers/notify/api/requests.go
darkodraskovic: Should it be exported (uppercase) in that case?

drasko:
@dusanb94 @nmarcetic regarding https://github.com/mainflux/mainflux/pull/1324#pullrequestreview-572404529 - I agree that `notifiers` would make more sense here, as we count on having several notification methods.

consumers/notifiers/tracing/subscriptions.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"context"
+
+	notifiers "github.com/mainflux/mainflux/consumers/notifiers"
+	opentracing "github.com/opentracing/opentracing-go"
+)
+
+const (
+	saveOp            = "save_op"
+	retrieveByEmailOp = "retrieve_by_email"
```
manuio: Should we use `retrieve_by_email_op` ?

consumers/notifiers/subscriptions.go
```diff
@@ -0,0 +1,46 @@
...
+	// error response.
+	Save(ctx context.Context, sub Subscription) (string, error)
+
+	// Retrieve retrieves the subscription for the given id.
+	Retrieve(ctx context.Context, id string) (Subscription, error)
+
+	// RetrieveAll retrieves all the subscriptions for the given metadata.
+	RetrieveAll(ctx context.Context, pm PageMetadata) (Page, error)
+
+	// Remove removes the subscription having the provided an ID.
```
manuio: Is `having the provided an ID.` ok? maybe `for the given ID`?

consumers/notifiers/api/logging.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		lm.logger.Info(fmt.Sprintf("%s without errors.", message))
+	}(time.Now())
+
+	return lm.svc.ViewSubscription(ctx, token, topic)
+}
+
+func (lm *loggingMiddleware) ListSubscriptions(ctx context.Context, token string, pm notifiers.PageMetadata) (res notifiers.Page, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method list_subscription for topic %s and token %s took %s to complete", pm.Topic, token, time.Since(begin))
```
manuio: `list_subscriptions`

consumers/notifiers/api/transport.go
```diff
@@ -0,0 +1,195 @@
...
+
+	mux.Get("/subscriptions", kithttp.NewServer(
+		kitot.TraceServer(tracer, "list_subscriptions")(listSubscriptionsEndpoint(svc)),
+		decodeList,
+		encodeResponse,
+		opts...,
+	))
+
+	mux.Delete("/subscriptions/:id", kithttp.NewServer(
+		kitot.TraceServer(tracer, "delete_group")(deleteSubscriptionEndpint(svc)),
```
manuio: `delete_subscription`

consumers/notifiers/notifier.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/mainflux/mainflux/pkg/messaging"
+)
+
+// ErrNotify wraps sending notification errors,
+var ErrNotify = errors.New("Error sending notification")
+
+// Notifier represents an API for sending notification.
+type Notifier interface {
+	Notify(from string, to []string, msg messaging.Message) error
```
manuio, line 19: This should have a comment

consumers/notifiers/openapi.yml
```diff
@@ -0,0 +1,206 @@
...
+          description: Total number of items.
+        offset:
+          type: integer
+          description: Number of items to skip during retrieval.
+        limit:
+          type: integer
+          description: Maximum number of items to return in one page.
+
+  parameters:
+    ID:
```
manuio: I think  that this should be `Id`. Maybe add a prefix like we did for `ThingId` ?

manuio (changes requested):

consumers/notifiers/api/requests.go
```diff
@@ -0,0 +1,64 @@
...
+
+package api
+
+import (
+	notifiers "github.com/mainflux/mainflux/consumers/notifiers"
+	"github.com/mainflux/mainflux/pkg/errors"
+)
+
+var (
+	errInvalidTopic   = errors.New("invalid Subscritpion topic")
```
mteodor: Subscription

consumers/notifiers/postgres/subscriptions.go
```diff
@@ -0,0 +1,165 @@
...
+	return sub.ID, nil
+}
+
+func (repo subscriptionsRepo) Retrieve(ctx context.Context, id string) (notifiers.Subscription, error) {
+	q := `SELECT id, owner_id, contact, topic FROM subscriptions WHERE id = $1`
+	sub := dbSubscription{}
+	if err := repo.db.QueryRowxContext(ctx, q, id).StructScan(&sub); err != nil {
+		if err == sql.ErrNoRows {
+			return notifiers.Subscription{}, errors.Wrap(notifiers.ErrNotFound, err)
+
```
mteodor: extra line

manuio (dismissed):
LGTM

mteodor (dismissed):

consumers/notifiers/README.md
```diff
@@ -0,0 +1,23 @@
+# Notifiers service
```
drasko, line 1: Wait... I was not expecting a full-blown service called `Notifiers`, as we do not have a special service called `Writers`. We have interface for Writers, but then each Writer is a separate service.

While `writers` is a folder that holed different writers, and not a special Writers service, here it looks like "the main boss in town" is a service (Notify service), and notifiers are only "executors" of this service, configured on deployment. It is not same for Writers - you do not configure `Writers` service to use a special writer - writer itself is a service.

So I think that we should re-discuss this naming. I find it also somehow misleading because it suggest that `writers` and `notifiers` have the same structure, when it looks that they do not.

drasko (changes requested):

consumers/notifiers/README.md
nmarcetic: @drasko For me, It's actually the same as a writer, a consumer interface impl. Since it is more complex than it looks like a `full-blown` service.
The difference is notifier service, repo, and obviously API (it can't use shared `consumers/api` as writers, cause it's not the same, it's a different API and domain level. Actually it's just the same pattern, nothing else). 
`Wait... I was not expecting a full-blown service called Notifiers, as we do not have a special service called Writers. We have an interface for Writers, but then each Writer is a separate service.` True, as you can see there is a separate service called `smtp-notifier`, it's just one of many notifiers, exactly the same as writers.
When the next notifier comes up, for example, `sms-notifier` it will impl Notifier interface (shooting our SMS message) but the domain level will be the same (transport, service, repository, models).  
@dusanb94 Maybe it will be easy to understand and follow the code if you move `consumers/api` into the `consumers/writers`? That way notifiers and writers would be aligned and easier to understand. I am right? :)

dborovcanin: They do not have the same structure. They only share the consumer part, that is: both writers and notifier service subscribe to the NATS channel, optionally transform, and consume received messages. While writers can store messages directly, the Notifier service needs to do more: it needs to check the topic, takes all the users who subscribed to that topic, and send them notifications. This flow is common for all the notifiers (just like "take a message and store it" is common for all the writers), but the Notifier itself requires some sort of wrapper to provide HTTP API and subscription management. For that reason, the code for sending notifications is extracted as a separate dependency (Notifier) for `Notifications service`: you implement repository and service only once and reuse it with different Notifiers. Fundamentally, the difference between Notifier and Writer is precisely what you wrote: `you do not configure Writers service to use a special writer - writer itself is a service`, while Notifier can't be a service per se, because it needs supporting layer for subscription handling and API part to make it work. Since that supporting layer is always the same, it can be extracted to the Notifier service.

nmarcetic: Nicely explained :clap: IMHO it's clean and I don't see any anti-patterns or pitfalls.

dborovcanin: Thanks! The only problem I see is starting up the service. If we make the Notifier type configurable from the env, it will get complicated because one Notifier service instance needs to handle env vars for all the Notifiers, and different notifiers are configured completely differently. For example, the Notifiers service needs to handle SMTP, SMPP, and WS configurations to support 3 Notifier types. There are 3 approaches:

1. Create multiple main methods for different Notifier types - a lot of duplicated code, but option to use different DBs for different Notifier types (in one DB contact is phone number, in another it's email... so we don't mix them)
2. Make Notifier type configurable in `main.go`: all the possible configs need to be handled, but only one Notifier will be created and passed to the service
3. Support multiple Notifiers; make dependency _a list of Notifiers you can pass to the Notification service_; this also implies that in `main` we handle all the configurations

nmarcetic: Option 1. Is how we did for writers, right? And IMHO it should be merged to a single main for writers, cause the storage layer (e.g different database configs and connection fn calls) can be easily wrapped and made generic. Same for readers, but anyway it's a different topic :)   

Option 2. Have the same cons as option 1. handling configs. 
Option 3. Same, handling configs. 

We fall slowly into env config hell and I would really consider config files (yaml, toml whatever). 
Regarding notifiers, if we don't want to refactor writers or move config to files, I think option 1. is what we are already doing and probably a way to go.  
If we are willing to refactor writers and merge main, we should then think about doing the same for notifiers (with config files).

dborovcanin: I agree. Regardless of this particular problem, we need to tackle configuration issues for just about all the services. We repeat NATS, Auth, HTTP(S) serve, and DB connection in almost every `main.go` and that's now quite a lot of code to maintain. It starts to hurt our codebase and I think that we should start resolving this issue ASAP.

nmarcetic: I will open a separate issue, for discussion that addressing this config problem.

darkodraskovic (approved):

drasko (approved):
LGTM

PR #1326, opened 2021-01-12
MF-1061 - Implement v, vb, vs, vd and from/to mongodb-reader filters

manuio:

Related to #1061


readers/mongodb/messages.go
```diff
@@ -109,6 +110,34 @@ func fmtCondition(chanID string, query map[string]string) *bson.D {
 			"name",
 			"protocol":
 			filter = append(filter, bson.E{Key: name, Value: value})
+		case "v":
```
dborovcanin, line 113: Can you try using values without parsing (because the value type in `bson.E` is `interface{}`)? Or at least consider returning conversion error.

mteodor (dismissed):

readers/mongodb/messages_test.go
```diff
@@ -56,31 +51,55 @@ func TestReadSenml(t *testing.T) {
 	db := client.Database(testDB)
 	writer := writer.New(db)
 
+	chanID, err := uuid.New().ID()
+	require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+	pubID, err := uuid.New().ID()
+	require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+	pub2ID, err := uuid.New().ID()
```
darkodraskovic: Why not `pubID2`? It's a second ID.

readers/mongodb/messages_test.go
```diff
@@ -117,13 +136,13 @@ func TestReadSenml(t *testing.T) {
 		},
 		"read message last page": {
 			chanID: chanID,
-			offset: 40,
-			limit:  10,
+			offset: 95,
+			limit:  limit,
```
darkodraskovic: Maybe we can remove this `limit` field from this struct type and use `tc.page.Limit` in line 274. It's an unecessary repetition.

darkodraskovic (changes requested):

readers/mongodb/messages.go
manuio: I agree, but as explained here #1061 the idea would be to replace `query map[string]string` by `rpm readers.PageMetadata`. That way the parsing would be done at the endpoint layer instead of here. If I return an error here I'll remove it in my next PR.

readers/mongodb/messages_test.go
manuio: It's also a second publisher. ID of publisher2

readers/mongodb/messages_test.go
manuio: `tc.page.limit` represents the response while `limit` represents a parameter of the request. We can't use it as the same value but what we can maybe add is a test like it's done for `Total`. Something like: `assert.Equal(t, tc.page.Limit, result.Limit, fmt.Sprintf("%s: expected %v got %v", desc, tc.page.Limit, result.Limit))`

darkodraskovic: I understand. As far as I can see, the value of `tc.limt` and `tc.Page.Limit` is always the same. If they are guaranteed to be same, one should be removed.

Besides, maybe I'm missing something, but the values of `tc.Page.Limit` and `tc.Page.Offset` are not used in tests. We should either remove them or use them.

readers/mongodb/messages.go
darkodraskovic: Does this mean that you will detect the invalid query at the endpoint layer? If yes, than this error checking should be removed.

manuio: Yes, it will be previously detected to properly set the structure `PageMetadata`. For now I check it and `continue` to avoid a bug. I'll remove it for all readers in the next PR.

readers/mongodb/messages_test.go
manuio: Yes, you are right, it's not used. It's also better if we pass all filters in the `page`

readers/mongodb/messages_test.go
drasko: I would also prefer `pubID2`, because this look like `pub-to-ID`. In functional languages we often see these vars with numbers in the end (because vars are immutable, and you have to add new with the practically same name).

blokovi (dismissed):

blokovi (dismissed):

readers/mongodb/messages.go
```diff
@@ -109,6 +110,34 @@ func fmtCondition(chanID string, query map[string]string) *bson.D {
 			"name",
```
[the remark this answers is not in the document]
dborovcanin: I think so. @manuio Can you remove the pointer here?

readers/mongodb/messages.go
```diff
@@ -109,8 +110,36 @@ func fmtCondition(chanID string, query map[string]string) *bson.D {
...
+			if err != nil {
+				continue
+			}
+			filter = append(filter, bson.E{Key: "time", Value: bson.M{"$gte": fVal}})
+		case "to":
+			fVal, err := strconv.ParseFloat(value, 64)
+			if err != nil {
+				continue
+			}
+			filter = append(filter, bson.E{Key: "time", Value: bson.M{"$lt": fVal}})
```
darkodraskovic, line 140: Is there any parrticular reason to use here `$lt` instead of `$lte`. I'm asking this, since you're using `$gte` for the `from` test.

readers/mongodb/messages_test.go
```diff
@@ -6,47 +6,42 @@ package mongodb_test
...
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
 const (
 	testDB      = "test"
 	collection  = "messages"
-	chanID      = "1"
 	subtopic    = "subtopic"
-	msgsNum     = 42
+	msgsNum     = 100
```
darkodraskovic, line 28: Maybe it's better to say `numMsgs` - number of messages (just an aesthetic remark).

darkodraskovic (dismissed):
LGTM

dborovcanin (dismissed):

readers/mongodb/messages.go
manuio: The reason was to do it like for other readers. After to think about I prefer `>=` and `<=`. I'll do it for all readers.

darkodraskovic (approved):
LGTM

dborovcanin (approved):

Issue #1286, opened 2020-11-18
Consider using Envconfig


drasko:
**ENHANCEMENT**
1. Describe the enhancement you are requesting. Enhancements include:
We could use @kelseyhightower's [Envconfig](https://github.com/kelseyhightower/envconfig) to simplify our implementation in each of he services.

2. Indicate the importance of this enhancement to you (must-have, should-have, nice-to-have).
Should-have

arvindh123:
This was done in https://github.com/absmach/magistrala/pull/1601  using another library https://github.com/caarlos0/env

PR #1331, opened 2021-01-16
MF-1265 - Make Transformer type configurable

### LINKED ISSUE #1265: Make Transformers configurable
**FEATURE REQUEST**

No.

Currently, Mainflux writers use transformers to transform the Mainflux message to the known format. However, those transformers are not configurable using Writer's configuration (env vars) but need to be changed programmatically. 

This is a must-have.

dborovcanin:

### What does this do?
This pull request enables configuring the Transformer type for the Mainflux writer through environment variables.
Resolves #1265.

There are no such breaks.

No.

Yes.


cmd/cassandra-writer/main.go
```diff
@@ -156,6 +162,21 @@ func newService(session *gocql.Session, logger logger.Logger) consumers.Consumer
 	return repo
 }
 
+func makeTransformer(cfg config, logger logger.Logger) transformers.Transformer {
```
mteodor, line 165: whould it make sense that this function is actually in Transformer interface?

cmd/cassandra-writer/main.go
```diff
@@ -156,6 +162,21 @@ func newService(session *gocql.Session, logger logger.Logger) consumers.Consumer
...
+	switch strings.ToUpper(cfg.transformer) {
+	case "SENML":
+		logger.Info("Using SenML transformer")
+		return senml.New(cfg.contentType)
+	case "JSON":
```
mteodor, line 170: should this be a constant?

cmd/influxdb-writer/main.go
```diff
@@ -150,6 +157,21 @@ func makeMetrics() (*kitprometheus.Counter, *kitprometheus.Summary) {
...
+	switch strings.ToUpper(cfg.transformer) {
+	case "SENML":
+		logger.Info("Using SenML transformer")
+		return senml.New(cfg.contentType)
+	case "JSON":
+		logger.Info("Using JSON transformer")
+		return json.New()
+	default:
+		logger.Error(fmt.Sprintf("Can't create transformer: unknown transformer type %s", cfg.transformer))
+		os.Exit(1)
```
mteodor, line 170: why exit here?

manuio: I agree, `SenML` should be also `default`

drasko: Default is probably defined by ENV constant, maybe it is not bad to report error here and give user / deployment engineer info that ENV var has not be set.

drasko (approved):
LGTM

cmd/influxdb-writer/main.go
dborovcanin: Yes, SenML is default and it makes sense to put it here, but the problem is that SenML requires format when instantiation, so defaulting (no env set) would also imply that we set a default format. I'm afraid it's too many implicit default values.

cmd/cassandra-writer/main.go
dborovcanin: This can't be extracted to the Transformer interface because it would cause circular dependency (implementation imports Transformer to implement it and Transformer imports implementations so that it can instantiate them).

cmd/cassandra-writer/main.go
dborovcanin: It's used only here, so I'd say it's not necessary (it only adds a couple of extra lines of code).

nmarcetic: I agree with @mteodor 
Maybe we can export it from `/pkg/messaging`? Currently is used only here yes, but also in all writers and probably will be used in other places too. In general, it would be cool to call it `messaging.format.senml`  or similar.

cmd/influxdb-writer/main.go
nmarcetic: Agree,  exit is ok here. Better to exit then trick me to use a format that I didn't want it. It's too critical IMHO

nmarcetic (dismissed):

cmd/cassandra-writer/main.go
dborovcanin: That makes sense, but it looks like it's related to  [this comment](https://github.com/mainflux/mainflux/pull/1331#discussion_r559023531). I'm not sure if we should put these constants in `pkg/messaging/transformer` and have implementation info on the same location where the interface is, or in a separate `configuration` package. We have a lot of repetitive code in `cmd` and we should probably fix this in a separate PR attached to issue #1286.

nmarcetic: @dusanb94 Yea I agree about transformers, but I don't think it's useful only for configuration, either related to transformers. 
IMHO it's a simple string, a format type definition that we support (in your PR used for config and related to transformers yes, but not neccearly), It actually belongs to messaging in general (it defines message format type). That's why I see it in `/pkg/messaging` as exported available format, currently a type string, easily changeable to any other format.

drasko (dismissed):
LGTM

manuio (dismissed):
LGTM

manuio (approved):
LGTM

mteodor (approved):

PR #1333, opened 2021-01-20
MF-1061 - Add PageMetadata to readers

### LINKED ISSUE #1061: Improve readers message retrival querying

manuio:

Resolves #1061


readers/mocks/messages.go
```diff
@@ -24,29 +24,27 @@ func NewMessageRepository(messages map[string][]readers.Message) readers.Message
...
+func (repo *messageRepositoryMock) ReadAll(chanID string, rpm readers.PageMetadata) (readers.MessagesPage, error) {
 	repo.mutex.Lock()
 	defer repo.mutex.Unlock()
 
-	end := offset + limit
+	end := rpm.Offset + rpm.Limit
 
 	numOfMessages := uint64(len(repo.messages[chanID]))
-	if offset < 0 || offset >= numOfMessages {
+	if rpm.Offset < 0 || rpm.Offset >= numOfMessages {
```
dborovcanin: Offset and Limit are `uint`, so comparison `<0` makes no sense.

readers/mocks/messages.go
```diff
@@ -24,29 +24,27 @@ func NewMessageRepository(messages map[string][]readers.Message) readers.Message
 	}
 }
 
-func (repo *messageRepositoryMock) ReadAll(chanID string, offset, limit uint64, query map[string]string) (readers.MessagesPage, error) {
+func (repo *messageRepositoryMock) ReadAll(chanID string, rpm readers.PageMetadata) (readers.MessagesPage, error) {
```
dborovcanin, line 33: This mock does not work properly because it ignores all the parameters other than offset and limit. Other parameters should probably be added into account and used for better API testing (Currently, we don't check returned messages, but only response code in API tests).

dborovcanin (changes requested):

readers/postgres/messages.go
```diff
@@ -37,33 +37,31 @@ func New(db *sqlx.DB) readers.MessageRepository {
...
+		"subtopic":     rpm.Subtopic,
+		"publisher":    rpm.Publisher,
+		"name":         rpm.Name,
+		"protocol":     rpm.Protocol,
+		"value":        rpm.Value,
+		"bool_value":   rpm.BoolValue,
+		"string_value": rpm.StringValue,
+		"data_value":   rpm.DataValue,
+		"from":         rpm.From,
+		"to":           rpm.To,
```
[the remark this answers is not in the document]
manuio: What do you mean? Maybe I can use a structure and Unmarshal.

readers/mocks/messages.go
dborovcanin: Looks like the `total` value is missing in the mock? Also, in this implementation, it looks like the operator between filters is `OR`, rather than `AND`.

drasko:
I guess @dusanb94 needs to look at this one - are these changes OK now?

readers/mocks/messages.go
manuio: Right, thank you. I just added Total check and modified the mock to allow combination of filters.

readers/mocks/messages.go
```diff
@@ -43,69 +43,95 @@ func (repo *messageRepositoryMock) ReadAll(chanID string, rpm readers.PageMetada
 	json.Unmarshal(meta, &query)
 
 	var msgs []readers.Message
+	filter := false
 	for _, m := range repo.messages[chanID] {
 		senml := m.(senml.Message)
+		filterOk := false
 
-		filter := false
+	forLoop:
```
dborovcanin: This looks unnecessarily complicated. Can you try to simplify with something like this:
```go

func (repo *messageRepositoryMock) ReadAll(chanID string, rpm readers.PageMetadata) (readers.MessagesPage, error) {
	repo.mutex.Lock()
	defer repo.mutex.Unlock()

	if rpm.Format != "" && rpm.Format != "messages" {
		return readers.MessagesPage{}, nil
	}

	var query map[string]interface{}
	meta, _ := json.Marshal(rpm)
	json.Unmarshal(meta, &query)

	var msgs []readers.Message
	for _, m := range repo.messages[chanID] {
		senml := m.(senml.Message)
		filter := true

		for name := range query {
			switch name {
			case "subtopic":
				if rpm.Subtopic != senml.Subtopic {
					filter = false
				}
			case "publisher":
				if rpm.Publisher != senml.Publisher {
					filter = false
				}
			case "name":
				if rpm.Name != senml.Name {
					filter = false
				}
			case "protocol":
				if rpm.Protocol != senml.Protocol {
					filter = false
				}
			case "v":
				if senml.Value != nil &&
					*senml.Value != rpm.Value {
					filter = false
				}
			case "vb":
				if senml.BoolValue != nil &&
					*senml.BoolValue != rpm.BoolValue {
					filter = false
				}
			case "vs":
				if senml.StringValue != nil &&
					*senml.StringValue != rpm.StringValue {
					filter = false
				}
			case "vd":
				if senml.DataValue != nil &&
					*senml.DataValue != rpm.DataValue {
					filter = false
				}
			case "from":
				if senml.Time < rpm.From {
					filter = false
				}
			case "to":
				if senml.Time >= rpm.To {
					filter = false
				}
			}
			if !filter {
				break
			}
		}

		if filter {
			msgs = append(msgs, m)
		}
	}

	numOfMessages := uint64(len(msgs))

	if rpm.Offset >= numOfMessages {
		return readers.MessagesPage{}, nil
	}

	if rpm.Limit < 1 {
		return readers.MessagesPage{}, nil
	}

	end := rpm.Offset + rpm.Limit
	if end > numOfMessages {
		end = numOfMessages
	}

	return readers.MessagesPage{
		PageMetadata: rpm,
		Total:        uint64(len(msgs)),
		Messages:     msgs[rpm.Offset:end],
	}, nil
}

```

readers/cassandra/messages.go
```diff
@@ -110,10 +105,17 @@ func (cr cassandraRepository) ReadAll(chanID string, offset, limit uint64, query
 	return page, nil
 }
 
-func buildQuery(chanID string, offset, limit uint64, query map[string]string) (string, []interface{}) {
+func buildQuery(chanID string, rpm readers.PageMetadata) (string, []interface{}) {
```
mteodor, line 108: this f is nice

mteodor (dismissed):

readers/mocks/messages.go
manuio: I agree that it's complex , I'll try to simplify it but your example is wrong because you only consider filtered messages while we have to consider messages without any filtering.

manuio: Sorry my bad, your example was wrong but because of `v`, `vb`, `vd` and `vs` checks (you can verify in my next commit) but not because of non filtered messages.

mteodor (approved):

dborovcanin (approved):

PR #1335, opened 2021-01-21
NOISSUE - Fix Postgres writer transaction handling

dborovcanin:

### What does this do?
This pull request fixes the Postgres writer transaction error handling bug discovered [here](https://github.com/mainflux/mainflux/pull/1324#pullrequestreview-572404529).

There is no such issue.

There are no such changes. This only fixes problem with handling the error in the deferred function.

No.

No.

### DIFF
--- a/consumers/writers/postgres/consumer.go
+++ b/consumers/writers/postgres/consumer.go
@@ -51,7 +51,7 @@ func (pr postgresRepo) Consume(message interface{}) (err error) {
 	}
 }
 
-func (pr postgresRepo) saveSenml(messages interface{}) error {
+func (pr postgresRepo) saveSenml(messages interface{}) (err error) {
 	msgs, ok := messages.([]senml.Message)
 	if !ok {
 		return errSaveMessage


nmarcetic (approved):

consumers/writers/postgres/consumer.go
```diff
@@ -51,7 +51,7 @@ func (pr postgresRepo) Consume(message interface{}) (err error) {
 	}
 }
 
-func (pr postgresRepo) saveSenml(messages interface{}) error {
+func (pr postgresRepo) saveSenml(messages interface{}) (err error) {
```
drasko, line 54: Why do we need this named param, when `err` is not used in the function?

dborovcanin: Because the deferred function is using `err` var. Before we named it, the deferred function was using the wrong error value - the one from the `BeginTxx` method, rather than the one returned from `saveSenml`.

mteodor (approved):

drasko (changes requested):
LGTM

drasko (approved):

PR #1336, opened 2021-01-21
NOISSUE - Fix run script and compiler warnings

dborovcanin:

### What does this do
This pull request addresses:
- naming of [NATS](https://nats.io/ binary
- unnecessary type conversions in Logger
- bugs with casting `int64` to `string` in Twins service.

There is no such issue.

There are no breaking changes.

No, because no new features are added.

No, there is no cush functionality.


drasko (changes requested):
This PR includes notify service as well, I guess it needs to be merged after

manuio (approved):
LGTM

readers/mongodb/messages.go
```diff
@@ -82,9 +82,6 @@ func (repo mongoRepository) ReadAll(chanID string, offset, limit uint64, query m
 	if err != nil {
 		return readers.MessagesPage{}, errors.Wrap(errReadMessages, err)
 	}
-	if total < 0 {
```
nmarcetic, line 85: It's not addressed in PR, not sure if it's accidentally removed?

dborovcanin: It's removed because `total` is `uint` type, so this can never be true (in fact, that's why we use `uint` instead of `int`).

nmarcetic (approved):

drasko (approved):
LGTM

PR #1341, opened 2021-01-28
NOISSUE - Upgrade influxdb and postgres docker images

manuio:

### DIFF
--- a/docker/addons/bootstrap/docker-compose.yml
+++ b/docker/addons/bootstrap/docker-compose.yml
@@ -17,7 +17,7 @@ volumes:
 
 services:
   bootstrap-db:
-    image: postgres:10.2-alpine
+    image: postgres:10.8-alpine
     container_name: mainflux-bootstrap-db
     restart: on-failure
     environment:
--- a/docker/addons/influxdb-writer/docker-compose.yml
+++ b/docker/addons/influxdb-writer/docker-compose.yml
@@ -18,7 +18,7 @@ volumes:
 
 services:
   influxdb:
-    image: influxdb:1.7-alpine
+    image: influxdb:1.8.3-alpine
     container_name: mainflux-influxdb
     restart: on-failure
     environment:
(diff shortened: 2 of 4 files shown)


docker/addons/bootstrap/docker-compose.yml
```diff
@@ -17,7 +17,7 @@ volumes:
 
 services:
   bootstrap-db:
-    image: postgres:10.2-alpine
+    image: postgres:10.8-alpine
```
nmarcetic, line 20: There is version 13 available, should with a try with the latest or there is a specific reason to use 10.x?

docker/addons/influxdb-writer/docker-compose.yml
```diff
@@ -18,7 +18,7 @@ volumes:
 
 services:
   influxdb:
-    image: influxdb:1.6.4-alpine
+    image: influxdb:1.7-alpine
```
nmarcetic: Same there is a version 1.8.x-alpine.

docker/addons/bootstrap/docker-compose.yml
manuio: Because version 13 doesn't support DBs initialised with version 10. We can probably do it for the next release

nmarcetic: Really? Damn. So how we will move from version 10 in the future? IMHO, it's docker-compose meant for local development, not really a breaking change (we can note it in release).

manuio: We plan to upgrade but probably a better moment would be when we replace User email by User ID.

nmarcetic: Ok. Btw looks really easy to do [safe upgrade](https://medium.com/better-programming/how-to-upgrade-your-postgresql-version-using-docker-d1e81dbbbdf9) with few commands, for existing databases that want to keep data (even if its development setup).

nmarcetic (approved):

mteodor (approved):

PR #1343, opened 2021-01-28
MF-1342 - Use environment variables in docker-compose to use tagged version of image

### LINKED ISSUE #1342: Use environment variables in docker-compose to create composition aligned with released versions.
**ENHANCEMENT**
  
Use `.env` and tagged versions of mainflux images in `docker-compose`, add ENVAR for enabling starting docker composition with tagged version which will be aligned with release tag.
  
mteodor:
Resolves #1342


provision/README.md
```diff
@@ -22,30 +22,31 @@ default values.
...
+| MF_PROVISION_X509_PROVISIONING      | Should X509 client cert be provisioned            | false                            |
+| MF_PROVISION_BS_CONFIG_PROVISIONING | Should thing config be saved in Bootstrap service | true                             |
+| MF_PROVISION_BS_AUTO_WHITELIST      | Should thing be auto whitelisted                  | true                             |
+| MF_PROVISION_BS_CONTENT             | Bootstrap service configs content, JSON format    | {}                               |
+| MF_PROVISION_CERTS_CA               | Mainflux CA cert for generating certificates      | ""                               |
+| MF_PROVISION_CERTS_CA_KEY           | Mainflux CA cert private key                      | ""                               |
+| MF_PROVISION_CERTS_RSA_BITS         | Certificate RSA bits parameter                    | 4096                             |
+| MF_PROVISION_CERTS_HOURS_VALID      | Number of days that certificate is valid          | "2400h"                          |
+| MF_RELEASE_TAG                      | Docker image version to be deployed               | latest                |
```
manuio: The last one :)

manuio (changes requested):

users/README.md
```diff
@@ -99,7 +100,12 @@ make users
...
+MF_USERS_LOG_LEVEL=[Users log level] MF_USERS_DB_HOST=[Database host address] MF_USERS_DB_PORT=[Database host port] MF_USERS_DB_USER=[Database user] MF_USERS_DB_PASS=[Database password] \
```
manuio: Add a new line for each envar: 
```
MF_USERS_LOG_LEVEL=[Users log level] \
MF_USERS_DB_HOST=[Database host address] \
MF_USERS_DB_PORT=[Database host port] \
MF_USERS_DB_USER=[Database user] \
MF_USERS_DB_PASS=[Database password]
```

manuio (changes requested):

auth/README.md
```diff
@@ -56,6 +56,7 @@ default values.
 | MF_AUTH_SERVER_KEY        | Path to server key in pem format                                         |               |
 | MF_AUTH_SECRET            | String used for signing tokens                                           | auth          |
 | MF_JAEGER_URL             | Jaeger server URL                                                        | localhost:6831|
+| MF_RELEASE_TAG            | Docker image version to be deployed                                      | latest         |
```
blokovi: space in last column

auth/README.md
```diff
@@ -34,7 +34,7 @@ The following actions are supported:
 
 ## Configuration
 
-The service is configured using the environment variables presented in the
+The service is configured using the envir onment variables presented in the
```
manuio: `envir onment`

users/README.md
```diff
@@ -55,34 +55,34 @@ locally:
 version: "3.7"
 services:
   users:
-    image: mainflux/users:[MF_RELEASE_TAG]
+    image: mainflux/users:[\MF_RELEASE_TAG]
```
manuio: `\MF_RELEASE_TAG`

users/README.md
```diff
@@ -55,34 +55,34 @@ locally:
...
+      \MF_JAEGER_URL: [Jaeger server URL]
+      \MF_EMAIL_DRIVER: [Mail server driver smtp]
+      \MF_EMAIL_HOST: [\MF_EMAIL_HOST]
+      \MF_EMAIL_PORT: [\MF_EMAIL_PORT]
+      \MF_EMAIL_USERNAME: [\MF_EMAIL_USERNAME]
+      \MF_EMAIL_PASSWORD: [\MF_EMAIL_PASSWORD]
+      \MF_EMAIL_FROM_ADDRESS: [\MF_EMAIL_FROM_ADDRESS]
+      \MF_EMAIL_FROM_NAME: [\MF_EMAIL_FROM_NAME]
+      \MF_EMAIL_TEMPLATE: [\MF_EMAIL_TEMPLATE]
+      \MF_TOKEN_RESET_ENDPOINT: [\MF_TOKEN_RESET_ENDPOINT]
```
manuio: Looka like a wrong find a replace. Remove these `\`

users/README.md
```diff
@@ -100,15 +100,33 @@ make users
...
+MF_EMAIL_PASSWORD=[Mail server password] \
+MF_EMAIL_FROM_ADDRESS=[Email from address] \
+MF_EMAIL_FROM_NAME=[Email from name] \
+MF_EMAIL_TEMPLATE=[Email template file] \
+MF_TOKEN_RESET_ENDPOINT=[Password reset token endpoint] \
+$GOBIN/mainflux-users
 ```
 
-If `MF_EMAIL_TEMPLATE` doesn't point to any file service will function but password reset functionality will not work.
+If `\MF_EMAIL_TEMPLATE` doesn't point to any file service will function but password reset functionality will not work.
```
manuio: same here

manuio (changes requested):

docker/README.md
```diff
@@ -21,3 +21,4 @@ docker-compose -f docker/docker-compose.yml up
 docker-compose -f docker/addons/<path>/docker-compose.yml  up
 ```
 
+To start specific release you need to change the value of `MF_RELEASE_TAG` in `.env` before running these commands.
```
manuio: I'd use `To pull docker images from a specific release...`

auth/README.md
```diff
@@ -67,7 +68,7 @@ locally:
 version: "3.7"
 services:
   auth:
-    image: mainflux/auth:[version]
+    image: mainflux/auth:[MF_RELEASE_TAG]
```
nmarcetic: I don't get this, why it's wrapped with those brackets `[MF_RELEASE_TAG]`? I was expecting the same syntax as in Yaml `${MF_RELEASE_TAG}` or at least `<MF_RELEASE_TAG>` a syntax used for variables in doc.

mteodor: or maybe `[Docker image version tag specified by MF_RELEASE_TAG]`

nmarcetic: I would go with `${MF_RELEASE_TAG}` because its env var and exact syntax. But yea, it's probably better because those brackets are everywhere.

dborovcanin (dismissed):

auth/README.md
```diff
@@ -56,39 +56,13 @@ default values.
...
-      MF_AUTH_DB_SSL_KEY: [Path to the PEM encoded key file]
-      MF_AUTH_DB_SSL_ROOT_CERT: [Path to the PEM encoded root certificate file]
-      MF_AUTH_HTTP_PORT: [Service HTTP port]
-      MF_AUTH_GRPC_PORT: [Service gRPC port]
-      MF_AUTH_SECRET: [String used for signing tokens]
-      MF_AUTH_SERVER_CERT: [String path to server certificate in pem format]
-      MF_AUTH_SERVER_KEY: [String path to server key in pem format]
-      MF_JAEGER_URL: [Jaeger server URL]
-```
+The service itself is distributed as Docker container. Check the [`auth`](https://github.com/mainflux/mainflux/blob/master/docker/docker-compose.yml#L71-L94) service section in
```
manuio, line 62: I'm not sure about using section here. If we change the something in thee docker-compose this will be wrong.

manuio (dismissed):
LGTM

drasko (changes requested):
I am generally not convinced by this PR.

I do not like removing documentation about ENV vars from README files and adding just links to docker-compose. I know this would be easier to maintain, but on the other hand hinders readability and quick understanding of deployment.

I also do not support adding docker-compose tag as an ENV var, which can be misleading that it is a part of a service, somebody might even think of using it in the Kubernetes. It is not, it is used only in docker-compose, and docker-compose is used ONLY in development.

I do not really want to support people giving them the wrong idea that docker-compose can be used for running stable releases. If they want to do it, I would rather let them change version by hand. We keep `latest` for a reason - to demotivate people using docker-compose for production, where Kubernetes should be used.

dborovcanin:
> I do not like removing documentation about ENV vars from README files and adding just links to docker-compose. I know this would be easier to maintain, but on the other hand hinders readability and quick understanding of deployment.

I don't agree with this.
We did not remove the env var table with all the explanation, we just removed unnecessary and quite useless "docker-compose-like" sections that we need to maintain with no benefit whatsoever (it does not explain the meaning of the variables better than the table in README, nor it explains usage better than the real compose file).

> I also do not support adding docker-compose tag as an ENV var, which can be misleading that it is a part of a service, somebody might even think of using it in the Kubernetes. It is not, it is used only in docker-compose, and docker-compose is used ONLY in development.

I do agree with this.
This env var can be misinterpreted as a service-related configuration, rather than deployment-related stuff. That's both dangerous and wrong. However, I don't see how providing a version for docker-compose in any way implies that it's production-ready.

mteodor:
> I also do not support adding docker-compose tag as an ENV var, which can be misleading that it is a part of a service, somebody might even think of using it in the Kubernetes. It is not, it is used only in docker-compose, and docker-compose is used ONLY in development. 

I dont think there is a place for confusion it just requires a basic knowledge of docker and kubernetes, it is clear that environment variables used for services are in the `environment` section of docker-compose.

People with such unclear knowledge of docker deployment will more likely checkout release, start it and be surprised that something isn't working as expected (since the services being deployed are latest rather than tagged release version ) and then ask for support, open an issue...

mteodor:
> I do not really want to support people giving them the wrong idea that docker-compose can be used for running stable releases. If they want to do it, I would rather let them change version by hand. We keep `latest` for a reason - to demotivate people using docker-compose for production, where Kubernetes should be used.  <

Latest stays on master, this is just for the people that are experimenting with releases, when they checkout the release tag `.env` will be populated with appropriate version tag to start appropriate docker versions

nmarcetic:
> I am generally not convinced by this PR.
> 
> I do not like removing documentation about ENV vars from README files and adding just links to docker-compose. I know this would be easier to maintain, but on the other hand hinders readability and quick understanding of deployment.
> 
> I also do not support adding docker-compose tag as an ENV var, which can be misleading that it is a part of a service, somebody might even think of using it in the Kubernetes. It is not, it is used only in docker-compose, and docker-compose is used ONLY in development.
> 
> I do not really want to support people giving them the wrong idea that docker-compose can be used for running stable releases. If they want to do it, I would rather let them change version by hand. We keep `latest` for a reason - to demotivate people using docker-compose for production, where Kubernetes should be used.

I also disagree, as its already really well-argued by @dusanb94 and @mteodor [here](https://github.com/mainflux/mainflux/pull/1343#issuecomment-770450415) and [here](https://github.com/mainflux/mainflux/pull/1343#issuecomment-770666880) I don't have much to add, except that both root README and documentation should be updated with note or section which will say that docker-compose is meant and designed to be used in development environments, even if [Docker is aiming that can be used in production](https://docs.docker.com/compose/production/) and clustered with Swarm, We highly recommend Kubernetes for production.
I didn't saw any notice/recommendation in our documentation or readme, more like `Mainflux can be easily deployed on Kubernetes platform` than` Should be`.  People who still want to use it to run it in production (even if it's possible), ignoring doc, readme, and recommendations, should be ignored also. 

IMHO this PR is a nice improvement and it elimitantes development pain.

drasko:
OK, I see that majority is for this PR, and makes sense. I would just need one more thought to consider: do we leave `MF_RELEASE_TAG` in the ENV var table (in the README), or we move it in the sentence below, adding information that docker-compose can be manipulated with this?

This was we could remove ambiguity that this ENV var is related to what service really accepts (in the table) and what is used for  deployment via docker-compose.

@nmarcetic @dusanb94 @mteodor please give your thoughts on this.

nmarcetic:
> OK, I see that majority is for this PR, and makes sense. I would just need one more thought to consider: do we leave `MF_RELEASE_TAG` in the ENV var table (in the README), or we move it in the sentence below, adding information that docker-compose can be manipulated with this?
> 
> This was we could remove ambiguity that this ENV var is related to what service really accepts (in the table) and what is used for deployment via docker-compose.
> 
> @nmarcetic @dusanb94 @mteodor please give your thoughts on this.

Service actually does accept `MF_RELEASE_TAG` which is a docker image tag. I would put this NOTE in root README.md, the section that explains how to run compose file (and here I will also mention with a bolded label at the section bottom that docker-compose is meant and designed to be used for development and testing environments, and Kubernetes for production, Linkedin to K8s section if official doc). Same for an official doc, running the Mainflux section.
IMHO this way, everything is super clear. Don't spare words, mention it everywhere where you are mentioning Running Mainflux.

dborovcanin:
@drasko @nmarcetic I think that we should remove this env var from the service README because it's explained elsewhere and has nothing to do with the service configuration. Root README sounds like a good candidate.

drasko:
OK, this makes sense - we can put it in the root README, under the [Install](https://github.com/mainflux/mainflux#install) section.

I am not sure if we need as an additional note in microservice READMEs (not in the table, but in a separate sentece). I think not, it is redundant - people will usually deploy looking at the Install section of central README (or by looking in the docker-compose and figuring out). I let @mteodor decide on this.

drasko:
On a second though - better just put it in the root README, because this is Mainflux-wide setting, not per-service setting.

nmarcetic (approved):

manuio (approved):
LGTM

drasko (approved):
LGTM

PR #1353, opened 2021-02-04
MF-1314 - Add value comparison filters for readers

### LINKED ISSUE #1314: Consider adding additional filters by SenML `value`
**ENHANCEMENT**
   - tests
   - code refactor
   - documentation
   - research
   - tooling

https://github.com/mainflux/mainflux/pull/1312 introduced filtering by SenML values.

It would probably be interesting to add a few more filters around `value`:
- Less than (`<some_value>`)
- Higher than (`<some_value>`)
- In range (less than `<value_max>` but higher than `<value_min>`)
- Not in range (less than `<value_min>` but higher than `<value_max>`)

Nice-to-have

manuio:

Resolves #1314


readers/messages.go
```diff
@@ -28,17 +28,40 @@ type MessagesPage struct {
...
+
+// ParseValueComparison convert comparison types to the mathematic anotation
+func ParseValueComparison(query map[string]interface{}) string {
+	comparison := "="
+	val, ok := query["comparison"]
+	if ok {
+		switch val.(string) {
+		case "equal":
+			comparison = "="
+		case "lower-than":
```
drasko: One question - do we need `lower-than`, or `lower` and `lower-equal` (or maybe `lower-or-equal`) would work?

@dusanb94 @nmarcetic @mteodor @darkodraskovic opinions please?

drasko (changes requested):

readers/messages.go
dborovcanin: Or maybe `lt`, `gt`, `le`, `ge`. While it sounds like abbreviations, it is common terminology.

drasko: You can borrow the naming from assembly conditional instructions: https://developer.arm.com/documentation/dui0068/b/arm-instruction-reference/conditional-execution

readers/cassandra/messages_test.go
```diff
@@ -215,7 +215,72 @@ func TestReadSenml(t *testing.T) {
...
+				Messages: fromSenml(valueMsgs[0:limit]),
+			},
+		},
+		"read message with value and lower-or-equal-than comparison": {
+			chanID: chanID,
+			pageMeta: readers.PageMetadata{
+				Offset:     0,
+				Limit:      limit,
+				Value:      v + 1,
+				Comparison: "lte",
```
drasko: I think we mentioned `le` instead of `lte`, according to ARM assembly spec.

@dusanb94 @nmarcetic opinions?

drasko (changes requested):

readers/openapi.yml
```diff
@@ -174,6 +174,20 @@ components:
...
+      name: comparison
+      description: Value comparison operator.
+      in: query
+      schema:
+        type: string
+        default: equal
+        enum:
+          - eq
+          - lt
+          - lte
```
dborovcanin: It's `le` and `ge`.

dborovcanin (changes requested):

readers/openapi.yml
manuio: good catch, thank you!

dborovcanin (dismissed):

dborovcanin (approved):

drasko (approved):
LGTM

PR #1355, opened 2021-02-04
MF-1317 - Configurable regexp rule for password

### LINKED ISSUE #1317: Minimal password length limitation
**ENHANCEMENT**

Password validation currently includes limitation of password to minimally 8 characters: https://github.com/mainflux/mainflux/search?q=minPassLen

This decision looks somewhat arbitrary, and it is certainly not documented or stressed enough, so it can lead to errors that are not intuitive to the user.

Password validation should avoid arbitrary limitation (for example 8 chars - why 8?), and be:
- Generic
- Configurable

This can eventually be achieved via user-configurable regexp in an ENV var that will be defined, per deployment, in a deploy-time.

If this turns out to be too complicated, consider removing limitation or at least document it well.

Should-have

blokovi:

Resolves #1317

### DIFF
--- a/cmd/users/main.go
+++ b/cmd/users/main.go
@@ -111,8 +111,6 @@ const (
 	envAuthTimeout = "MF_AUTH_GRPC_TIMEOUT"
 )
 
-// var passRegex = regexp.MustCompile(mainflux.Env(envPassRegex, defPassRegex))
-
 type config struct {
 	logLevel      string
 	dbConfig      postgres.Config
@@ -345,11 +343,9 @@ func startHTTPServer(tracer opentracing.Tracer, svc users.Service, port string, 
 	p := fmt.Sprintf(":%s", port)
 	if certFile != "" || keyFile != "" {
 		logger.Info(fmt.Sprintf("Users service started using https, cert %s key %s, exposed port %s", certFile, keyFile, port))
-		// errs <- http.ListenAndServeTLS(p, certFile, keyFile, api.MakeHandler(svc, tracer, passRegex))
 		errs <- http.ListenAndServeTLS(p, certFile, keyFile, api.MakeHandler(svc, tracer))
 	} else {
 		logger.Info(fmt.Sprintf("Users service started using http, exposed port %s", port))
-		// errs <- http.ListenAndServe(p, api.MakeHandler(svc, tracer, passRegex))
 		errs <- http.ListenAndServe(p, api.MakeHandler(svc, tracer))
 	}
 }
--- a/cmd/users/main.go
+++ b/cmd/users/main.go
@@ -111,6 +111,8 @@ const (
 	envAuthTimeout = "MF_AUTH_GRPC_TIMEOUT"
 )
 
+var passRegexVar = regexp.MustCompile(mainflux.Env(envPassRegex, defPassRegex))
+
 type config struct {
 	logLevel      string
 	dbConfig      postgres.Config
@@ -343,9 +345,9 @@ func startHTTPServer(tracer opentracing.Tracer, svc users.Service, port string, 
 	p := fmt.Sprintf(":%s", port)
 	if certFile != "" || keyFile != "" {
 		logger.Info(fmt.Sprintf("Users service started using https, cert %s key %s, exposed port %s", certFile, keyFile, port))
-		errs <- http.ListenAndServeTLS(p, certFile, keyFile, api.MakeHandler(svc, tracer))
+		errs <- http.ListenAndServeTLS(p, certFile, keyFile, api.MakeHandler(svc, tracer, passRegexVar))
 	} else {
 		logger.Info(fmt.Sprintf("Users service started using http, exposed port %s", port))
-		errs <- http.ListenAndServe(p, api.MakeHandler(svc, tracer))
+		errs <- http.ListenAndServe(p, api.MakeHandler(svc, tracer, passRegexVar))
 	}
 }
--- a/cmd/users/main.go
+++ b/cmd/users/main.go
@@ -128,7 +128,6 @@ type config struct {
 	authTimeout   time.Duration
 	adminEmail    string
 	adminPassword string
-	// passRegex     *regexp.Regexp
 }
 
 func main() {
@@ -181,11 +180,6 @@ func loadConfig() config {
 		log.Fatalf("Invalid value passed for %s\n", envAuthTLS)
 	}
 
-	// passRegex, err := regexp.Compile(mainflux.Env(envPassRegex, defPassRegex))
-	// if err != nil {
-	// 	log.Fatalf("Invalid value passed for %s\n", envPassRegex)
-	// }
-
 	dbConfig := postgres.Config{
 		Host:        mainflux.Env(envDBHost, defDBHost),
 		Port:        mainflux.Env(envDBPort, defDBPort),
@@ -224,7 +218,6 @@ func loadConfig() config {
 		authTimeout:   authTimeout,
 		adminEmail:    mainflux.Env(envAdminEmail, defAdminEmail),
 		adminPassword: mainflux.Env(envAdminPassword, defAdminPassword),
-		// passRegex:     passRegex,
 	}
 
 }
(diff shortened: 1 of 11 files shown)


users/users_test.go
```diff
@@ -72,13 +72,6 @@ func TestValidate(t *testing.T) {
 			},
 			err: users.ErrMalformedEntity,
 		},
-		"validate user with empty password": {
-			user: users.User{
-				Email:    email,
-				Password: "",
-			},
-			err: users.ErrMalformedEntity,
-		},
```
manuio, line 81: There is a reason to remove this test?

users/api/requests.go
```diff
@@ -17,6 +16,9 @@ type userReq struct {
 }
 
 func (req userReq) validate() error {
+	if !passRegex.MatchString(req.user.Password) {
+		return users.ErrMalformedEntity
```
mteodor: maybe the error should state exactly what the problem is `password doesn't match policy`, something like that

users/users.go
```diff
@@ -47,10 +46,6 @@ func (u User) Validate() error {
 		return ErrMalformedEntity
 	}
 
-	if len(u.Password) < minPassLen {
```
mteodor, line 50: why we dont check password here?

blokovi: two reasons:
- we already check it on other places (validating requests for registering and password changing)
- can't import passRegex from api package here

manuio: @blokovi I would check it here as it was done before and simply do: 
```
func (req userReq) validate() error {
	return req.user.Validate()
}
```

manuio (changes requested):

drasko (changes requested):

cmd/users/main.go
```diff
@@ -108,6 +111,8 @@ const (
 	envAuthTimeout = "MF_AUTH_GRPC_TIMEOUT"
 )
 
+var passRegex = regexp.MustCompile(mainflux.Env(envPassRegex, defPassRegex))
```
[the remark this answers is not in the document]
blokovi: I have used `MustCompile` because I saw it's used on other places:
https://github.com/mainflux/mainflux/blob/master/users/users.go#L26-L28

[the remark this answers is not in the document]
manuio: @blokovi I would use `Compile` instead of `MustCompile` and check the errors:  https://golang.org/pkg/regexp/#Compile 
You can maybe also fix it in users service in this PR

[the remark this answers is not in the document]
manuio: I'd do something like that

```
logger, err := logger.New(os.Stdout, cfg.logLevel)
if err != nil {
     log.Fatalf(err.Error())
}

passRegex, err := regexp.Compile(mainflux.Env(envPassRegex, defPassRegex))
if err != nil {
     passRegex, err := regexp.MustCompile(defPassRegex)
}
```

users/users.go
blokovi: @manuio  can't use `passRegex` from api package here in users (because api imports users)

manuio: @blokovi first remove it from the api layer :)

cmd/users/main.go
```diff
@@ -111,7 +111,7 @@ const (
 	envAuthTimeout = "MF_AUTH_GRPC_TIMEOUT"
 )
 
-var passRegex = regexp.MustCompile(mainflux.Env(envPassRegex, defPassRegex))
+// var passRegex = regexp.MustCompile(mainflux.Env(envPassRegex, defPassRegex))
```
manuio: Remove this please

users/service.go
```diff
@@ -164,17 +170,20 @@ type usersService struct {
...
+func New(users UserRepository, groups GroupRepository, hasher Hasher, auth mainflux.AuthServiceClient, m Emailer, idp mainflux.IDProvider, passRegex *regexp.Regexp) Service {
+	PassRegex = passRegex
 	return &usersService{
 		users:      users,
 		groups:     groups,
 		hasher:     hasher,
 		auth:       auth,
 		email:      m,
 		idProvider: idp,
+		// passRegex:  passRegex,
```
manuio: Uncomment and use this one

users/service.go
```diff
@@ -164,17 +170,20 @@ type usersService struct {
 	email      Emailer
 	auth       mainflux.AuthServiceClient
 	idProvider mainflux.IDProvider
+	// passRegex  *regexp.Regexp
 }
 
 // New instantiates the users service implementation
-func New(users UserRepository, groups GroupRepository, hasher Hasher, auth mainflux.AuthServiceClient, m Emailer, idp mainflux.IDProvider) Service {
+func New(users UserRepository, groups GroupRepository, hasher Hasher, auth mainflux.AuthServiceClient, m Emailer, idp mainflux.IDProvider, passRegex *regexp.Regexp) Service {
+	PassRegex = passRegex
```
manuio: Remove this one

users/api/requests.go
```diff
@@ -16,10 +16,17 @@ type userReq struct {
...
+	// 	return users.ErrPasswordPolicy
+	// }
 	return req.user.Validate()
+	// if err := req.user.Validate(); err != nil {
+	// 	return err
+	// }
+	// if !passRegex.MatchString(req.user.Password) {
+	// 	return users.ErrPasswordPolicy
+	// }
+	// return nil
```
manuio: Remove this comments please

manuio (changes requested):

cmd/users/main.go
```diff
@@ -175,6 +179,11 @@ func loadConfig() config {
 		log.Fatalf("Invalid value passed for %s\n", envAuthTLS)
 	}
 
+	passRegex, err := regexp.Compile(mainflux.Env(envPassRegex, defPassRegex))
+	if err != nil {
+		passRegex = regexp.MustCompile(defPassRegex)
+	}
```
dborovcanin: No need to fail silently. We can do it similar to the other env var handling:
```go
passRegex, err := regexp.Compile(mainflux.Env(envPassRegex, defPassRegex))
if err != nil {
    log.Fatalf("Invalid password validation rules %s\n", envPassRegex)
 }
 ```

File: users/service.go [hunk 10]
```diff
@@ -66,6 +66,9 @@ var (
 
 	// ErrAssignUserToGroup indicates an error in assigning user to a group.
 	ErrAssignUserToGroup = errors.New("failed assigning user to a group")
+
+	// ErrPasswordPolicy indicates weak password.
+	ErrPasswordPolicy = errors.New("password doesn't match policy")
```
dborovcanin: Using `Policy` terminology can be confusing; maybe `ErrPasswordFormat`.

users/service.go
```diff
@@ -302,6 +310,9 @@ func (svc usersService) ChangePassword(ctx context.Context, authToken, password,
 	if err != nil {
 		return errors.Wrap(ErrUnauthorizedAccess, err)
 	}
+	if !svc.passRegex.MatchString(password) {
```
dborovcanin, line 316: Should we validate password format in the `ResetPassword` metod?

dborovcanin (changes requested):
While I see why we added password regex as a field to the Users service, I find it a bit cleaner to use a package-level variable in the `api` package because we offload data validation to the transport layer (that's what we usually do). While package-level variables are generally not a good idea, this one is 1) safe for concurrent use and 2) set only on call of `MakeHandler` and never changed (i.e. does not hold the state and is easy to test and mock).

users/service.go
manuio: Definitely

users/users_test.go
blokovi: Removed because password validation is removed from users/api/requests.go

manuio (dismissed):
LGTM

mteodor (approved):

mteodor (dismissed):

dborovcanin (dismissed):

manuio (dismissed):
LGTM

mteodor (dismissed):

drasko (approved):
LGTM

Issue #963, opened 2019-11-24
Consider replacing Bone mux with Chi


drasko:
**FEATURE REQUEST**

1. Is there an open issue addressing this request? If it does, please add a "+1" reaction to the
   existing issue, otherwise proceed to step 2.

2. Describe the feature you are requesting, as well as the possible use case(s) for it.
Currently Mainflux HTTP and WS use [Bone mux](https://github.com/go-zoo/bone).

New mux [Chi](https://github.com/go-chi/chi) that appeared in the meanwhile gives more modern impression and should be evaluated as a Bone replacement.

3. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).
nice-to-have

nmarcetic:
Chi looks really interesting, light ( stdlib only), and performant. Yet I would mark this as a low priority, cause it does not bring real benefits (its control plane API, not performance-critical, bone served us quite well, except now with #1357 which will be fixed with #1358 and the issue will be opened (@mteodo you should probably [raise the issue](https://github.com/go-zoo/bone/issues) ).

felixgateru:
This was addressed by this pr: https://github.com/absmach/magistrala-old/pull/47

PR #1358, opened 2021-02-07
MF-1357 - Metadata in URL query not being properly parsed

### LINKED ISSUE #1357: Metadata in URL query not being properly parsed
**BUG REPORT**

Search things/users by metadata

List of things that match the metadata query

```
curl -i  -s -X GET http://localhost:8182/things\?name\=test\&metadata\=\{\"role\":\"manager\"\} -H "Authorization: $TOK" -H 'Content-Type: application/json'

HTTP/1.1 400 Bad Request
Date: Fri, 05 Feb 2021 20:04:15 GMT
Content-Length: 33

{"error":"invalid query params"}
```

dborovcanin:

### What does this do?
This pull request fixes parsing of the metadata query parameter in list requests.

This pull request fixes #1357.

There are no such changes.

No.

No, now it works as expected and as documented.

### DIFF
--- a/internal/groups/api/transport.go
+++ b/internal/groups/api/transport.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
-	"net/url"
 	"strconv"
 	"strings"
 
@@ -191,19 +190,12 @@ func readStringQuery(r *http.Request, key string) (string, error) {
 }
 
 func readMetadataQuery(r *http.Request, key string) (map[string]interface{}, error) {
-	q, err := url.ParseQuery(r.URL.RawQuery)
-	if err != nil {
-		return nil, errInvalidQueryParams
-	}
-	vals, ok := q[key]
-	if !ok {
+	md := r.URL.Query().Get(key)
+	if md == "" {
 		return nil, nil
 	}
-	if len(vals) > 1 {
-		return nil, errInvalidQueryParams
-	}
 	m := make(map[string]interface{})
-	if err := json.Unmarshal([]byte(vals[0]), &m); err != nil {
+	if err := json.Unmarshal([]byte(md), &m); err != nil {
 		return nil, errors.Wrap(errInvalidQueryParams, err)
 	}
 	return m, nil
--- a/internal/groups/api/transport.go
+++ b/internal/groups/api/transport.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


internal/groups/api/transport.go
```diff
@@ -190,21 +191,21 @@ func readStringQuery(r *http.Request, key string) (string, error) {
 }
 
 func readMetadataQuery(r *http.Request, key string) (map[string]interface{}, error) {
-	vals := bone.GetQuery(r, key)
-	if len(vals) > 1 {
+	q, err := url.ParseQuery(r.URL.RawQuery)
```
drasko: What seems to be a problem with `bone`'s query and why are we switching to Go's stdlib `url`?

drasko (changes requested):

internal/groups/api/transport.go
dborovcanin: As @mteodor explained [here](https://github.com/mainflux/mainflux/issues/1357#issuecomment-774448180), Bone is not aware that the provided value is JSON and it's splitting the underlying string by `,` creating an array ([here is the code](https://github.com/go-zoo/bone/blob/31c3a0bb520c6d7a63dbb942459a3067787a975e/helper.go#L148)) instead of returningng a single string that contains JSON query.

drasko: Before we go and use Go stdlib, I would like that we consider 2 things:
1. Open an issue to Bone
2. Examine how [Chi](https://github.com/go-chi/chi) is handling this, and maybe implement https://github.com/mainflux/mainflux/issues/963 first

dborovcanin: 1. I would say no since this is by design (and will potentially cause a lot of code running Bone to break).
2. We should examine this multiplexer (it looks promising) but that it's a separate issue while this PR is only a fix for the reported bug.

mteodor (dismissed):

internal/groups/api/transport.go
nmarcetic: Agree with @dusanb94 , this PR resolves the current bug. Let's fix it ASAP, by merging a fix. Then consider #963 which is a big change.

nmarcetic (approved):

manuio (approved):
LGTM

drasko (changes requested):
This should be solved with `POST` on new API endpoint (for example `/search`). @blokovi will take this implementation.

This is needed because we can notput JSON in the URL query params, not all JSON characters are supported in URL query. Real solution will be to put JSON in the body of the request, and this is wy we unfortunately need a separate dedicated endpoint for this.

At this moment I would avoid merging this PR, because it removes `bone` router commands and replaces them with underlying Go stdlib functions. This is why I will close this PR and open a dedicated issue.

drasko:
Closing this one as explained. We'll use the same https://github.com/mainflux/mainflux/issues/1357 for the new PR that will implement new endpoint and JSON in a body.

PR #1359, opened 2021-02-08
MF-1317

### LINKED ISSUE #1317: Minimal password length limitation

kasraabdollahi:
### What does this do?
It checks if user has defined a pattern for passwords with an environment var (envpassregex)
If this environment var is set, passwords should match with it
If this is not set, for backward compatibility a default regex will be used
This default regex only checks that password length be greater than or equal to 8 chars

Resolves #1317

It doesn't break any functionality and it is completely backward compatible

### DIFF
--- a/users/api/requests.go
+++ b/users/api/requests.go
@@ -4,12 +4,16 @@
 package api
 
 import (
+	"os"
+	"regexp"
+
 	"github.com/mainflux/mainflux/users"
 )
 
 const (
-	minPassLen  = 8
-	maxNameSize = 1024
+	minPassLen       = 8
+	maxNameSize      = 1024
+	defaultPassRegex = "^.{8,}$"
 )
 
 type userReq struct {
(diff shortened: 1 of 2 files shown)


manuio:
@kasraabdollahi There was a PR already open to resolve this issue: https://github.com/mainflux/mainflux/pull/1355
Apart from this, any commit must be signed to pass the DCO. Also a title would be welcome. 

I propose you to follow and help @blokovi in his PR and to close this one.

PR #1383, opened 2021-03-05
MF-1357 - Add new endpoint for searching things

### LINKED ISSUE #1357: Metadata in URL query not being properly parsed

blokovi:

### What does this do?
Add /things/search POST endpoint with filter parameters for name and metadata in json body

Resolves #1357 

Yes


things/api/things/http/transport.go
```diff
@@ -105,6 +105,13 @@ func MakeHandler(tracer opentracing.Tracer, svc things.Service) http.Handler {
 		opts...,
 	))
 
+	r.Post("/things/list", kithttp.NewServer(
```
nmarcetic: Can we add a search here `/things/search`? Like search is a resource itself. Then  `POST` makes sense because you are creating a new search resource with a req body (containing search criteria).

things/api/things/http/requests.go
```diff
@@ -180,6 +180,13 @@ type listResourcesReq struct {
 	pageMetadata things.PageMetadata
 }
 
+type listResourcesMetaReq struct {
+	token        string
```
nmarcetic: Which token is this? 
User token is in req header, thing doesn't have a token, it has key (in domain language we don't call it token).

nmarcetic (changes requested):

things/openapi.yml
```diff
@@ -62,6 +62,38 @@ paths:
...
+        '200':
+          $ref: "#/components/responses/ThingsPageRes"
+        '400':
+          description: Failed due to malformed query parameters.
+        '401':
+          description: Missing or invalid access token provided.
+        '404':
+          description: A non-existent entity request.
+        '422':
+          description: Database can't process request.
```
nmarcetic: I would prefer just ` Unprocessable Entity` here a "standard" `422` status message, if you can't say more (like which field etc...). Without mentioning Database...

things/service_test.go
```diff
@@ -711,13 +711,13 @@ func TestListChannels(t *testing.T) {
...
 		},
-		"list with zero limit": {
+		"list with zero limit and offset 1": {
 			token: token,
 			pageMetadata: things.PageMetadata{
 				Offset: 1,
 				Limit:  0,
 			},
-			size: 0,
+			size: 9,
```
manuio: Use something like `n - 1` here

things/redis/streams_test.go
```diff
@@ -72,7 +72,7 @@ func TestCreateThings(t *testing.T) {
 			key: token,
 			err: nil,
 			event: map[string]interface{}{
-				"id":        "1",
+				"id":        "001",
```
manuio, line 75: We should try to use real IDs

things/openapi.yml
```diff
@@ -569,6 +597,30 @@ components:
...
+          description: Number of items to skip during retrieval.
+        limit:
+          type: integer
+          description: Maximum number of items to return in one page.
+        order:
+          type: string
+          description: Order type.
+        dir:
+          type: string
+          description: Order direction.
```
manuio: This needs to be described better.
Follow this existent example and adapt it to use params from the body:
```
    Limit:
      name: limit
      description: Size of the subset to retrieve.
      in: query
      schema:
        type: integer
        default: 10
        maximum: 100
        minimum: 1
      required: false
    Offset:
      name: offset
      description: Number of items to skip during retrieval.
      in: query
      schema:
        type: integer
        default: 0
        minimum: 0
      required: false
    Connected:
      name: connected
      description: Connection state of the subset to retrieve.
      in: query
      schema:
        type: boolean
        default: true
      required: false
    Name:
      name: name
      description: Name filter. Filtering is performed as a case-insensitive partial match.
      in: query
      schema:
        type: string
      required: false
    Order:
      name: order
      description: Order type.
      in: query
      schema:
        type: string
        default: id
        enum:
          - name
          - id
      required: false
    Direction:
      name: dir
      description: Order direction.
      in: query
      schema:
        type: string
        default: desc
        enum:
          - asc
          - desc
      required: false
    Metadata:
      name: metadata
      description: Metadata filter. Filtering is performed matching the parameter with metadata on top level. Parameter is json.
      in: query
      required: false
      schema:
        type: object
        additionalProperties: {}
 ```

File: things/mocks/things.go [hunk 7]
```diff
@@ -115,7 +115,7 @@ func (trm *thingRepositoryMock) RetrieveAll(_ context.Context, owner string, pm
 	trm.mu.Lock()
 	defer trm.mu.Unlock()
 
-	if pm.Limit <= 0 {
+	if pm.Limit < 0 {
```
manuio, line 118: Any reason to do this?

things/mocks/channels.go
```diff
@@ -78,30 +78,40 @@ func (crm *channelRepositoryMock) RetrieveByID(_ context.Context, owner, id stri
 }
 
 func (crm *channelRepositoryMock) RetrieveAll(_ context.Context, owner string, pm things.PageMetadata) (things.ChannelsPage, error) {
-	if pm.Limit <= 0 {
+	if pm.Limit < 0 {
```
manuio, line 81: It's 0 limit valid?

things/api/things/http/transport.go
```diff
@@ -105,6 +105,13 @@ func MakeHandler(tracer opentracing.Tracer, svc things.Service) http.Handler {
 		opts...,
 	))
 
+	r.Post("/things/search", kithttp.NewServer(
+		kitot.TraceServer(tracer, "list_things")(listThingsEndpoint(svc)),
```
manuio: `search_things`

pkg/sdk/go/things_test.go
```diff
@@ -36,7 +35,7 @@ const (
 var (
 	metadata   = map[string]interface{}{"meta": "data"}
 	metadata2  = map[string]interface{}{"meta": "data2"}
-	thing      = sdk.Thing{ID: "1", Name: "test_device", Metadata: metadata}
+	thing      = sdk.Thing{ID: "001", Name: "test_device", Metadata: metadata}
```
manuio, line 38: Try to use UUID

pkg/sdk/go/channels_test.go
```diff
@@ -16,7 +15,7 @@ import (
 )
 
 var (
-	channel      = sdk.Channel{ID: "1", Name: "test"}
+	channel      = sdk.Channel{ID: "001", Name: "test"}
```
manuio, line 18: Can we use UUID?

manuio (changes requested):

pkg/sdk/go/things_test.go
blokovi: Point is that here we mock id so we can test "order by id" functionalities.
That is the reason why I have change mock IDs to use three chars instead one, because ordering with just one char, like it is now, is `1`, `10`, `100`, `2`, `20`...
With this change order would be `001`, `002`, `003`... and you can write tests that check order by id functionality.

things/mocks/channels.go
blokovi: yes, limit 0 would not throw error, and in that case you will get default limit (10)

nmarcetic (dismissed):

things/mocks/things.go
blokovi: Reason is to be consistent with all other endpoints. Limit and other page parameters (offset, order, dir) are everywhere optional parameters, and if we want to keep them optional here we need default value if they are missing. When missing in request body, limit is zero (`omitempty` option) so we need to keep not throwing error in that case, that is why we change validation to allow it zero and use default value in that case. Same like you omit `limit` in all other endpoints, it's not `0` if you omit it, it's default `10`

nmarcetic (dismissed):

things/api/things/http/endpoint_test.go
```diff
@@ -49,6 +49,10 @@ var (
 	invalidName = strings.Repeat("m", maxNameSize+1)
 	notFoundRes = toJSON(errorRes{things.ErrNotFound.Error()})
 	unauthRes   = toJSON(errorRes{things.ErrUnauthorizedAccess.Error()})
+	searchThReq = things.PageMetadata{
```
mteodor: `searchThingReq`

manuio (approved):
LGTM

mteodor (approved):

PR #1368, opened 2021-02-22
NOISSUE - Fix certs and vault deployment, reorganize and remove unnecessary vars

mteodor:
### What does this do?
Reorganize cert enviroment variables and correct scripts for initializing vault PKI

Issue #1263, opened 2020-10-23
Move repeating errors to the separate package


dborovcanin:
**FEATURE REQUEST**

1. Is there an open issue addressing this request?
No.

2. Describe the feature you are requesting, as well as the possible use case(s) for it.
Consider moving errors that repeat such as Unauthorized, MalformedEntity, and Conflict to the `errors ` package so that we have it in one place, instead of repeating the code for each service.

3. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).
This is a nice-to-have.

_Originally posted by @mteodor in https://github.com/mainflux/mainflux/pull/1262#discussion_r510772637_

PR #1384, opened 2021-03-05
MF-1368 - Add internal http api package for query params reading 

manuio:

What does this do?

Resolves #1368 
Involuntarily starts #1263

List any changes that modify/break current functionality
N/A

Have you included tests for your changes?
N/A

Did you document any new/modified functionality?
N/A

Notes
N/A


blokovi (dismissed):

internal/api/http/http.go
```diff
@@ -0,0 +1,107 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package errors
```
dborovcanin: This seems like the wrong package name.

internal/errors/errors.go
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package http
```
dborovcanin: Same here.

provision/api/requests.go
```diff
@@ -1,5 +1,9 @@
 package api
 
+import (
```
dborovcanin: Use single import in format `import alias "pkg"`.

dborovcanin (changes requested):

internal/api/http/http.go
manuio: right :) good catch

certs/api/transport.go
```diff
@@ -8,14 +8,14 @@ import (
...
-	"strconv"
-
-	"github.com/mainflux/mainflux/certs"
-	"github.com/mainflux/mainflux/pkg/errors"
 
 	kithttp "github.com/go-kit/kit/transport/http"
 	"github.com/go-zoo/bone"
 	"github.com/mainflux/mainflux"
+	"github.com/mainflux/mainflux/certs"
+	intapihttp "github.com/mainflux/mainflux/internal/api/http"
```
nmarcetic: I like the prefix `internal` it's so obvious and I don't think it's too long (would rather sacrifice few more charts for readability). I would keep it everywhere and remove middle folder names like `api`. I think it's not needed `internalhttp` would be ok, even better `internalHTTP` or `internalErr` so it's super easy to spot.
Also regarding structure, I would remove `api` and keep `internal/http` only not sure why we have it in service at all... Anyway for an internal package, I would consider removing `api`.

bootstrap/api/transport.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/go-zoo/bone"
 	"github.com/mainflux/mainflux"
 	"github.com/mainflux/mainflux/bootstrap"
+	internalerr "github.com/mainflux/mainflux/internal/errors"
```
nmarcetic: I would rather move errors in `pkg` so people can import it and use it on their Mainflux Apps (Apps on top of Mainflux, compere/check errors, etc...). It's something that I see as usable for sharing, not for internal usage only.

nmarcetic (changes requested):

certs/api/transport.go
manuio: done!

bootstrap/api/transport.go
manuio: @drasko @nmarcetic I'm not sure about this one. We have our mainflux/errors package there and the alias wouldn't be `internalerr` but probably `pkgerr`. Any suggestion?

nmarcetic: Yea IMHO `pkgErr` sounds good. Why don't you move it there then? in `pkg/errros` put a new file like `pkt/errors/types.go` or something like that so those errors that you are sharing will be part of our `pkg/errors` package. 
Even if you leave everything as you initially did, in `internal/errors` IMHO it doesn't make sense if we already have shared `pkg/errors` , merge it. What do you think @manuio  ?

nmarcetic (dismissed):

internal/http/http.go
```diff
@@ -0,0 +1,106 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package http
```
dborovcanin: Consider renaming to `httputil`. It just sounds like a better name for this package (also, no need for import aliases).

pkg/errors/types.go
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package errors
```
dborovcanin, line 4: Consider moving HTTP errors to `pkg/http/errors.go`

dborovcanin (changes requested):

pkg/errors/types.go
manuio: This package is supposed to contain all types of errors not only http

dborovcanin (dismissed):

blokovi (dismissed):

pkg/errors/types.go
```diff
@@ -0,0 +1,15 @@
...
+
+var (
+	// ErrUnsupportedContentType indicates unacceptable or lack of Content-Type
+	ErrUnsupportedContentType = New("unsupported content type")
```
mteodor, line 8: don't we usually add blank line after error declaration

mteodor (dismissed):

nmarcetic (dismissed):

internal/httputil/query.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	return m, nil
+}
+
+// ReadBoolQuery reads boolean query parameters in a given http request
+func ReadBoolQuery(r *http.Request, key string) (bool, error) {
+	vals := bone.GetQuery(r, key)
+	if len(vals) == 0 {
+		return true, errors.ErrNotInQuery
```
dborovcanin: Why `true` if not in the query? The default for bool is `false`.

internal/httputil/query.go
```diff
@@ -0,0 +1,106 @@
...
+
+// ReadFloatQuery reads the value of float64 http query parameters for a given key
+func ReadFloatQuery(r *http.Request, key string) (float64, error) {
+	vals := bone.GetQuery(r, key)
+	if len(vals) > 1 {
+		return 0, errors.ErrInvalidQueryParams
+	}
+
+	if len(vals) == 0 {
+		return 0, nil
```
dborovcanin: Same here, why not: `return 0, errors.NotInQuery`? The util methods should be self-contained, and errors need to be handled elsewhere. Of, simpler: do we return the error in case the query does not contain the key? And keep it consistent. Anyways, in case of an error, we need to return the default value.

dborovcanin (changes requested):

dborovcanin (dismissed):

nmarcetic (dismissed):

consumers/notifiers/api/transport.go
```diff
@@ -21,7 +21,11 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
-const contentType = "application/json"
+const (
+	contentType = "application/json"
+
```
dborovcanin: No need for an empty line. Please extract `offset` and `limit` keys to constants, it's like that in the rest of the codebase.

readers/api/transport.go
```diff
@@ -23,10 +23,11 @@ import (
 
 const (
 	contentType = "application/json"
-	defLimit    = 10
-	defOffset   = 0
-	format      = "format"
-	defFormat   = "messages"
+
```
dborovcanin: There is no need for empty lines.

dborovcanin (approved):

blokovi (approved):

PR #1414, opened 2021-05-04
MF 1413 - Use per-service URL in SDK

### LINKED ISSUE #1413: Use per-service URL in SDK
**ENHANCEMENT**
Currently SDK uses `BaseURL` field, as it was supposed that will be used by a client that will be hitting Mainflux in the cloud, behind the proxy and with the single URL.

However, this prevents SDK being used by microservices internally, in the private Docker network, as then need to communicate in the service mesh fashion, so each microservice must be capable to address any other - we need SDK not to keep only `BaseURL`, but URLs of each microservice.

Must have

drasko:

### What does this do?
Fixes SDK to use per-service URL instead of `BaseURL`.

Resolves #1413


cmd/cli/main.go
```diff
@@ -11,21 +11,19 @@ import (
...
-		HTTPAdapterPrefix: "http",
-		BootstrapPrefix:   "things",
-		MsgContentType:    sdk.ContentType(msgContentType),
-		TLSVerification:   false,
+		AuthURL:         defURL,
+		ThingsURL:       defURL,
+		UsersURL:        defURL,
+		ReaderURL:       defURL,
+		BootstrapURL:    defURL,
+		CertsURL:        defURL,
```
manuio, line 24: Are you sure here?

bootstrap/api/endpoint_test.go
```diff
@@ -160,7 +160,7 @@ func dec(in []byte) ([]byte, error) {
 func newService(auth mainflux.AuthServiceClient, url string) bootstrap.Service {
 	things := mocks.NewConfigsRepository()
 	config := mfsdk.Config{
-		BaseURL: url,
+		BootstrapURL: url,
```
mteodor, line 163: you dont have ThingsURL here

bootstrap/api/endpoint_test.go
```diff
@@ -238,7 +238,7 @@ func TestAdd(t *testing.T) {
 			auth:        validToken,
 			contentType: contentType,
 			status:      http.StatusCreated,
-			location:    "/things/configs/1",
+			location:    "/configs/1",
```
mteodor, line 241: have you changed `responses.go`

bootstrap/api/endpoint_test.go
drasko: Why do we need it and where is defined? In an ENV var?

PR #1432, opened 2021-06-28
MF-1421 - Make flattening of JSON transformer only available on InfluxDB

### LINKED ISSUE #1421: Make "flattening" of JSON in JSON transformer optional
When I publish a message via http/mqtt/... I want it to be stored via the mongodb-writer into a mongodb database.
The mongodb-writer is configured to accept json, what means that the json transformer has to do the job.

I want the sent json to be written as-is in the database

The json is "flattened". According to the documentation, this should happen, but this is definitely a problem, because the data-structures are gone. This makes sense if the data is read out only with the mongodb-reader, and makes it impossible to use other tools for data-evaluation

It would be better to make the "flattening" in the transformer optional, with default value - "flattening off".

buraksekili:

### What does this do?
This PR introduces optional "flattening" of JSON based on the DB. So that "flattening" of JSON is only allowed for InfluxDB.

Resolves #1421 

None

No

No

### DIFF
--- a/pkg/transformers/json/transformer.go
+++ b/pkg/transformers/json/transformer.go
@@ -17,7 +17,7 @@ const sep = "/"
 var keys = [...]string{"publisher", "protocol", "channel", "subtopic"}
 
 var (
-	// ErrTransform reprents an error during parsing message.
+	// ErrTransform represents an error during parsing message.
 	ErrTransform         = errors.New("unable to parse JSON object")
 	errInvalidKey        = errors.New("invalid object key")
 	errUnknownFormat     = errors.New("unknown format of JSON message")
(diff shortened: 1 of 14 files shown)


pkg/transformers/json/transformer_test.go
```diff
@@ -22,7 +22,7 @@ const (
 
 func TestTransformJSON(t *testing.T) {
 	now := time.Now().Unix()
-	tr := json.New()
+	tr := json.New(true)
```
dborovcanin: Test cases for JSON Transformer without flattening are missing. Other than that, this looks good.

dborovcanin (changes requested):

nmarcetic (dismissed):

pkg/transformers/json/transformer.go
```diff
@@ -28,65 +28,77 @@ var (
...
+		default:
+			return nil, errors.Wrap(ErrTransform, errInvalidFormat)
 		}
-		return Messages{res, format}, nil
-	default:
-		return nil, errors.Wrap(ErrTransform, errInvalidFormat)
 	}
 }
 
-// ParseFlat receives flat map that reprents complex JSON objects and returns
```
manuio, line 89: If this method is exported you should add a comment

manuio (changes requested):

manuio (dismissed):
LGTM

pkg/transformers/json/transformer_test.go
```diff
@@ -91,7 +94,8 @@ func TestTransformJSON(t *testing.T) {
 		Format: msg.Subtopic,
 	}
 
-	cases := []struct {
+	// Test cases with JSON flattening.
+	flatCases := []struct {
```
dborovcanin: Since the function we test (`Transform`) acts differently depending on the flag, it would be best to separate it into two tests. This usually implies a bad design, but it's only because we implemented interface as a function, rather than attaching a function to the struct with a `flat` flag. It's usually not a practice to have two tests for a single unit, but IMHO it would be the best in this particular case. Something like `TestTransformerFlat` and `TestTransformerNested`. We'll have all the same cases (of course, with different values) in both tests, but with a flag set differently. It also helps with documenting the behavior of the `Transformer`.

drasko:
@buraksekili please update the branch

@dusanb94 please review

manuio (dismissed):
LGTM

mteodor (dismissed):

dborovcanin (changes requested):
Please take a look at Readers and update them, as well. Other than that, this looks OK.

manuio (approved):
LGTM

drasko:
@dusanb94 please review and let's move with this one

dborovcanin (approved):

Issue #1347, opened 2021-02-01
Add policies


drasko:
**FEATURE REQUEST**

1. Is there an open issue addressing this request? If it does, please add a "+1" reaction to the
   existing issue, otherwise proceed to step 2.

2. Describe the feature you are requesting, as well as the possible use case(s) for it.
Following the https://github.com/mainflux/mainflux/issues/931 and https://github.com/mainflux/mainflux/issues/1346, we need to add policies as super-important feature.

3. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).
Must-have

drasko:
Closed in favor of https://github.com/mainflux/mainflux/issues/1443

PR #1440, opened 2021-07-22
MF-1443 - Add policies

buraksekili:
> This PR is still in progress and is just a draft PR for now.

### What does this do?

This PR adds policies but it is still WIP. All policies will be added subsequently.

Related to #1443 and #1347 

- Adds new dependencies related to `Keto`.
- Adds a new gRPC method called `AddPolicy` for Auth.
- Adds a new endpoint for `Users` service, `/selfsignon`
- Updated current `POST /users` endpoint. Now, it takes a `token` for checking user authorization.

Yes

Not yet.


auth/service.go
```diff
@@ -170,7 +191,58 @@ func (svc service) Identify(ctx context.Context, token string) (Identity, error)
...
+			Relation:  act,
+			Subject: &acl.Subject{Ref: &acl.Subject_Id{
+				Id: sub,
+			}},
+		})
+		if err != nil {
+			return false, err
+		}
+
+		if res.Allowed {
```
tzzed: Good job Burak!
what about `return res.Allowed, nil`

buraksekili: Hi @tzzed, thanks. 
Your suggestion seems more appropriate than my solution. I guess it is better to return with `res.Allowed`. I will update the PR.

drasko:
@buraksekili @mteodor  I think a dedicated issue should be opened for this one - it's a huge feature. Maybe it should be correlated to some of the existing openes issues - if such exists.

Issue should explain in details what we are doing here, give link to [Keto](https://github.com/ory/keto) and [Google Zanzibar](https://www.usenix.org/conference/atc19/presentation/pang), and so on.

buraksekili:
> @buraksekili @mteodor I think a dedicated issue should be opened for this one - it's a huge feature. Maybe it should be correlated to some of the existing openes issues - if such exists.
> 
> Issue should explain in details what we are doing here, give link to [Keto](https://github.com/ory/keto) and [Google Zanzibar](https://www.usenix.org/conference/atc19/presentation/pang), and so on.

Okay @drasko, I will open a new issue based on your suggestions.

cmd/users/main.go
```diff
@@ -313,31 +314,79 @@ func newService(db *sqlx.DB, tracer opentracing.Tracer, auth mainflux.AuthServic
...
+
 	user := users.User{
 		Email:    c.adminEmail,
 		Password: c.adminPassword,
 	}
 
-	if _, err := userRepo.RetrieveByEmail(context.Background(), user.Email); err == nil {
-		// Exiting if user already exists
+	if u, err := userRepo.RetrieveByEmail(context.Background(), user.Email); err == nil {
+		// If the Admin is existing, check whether it has proper policies or not.
```
mteodor: If admin is created but policies failed maybe we can recreate admin on next startup
check this https://github.com/mainflux/mainflux/blob/0fe9f55bae8d0887d5ea2d7be9fb483c6b0bfafc/provision/service.go#L103
`recover` function does cleanup in case of error

dborovcanin:
To simplify PRs and reviews, let's split the solutions into multiple steps:

1. Introduce Policy entity and support ownership-based access control, but using policies.
2. Enrich Policies to support any-to-any User-Thing relation rather than only an ownership-based approach.
3. Add administrators to prevent any User to be able to create any Policy.
4. Introduce Policies that operate over Things groups.
5. Introduce Policies that operate over Users groups.

How policies work with groups is a very complex topic and we will possibly need to change our approach to groups, but let's discuss that once we resolve the first 3 steps. Adding administrators can also be tricky since administrators are - a user group.

docker/docker-compose.yml
```diff
@@ -11,11 +11,81 @@ volumes:
...
+        target: /home/ory
+    command: relation-tuple create /home/ory/relation-tuples
+    restart: on-failure
+    networks:
+      - mainflux-base-net
+    depends_on:
+      - keto-db
+      - keto-migrate
+
+  keto-migrate:
```
mteodor, line 39: do we need to have `keto-init` and `keto-migrate` to be started every time composition is up

things/service.go
```diff
@@ -180,6 +199,9 @@ func (ts *thingsService) CreateThings(ctx context.Context, token string, things
 				return []Thing{}, errors.Wrap(ErrCreateUUID, err)
 			}
 		}
+		if err := ts.addPolicy(ctx, res.GetId(), things[i].ID, ownerRelationKey); err != nil {
```
mteodor: we can put logic for creation of a thing and adding a policy in additionl function that would be called from within a loop
also, what happens with a created thing when adding policy has failed?

things/service.go
```diff
@@ -213,6 +243,10 @@ func (ts *thingsService) ViewThing(ctx context.Context, token, id string) (Thing
 		return Thing{}, errors.Wrap(ErrUnauthorizedAccess, err)
 	}
 
+	if err := ts.authorize(ctx, res.GetId(), id, ownerRelationKey, accessRelationKey); err != nil {
```
mteodor: why both owner and access relation?

things/service.go
```diff
@@ -213,6 +243,10 @@ func (ts *thingsService) ViewThing(ctx context.Context, token, id string) (Thing
...
+		return Thing{}, err
+	}
+
 	return ts.things.RetrieveByID(ctx, res.GetEmail(), id)
 }
```
mteodor, line 277: do we need authorize in `ListThings`

users/api/endpoint.go
```diff
@@ -31,6 +31,25 @@ func registrationEndpoint(svc users.Service) endpoint.Endpoint {
 	}
 }
 
+func createUserEndpoint(svc users.Service) endpoint.Endpoint {
```
mteodor, line 34: we dont need this in this PR, right?

users/api/transport.go
```diff
@@ -39,13 +39,20 @@ func MakeHandler(svc users.Service, tracer opentracing.Tracer) http.Handler {
...
-	mux.Post("/users", kithttp.NewServer(
+	mux.Post("/users/create", kithttp.NewServer(
 		kitot.TraceServer(tracer, "register")(registrationEndpoint(svc)),
 		decodeCredentials,
 		encodeResponse,
 		opts...,
 	))
 
+	mux.Post("/users", kithttp.NewServer(
+		kitot.TraceServer(tracer, "create_user")(createUserEndpoint(svc)),
```
mteodor: this looks reversed

pkg/sdk/go/users.go
```diff
@@ -15,10 +15,11 @@ import (
 )
 
 const (
-	usersEndpoint    = "users"
-	tokensEndpoint   = "tokens"
-	passwordEndpoint = "password"
-	membersEndpoint  = "members"
+	usersEndpoint      = "users"
+	createUserEndpoint = "users/create"
```
drasko: Is this maybe `usersEndpoint + "/create"`?

things/service.go
```diff
@@ -439,3 +477,38 @@ func (ts *thingsService) members(ctx context.Context, token, groupID, groupType
...
+			return errors.Wrap(ErrAuthorization, err)
+		}
+		if res.Authorized {
+			return nil
+		}
+	}
+	return ErrAuthorization
+}
+
+func (ts *thingsService) addPolicy(ctx context.Context, subject, object, relation string) error {
```
drasko: Why both Things service and Auth service have `AddPolicy()`? Should not we add policies only on one place?

users/api/endpoint.go
drasko: What's this? Don't we already have an endpoint to create users?

users/service.go
```diff
@@ -321,12 +399,44 @@ func (svc usersService) issue(ctx context.Context, id, email string, keyType uin
...
+		Act: relation,
+	}
+	res, err := svc.auth.Authorize(ctx, req)
+	if err != nil {
+		return false, errors.Wrap(ErrAuthorization, err)
+	}
+	return res.Authorized, nil
+}
+
+func (svc usersService) addPolicy(ctx context.Context, subject, object, relation string) (bool, error) {
```
drasko: Again - why `AddPolicy()` in this microservice, why we do not isolate adding policies in only one microservice?

drasko (changes requested):

docker/docker-compose.yml
buraksekili: Actually, successful `keto-init` and `keto-migrate` are sufficient for once. We do not need them to build every composition. They prepare the database for Keto.

things/service.go
buraksekili: When adding policy fails, the service returns an error; although, the things policies are saved into the DB. So that, the user might have an `ownerRelation` over non-existing things. It should be handled as well.

things/service.go
buraksekili: Because I've thought that when a user shares a thing with other users, other users will have an `access` relation on the thing. Therefore, I checked that relation here but actually, this part does not belong to this PR.

things/service.go
buraksekili: I've thought only admin fetches all things. However, for the sake of this PR, we only have ownership-based access. Therefore, we do not need to authorize here.

auth/api/grpc/client.go
```diff
@@ -129,7 +138,7 @@ func (client grpcClient) Authorize(ctx context.Context, req *mainflux.AuthorizeR
 	ctx, close := context.WithTimeout(ctx, client.timeout)
 	defer close()
 
-	res, err := client.authorize(ctx, authReq{Act: req.Act, Obj: req.Obj, Sub: req.Sub})
+	res, err := client.authorize(ctx, authReq{Act: req.GetAct(), Obj: req.GetObj(), Sub: req.GetSub()})
```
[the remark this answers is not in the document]
buraksekili: Yes, I agree. But those getter functionalities come from auto-generated proto files. Therefore, I've used them instead of calling struct members (as `res.Obj`) directly. . Should it be called `res.Obj` instead of `res.GetObj()`?

users/api/endpoint.go
buraksekili: Yes, we have an endpoint to create users. However, we thought that only admin users should be able to create users. So, instead of `Register`, I created a `CreateUser` endpoint which verifies if the user who called `CreateUser` has rights to create a new user, and writes a default policy for the newly created user. The old endpoint still remains for the admin creation as here https://github.com/mainflux/mainflux/blob/19f0437f577fa24ae0fe280fda3a0cf0a6a1c674/cmd/users/main.go#L323.

things/service.go
buraksekili: Yes, AddPolicy should belong to Auth service. However, I wrote this `addPolicy` to prevent code-duplication while calling `auth.AddPolicy`. So, `addPolicy` just calls the auth's AddPolicy. Although `addPolicy` here is only used once, we will need to call it more than once for further policies (e.g. creating a channel). Should it be separated from `ts *thingsService`?

users/api/transport.go
buraksekili: Actually, I endeavor to replace `registrationEndpoint` with the new `createUserEndpoint`. Therefore, I used `/users` endpoint for `createUserEndpoint` and used `/users/create` for `registrationEndpoint`. However, it seems like this approach creates inconsistency about naming.

pkg/sdk/go/users.go
```diff
@@ -15,10 +15,11 @@ import (
 )
 
 const (
-	usersEndpoint    = "users"
-	tokensEndpoint   = "tokens"
-	passwordEndpoint = "password"
-	membersEndpoint  = "members"
+	usersEndpoint      = "users"
```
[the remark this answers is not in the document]
buraksekili: In order to restrict user creation, we've thought only admin should be able to create users. Therefore, I created `createUserEndpoint`. Unlike `usersEndpoint`, `createUserEndpoint` checks whether the user who made a request to `createUserEndpoint` has the policy to create a new user. Furthermore, `createUserEndpoint` adds a default policy for the newly created user. This default policy allows new users to create `things`.

Since I updated user creation endpoint `/users` with the new `CreateUser` (previously, the `/users` was using `Register`), I've added `createUserEndpoint = "users/create"`. So, `createUserEndpoint` calls the `Register`. However, I realized that the naming is bad for these variables. It'd be better if I rename it as `registerEndpoint = "users/create"`

auth/policies.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"context"
+
+	acl "github.com/ory/keto/proto/ory/keto/acl/v1alpha1"
+)
+
+// PolicyService represents a authorization service. It exposes
+// functionalities through `auth` to perform authorization.
+type PolicyService interface {
```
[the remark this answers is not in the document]
buraksekili: For `PolicyCommunicator`, `Policy` prefix is not necessary. It can be omitted.

For `PolicyService`, if I omit the Policy prefix, it would conflict with the [auth package's Service ](https://github.com/mainflux/mainflux/blob/f212b064c5de6438746cf180589a6afdf193ced0/auth/service.go#L95). Maybe it'd be better if I rename to `Authz` instead of `PolicyService`.

things/service.go
mteodor: @buraksekili right, this should be in subsequent PR, and I think that we need to solve that with policy design

things/service.go
mteodor: User should be able to fetch all of the things he owns, and authorization is on the Keto ( though it is duplicated for now in the service itself)

things/service.go
```diff
@@ -439,3 +477,38 @@ func (ts *thingsService) members(ctx context.Context, token, groupID, groupType
 	}
 	return res.Members, nil
 }
+
+func (ts *thingsService) authorize(ctx context.Context, subject, object string, relations ...string) error {
+	req := &mainflux.AuthorizeReq{
+		Sub: subject,
+		Obj: object,
+	}
+	for _, relation := range relations {
```
mteodor: since, for now, we will check only one relation and we should strive that `keto` does evaluation for more complex relations by carefuly designing policies you should remove this loop
also this can slow server to have relations in evaluated in loop

cmd/auth/main.go
```diff
@@ -129,6 +139,11 @@ func loadConfig() config {
 		SSLRootCert: mainflux.Env(envDBSSLRootCert, defDBSSLRootCert),
 	}
 
+	kc := auth.KetoConfig{
+		WritePort: mainflux.Env(envKetoWritePort, defKetoWritePort),
+		ReadPort:  mainflux.Env(envKetoReadPort, defKetoReadPort),
```
dborovcanin: Write and Read ports are only used in `main`, so there is no need to have them in the KetoConfig. We can completely remove the KetConfig struct if we pass `Checker` and `Writer` to the fabric method (`NewPolicyCommunicator`).

cmd/auth/main.go
```diff
@@ -47,6 +49,9 @@ const (
 	defServerCert    = ""
 	defServerKey     = ""
 	defJaegerURL     = ""
+	ketoContainer    = "mainflux-keto"
```
dborovcanin: Should this be configurable?

auth/policies.go
```diff
@@ -0,0 +1,108 @@
...
+	// AddPolicy creates a policy for the given subject. So that, after
+	// AddPolicy, `subject`has a `relation` on `object`. Returns non-nil
+	// error in case of failures.
+	AddPolicy(ctx context.Context, subject, object, relation string) error
+}
+
+// PolicyCommunicator facilitates the communication to authorization
+// services and implements PolicyService functionalities for certain
+// authorization services (e.g. ORY Keto).
+type PolicyCommunicator interface {
```
dborovcanin: Consider renaming to `PolicyAgent`.

auth/policies.go
```diff
@@ -0,0 +1,108 @@
...
+	// AddPolicy, `subject` has a `relation` on `object`. Returns non-nil
+	// error in case of failures.
+	AddPolicy(ctx context.Context, subject, object, relation string) error
+}
+
+// PolicyResult shows the result of the authorization check operations.
+// Authorized field is true when the operation is allowed. If Authorized
+// is false, the operation is denied.
+type PolicyResult struct {
+	Authorized bool
```
dborovcanin: Consider renaming to `AuthorizationResult`. Also, having an error instead of bool provides more info in case of request failure. Nil error, as usual, indicates success.

auth/api/grpc/client.go
dborovcanin, line 141: One thought for the future: Authorization (gRPC) client **does not have to use Auth service**. We can use it to directly communicate with Keto - our Auth service acts as a mere proxy when it comes to policy evaluation and we can speed and simplify the process by calling Keto gRPC API directly. What do you think? For this initial version, I'd leave it as it is, but this is a remark for a general approach.

auth/api/logging.go
```diff
@@ -82,17 +82,28 @@ func (lm *loggingMiddleware) Identify(ctx context.Context, key string) (id auth.
...
 		}
 		lm.logger.Info(fmt.Sprintf("%s without errors.", message))
 	}(time.Now())
+	return lm.svc.Authorize(ctx, subject, object, relation)
+}
 
-	return lm.svc.Authorize(ctx, token, sub, obj, act)
+func (lm *loggingMiddleware) AddPolicy(ctx context.Context, subject, object, relation string) (err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method add_policy %s#%s@%s took %s to complete", object, relation, subject, time.Since(begin))
```
dborovcanin: There is no need to translate `(object, relation, subject)` tuple to Keto syntax.

dborovcanin (changes requested):

cmd/auth/main.go
buraksekili: No, it should not be configurable. I've only declared it because I am using `mainflux-keto` in multiple places across `cmd/auth/main.go`. Should `ketoContainer` constant usage be replaced with plain string?

auth/policies.go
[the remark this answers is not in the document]
drasko: Avoid `PolicyService` and find a shorter name for that inteface. Also, Interface names should be something like `Authorizer` or something...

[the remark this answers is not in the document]
drasko: @dusanb94 what do you think about this?

users/api/endpoint.go
drasko: Ah I see. I would demand one more significant change then - now that we are introducing policies and admins, I think we should by default support only creating users via admin, and not self-sign-on. Self-sign-on should be an option, that can be activated via ENV var, and should have a dedicated endpoint - something like `/selfsignon` or something. And we should keep simple `POST /users` user creation endpoint to be used by the admin - i.e. we should add authorization to it (currently it is not authorized, because it is self-sign-on).

things/service.go
drasko: If this is just calling Auth service, than a gRPC client for Auth service should be made and we can call directly Auth service with it.

auth/policies.go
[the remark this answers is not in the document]
dborovcanin: Since there is more than one method (and will probably be even more), `Authorizer` is not necessarily the best name. I think that the old name `Authz` (since this interface is really all about authorization) fits the best.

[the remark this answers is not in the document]
drasko: OK, good. `Authz` it is.

auth/service.go
```diff
@@ -104,17 +102,19 @@ type service struct {
 	groups       GroupRepository
 	idProvider   mainflux.IDProvider
 	ulidProvider mainflux.IDProvider
+	keto         PolicyAgent
 	tokenizer    Tokenizer
 }
 
 // New instantiates the auth service implementation.
-func New(keys KeyRepository, groups GroupRepository, idp mainflux.IDProvider, tokenizer Tokenizer) Service {
+func New(keys KeyRepository, groups GroupRepository, idp mainflux.IDProvider, tokenizer Tokenizer, policy PolicyAgent) Service {
```
drasko: I would prefer `pa` or `policyAgent` for `PolicyAgent` var then `policy`

auth/service.go
```diff
@@ -104,17 +102,19 @@ type service struct {
 	groups       GroupRepository
 	idProvider   mainflux.IDProvider
 	ulidProvider mainflux.IDProvider
+	keto         PolicyAgent
```
drasko: Should this one be called `keto`? On one hand - it is readable. On the other hand, Keto is just a policy provider, or authorization system, and can be potentially changed in the future in Mainflux.

Maybe `policyAgent` here would be a better naming?

@dusanb94 @mteodor @nmarcetic opinions please

cmd/auth/main.go
drasko: It should be configurable, container names (and URLs) can change easily. From what I see you are dialing this URL from the init function - changing docker-compose, renaming containers, using K8s - or whatever similar - can change container name (and thus hostname, or URL in a private network) and then your code will stop working, because you hard-coded it.

drasko (changes requested):
Great progress here, @buraksekili. You are getting there.

@mteodor @dusanb94 @nmarcetic please review this in details, and let's merge this important one! After that it will be easier to develop policies further.

auth/api/grpc/client.go
drasko: As discussed before, we'll avoid calling Keto directly from clients for several reasons:
- Keto is currently used, but we wrap it's usage so it can be potentially changed in the future
- When something in Keto changes (version), you do not have to update every client and app out there, we can just adapt our wrapper to keep backward compatibility for example
- And so on

auth/service.go
dborovcanin: I vote for `policyAgent`.

auth/policies.go
```diff
@@ -0,0 +1,101 @@
...
+// Authz represents a authorization service. It exposes
+// functionalities through `auth` to perform authorization.
+type Authz interface {
+	// Authorize checks authorization of the given `subject`. Basically,
+	// Authorize verifies that Is `subject` allowed to `relation` on
+	// `object`. Authorize returns the bool indicating the authorization
+	// of the given subject and the error. For example, the response as:
+	// `false, nil` means that `subject` has not `relation` on the `object`.
+	// Therefore, this response means that incoming request is denied.
+	Authorize(ctx context.Context, subject, object, relation string) (bool, error)
```
dborovcanin: There is no need to return both bool and error. Error is enough. Nil error indicates success.

auth/policies.go
```diff
@@ -0,0 +1,101 @@
...
+						Id: subject,
+					}},
+				},
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+	return nil
```
dborovcanin: You can replace this with `return err`.

auth/policies.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	if !res.GetAllowed() {
+		return AuthorizationResult{ErrAuthorization}, err
+	}
+
+	return AuthorizationResult{}, nil
+}
+
+func (c policyAgent) AddPolicy(ctx context.Context, subject, object, relation string) error {
+	_, err := c.writer.TransactRelationTuples(context.Background(), &acl.TransactRelationTuplesRequest{
```
dborovcanin: I would extract this `TransactRelationTuplesRequest` to a separate local variable to make a function call a bit cleaner, but if you prefer it this way, feel free to resolve this comment.

things/service.go
```diff
@@ -42,6 +47,16 @@ var (
 	ErrFailedToRetrieveThings = errors.New("failed to retrieve group members")
 )
 
+const (
+	thingObjectKey    = "thing"
+	authoritiesObjKey = "authorities"
+
```
dborovcanin: No need for the empty line.

things/service.go
dborovcanin: Authorization of list requests is very tricky, so I don't mind if we have simple filtering on the service layer in the initial version. We need to authenticate the user, though.

dborovcanin (changes requested):
Great progress, I think we're close to the first version.

docker/docker-compose.yml
mteodor: I would move `keto-init` and `keto-migrate` to separae docker-compose file and add explanation how to prepare DB for keto into README.md

users/api/transport.go
```diff
@@ -41,11 +41,20 @@ func MakeHandler(svc users.Service, tracer opentracing.Tracer) http.Handler {
 
 	mux.Post("/users", kithttp.NewServer(
 		kitot.TraceServer(tracer, "register")(registrationEndpoint(svc)),
-		decodeCredentials,
+		decodeCreateUserReq,
 		encodeResponse,
 		opts...,
 	))
 
+	if v := mainflux.Env("MF_SELFREGISTER", ""); v != "" {
```
mteodor: I think that for the sake of clean code reading of environment variable should be in the same place as for other variables ( users/main.go ), than you pass that to `users := service.New(...., selfergister )` and in `SelfRegister` in `service.go:users.go` you proceed or return an message if `selfregister = false`, @dusanb94 , what do you think

auth/policies.go
```diff
@@ -0,0 +1,101 @@
...
+		RelationTupleDeltas: []*acl.RelationTupleDelta{
+			{
+				Action: acl.RelationTupleDelta_INSERT,
+				RelationTuple: &acl.RelationTuple{
+					Namespace: ketoNamespace,
+					Object:    object,
+					Relation:  relation,
+					Subject: &acl.Subject{Ref: &acl.Subject_Id{
```
mteodor: this is ok for this PR, but just to have it in mind that we can have `SubjectSet` here also

users/api/transport.go
buraksekili: `SelfRegister` in `users/service.go` is also used for admin creation in [`cmd/users/main.go`](https://github.com/mainflux/mainflux/blob/e3f90ee936d0ada0b7ed392203bf86685e6d93b5/cmd/users/main.go#L334). In order to create an admin, `SelfRegister` should be enabled. If `SelfRegister` is disabled via the `.env` file, admin creation will be failed. So, in my opinion, `SelfRegister` as a method of `users.Service` should be allowed to call but serving it through HTTP endpoint should not. 

Maybe, I can read the environment variable in the `cmd/user/main.go` and pass the environment variable to [api.MakeHandler](https://github.com/mainflux/mainflux/blob/e3f90ee936d0ada0b7ed392203bf86685e6d93b5/users/api/transport.go#L35)

drasko (changes requested):

docker/docker-compose.yml
drasko: Agreed. This should be solved in the better way. Do we need these containers at all - are those maybe just binary tools that we can download in the script and execute? Anyway - they are burdening the main docker-compose and we need to find an elegant way to introduce Keto.

users/service.go
drasko: @dusanb94 opinions on this?

drasko (changes requested):

auth/api/grpc/endpoint.go
```diff
@@ -62,17 +62,26 @@ func authorizeEndpoint(svc auth.Service) endpoint.Endpoint {
 			return authorizeRes{}, err
 		}
 
-		_, err := svc.Identify(ctx, req.token)
+		authorized, err := svc.Authorize(ctx, req.Sub, req.Obj, req.Act)
```
nmarcetic: Can we pass AuthReq here, as everywhere else?

auth/api/grpc/endpoint.go
```diff
@@ -62,17 +62,26 @@ func authorizeEndpoint(svc auth.Service) endpoint.Endpoint {
...
 		if err != nil {
 			return authorizeRes{}, err
```
nmarcetic, line 67: No need for `authorizeRes{authorized: false}` ? Like [here](https://github.com/mainflux/mainflux/pull/1440/files#diff-8fcf53605bc54615ea033c69bbba7d42b516e64c236776073c358c3b420e9189R170) 
If you are considering it as the default value for bool, that's ok just make it consistent.

auth/mocks/policies.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/mainflux/mainflux/auth"
+)
+
+type MockSubjectSet struct {
+	Object   string
+	Relation string
+}
+
+type ketoMock struct {
```
nmarcetic: I would avoid using the keto word, it's impl specific.

auth/policies.go
```diff
@@ -0,0 +1,101 @@
...
+	// `object`. Authorize returns the bool indicating the authorization
+	// of the given subject and the error. For example, the response as:
+	// `false, nil` means that `subject` has not `relation` on the `object`.
+	// Therefore, this response means that incoming request is denied.
+	Authorize(ctx context.Context, subject, object, relation string) (bool, error)
+
+	// AddPolicy creates a policy for the given subject. So that, after
+	// AddPolicy, `subject`has a `relation` on `object`. Returns non-nil
+	// error in case of failures.
+	AddPolicy(ctx context.Context, subject, object, relation string) error
```
nmarcetic: Same here, can we pass policy req obj? Instead of 3 strings?

auth/policies.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// PolicyAgent facilitates the communication to authorization
+// services and implements Authz functionalities for certain
+// authorization services (e.g. ORY Keto).
+type PolicyAgent interface {
+	// CheckPolicy checks if the subject has a relation on the object.
+	// It returns AuthorizationResult consisting of one boolean which indicates
+	// result of the authorization (true: Allowed, false: Denied) and error.
```
nmarcetic: Similar to @dusanb94 [remark](https://github.com/mainflux/mainflux/pull/1440/files#r686922624) do we need bool here or error that has none nill value is enough (can contain error message also, would be really helpful for logs)?

auth/policies.go
```diff
@@ -0,0 +1,101 @@
...
+	// AddPolicy creates a policy for the given subject. So that, after
+	// AddPolicy, `subject` has a `relation` on `object`. Returns non-nil
+	// error in case of failures.
+	AddPolicy(ctx context.Context, subject, object, relation string) error
+}
+
+// AuthorizationResult shows the result of the authorization check operations.
+// AuthzError field is nil when the operation is allowed. If AuthzError is
+// non-nil, the operation is denied.
+type AuthorizationResult struct {
```
nmarcetic: Why did you wrap the error? Is it not easier to just return an error? 
For example [here](https://github.com/mainflux/mainflux/pull/1440/files#diff-d60a65301395efdc368c161a4097c96d6f3f1b4500dbcd15de618be7fc9b72a1R72) Looks like a double error for me ?

auth/policies.go
nmarcetic: I agree with @dusanb94 but it's already mixed everywhere, constructors are all over the fn calls. Not really bad IMHO, but for me personally it's cleaner and easier to read if we extract to separate variables.

auth/service.go
```diff
@@ -170,10 +170,21 @@ func (svc service) Identify(ctx context.Context, token string) (Identity, error)
...
+		return false, errors.Wrap(ErrAuthorization, err)
+	}
+	if ar.AuthzError != nil {
+		return false, errors.Wrap(ErrAuthorization, err)
+	}
 	return true, nil
 }
 
+func (svc service) AddPolicy(ctx context.Context, subject, object, relation string) error {
+	return svc.policyAgent.AddPolicy(ctx, subject, object, relation)
```
nmarcetic: Is there are other agents also? Like Auth agent or identity agent? If no, maybe just an agent makes more sense? 
Like svc.agent.addPolicy ? If its only one agent , method name is already self-documented

bootstrap/mocks/users.go
```diff
@@ -44,6 +44,10 @@ func (svc serviceMock) Authorize(ctx context.Context, req *mainflux.AuthorizeReq
 	panic("not implemented")
 }
 
+func (svc serviceMock) AddPolicy(ctx context.Context, in *mainflux.AddPolicyReq, opts ...grpc.CallOption) (*mainflux.AddPolicyRes, error) {
+	panic("not implemented")
```
nmarcetic, line 48: Is it smart to panic? I know it all over the not impl methods , just checking

docker/keto/relation-tuples/authorities.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "namespace":"members",
```
nmarcetic: What does this mean by "members"? Is it keto domain or ?

cli/provision.go
```diff
@@ -124,7 +124,7 @@ var cmdProvision = []cobra.Command{
 				Email:    un,
 				Password: "12345678",
 			}
-			if _, err := sdk.CreateUser(user); err != nil {
+			if _, err := sdk.CreateUser("t", user); err != nil {
```
drasko: What's this `"t"` string?

drasko (changes requested):

cli/provision.go
buraksekili: CreateUser takes a token to verify the user. If a token is an empty string, validation of the [`createUserReq`](https://github.com/buraksekili/mainflux/blob/ee866f1fc7dd5a2bfa54651f1e6a0729d2ca8ea5/users/api/endpoint.go#L18) will be failed. Therefore, `"t"` is an arbitrary value.

docker/keto/relation-tuples/authorities.json
buraksekili: It is a kind of scope for ACLs in Keto. Keto organizes relation tuples based on the `namespace`. But for the sake of this PR, a separate JSON file for relation tuples is not needed. It is deleted for now.

auth/policies.go
buraksekili: In the beginning (before splitting the PR as indicated [here](https://github.com/mainflux/mainflux/pull/1440#issuecomment-892756436)), I've thought that I might need to return more than one field. However, it is not needed for now. Returning an error is enough.

users/service.go
```diff
@@ -197,19 +229,19 @@ func (svc usersService) ViewUser(ctx context.Context, token, id string) (User, e
...
+	dbUser, err := svc.users.RetrieveByEmail(ctx, ir.email)
 	if err != nil {
 		return User{}, errors.Wrap(ErrUnauthorizedAccess, err)
 	}
 
 	return User{
 		ID:       dbUser.ID,
-		Email:    email,
+		Email:    ir.email,
 		Password: "",
```
nmarcetic: You can just omit this field.

users/service.go
```diff
@@ -249,11 +281,11 @@ func (svc usersService) GenerateResetToken(ctx context.Context, email, host stri
...
 
 func (svc usersService) ResetPassword(ctx context.Context, resetToken, password string) error {
-	email, err := svc.identify(ctx, resetToken)
+	ir, err := svc.identify(ctx, resetToken)
 	if err != nil {
 		return errors.Wrap(ErrUnauthorizedAccess, err)
 	}
-	u, err := svc.users.RetrieveByEmail(ctx, email)
+	u, err := svc.users.RetrieveByEmail(ctx, ir.email)
 	if err != nil || u.Email == "" {
```
nmarcetic: shouldn't it be here and the operator not or? 
Anyway, the next line is error checking IMHO it would be better to first check for the error then check if user email is empty, more readable.

auth/service.go
buraksekili: Under `auth` package, there is no other service called `agent`. Auth Service has `Authn`, `Authz`, and `GroupService`. I will rename `policyAgent` to `agent`

auth/api/grpc/endpoint.go
buraksekili: Yes, we can pass `AuthReq` but it causes `import cycle not allowed` since `"github.com/mainflux/mainflux/auth"` and [`"github.com/mainflux/mainflux/auth/api/grpc"`](https://github.com/buraksekili/mainflux/blob/ee866f1fc7dd5a2bfa54651f1e6a0729d2ca8ea5/auth/api/grpc/requests.go#L7) call each other after importing `AuthReq`. What if we create a new generic struct (inside of `auth/policies.go`) instead of passing 3 strings?

nmarcetic: Yea new local struct which will wrap req params sounds great.

cli/provision.go
drasko: I would prefer than a defined constant string with some self-describing value - like `"foobar"` and a comment that this is to avoid validation fail.

BTW, can we maybe permit empty strings, and fail validation only when policies do not allow self-registration? That would make more sense and make it cleaner.

@mteodor @nmarcetic what do you think?

pkg/sdk/go/sdk.go
```diff
@@ -116,8 +116,8 @@ type Channel struct {
 
 // SDK contains Mainflux API.
 type SDK interface {
-	// CreateUser registers mainflux user.
-	CreateUser(user User) (string, error)
+	// CreateUser registers mainflux user. It is only allowed to perform by admins.
```
drasko: This is not really true, If policies are set to accept `*`, then it can be performed by anybody (self-registration).

drasko (changes requested):

manuio (changes requested):

cli/provision.go
```diff
@@ -19,8 +19,12 @@ import (
 
 var errMalformedCSV = errors.New("malformed CSV")
 
-const jsonExt = ".json"
-const csvExt = ".csv"
+const (
+	jsonExt = ".json"
+	csvExt  = ".csv"
+	// tmpToken is defined to avoid validation failure while user registration.
+	tmpToken = "token"
```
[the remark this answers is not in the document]
buraksekili: No, it should be removed

auth/policies.go
```diff
@@ -0,0 +1,99 @@
...
+	// Authorize checks authorization of the given `subject`. Basically,
+	// Authorize verifies that Is `subject` allowed to `relation` on
+	// `object`. Authorize returns a non-nil error if the subject has
+	// no relation on the object (which simply means the operation is
+	// denied).
+	Authorize(ctx context.Context, pr PolicyReq) error
+
+	// AddPolicy creates a policy for the given subject. So that, after
+	// AddPolicy, `subject`has a `relation` on `object`. Returns non-nil
+	// error in case of failures.
```
dborovcanin: A couple of typos:


```
// AddPolicy creates a policy for the given subject, so that, after
// AddPolicy, `subject` has a `relation` on `object`. Returns non-nil
// error in case of failures.
```

auth/policies.go
```diff
@@ -0,0 +1,99 @@
...
+	AddPolicy(ctx context.Context, pr PolicyReq) error
+}
+
+type policyAgent struct {
+	writer  acl.WriteServiceClient
+	checker acl.CheckServiceClient
+}
+
+// NewPolicyAgent returns a gRPC communication functionalities
+// to communicate with the authorization services, e.g. ORY Keto.
```
dborovcanin: `NewPolicyAgent` returns a gRPC communication functionalities to communicate with ORY Keto (not `with the authorization services, e.g. ORY Keto`). This is a fabric method for this particular implementation of the PolicyAgent.

auth/policies.go
dborovcanin: I also assumed that the signature of this method will change, and that's one of the reasons why wrapping params into `PolicyReq` makes even more sense.

auth/service.go
```diff
@@ -170,8 +170,13 @@ func (svc service) Identify(ctx context.Context, token string) (Identity, error)
 	}
 }
 
-func (svc service) Authorize(ctx context.Context, token, sub, obj, act string) (bool, error) {
-	return true, nil
+// func (svc service) Authorize(ctx context.Context, subject, object, relation string) error {
```
dborovcanin: Remove commented code.

users/service.go
```diff
@@ -142,30 +152,56 @@ func New(users UserRepository, hasher Hasher, auth mainflux.AuthServiceClient, e
...
-	}
-	user.Password = hash
+
 	uid, err := svc.idProvider.ID()
 	if err != nil {
 		return "", errors.Wrap(ErrCreateUser, err)
 	}
 	user.ID = uid
+
+	if err := svc.claimOwnership(ctx, user.ID, usersObjKey, memberRelationKey); err != nil {
```
dborovcanin, line 173: What does `claimOwnership` over the user do? Does it mean that only the admin will be allowed to manipulate user data?

mteodor (approved):

auth/policies.go
```diff
@@ -0,0 +1,99 @@
...
+	// the object (which simply means the operation is denied).
+	CheckPolicy(ctx context.Context, pr PolicyReq) error
+
+	// AddPolicy creates a policy for the given subject. So that, after
+	// AddPolicy, `subject` has a `relation` on `object`. Returns non-nil
+	// error in case of failures.
+	AddPolicy(ctx context.Context, pr PolicyReq) error
+}
+
+type policyAgent struct {
```
dborovcanin: Please move implementation to a separate `keto` package, similar to the repository and `postgres`. We keep only the service interface and implementation together because there is no better place for implementation.

nmarcetic (approved):

dborovcanin (approved):

users/service.go
```diff
@@ -142,30 +152,56 @@ func New(users UserRepository, hasher Hasher, auth mainflux.AuthServiceClient, e
...
-	}
-	user.Password = hash
+
 	uid, err := svc.idProvider.ID()
 	if err != nil {
 		return "", errors.Wrap(ErrCreateUser, err)
 	}
 	user.ID = uid
+
+	if err := svc.claimMembership(ctx, user.ID, usersObjKey, memberRelationKey); err != nil {
```
drasko: Finally we settled down on `claimOwnership()`, not `claimMembership()`. Please correct.

drasko (changes requested):

drasko (approved):
LGTM

drasko:
Merged! Huge one! Congrats @buraksekili

PR #1444, opened 2021-08-08
MF 1413 - Use per-service URL in SDK

### LINKED ISSUE #1413: Use per-service URL in SDK

drasko:

### What does this do?
Fixes SDK to use per-service URL instead of `BaseURL`.

Resolves #1413


bootstrap/api/endpoint_test.go
```diff
@@ -160,7 +160,7 @@ func dec(in []byte) ([]byte, error) {
 func newService(auth mainflux.AuthServiceClient, url string) bootstrap.Service {
 	things := mocks.NewConfigsRepository()
 	config := mfsdk.Config{
-		BootstrapURL: url,
+		ThingsURL: url,
```
drasko, line 163: Why `ThingsURL`? It looks that caller is calling this `newService()` funtion with the URL of Bootstrap server, right? And we are giving this to ThingsURL.

certs/service_test.go
```diff
@@ -65,7 +65,7 @@ func newService(tokens map[string]string) (certs.Service, error) {
 
 	auth := thmocks.NewAuthService(tokens)
 	config := mfsdk.Config{
-		CertsURL: server.URL,
+		ThingsURL: server.URL,
```
drasko, line 67: What is this `server`? Is it Bootstrap server? If yes - why we are giving this URL to Things service configuration - in realistic deployment these will be two completely different URLs.

bootstrap/service_test.go
```diff
@@ -58,6 +58,7 @@ func newService(auth mainflux.AuthServiceClient, url string) bootstrap.Service {
 	things := mocks.NewConfigsRepository()
 	config := mfsdk.Config{
 		BootstrapURL: url,
+		ThingsURL:    url,
```
drasko: Same question here.

bootstrap/api/endpoint_test.go
mteodor: `ThingsURL` is used to reach `things` service API endpoints.
Bootstrap test server is created after `newService` with
`bs := newBootstrapServer(svc)`
and then you can test bootstrap endpoints using bs.URL
```
	for _, tc := range cases {
		req := testRequest{
			client:      bs.Client(),
			method:      http.MethodPost,
			url:         fmt.Sprintf("%s/configs", bs.URL),```

File: bootstrap/service_test.go (hunk 3 above)
Comment (mteodor) [reply bootstrap/service_test.go]: answer is similar as above, additionaly providing `ThingsURL` here is configuring bootstrap service so that it can reach `things` service API endpoints to create things, channels, check connections

File: bootstrap/api/endpoint_test.go (hunk 1 above)
Comment (drasko) [reply bootstrap/api/endpoint_test.go:163]: Naming of these functions is confusing. Calling something called `newService()` inside of Boostrap code, I would expect that it gives me back Boostrap service.

Also, take a look at the callers - what URLs do they provide? URLs of Things? How do you know where Things is - this is something that user should tell you via ENV var, no?

File: certs/service_test.go (hunk 2 above)
Comment (mteodor) [reply certs/service_test.go:67]: `server` here is http server serving API calls to things service

Comment (drasko) [reply certs/service_test.go:67]: Yes, serving HTTP API of a *Bootstrap* service, i.e. it is a Bootstrap service `httptest` server. This is why I am saying that this is fundamentally wrong.

File: Makefile [hunk 4]
```diff
@@ -102,12 +102,3 @@ rundev:
 run:
 	docker-compose -f docker/docker-compose.yml up
 
-runlora:
```
drasko, line 105: I removed this clutter because it is rarely launched and should be documented somewhere and not burden the main Makefile.

Makefile
```diff
@@ -102,12 +102,3 @@ rundev:
...
-	docker-compose \
-		-f docker/docker-compose.yml \
-		-f docker/addons/influxdb-writer/docker-compose.yml \
-		-f docker/addons/lora-adapter/docker-compose.yml up \
-
-# Run all Mainflux core services except distributed tracing system - Jaeger. Recommended on gateways:
-rungw:
```
drasko, line 112: This might be useful, but again - rarely used. Should be rather documented somewhere and executd manually when needed.

bootstrap/service_test.go
```diff
@@ -57,7 +57,8 @@ var (
 func newService(auth mainflux.AuthServiceClient, url string) bootstrap.Service {
 	things := mocks.NewConfigsRepository()
 	config := mfsdk.Config{
-		BaseURL: url,
+		BootstrapURL: url,
```
mteodor: we dont need this BootstrapURL

certs/service_test.go
mteodor: not bootstrap, but things

cmd/cli/main.go
```diff
@@ -4,28 +4,28 @@
...
-		CertsURL:          "http://localhost:8204",
-		ReaderPrefix:      "",
-		UsersPrefix:       "",
-		GroupsPrefix:      "",
-		ThingsPrefix:      "",
-		HTTPAdapterPrefix: "http",
-		BootstrapPrefix:   "things",
-		MsgContentType:    sdk.ContentType(msgContentType),
-		TLSVerification:   false,
+		AuthURL:         defURL,
```
mteodor, line 20: I think that we should have preserved BaseURL

pkg/sdk/go/things_test.go
```diff
@@ -737,13 +705,9 @@ func TestConnect(t *testing.T) {
...
-		BaseURL:           ts.URL,
-		UsersPrefix:       "",
-		GroupsPrefix:      "",
-		ThingsPrefix:      "",
-		HTTPAdapterPrefix: "",
-		MsgContentType:    contentType,
-		TLSVerification:   false,
+		ThingsURL:       ts.URL,
+		MsgContentType:  contentType,
+		TLSVerification: false,
```
mteodor, line 710: can we perform tests with TLSVerification = true

provision/configs/config.toml
```diff
@@ -21,9 +21,9 @@ file = "config.toml"
   port = ""
   server_cert = ""
   server_key = ""
-  things_location = "http://localhost:8182"
+  things_url = "http://localhost:8182"
   tls = true
-  users_location = ""
+  users_url = ""
```
dborovcanin: Why not defaulting to the correct Users' service URL? Also, TLS is on, but URLs are `http`.

dborovcanin (approved):

PR #1456, opened 2021-09-03
MF-1454 - Add HTTP endpoint and service for sharing things

### LINKED ISSUE #1454: Policies for Things sharing
**FEATURE REQUEST**

Instead of an ownership-based approach, `Things` should be shared among entities via Policies. So that, each entity that has the policy to access a `Thing` should be able to access the `Thing`. Or, if the entity has no policy to access a Thing, the entity should not access a Thing.

must-have

Some tasks for further steps:
- [ ] After deleting a `thing`, delete the policy related to that particular `thing`. It requires adding a new gRPC endpoint.

buraksekili:

### What does this do?

This PR adds an HTTP endpoint and service for sharing a `Thing` with Policies.

Resolves #1454 

This PR updates `thingRepository`. Instead of the ownership-based approach, it uses policies.

Yes

No


things/service.go
```diff
@@ -247,13 +241,41 @@ func (ts *thingsService) UpdateThing(ctx context.Context, token string, thing Th
...
+			if err != nil {
+				return errors.Wrap(ErrAuthorization, err)
+			}
+			if !apr.GetAuthorized() {
+				return ErrAuthorization
+			}
+		}
+	}
+	return nil
+}
```
dborovcanin: Wrapping multiple requests in a single method that receives a list is convenient for the user, but it comes at the cost of losing atomic operation: this can fail for any of the provided user IDs. I'd say that this way is OK, but we need to improve error handling. If sharing with some users fails, we need to return it in the error response and add policies for all the other users (so keep iterating even if we fail). The other option is to create a bulk endpoint on the Auth service, which will result in similar behavior (only on the Auth side) unless there is a bulk option provided by Keto.

dborovcanin (changes requested):

things/mocks/auth.go
```diff
@@ -42,15 +48,18 @@ func (svc authServiceMock) Issue(ctx context.Context, in *mainflux.IssueReq, opt
...
+		if policy.Relation == req.GetAct() && policy.Object == req.GetObj() {
 			return &mainflux.AuthorizeRes{Authorized: true}, nil
 		}
 	}
 	return nil, things.ErrAuthorization
 }
 
 func (svc authServiceMock) AddPolicy(ctx context.Context, in *mainflux.AddPolicyReq, opts ...grpc.CallOption) (*mainflux.AddPolicyRes, error) {
+	obj := in.GetObj()
+	obj = obj[len(obj)-3:]
```
drasko, line 65: Magic numbers. What are these constant numbers? Should they be declared as a constant somewhere. Also - please add a comment of what's going on here.

things/service.go
drasko: @buraksekili waht is the best way to proceed with this? should we have a meeting on this?

drasko (changes requested):
Swagger documentation needs to be updated as well

things/mocks/auth.go
buraksekili: While creating a `Thing`, `CreateThings` on the service layer assigns a new ID for the thing [as here.](https://github.com/mainflux/mainflux/blob/f4312aef8ac288f97c4db0e2dcb4e70475e620bf/things/service.go#L170)

Then, the `mock Things repository` saves this thing ID by padding [as here](https://github.com/mainflux/mainflux/blob/f4312aef8ac288f97c4db0e2dcb4e70475e620bf/things/mocks/things.go#L58)

Since we are adding policies within the Service layer, we are storing them as a full ID which is eventually not compatible with the one inside of the `mock things repository`.

Therefore, I used the padded part while adding a policy. That's why I used 
```go
obj = obj[len(obj)-3:]
```

things/service.go
buraksekili: I would say `If sharing with some users fails, we need to return it in the error response and add policies for all the other users (so keep iterating even if we fail). `

dborovcanin: I would say so, as well. But we need to design a nice error response for this.  Something like:
```go
func (ts *thingsService) claimOwnership(ctx context.Context, thingID string, policies, userIDs []string) error {
	var errors Error
        for _, userID := range userIDs {
		for _, policy := range policies {
			apr, err := ts.auth.AddPolicy(ctx, &mainflux.AddPolicyReq{Obj: thingID, Act: policy, Sub: userID})
			if err != nil {
				accumulateError(errors, errors.Wrap(ErrAuthorization, err), thingID, userID)
			}
			if !apr.GetAuthorized() {
				accumulatError(errors, ErrAuthorization, thingID, userID)
			}
		}
	}
	return errors
}
```

drasko (changes requested):
I must say I did not understand why we need to share with one by one user - why can't we share with the list of users?

things/mocks/auth.go
drasko: OK, but then needs a comment.

drasko:
Out of sync, and difficult to put back in sync.

Closing this one and waiting for new PR from the freshly forked branch (in sync).

PR #1458, opened 2021-09-08
WIP:influxdb writer json msg

qs-wang:


qs-wang:
Sorry, this is for our own internal logic. Created in wrong repo

PR #1463, opened 2021-09-17
MF-1454 - Add Policies for sharing a Thing

### LINKED ISSUE #1454: Policies for Things sharing

buraksekili:

### What does this do?

This PR adds an HTTP endpoint and service for sharing a Thing via Policies.

Resolves #1454 

This PR updates `thingRepository`. Instead of the ownership-based approach, it uses policies.
Yes
Yes
### Notes

The HTTP endpoint for sharing a thing can be tested via:

```bash
curl -isSX POST http://localhost/things/<thing_id>/share -d '{"user_ids":["<user_id>"], "policies": ["read", "write", "delete"]}' -H "Authorization: <user_auth_token>" -H 'Content-Type: application/json'
```

### DIFF
--- a/things/service.go
+++ b/things/service.go
@@ -64,8 +64,10 @@ type Service interface {
 	// belongs to the user identified by the provided key.
 	UpdateThing(ctx context.Context, token string, thing Thing) error
 
-	// ShareThing gives an access policy of thingID to the given user IDs.
-	ShareThing(ctx context.Context, token, thingID string, policies, userIDs []string) error
+	// ShareThing gives actions associated with the thing to the given user IDs.
+	// The requester user identified by the token has to have a "write" relation
+	// on the thing in order to share the thing.
+	ShareThing(ctx context.Context, token, thingID string, actions, userIDs []string) error
 
 	// UpdateKey updates key value of the existing thing. A non-nil error is
 	// returned to indicate operation failure.
@@ -242,7 +244,7 @@ func (ts *thingsService) UpdateThing(ctx context.Context, token string, thing Th
 	return ts.things.Update(ctx, thing)
 }
 
-func (ts *thingsService) ShareThing(ctx context.Context, token, thingID string, policies, userIDs []string) error {
+func (ts *thingsService) ShareThing(ctx context.Context, token, thingID string, actions, userIDs []string) error {
 	res, err := ts.auth.Identify(ctx, &mainflux.Token{Value: token})
 	if err != nil {
 		return errors.Wrap(ErrUnauthorizedAccess, err)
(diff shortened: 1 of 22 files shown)


things/service.go
```diff
@@ -247,13 +242,46 @@ func (ts *thingsService) UpdateThing(ctx context.Context, token string, thing Th
...
+				errs = accumulateError(errs, ErrAuthorization, thingID, userID)
+			}
+		}
+	}
+	return errs
+}
+
+func accumulateError(errs, newErr error, thingID, userID string) error {
+	return errors.Wrap(newErr, errors.Wrap(fmt.Errorf("cannot claim ownership on thing: %s by user: %s", thingID, userID), errs))
+}
```
dborovcanin: No need for a separate function and there is no need to wrap each error to `ErrAuthorization`. Something like:
```go
func (ts *thingsService) claimOwnership(ctx context.Context, thingID string, policies, userIDs []string) error {
	var errs error
	for _, userID := range userIDs {
		for _, policy := range policies {
			apr, err := ts.auth.AddPolicy(ctx, &mainflux.AddPolicyReq{Obj: thingID, Act: policy, Sub: userID})
			if err != nil {
                                errs = errors.Wrap(fmt.Errorf("Cannot claim ownership on thing '%s' by user '%s': %s", thingID, userID, err), errs))
			}
			if !apr.GetAuthorized() {
				errs = errors.Wrap(fmt.Errorf("Cannot claim ownership on thing '%s' by user '%s': unauthorized", thingID, userID), errs))
			}
		}
	}
        if errs != nil {
                return errors.Wrap(ErrAuthorization, errs)
        }
	return nil
}
```
will do.

things/service_test.go
```diff
@@ -154,6 +155,55 @@ func TestUpdateKey(t *testing.T) {
...
 
+func TestShareThing(t *testing.T) {
+	email2 := "user2@example.com"
+	svc := newService(map[string]string{token: email, token2: email2})
+	ths, err := svc.CreateThings(context.Background(), token, thing)
+	require.Nil(t, err, fmt.Sprintf("unexpected error: %s\n", err))
+	th := ths[0]
+	policies := []string{"read"}
+
+	cases := []struct {
```
dborovcanin, line 166: Test for partial success (some failed, some successfully shared) is missing.

dborovcanin (changes requested):

things/service_test.go
```diff
@@ -154,6 +155,63 @@ func TestUpdateKey(t *testing.T) {
 	}
 }
 
+func TestShareThing(t *testing.T) {
+	email2 := "user2@example.com"
+	svc := newService(map[string]string{token: email, token2: email2})
+	ths, err := svc.CreateThings(context.Background(), token, thing)
+	require.Nil(t, err, fmt.Sprintf("unexpected error: %s\n", err))
+	th := ths[0]
+	policies := []string{"read"}
```
dborovcanin: Change order; prepare test (create service, create things, require nil erro) and declare variables for different cases after:
```go
	svc := newService(map[string]string{token: email, token2: email2})
	ths, err := svc.CreateThings(context.Background(), token, thing)
	require.Nil(t, err, fmt.Sprintf("unexpected error: %s\n", err))
	th := ths[0]
        email2 := "user2@example.com"
	policies := []string{"read"}
```

things/service.go
```diff
@@ -247,13 +242,42 @@ func (ts *thingsService) UpdateThing(ctx context.Context, token string, thing Th
 	return ts.things.Update(ctx, thing)
 }
 
+func (ts *thingsService) ShareThing(ctx context.Context, token, thingID string, policies, userIDs []string) error {
```
dborovcanin: Is `policies` parameter named ambiguously? It represents the action, not a list of `policies` per se, it's not "policy ids" or "policy names"?

buraksekili: Yes, it represents actions related to Things policies: `access`, `write` and `delete`. However, in the essence, they are `relations (actions)`, not policies. I'll update the naming.

dborovcanin (approved):

blokovi (approved):

PR #1466, opened 2021-09-24
MF-1454 - Policy Removal 

buraksekili:
### What does this do?

This PR adds a new gRPC endpoint for policy removal. It is still WIP.

Resolves a task described in #1454 
Tests are not fully completed

No


docker/.env
```diff
@@ -56,6 +56,7 @@ MF_USERS_ADMIN_EMAIL=admin@example.com
 MF_USERS_ADMIN_PASSWORD=12345678
 MF_USERS_RESET_PWD_TEMPLATE=users.tmpl
 MF_USERS_PASS_REGEX=^.{8,}$
+MF_ONLY_ADMIN_CREATES_USER=
```
drasko, line 59: Is `MF_ONLY_ADMIN_CREATES_USER` a good name for ENV var? Would it better to call it `MF_SELF_SIGNON`, and if this one is false, then only admin can create users. Or am I missing something?

drasko (changes requested):

docker/.env
buraksekili: `MF_SELF_SIGNON` is okay for me. I'll rename the environment variable.

By default, everybody can create a new user. If `MF_ONLY_ADMIN_CREATES_USER` is empty, everybody is allowed to create a new user. Otherwise (`MF_ONLY_ADMIN_CREATES_USER` is not empty, i.e., an arbitrary string), only the admin is allowed to create a new user.

drasko: Ah, then this is not the same as self-sign-on. OK, we can keep this one.

dborovcanin: I still don't like this naming, but let's fix it in the next PR.

dborovcanin (approved):

docker/.env
drasko: Yes, @buraksekili please do not forget to fix one in the next PR

things/users/users.go
```diff
@@ -66,6 +66,10 @@ func (repo singleUserRepo) AddPolicy(ctx context.Context, in *mainflux.AddPolicy
 	return &mainflux.AddPolicyRes{}, errUnsupported
 }
 
+func (repo singleUserRepo) DeletePolicy(ctx context.Context, in *mainflux.DeletePolicyReq, opts ...grpc.CallOption) (*mainflux.DeletePolicyRes, error) {
+	return &mainflux.DeletePolicyRes{}, errUnsupported
```
drasko, line 70: @buraksekili please do not forget to check this one with @dusanb94

PR #1470, opened 2021-10-06
MF-1257 - Access messages from readers endpoint with user access token

### LINKED ISSUE #1257: Reader endpoint that retrieves messages related to user
**FEATURE REQUEST**

Currently there is endpoint that retrieves messages from channels that are connected to thing passed in request.
There should be endpoint where you can retrieve messages for channels that belong to user filtered by some criteria.

Must have

mteodor:
### What does this do?
Enables user to read messages from channels he owns

Resolves #1257.

None
No


cmd/cassandra-reader/main.go
```diff
@@ -112,6 +127,32 @@ func main() {
 	logger.Error(fmt.Sprintf("Cassandra reader service terminated: %s", err))
 }
 
+func connectToAuth(cfg config, logger logger.Logger) *grpc.ClientConn {
+	var opts []grpc.DialOption
+	logger.Info("connecting to auth via gRPC")
```
manuio: Here and for all readers I'd move this log after to succeed the connection. something like `Established gRPC connection with to auth service`

manuio (changes requested):

readers/api/transport.go
```diff
@@ -210,25 +212,44 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 	}
 }
 
-func authorize(r *http.Request, chanID string) error {
+func authorize(r *http.Request, chanID string) (err error) {
 	token := r.Header.Get("Authorization")
 	if token == "" {
 		return errUnauthorizedAccess
 	}
+	// if strings.Contains(token, "Bearer ") {
```
manuio: Dead code

readers/api/transport.go
```diff
@@ -210,25 +212,44 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
...
 
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
 
-	_, err := auth.CanAccessByKey(ctx, &mainflux.AccessByKeyReq{Token: token, ChanID: chanID})
+	t, err := auth.CanAccessByKey(ctx, &mainflux.AccessByKeyReq{Token: token, ChanID: chanID})
+	if err == nil {
+		return nil
+	}
+	fmt.Println(t)
```
manuio: Remove this log

manuio (changes requested):

readers/auth.go
```diff
@@ -0,0 +1,44 @@
...
+	ts mainflux.ThingsServiceClient
+	us mainflux.AuthServiceClient
+}
+
+type Auth interface {
+	Identify(ctx context.Context, in *mainflux.Token, opts ...grpc.CallOption) (*mainflux.UserIdentity, error)
+	CanAccessByKey(ctx context.Context, in *mainflux.AccessByKeyReq, opts ...grpc.CallOption) (*mainflux.ThingID, error)
+	IsChannelOwner(ctx context.Context, in *mainflux.ChannelOwnerReq, opts ...grpc.CallOption) (*empty.Empty, error)
+	CanAccessByID(ctx context.Context, in *mainflux.AccessByIDReq, opts ...grpc.CallOption) (*empty.Empty, error)
+}
```
dborovcanin: I would say this abstraction is unnecessary. We can simply have two clients (`mainflux.ThingsServiceClient` and `mainflux.AuthServiceClient`) and chose between them depending on the context.

readers/api/transport.go
```diff
@@ -210,25 +210,35 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
...
 	token := r.Header.Get("Authorization")
 	if token == "" {
 		return errUnauthorizedAccess
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
-	defer cancel()
+	if _, err := auth.CanAccessByKey(ctx, &mainflux.AccessByKeyReq{Token: token, ChanID: chanID}); err == nil {
+		return nil
+	}
```
dborovcanin: I'd prefer if we put authorization type somewhere ( `Authorization` header prefix or in a separate header or something like that) so that we know which method to use, rather than implicitly fallback to user if thing authorization fails.

dborovcanin (changes requested):

readers/api/transport.go
mteodor: That sounds ok to me, I propose the following 
`Authorization: <type> <value>`
So for thing key - `Authorization: Thing <value>`
And for user token - `Authorization: Bearer <value>`
What do you say?
Seems like the right way - https://stackoverflow.com/a/43164958/2873439
@drasko ?

dborovcanin: Makes sense. The problem is that the prefix will break compatibility with the current authorization header. We can support both (`Things` prefix and no prefix for Thing authorization, `Bearer` for user) to give some space for migration and make prefix required in the next release. Despite the most popular schemas being `Basic` and `Bearer`, I don't see anything wrong with the custom `Thing` schema for thing authorization.

readers/api/transport.go
```diff
@@ -210,25 +218,42 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
...
 }
 
-func authorize(r *http.Request, chanID string) error {
+func authorize(ctx context.Context, r *http.Request, chanID string) (err error) {
 	token := r.Header.Get("Authorization")
 	if token == "" {
 		return errUnauthorizedAccess
 	}
+	if strings.HasPrefix(token, thingToken) {
+		token = strings.ReplaceAll(token, thingToken, "")
```
dborovcanin: Use `strings.TrimPrefix`.

readers/api/transport.go
```diff
@@ -210,25 +218,42 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 	}
 }
 
-func authorize(r *http.Request, chanID string) error {
+func authorize(ctx context.Context, r *http.Request, chanID string) (err error) {
```
dborovcanin: Consider using an empty `switch-case`, like:
```go
	switch {
	case strings.HasPrefix(token, thingToken):
		.....
	case strings.HasPrefix(token, userToken):
		.....
	default:
		return errors.Wrap(errUnauthorizedAccess, errWrongToken)
	}
```
Looks like a nice fit.

readers/messages.go
```diff
@@ -3,7 +3,9 @@
 
 package readers
 
-import "errors"
+import (
+	"errors"
+)
```
dborovcanin:
```go
import "errors"
```

dborovcanin (changes requested):

dborovcanin (dismissed):

api/readers.yml
```diff
@@ -99,7 +99,9 @@ components:
   parameters:
     Authorization:
       name: Authorization
-      description: Thing access token.
+      description: Thing or User access token. 
+      For thing access use "Authorization: Thing <thing_key>"
+      For user access use "Authorization: Bearer <user_token>"
```
manuio: Are you sure about this section? It shoudl be
```
      description: | 
          Thing or User access token: 
          * For thing access use "Authorization: Thing <thing_key>"
          * For user access use "Authorization: Bearer <user_token>"
```

readers/api/transport.go
```diff
@@ -39,16 +39,24 @@ const (
...
+	errUserAccess          = errors.New("user has no permission")
+	errWrongToken          = errors.New("incorrect or missing Authorization header")
+	thingsAuth             mainflux.ThingsServiceClient
+	usersAuth              mainflux.AuthServiceClient
 )
 
 // MakeHandler returns a HTTP handler for API endpoints.
-func MakeHandler(svc readers.MessageRepository, tc mainflux.ThingsServiceClient, svcName string) http.Handler {
-	auth = tc
+func MakeHandler(svc readers.MessageRepository, tc mainflux.ThingsServiceClient, uc mainflux.AuthServiceClient, svcName string) http.Handler {
```
drasko: I would call it `ac` not `uc`

drasko (changes requested):

readers/api/transport.go
```diff
@@ -39,16 +39,24 @@ const (
...
+	thingsAuth             mainflux.ThingsServiceClient
+	usersAuth              mainflux.AuthServiceClient
 )
 
 // MakeHandler returns a HTTP handler for API endpoints.
-func MakeHandler(svc readers.MessageRepository, tc mainflux.ThingsServiceClient, svcName string) http.Handler {
-	auth = tc
+func MakeHandler(svc readers.MessageRepository, tc mainflux.ThingsServiceClient, ac mainflux.AuthServiceClient, svcName string) http.Handler {
+	thingsAuth = tc
+	usersAuth = ac
```
drasko, line 56: Aren't users authenticated via Users service? Why using `AuthServiceClient` then, which b the name should connect to Auth service (which can't identify the user, just read the policies)?

drasko (changes requested):

readers/api/transport.go
mteodor: we are using `AuthServiceClient`  to check the token passed as `Authorization` header in HTTP request, token that was issued by `auth` when user was authenticated

drasko:
@mteodor what would be the best use-case where user, with his account, needs to read data, and he cannot create one app to which he (or many other users) can give api keys?

I am asking this, because in my opinion this can be solved with one app (Thing) and API keys features that we already have.

cmd/cassandra-reader/main.go
```diff
@@ -75,7 +78,9 @@ type config struct {
 	serverKey         string
 	jaegerURL         string
 	thingsAuthURL     string
+	usersAuthURL      string
```
dborovcanin, line 83: How do we set this, since there is no env var added? Also, when you load config, this field should be populated. Please check this for all the other readers.

dborovcanin (changes requested):

mteodor:
>

mteodor:
> I am asking this, because in my opinion this can be solved with one app (Thing) and API keys features that we already have.

@drasko, API keys will not work without this change

readers/api/endpoint_test.go
```diff
@@ -138,7 +145,298 @@ func TestReadAll(t *testing.T) {
...
+			url:    fmt.Sprintf("%s/channels/%s/messages?offset=0", ts.URL, chanID),
+			token:  fmt.Sprintf("Thing %s", thingToken),
+			status: http.StatusOK,
+			res: pageRes{
+				Total:    uint64(len(messages)),
+				Messages: messages[0:10],
+			},
+		},
+		{
+			desc:   "read page with senml fornat",
```
dborovcanin: Fix typo: `format`.

readers/api/transport.go
```diff
@@ -210,25 +218,41 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
...
-	defer cancel()
-
-	_, err := auth.CanAccessByKey(ctx, &mainflux.AccessByKeyReq{Token: token, ChanID: chanID})
-	if err != nil {
-		e, ok := status.FromError(err)
-		if ok && e.Code() == codes.PermissionDenied {
-			return errUnauthorizedAccess
+	switch {
+	case strings.HasPrefix(token, thingToken):
+		token = strings.TrimPrefix(token, thingToken)
```
dborovcanin: The problem with this approach is that both Thing and User are `Bearers`, so the `Bearer` prefix makes sense in both cases.

mteodor: so what do you propose?
I'll remind you on this conversation https://github.com/mainflux/mainflux/pull/1470#discussion_r730734378

drasko:
@dusanb94 please let's move this one forward

dborovcanin (changes requested):
Please don't forget to update README files with the new env variable.

readers/influxdb/README.md
```diff
@@ -23,6 +23,9 @@ default values.
 | MF_JAEGER_URL                | Jaeger server URL                                   | localhost:6831 |
 | MF_THINGS_AUTH_GRPC_URL      | Things service Auth gRPC URL                        | localhost:8181 |
 | MF_THINGS_AUTH_GRPC_TIMEOUT  | Things service Auth gRPC request timeout in seconds | 1s             |
+| MF_AUTH_GRPC_URL             | Auth service gRPC URL                               | localhost:8181 |
+| MF_AUTH_GRPC_TIMEOUT         | Auth service gRPC request timeout in seconds        | 1s             |
```
manuio: Do we need `MF_THINGS_AUTH_GRPC_URL` and `MF_THINGS_AUTH_GRPC_TIMEOUT` anymore? I'd even replace it in other services.

mteodor: yes we do

dborovcanin (dismissed):

api/readers.yml
```diff
@@ -99,7 +99,10 @@ components:
   parameters:
     Authorization:
       name: Authorization
-      description: Thing access token.
+      description: | 
+        Thing or User access token: 
+        * For thing access use "Authorization: Thing <thing_key>"
```
drasko: This is something that should be changed, probably.

readers/api/transport.go
```diff
@@ -39,16 +39,24 @@ const (
 	defLimit       = 10
 	defOffset      = 0
 	defFormat      = "messages"
+	thingToken     = "Thing "
```
drasko: Why there is an empty space after the word? Does not seem quite right... Anyway, we should probably not use custom header tagsm as authentication schemes are limited to [these](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authorization_and_proxy-authorization_headers).

drasko: On the other hand - there are [variants](https://www.loginradius.com/blog/async/everything-you-want-to-know-about-authorization-headers/), and AWS is using [this](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html) for example.

drasko (changes requested):

readers/api/transport.go
mteodor: @dusanb94 and I already discussed this and I think we agreed with custom header authentication as I saw that it is widely being used ( and looks like AWS example testifies that ).
My initial idea was just to use `Authorization: <token_value|thing_key>` or  `Authorization:Bearer <token_value|thing_key>`, if one fails try other. If discard that solution I don't know any other way than to use different prefix to specify different authentication.

readers/api/transport.go
```diff
@@ -39,16 +39,24 @@ const (
...
 )
 
 var (
-	errUnauthorizedAccess = errors.New("missing or invalid credentials provided")
-	auth                  mainflux.ThingsServiceClient
+	errUnauthorizedAccess  = errors.New("missing or invalid credentials provided")
+	errCannotAuthorizeUser = errors.New("authorization failed")
+	errThingAccess         = errors.New("thing has no permission")
+	errUserAccess          = errors.New("user has no permission")
+	errWrongToken          = errors.New("incorrect or missing Authorization header")
```
dborovcanin: This one looks like it's not used.

dborovcanin: It would be good to align with this: https://github.com/mainflux/mainflux/pull/1538#discussion_r784783227 (these errors will be a part of `pkg/errors/types.go` once #1538 is merged).

dborovcanin (changes requested):

readers/api/transport.go
```diff
@@ -44,7 +44,7 @@ const (
 )
 
 var (
-	errUnauthorizedAccess  = errors.New("missing or invalid credentials provided")
+	errEmptyToken          = errors.New("missing token")
```
manuio: You can use `errors.ErrAuthentication` and remove this one

readers/api/transport.go
```diff
@@ -82,7 +82,7 @@ func decodeList(ctx context.Context, r *http.Request) (interface{}, error) {
 	}
 
 	if err := authorize(ctx, r, chanID); err != nil {
-		return nil, err
+		return nil, errors.ErrAuthorization
```
manuio: You are ignoring the custom errors returned by this method and they are probably already wrapped with `errors.Authorization`. Can you check it please? You can probably return `err` here.

readers/api/transport.go
```diff
@@ -229,23 +229,23 @@ func authorize(ctx context.Context, r *http.Request, chanID string) (err error)
 		if err != nil {
 			e, ok := status.FromError(err)
 			if ok && e.Code() == codes.PermissionDenied {
```
manuio: You can inline here

readers/api/transport.go
```diff
@@ -229,23 +229,23 @@ func authorize(ctx context.Context, r *http.Request, chanID string) (err error)
...
-				return errUnauthorizedAccess
+				return errCannotAuthorizeUser
 			}
-			return errors.Wrap(errUnauthorizedAccess, errCannotAuthorizeUser)
+			return errCannotAuthorizeUser
```
manuio: You are always returning `errCannotAuthorizeUser`. Do we need to check if `e.Code() == codes.PermissionDenied`? Also, can we use `errors.ErrAuthorization` instead of the custom one?

manuio (changes requested):

readers/api/transport.go
```diff
@@ -210,25 +217,38 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 	}
 }
 
-func authorize(r *http.Request, chanID string) error {
+func authorize(ctx context.Context, r *http.Request, chanID string) (err error) {
 	token := r.Header.Get("Authorization")
 	if token == "" {
```
manuio: Consider moving this to validation in resquests.go

readers/api/transport.go
```diff
@@ -210,25 +217,38 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
...
-
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
-	defer cancel()
-
-	_, err := auth.CanAccessByKey(ctx, &mainflux.AccessByKeyReq{Token: token, ChanID: chanID})
-	if err != nil {
-		e, ok := status.FromError(err)
-		if ok && e.Code() == codes.PermissionDenied {
-			return errUnauthorizedAccess
+	switch {
```
manuio, line 212: This switch case should be probaly moved  after the validation in endpoint layer

readers/api/requests.go
```diff
@@ -14,10 +14,14 @@ type apiReq interface {
...
 type listMessagesReq struct {
 	chanID   string
+	token    string
 	pageMeta readers.PageMetadata
 }
 
 func (req listMessagesReq) validate() error {
+	if req.token == "" {
+		return errors.ErrAuthorization
+	}
```
manuio, line 24: Looks like `chanID` check is missing

manuio (changes requested):

readers/api/transport.go
```diff
@@ -207,25 +209,34 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
...
+		if err != nil {
+			e, ok := status.FromError(err)
+			if ok && e.Code() == codes.PermissionDenied {
+				return errCannotAuthorizeUser
+			}
+			return errCannotAuthorizeUser
+		}
+		_, err = thingsAuth.IsChannelOwner(ctx, &mainflux.ChannelOwnerReq{Owner: user.Email, ChanID: req.chanID})
+		if err != nil {
+			e, ok := status.FromError(err)
```
drasko, line 224: Move this line of initialization below, in `if` statement.

readers/api/transport.go
```diff
@@ -207,25 +209,34 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
...
-		e, ok := status.FromError(err)
-		if ok && e.Code() == codes.PermissionDenied {
-			return errors.ErrAuthorization
+func authorize(ctx context.Context, req listMessagesReq, tc mainflux.ThingsServiceClient, ac mainflux.AuthServiceClient) (err error) {
+	switch {
+	case strings.HasPrefix(req.token, userToken):
+		token := strings.TrimPrefix(req.token, userToken)
+		user, err := usersAuth.Identify(ctx, &mainflux.Token{Value: token})
+		if err != nil {
+			e, ok := status.FromError(err)
```
drasko, line 217: Inline with the line below

drasko (changes requested):

readers/api/transport.go
```diff
@@ -195,6 +195,8 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 	case errors.Contains(err, nil):
 	case errors.Contains(err, errors.ErrInvalidQueryParams):
 		w.WriteHeader(http.StatusBadRequest)
+	case errors.Contains(err, errors.ErrMalformedEntity):
+		w.WriteHeader(http.StatusBadRequest)
```
manuio: You can do:
```
case errors.Contains(err, errors.ErrInvalidQueryParams),
	errors.Contains(err, errors.ErrMalformedEntity):
```

bootstrap/service_test.go
```diff
@@ -60,6 +61,9 @@ func newService(auth mainflux.AuthServiceClient, url string) bootstrap.Service {
 		ThingsURL: url,
 	}
 
+	t := map[string]interface{}{}
+	t["test"]= "test"
+	json.Marshal(t)
```
dborovcanin: What's this?

readers/api/transport.go
dborovcanin: Actually, since `user` is used later, there is no need to inline. This is just fine:
```go

user, err = usersAuth.Identify(ctx, &mainflux.Token{Value: token})
if err != nil {
...}
use user here
....
```

readers/api/endpoint_test.go
```diff
@@ -138,7 +145,298 @@ func TestReadAll(t *testing.T) {
...
-			token:  token,
+			token:  fmt.Sprintf("Thing %s", thingToken),
+			status: http.StatusOK,
+			res: pageRes{
+				Total:    uint64(len(messages)),
+				Messages: messages[0:10],
+			},
+		},
+		{
+			desc:   "read page with valid offset and limit",
```
dborovcanin: It's a good idea to introduce the type of client (thing or user) in the case description.

dborovcanin (changes requested):

readers/api/transport.go
dborovcanin: For code readability, I would maybe keep it this way. I leave it to @mteodor to decide.

readers/api/transport.go
dborovcanin: @mteodor Please address this remark.

readers/api/transport.go
dborovcanin: @mteodor Please address this remark.

dborovcanin (changes requested):

readers/api/transport.go
mteodor: I would leave it as it is

readers/api/transport.go
```diff
@@ -39,21 +39,29 @@ const (
...
 	defOffset      = 0
 	defFormat      = "messages"
+	thingToken     = "Thing "
+	userToken      = "Bearer "
 )
 
-var auth mainflux.ThingsServiceClient
+var (
+	errThingAccess = errors.New("thing has no permission")
+	errUserAccess  = errors.New("user has no permission")
```
dborovcanin, line 48: These errors should be handled in the `encodeError` to return `StatuForbidden`, not `InternalServerError`(by default).

dborovcanin (changes requested):

readers/api/transport.go
mteodor: these errors wrapped intot `ErrAuthorization` which results in `StatusUnauthorized`
```
if err := authorize(ctx, req, tc, ac); err != nil {
			return nil, errors.Wrap(errors.ErrAuthorization, err)
		}
```
```
	case errors.Contains(err, errors.ErrAuthentication):
```

dborovcanin: Missed that one, thanks. 👍🏼

dborovcanin (approved):

manuio (approved):
LGTM

manuio (dismissed):

dborovcanin (dismissed):

readers/api/transport.go
drasko: This is OK, but it should be `"Thing"`, not `"Thing "`.

readers/api/transport.go
```diff
@@ -39,21 +39,29 @@ const (
 	defLimit       = 10
 	defOffset      = 0
 	defFormat      = "messages"
+	thingToken     = "Thing "
+	userToken      = "Bearer "
```
drasko: Same here - `"Bearer"`, not `"Bearer "`.

drasko (changes requested):

dborovcanin (approved):

manuio (approved):
LGTM

drasko (approved):
LGTM

drasko:
Thanks @mteodor - this was a big one! Great contribution.

PR #1471, opened 2021-10-07
NOISSUE - Refactor single-user mode

dborovcanin:
Signed-off-by: dusanb94 <dusan.borovcanin@mainflux.com

### What does this do?
This pull request refactors single-user mode for Things service.

There is no such issue.

There are no such changes.

Yes.

Yes. Docs explaining single-user mode are added.

### Notes
N/A


cmd/twins/main.go
```diff
@@ -23,7 +23,7 @@ import (
 	"github.com/mainflux/mainflux/pkg/messaging"
 	"github.com/mainflux/mainflux/pkg/messaging/nats"
 	"github.com/mainflux/mainflux/pkg/uuid"
-	localusers "github.com/mainflux/mainflux/things/users"
+	localusers "github.com/mainflux/mainflux/things/auth"
```
drasko: My only cocncern is should this one be called `auth`, because again when we take a look at `things` folder in a couple of months and when we see `auth` - will we know it is a Single-User Mode related module? I was confused like this with previous naming, which was `user`.

So just one thing to conside - should it be something like `singleuser` (or something like this, self-documenting), or we leave `auth` (I am fine with `auth` also, just challenging the solution a bit).

things/README.md
```diff
@@ -101,6 +101,10 @@ $GOBIN/mainflux-things
 
 Setting `MF_THINGS_CA_CERTS` expects a file in PEM format of trusted CAs. This will enable TLS against the Users gRPC endpoint trusting only those CAs that are provided.
 
+In constrained environments, sometimes it makes sense to run Things service as a standalone to reduce network traffic and simplify deployment. This means that Things service
```
drasko, line 104: I think a more detailed documentation in the `mainflux/docs` repo is needed, with detailed explication how this single-user mode is deployed (Dockerfile) and used.

I.e. - there should be a dedicated chapter (or subsection in Authorization) for Single-User Mode, so that people see and understand that this feature exists (otherwise, it is unnoticed).

drasko (changes requested):

things/README.md
mteodor: yes, but one paragraph is sufficient for this, since you just set the environment variables and one example how `things` api is used then, additionally advise to remove auth service from compose when using this mode

cmd/twins/main.go
dborovcanin: I actually spent some time thinking about the naming and I decided to go with `auth` and `single_user` for go files. The reason is that this **is** implementation of `auth` for `single_user`. It is not ideal, but I find it a bit less confusing than `users`. Also, `standalone` sounds a bit better to me than `single_user`, but I did not want to rename env vars to provide backward compatibility. @drasko What do you think about that `auth` and `standalone`?

things/README.md
dborovcanin: I agree with @mteodor, I'd say this is good enough for service README. I'll add suggestion for removing `auth` from compose and we can add more details in a separate section in docs.

dborovcanin: I renamed `single-user` to `standalone`, please review.

drasko (approved):
LGTM

PR #1474, opened 2021-10-11
MF-1468 - Fix ThingsURL in Certs Service

### LINKED ISSUE #1468: certs can't contact things svc lacking proper configuration
Hi,
I'm trying to run certs with vault to generate certificates

calling 
curl -s -S  -X POST  http://localhost:8204/certs -H "Authorization: $TOK" -H 'Content-Type: application/json'   -d '{"thing_id":"5ed91a85-a325-458c-b816-aacde1472b38", "rsa_bits":2048, "key_type":"rsa"}'

getting the below error
{"level":"warn","message":"Method issue_cert for token: xxxxx and thing: 5ed91a85-a325-458c-b816-aacde1472b38 took 1.6573ms to complete with error: failed to create client certificate : Get \"/things/5ed91a85-a325-458c-b816-aacde1472b38\": unsupported protocol scheme \"\".","ts":"2021-10-05T22:47:26.8411872Z"}

after reviewing the source code for certs it looks like initializing the sdk doesn't pass the thingsURL

func newService(auth mainflux.AuthServiceClient, db *sqlx.DB, logger mflog.Logger, esClient *redis.Client, tlsCert tls.Certificate, x509Cert *x509.Certificate, cfg config, pkiAgent vault.Agent) certs.Service {
........
	config := mfsdk.Config{
		CertsURL: cfg.certsURL,
	}
i fixed to be, also had to add a few lines to get the ThingsURL from Environment Variables
	config := mfsdk.Config{
		CertsURL: cfg.certsURL,
		ThingsURL: cfg.thingsURL,
	}
and it works now , is this a bug ? should i submit a PR to fix ?

smartyfi:
### What does this do?
Certs Service  fails to contact the Things Service
Resolves #1468

no breaks as default value for ThingsURL is http://things:8182
while new MF_CERTS_THINGS_LOCATION env variable added


cmd/certs/main.go
```diff
@@ -88,7 +89,7 @@ const (
 	envJaegerURL     = "MF_JAEGER_URL"
 	envAuthURL       = "MF_AUTH_GRPC_URL"
 	envAuthTimeout   = "MF_AUTH_GRPC_TIMEOUT"
-
+	envThingsURL      = "MF_CERTS_THINGS_LOCATION"
```
drasko: I do not like the fact that this is called `MF_CERTS_THINGS_LOCATION` and not `MF_CERTS_THINGS_URL` like in other cases. Also - if this is just Things service URL, should it not be just MF_THINGS_URL.

Can you please check how this is handled in other services and align.

drasko (changes requested):

smartyfi:
i tried to match the provision service same syntax 
MF_PROVISION_USERS_LOCATION=http://users:8180
MF_PROVISION_THINGS_LOCATION=http://things:8182

but digging around more i noticed it is related to initaite the SDK, other services looks like MF_SDK_BASE_URL
so i would recommend to go with MF_SDK_THINGS_URL
what do you think ?

cmd/certs/main.go
drasko: I think that it actually should be `MF_CERTS_URL`, but I let @mteodor and @dusanb94 confirm.

dborovcanin:
> i tried to match the provision service same syntax MF_PROVISION_USERS_LOCATION=http://users:8180 MF_PROVISION_THINGS_LOCATION=http://things:8182
> 
> but digging around more i noticed it is related to initaite the SDK, other services looks like MF_SDK_BASE_URL so i would recommend to go with MF_SDK_THINGS_URL what do you think ?

Since the usage of SDK is transparent from the end-user perspective, `MF_SDK_*` does not bring too much value. We should probably change it in the Bootstrap service, as well. The same goes for Provision svc: instead of `MF_PROVISION_USERS_LOCATION` simply `MF_THINGS_URL` I understand that some of these env vars can overlap when running locally (and that's why we added these prefixes), but I'd say that we should simplify. You'll have to specify service config when running it anyway, and simplification wins. So, I'd always use `MF_THINGS_URL`, `MF_USERS_URL`, `MF_<SVC>_URL`... What do you think, @mainflux/maintainers?

drasko:
@smartyfi can you please apply suggestions that @dusanb94 proposed and push forward this useful PR.

smartyfi:
updated env variable to MF_THINGS_URL

drasko (approved):
LGTM

drasko:
Merged! Thanks @smartyfi for the contribution.

cmd/certs/main.go
```diff
@@ -54,6 +54,7 @@ const (
 	defServerCert    = ""
 	defServerKey     = ""
 	defCertsURL      = "http://localhost"
+	defThingsURL      = "http://things:8182"
```
manuio, line 57: Tabulation is not good here

cmd/certs/main.go
```diff
@@ -88,7 +89,7 @@ const (
 	envJaegerURL     = "MF_JAEGER_URL"
 	envAuthURL       = "MF_AUTH_GRPC_URL"
 	envAuthTimeout   = "MF_AUTH_GRPC_TIMEOUT"
-
+	envThingsURL     = "MF_THINGS_URL"
```
manuio, line 92: Same here

cmd/certs/main.go
```diff
@@ -220,6 +222,7 @@ func loadConfig() config {
 		serverCert:  mainflux.Env(envServerCert, defServerCert),
 		serverKey:   mainflux.Env(envServerKey, defServerKey),
 		certsURL:    mainflux.Env(envCertsURL, defCertsURL),
+		thingsURL:    mainflux.Env(envThingsURL, defThingsURL),
```
manuio, line 225: and here

PR #1480, opened 2021-10-20
NOISSUE - Add tests for AddPolicies

buraksekili:

### What does this do?

This PR introduces an endpoint and service test for `AddPolicies`. Also, it adds verification for the request body for adding new policies. So that, arbitrary policies are not allowed and It returns `400` with a `malformed entity` error in case of unknown policy requests.

Yes

### DIFF
--- a/auth/api/http/policies/endpoint_test.go
+++ b/auth/api/http/policies/endpoint_test.go
@@ -1,0 +1,205 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package policies_test
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/mainflux/mainflux/auth"
+	httpapi "github.com/mainflux/mainflux/auth/api/http"
+	"github.com/mainflux/mainflux/auth/jwt"
+	"github.com/mainflux/mainflux/auth/mocks"
+	"github.com/mainflux/mainflux/pkg/uuid"
+	"github.com/opentracing/opentracing-go/mocktracer"
+)
+
+const (
+	secret       = "secret"
+	contentType  = "application/json"
+	id           = "123e4567-e89b-12d3-a456-000000000001"
+	email        = "user@example.com"
+	unauthzID    = "123e4567-e89b-12d3-a456-000000000002"
+	unauthzEmail = "unauthz@example.com"
+)
+
+type issueRequest struct {
+	Duration time.Duration `json:"duration,omitempty"`
+	Type     uint32        `json:"type,omitempty"`
+}
+
+type testRequest struct {
+	client      *http.Client
+	method      string
+	url         string
+	contentType string
+	token       string
+	body        io.Reader
+}
+
+func (tr testRequest) make() (*http.Response, error) {
+	req, err := http.NewRequest(tr.method, tr.url, tr.body)
+	if err != nil {
+		return nil, err
+	}
+	if tr.token != "" {
+		req.Header.Set("Authorization", tr.token)
+	}
+	if tr.contentType != "" {
+		req.Header.Set("Content-Type", tr.contentType)
+	}
+
+	req.Header.Set("Referer", "http://localhost")
+	return tr.client.Do(req)
+}
+
+func newService() auth.Service {
+	repo := mocks.NewKeyRepository()
+	groupRepo := mocks.NewGroupRepository()
+	idProvider := uuid.NewMock()
+	t := jwt.New(secret)
+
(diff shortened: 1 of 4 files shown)


auth/api/http/policies/endpoint_test.go
```diff
@@ -0,0 +1,205 @@
...
+	assert.Nil(t, err, fmt.Sprintf("Issuing user key expected to succeed: %s", err))
+
+	_, userLoginSecret, err := svc.Issue(context.Background(), "", auth.Key{Type: auth.UserKey, IssuedAt: time.Now(), IssuerID: unauthzID, Subject: unauthzEmail})
+	assert.Nil(t, err, fmt.Sprintf("Issuing unauthorized user's key expected to succeed: %s", err))
+
+	ts := newServer(svc)
+	defer ts.Close()
+	client := ts.Client()
+
+	valid := addPolicyRequest{Object: "obj", Policies: []string{"read"}, SubjectIDs: []string{"user1", "user2"}}
```
dborovcanin, line 102: Can you provide the case with multiple policies (such as `valid := addPolicyRequest{Object: "obj", Policies: []string{"read", "write"}, SubjectIDs: []string{"user1", "user2"}}`)?

auth/api/http/policies/requests.go
```diff
@@ -1,6 +1,18 @@
...
+)
+
+const (
+	readPolicy   = "read"
+	writePolicy  = "write"
+	deletePolicy = "delete"
+	accessPolicy = "access"
+	memberPolicy = "member"
+	createPolicy = "create"
+)
```
dborovcanin: Consider creating a new type:
```go

type Action int

const (
	Create Action = iota
	Write
	Read
	Access
	Member
	Delete
)

func (a Action) String() string {
	switch a {
	case Create:
		return "create"
	case Write:
		return "write"
	case Read:
		return "read"
	case Access:
		return "access"
	case Member:
		return "member"
	case Delete:
		return "delete"
	default:
		return "unknown"
	}
}
```
It will be more complex to parse (to Action and back to string), but the code will be a bit cleaner and less error-prone (and easier to validate: `if action >= Create && action <= Delete`).

auth/api/http/policies/endpoint_test.go
buraksekili: I added new test cases

auth/api/http/policies/requests.go
buraksekili: I tried to utilize the new type called `Action` as you suggested. However, I am not sure that the implementation is what you want.

auth/service_test.go
```diff
@@ -987,3 +987,58 @@ func TestUnassign(t *testing.T) {
...
+	err = svc.Authorize(context.Background(), auth.PolicyReq{Object: thingID, Relation: deletePolicy, Subject: id})
+	require.Nil(t, err, fmt.Sprintf("authorizing valid 'delete' policy for '%s' expected to succeed: %s", id, err))
+
+	// Verify policies of the user with tmpid.
+	err = svc.Authorize(context.Background(), auth.PolicyReq{Object: thingID, Relation: readPolicy, Subject: tmpID})
+	require.Nil(t, err, fmt.Sprintf("authorizing valid 'read' policy for '%s' expected to succeed: %s", tmpID, err))
+	err = svc.Authorize(context.Background(), auth.PolicyReq{Object: thingID, Relation: writePolicy, Subject: tmpID})
+	require.Nil(t, err, fmt.Sprintf("authorizing valid 'write' policy for '%s' expected to succeed: %s", tmpID, err))
+	err = svc.Authorize(context.Background(), auth.PolicyReq{Object: thingID, Relation: deletePolicy, Subject: tmpID})
+	require.Nil(t, err, fmt.Sprintf("authorizing valid 'delete' policy for '%s' expected to succeed: %s", tmpID, err))
```
dborovcanin: You can create a list of cases for these to be aligned with other tests. Also, use `assert` rather than `require`.

auth/api/http/policies/endpoint_test.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/mainflux/mainflux/auth/jwt"
+	"github.com/mainflux/mainflux/auth/mocks"
+	"github.com/mainflux/mainflux/pkg/uuid"
+	"github.com/opentracing/opentracing-go/mocktracer"
+)
+
+const (
+	secret       = "secret"
+	contentType  = "application/json"
+	id           = "123e4567-e89b-12d3-a456-000000000001"
```
manuio: You can maybe use the prefix from `pkg/mocks/uuid` here

auth/api/http/policies/endpoint_test.go
```diff
@@ -0,0 +1,213 @@
...
+	email        = "user@example.com"
+	unauthzID    = "123e4567-e89b-12d3-a456-000000000002"
```
manuio: same here

dborovcanin (approved):

PR #1481, opened 2021-10-20
NOISSUE - Improve Vault initialization

mteodor:
### What does this do?
Makes `vault` setup more easy. 

No
Not needed

Yes


docker/addons/vault/Dockerfile
```diff
@@ -0,0 +1,36 @@
+# base image
```
drasko, line 1: I do not like the idea that we are insert our own Dockerfile insted of being aligned to the official one. This will also mean that we'll have to build and maintain our Vault images and upload them somewhere.

drasko (changes requested):

docker/addons/vault/Dockerfile
mteodor: We are aligned to a official for deployment of vault we are using official image as you can see here https://github.com/mainflux/mainflux/blob/master/docker/addons/vault/docker-compose.yml#L22

Custom image is just  for a helper container here https://github.com/mainflux/mainflux/pull/1481/files#diff-d64c65e65debd3ed794b7ab5da07a691b56ff2458829e5c53102ddfd46e18dfeR41
This helper container is only for development, not to be used in production as stated here https://github.com/mainflux/mainflux/pull/1481/files#diff-d4c42ad23a7e26d2fdfc3b9801ba5512ba20fac0438021badc3f974aee56bc12R39

drasko: What is the benefit having this for development in the container, forcing the user to pack into container the inti sctips and each time build Docker image? Cannot these be mapped to an existing official container?

Also - if we develop something, and this is development-only, how related than is it to a production? Don't we want the same procedures, as we tested using these steps/processes?

docker/addons/vault/README.md
```diff
@@ -27,57 +27,25 @@ When the Vault service is started, some initialization steps need to be done to
...
-2. `vault-unseal.sh`
-
-This can be run after the initialization to unseal Vault, which is necessary for it to be used to store and/or get secrets.
-This can be used if you don't want to restart the service.
+Script calls `vault operator init` to perform the initial vault initialization and generates
+a `data/secrets` file which contains the Vault unseal keys and root tokens.
 
-The unseal environment variables need to be set in `.env` for the script to work (`MF_VAULT_TOKEN`, `MF_VAULT_UNSEAL_KEY_1`,
-`MF_VAULT_UNSEAL_KEY_2`, `MF_VAULT_UNSEAL_KEY_3`).
+This procedure is not production safe, this is only for development.
```
drasko, line 39: Why should we merge a solution that is developement-only? Why not publish it on a blog or something. Code in the core repo should contain production code. It can be immature, and users should be warned, but it should target production, not development.

docker/addons/vault/Dockerfile
drasko: @mteodor are you planning to provide some answers here, so we can move this on? It looks to me that this PR is some helper code, without ambition to become production-pready, and I am failing to understand why this is the case.

drasko (changes requested):

docker/addons/vault/Dockerfile
mteodor: I'll see to create a clean development flow without all the steps needed in production when find a time

mteodor:
closing this one https://github.com/mainflux/mainflux/pull/1481#discussion_r757006318

drasko:
> closing this one https://github.com/mainflux/mainflux/pull/1481#discussion_r757006318

Why? There should be an appropriate
closing report/reasoning.

PR #1482, opened 2021-10-25
MF-1443 - Add policies

### LINKED ISSUE #1443: Add policies
**FEATURE REQUEST**

This issue is related to #1347 

The `Auth` service manages access control of entities. However, the current service is not feasible to establish an advanced access control mechanism. Therefore, we need an authorization system to check whether Mainflux entities (e.g. `Users`) are authorized to do certain operations (e.g. `creating a group`) on the objects (e.g. `Groups`). 

There are two widely used policy-based authorization systems called RBAC and ABAC. Besides these approaches, Google introduced its own authorization system, called Zanzibar. Zanzibar does not use RBAC or ABAC, but it does take a similar `Access Control Lists (ACLs)` approach. 

First of all, Zanzibar has to be flexible to control access control of millions (maybe trillions?) Google's users. This flexibility comes with Simple Data Model and Configuration Language for policy relations.
> "ACLs are collections of object-user or object-object relations represented as *relation tuples*." [1]

So, relation tuples are the data type for the ACLs. Moreover: 

> this data type includes set-algebraic operators such as intersection and union for specifying potentially complex access control policies in terms of those user-object relations. [1]

For example, ACLs can refer to other ACLs to establish sophisticated relations (`Members of a <group_1>` are allowed to modify `<group_1>` where being a `members of a <group_1>` is another policy relation of itself). 

For more details of Google's Zanzibar, refer to [1].

<br/>

> **Ory Keto** is the first and only open-source implementation of "Zanzibar: Google's Consistent, Global Authorization System" [2]

All relation tuples can be represented in JSON format. The example relation tuples can be found via https://github.com/ory/keto/tree/master/contrib/cat-videos-example/relation-tuples.

ORY Keto provides REST API and Protocol Buffers API  for interacting with the Keto. 
> The APIs are separated based on privileges into a read and write endpoint. 
...
All APIs are available to gRPC and REST clients, although feature parity is not always given. [3]

The Keto APIs consist of two main parts; `Read` and `Write`. `Read` APIs allow us to `check`, `list`, and `expand` the relation tuples. On the other hand, `Write` APIs allow us to change (e.g. creating or deleting) relation tuples. For a more detailed overview, please refer to [3].

To sum up; Keto is the only open-source alternative for Google's Zanzibar at the moment. Although it is not fully enriched Zanzibar's features, it provides adequate APIs for managing and checking relations ("permissions") with HTTP and gRPC [2].

**References**

[1] https://www.usenix.org/conference/atc19/presentation/pang
[2] https://www.ory.sh/keto/docs/
[3] https://www.ory.sh/keto/docs/concepts/api-overview/

must-have

dborovcanin:
### What does this do?
This pull request introduces the initial version of policy support.

This pull request resolves #1443.

Current access control is rewritten to use [Keto](https://github.com/ory/keto) implementation of [Google Zanzibar](https://research.google/pubs/pub48190/) authorization system. Please take a look at the issue description for more information. 

Yes.

Yes.

### Notes
This is the initial version that introduces new access control to Mainflux and breaks compatibility with old versions. That is, we're yet to provide migration tools for a seamless transition to using policies. Manual transition can be done using the API provided in this pull request.


auth/api/http/policies/requests.go
```diff
@@ -1,6 +1,40 @@
...
+	Create Action = iota
+	Read
+	Write
+	Delete
+	Access
+	Member
+	unknown
+)
+
+var actions = [...]string{
```
tzzed: You can use a map, `map[string]Action` that allows you to have a O(1) search a policy and create an efficient lookup function.

auth/api/grpc/endpoint.go
```diff
@@ -62,17 +62,41 @@ func authorizeEndpoint(svc auth.Service) endpoint.Endpoint {
...
+		if err := req.validate(); err != nil {
+			return deletePolicyRes{}, err
 		}
 
-		return authorizeRes{authorized: authorized}, err
+		err := svc.DeletePolicy(ctx, auth.PolicyReq{Subject: req.Sub, Object: req.Obj, Relation: req.Act})
+		if err != nil {
+			return deletePolicyRes{}, err
+		}
+		return deletePolicyRes{deleted: true}, err
```
tzzed: you can return `deletePolicyRes{deleted: true}, nil` instead of err.

auth/api/http/policies/requests.go
```diff
@@ -0,0 +1,68 @@
...
+	if req.token == "" {
+		return auth.ErrUnauthorizedAccess
+	}
+
+	if len(req.SubjectIDs) == 0 || len(req.Policies) == 0 || req.Object == "" {
+		return auth.ErrMalformedEntity
+	}
+
+	for _, policy := range req.Policies {
+		if action := parsePolicy(policy); action > Member || action < Create {
```
tzzed: If you use map as I suggested to you above it will let you to avoid this kind of check.

auth/service.go
```diff
@@ -170,8 +169,37 @@ func (svc service) Identify(ctx context.Context, token string) (Identity, error)
...
+	user, err := svc.Identify(ctx, token)
+	if err != nil {
+		return errors.Wrap(ErrUnauthorizedAccess, err)
+	}
+
+	if err := svc.Authorize(ctx, PolicyReq{Object: "authorities", Relation: "member", Subject: user.ID}); err != nil {
+		return err
+	}
+
+	var errs error
```
tzzed, line 193: I think it could be interesting to use [uber multierr](https://github.com/uber-go/multierr)

cmd/users/main.go
```diff
@@ -313,28 +319,70 @@ func newService(db *sqlx.DB, tracer opentracing.Tracer, auth mainflux.AuthServic
...
 
-	if _, err := userRepo.RetrieveByEmail(context.Background(), user.Email); err == nil {
-		// Exiting if user already exists
+	if admin, err := userRepo.RetrieveByEmail(context.Background(), user.Email); err == nil {
+		// The admin is already created. Check existence of the admin policy.
+		_, err := auth.Authorize(context.Background(), &mainflux.AuthorizeReq{Obj: "authorities", Act: "member", Sub: admin.ID})
+		if err != nil {
+			apr, err := auth.AddPolicy(context.Background(), &mainflux.AddPolicyReq{Obj: "authorities", Act: "member", Sub: admin.ID})
+			if err != nil {
+				return err
```
tzzed, line 354: It will be more efficient to have a multi err here by chaining `err` from `auth.Authorize` and `auth.AddPolicy()`

tzzed (commented):
Hi @buraksekili and @dusanb94 kudos to you both.
I let myself to make a review about your PR I have some suggestions.

auth/api/grpc/endpoint.go
drasko: I agree with this one - please put `nil` here because it indicates to the reader that if you arrived to this point in the code, there is no error. Code will be more self-documenting.

drasko (changes requested):

cmd/users/main.go
```diff
@@ -313,28 +319,70 @@ func newService(db *sqlx.DB, tracer opentracing.Tracer, auth mainflux.AuthServic
 			Help:      "Total duration of requests in microseconds.",
 		}, []string{"method"}),
 	)
-	if err := createAdmin(svc, userRepo, c); err != nil {
+	if err := createAdmin(svc, userRepo, c, auth); err != nil {
 		logger.Error("failed to create admin user: " + err.Error())
 		os.Exit(1)
 	}
+
+	if c.adminCreatesUser != "" {
```
arvindh123, line 327: It will be better if `c.adminCreatesUser` is compare with some constant string rather compare with `""` empty string
Example :
```suggestion
	if c.adminCreatesUser != "true" {
```

arvindh123, line 327: It will be better if `c.adminCreatesUser` is compare with some constant string instead of comparing with `""` empty string
Example :
```suggestion
	if c.adminCreatesUser != "true" {
```

arvindh123: So if admin only need to create users , 
then variable `MF_ONLY_ADMIN_CREATES_USER` should set to `true`
Or Can be make its as `Boolean`.?

docker/docker-compose.yml
```diff
@@ -11,11 +11,62 @@ volumes:
...
 
 services:
+  keto:
+    image: oryd/keto:v0.6.0-alpha.3
+    container_name: mainflux-keto
+    ports:
+      - ${MF_KETO_READ_REMOTE_PORT}:4466
+      - ${MF_KETO_WRITE_REMOTE_PORT}:4467
+    environment:
+      - DSN=postgresql://${MF_KETO_DB_USER}:${MF_KETO_DB_PASS}@keto-db:${MF_KETO_DB_PORT}/${MF_KETO_DB}?sslmode=disable
```
arvindh123, line 27: I think sslmode might required in future ,can be made it as env var 
similar to auth DB, 
Its possible to make like this env var
`MF_KETO_DB_SSL_MODE`
`MF_KETO_DB_SSL_CERT`
`MF_KETO_DB_SSL_KEY`
`MF_KETO_DB_SSL_ROOT_CERT`

cmd/users/main.go
buraksekili: The user configures the environment variables on the `.env` file and can write any arbitrary string as a value for a particular environment variable. I think, forcing a user to write a fixed value for the environment variable is not convenient. Of course, both ways will work in the end.

What I've done here is that if the user writes any value for the MF_ONLY_ADMIN_CREATES_USER (which eventually makes `c.adminCreatesUser != ""` true), it means that the only admin is capable to create new users.

cmd/users/main.go
dborovcanin: In this case, i's OK to ignore the first error - it just indicates that the admin user does not have an admin policy.

docker/docker-compose.yml
dborovcanin: Created an [issue](https://github.com/mainflux/mainflux/issues/1485) for this.

drasko (approved):
LGTM

drasko:
WOW! Merged this huge one!

Thanks a lot @buraksekili and @dusanb94 for this wonderful contribution!

PR #1483, opened 2021-10-26
NOISSUE - Add view and list serials endpoints in certs service

manuio:


docker/addons/vault/vault-unseal.sh
```diff
@@ -1,5 +1,4 @@
-
-#!/bin/bash
+#!/usr/bin/bash
```
mteodor, line 1: there is other pr that changes the way the vault is initialized no need to make this change

certs/api/endpoint.go
```diff
@@ -20,12 +20,14 @@ func issueCert(svc certs.Service) endpoint.Endpoint {
 		if err != nil {
```
mteodor, line 20: thanks @manuio for tightening this service, this change is not related to the stuff you did but if you  could change `req.Valid` to `req.TTL` as that would be more appropriate ( and `days_valid` to `TTL` in openapi doc)

docker/addons/vault/vault-unseal.sh
manuio: I know but while we decide what to do with the other PR this fix alows you to init properly vault.

mteodor (dismissed):

drasko (commented):
This PR looks OK to me, but I would let @dusanb94 review and approve before we merge.

mteodor (approved):
LGTM

mteodor (approved):

mteodor (dismissed):

drasko:
I think there should be a documentation added for this

dborovcanin (changes requested):
Please add tests for new service methods.

mteodor (dismissed):

blokovi (dismissed):

blokovi (dismissed):

mteodor (dismissed):

api/certs.yml
```diff
@@ -22,29 +22,64 @@ paths:
...
+        '404':
+          description: |
+            Failed to retrieve corresponding certificates.
+        '500':
+          $ref: "#/components/responses/ServiceError"
+  /certs/{thingID}/serials:
+    get:
+      summary: Retrieves certificates serial IDs
+      description: |
+        Retrieves a list of issued certificates serial IDs for a given thing ID.
```
drasko: `issued certificates'`

api/certs.yml
```diff
@@ -22,29 +22,64 @@ paths:
           description: Failed due to malformed JSON.
         '500':
           description: Unexpected server-side error ocurred.
-  /certs/{thingId}:
+  /certs/{thingID}/list:
```
drasko: Shouold here `thingID` be part of the URL, or rather a param? @nmarcetic @dusanb94 opinion?

drasko (changes requested):
@manuio this implementation basically looks good, I am just confused a bit with endpoints - you use interchangeably `/certs/{certID}` and `/certs/{thingID}` and I do not feel that this is the best approach.

I would like to hear opinion of others on this.

manuio:
> @manuio this implementation basically looks good, I am just confused a bit with endpoints - you use interchangeably `/certs/{certID}` and `/certs/{thingID}` and I do not feel that this is the best approach.
> 
> I would like to hear opinion of others on this.

@drasko It's because you can issue several certificates for the same ThingID and we need to list certificates for a thing but also certIDs (serials) to view or revoke. Usinf HTTP params as you proposed could simplify it. Something like that:
```
post  /certs (create cert using thingID from the request body)
get  /certs/{thingID} (list of certs issued by thingID)
get  /certs/{thingID}?serial={serialID} (view cert by serialID)
delete  /certs/{thingID} (revoke all certs issued by thingID)
delete  /certs/{thingID}?serial={serialID} (revoke cert by serialID)
```

certs/api/endpoint.go
drasko: I like this proposition.

drasko (changes requested):

mteodor (dismissed):

blokovi (dismissed):

manuio:
@drasko @dusanb94 Can you review please?

api/certs.yml
```diff
@@ -58,6 +57,24 @@ paths:
             Failed to revoke corresponding certificate.
         '500':
           $ref: "#/components/responses/ServiceError"
+  /serials/{thingID}:
+    get:
+      summary: Retrieves certificates's serial IDs
```
dborovcanin: It's probably a good idea to use a neutral form here: `Retrieve certificates serial IDs`. The same goes for all the other `summary` sections.

certs/api/endpoint.go
dborovcanin: Is this resolved?

certs/mocks/certs.go
```diff
@@ -15,44 +15,62 @@ var _ certs.Repository = (*certsRepoMock)(nil)
...
+		Serial:  cert.Serial,
+		Expire:  cert.Expire,
+	}
+
+	_, ok := c.certsByThingID[cert.OwnerID][cert.ThingID]
+	if !ok {
+		c.certsByThingID[cert.OwnerID] = map[string][]certs.Cert{
+			cert.ThingID: []certs.Cert{crt},
+		}
+	} else {
```
dborovcanin: Please avoid using `else`.

dborovcanin (changes requested):

certs/api/endpoint.go
manuio: I'm not sure but I think that this remark was previous to the renaming of `days_valid` into `hours_valid`. IMO it's better to use `hours_valid`, it's self documenting. `ttl` is a Vault therminology.

api/certs.yml
manuio: We are using this form in all other api docs. Should we change all?

certs/api/endpoint.go
drasko: [`TTL`](https://en.wikipedia.org/wiki/Time_to_live) is not a Vault terminology.

api/certs.yml
dborovcanin: Probably yes, but let's do it in a separate PR.

certs/api/endpoint.go
dborovcanin: I agree that `hours_valid`, while a little clumsy, is more self-documenting. However, since we are using it as a string, we can include a time determinant such as "h", "m" or "s", so TTL is probably a better naming. Also, since we don't control the aforementioned time determinant, `hours` in the name is actually misleading.

manuio: Right. What I mean is that we are using `TTL`  in the pki layer but we don't necessaraly have to follow this therminology for the HTTP API. However we are using `KeyBytes` and `KeyType` so maybe I can change it.

manuio: @drasko @dusanb94 ok, let me rename then

dborovcanin (approved):

drasko (approved):
LGTM

PR #1484, opened 2021-10-26
NOISSUE - Update naming of env variable

buraksekili:

### What does this do?
This PR updates name of the `MF_ONLY_ADMIN_CREATES_USER` environment variable to `MF_USERS_ALLOW_SELF_SIGNED` that is responsible for controlling who can create a new user. 

By default, it has a `"true"` value which means everybody can create a new user. If the user sets it to another string (other than an empty string), only the admin rather than everybody can create a new user. 


cmd/users/main.go
```diff
@@ -112,26 +112,26 @@ const (
...
+	jaegerURL      string
+	resetURL       string
+	authTLS        bool
+	authCACerts    string
+	authURL        string
+	authTimeout    time.Duration
+	adminEmail     string
+	adminPassword  string
+	passRegex      *regexp.Regexp
+	userSelfSigned string
```
dborovcanin: This should be bool.

dborovcanin (changes requested):

cmd/users/main.go
buraksekili: okay, i will update it now

dborovcanin (approved):

cmd/users/main.go
```diff
@@ -74,7 +74,7 @@ const (
 	defAuthURL     = "localhost:8181"
 	defAuthTimeout = "1s"
 
-	defOnlyAdminCreatesUser = "" // If empty, everybody can create a user. Otherwise, only admin can create a user.
+	defUserSelfSigned = "true" // By default, everybody can create a user. Otherwise, only admin can create a user.
```
drasko: Instead of `selfSigned` please use `selfRegister`, so avoid confusion with signatures.

drasko (changes requested):

cmd/users/main.go
buraksekili: Updated to `selfRegister`.

drasko (approved):
LGTM

PR #1491, opened 2021-11-02
MF-1489 - Add API for deleting policies

### LINKED ISSUE #1489: API for deleting policies
**FEATURE REQUEST**

Admin can add policies via `POST /policies` API (defined under `auth` package). However, there is no option for the admin to delete these policies. It will be nice to implement an HTTP API for deleting policies similar to the `POST /policies` API.

should-have

buraksekili:

### What does this do?
This PR adds a new HTTP endpoint that allows the admin to delete policies. Like `POST /policies`, only policies defined under Mainflux are allowed to be deleted.

Only the admin can make this request. Other users are not allowed. Therefore, a token issued to the admin is required while making a request. Example usage is provided under the **Notes** section.

Resolves #1489 

The mock for the Keto Agent is updated.

Yes
Yes, I added API documentation (swagger).
### Notes

For example, you can delete one of the default policies that allow anybody to create a new user, as follows:
```bash
curl -isSX PUT http://localhost:8189/policies -d '{"subjects": ["*"], "object": "user", "policies": ["create"]}' -H "Authorization: $token" -H 'Content-Type: application/json'
```
where `$token` belongs to admin.


auth/api/http/policies/transport.go
```diff
@@ -33,6 +33,13 @@ func MakeHandler(svc auth.Service, mux *bone.Mux, tracer opentracing.Tracer) *bo
 		opts...,
 	))
 
+	mux.Put("/policies", kithttp.NewServer(
```
drasko, line 36: Why PUT for deletion, and not DELETE?

auth/api/http/policies/transport.go
```diff
@@ -33,6 +33,13 @@ func MakeHandler(svc auth.Service, mux *bone.Mux, tracer opentracing.Tracer) *bo
...
+		kitot.TraceServer(tracer, "delete_policies")(deletePoliciesEndpoint(svc)),
+		decodeCreatePolicyRequest,
```
drasko: Why CreatePolicy and not DeletePolicy?

drasko (changes requested):

auth/api/http/policies/transport.go
buraksekili: Since OpenAPI spec 3.0 doesn't accept request body for DELETE method type, I picked PUT over DELETE.

auth/api/http/policies/transport.go
buraksekili: Both creating and deleting operations use same structure which consists of `token` and same request body. Therefore, I used the same function. But I can make separate function to prevent anyd inconvenience or inconsistency

dborovcanin (dismissed):

auth/api/http/policies/transport.go
drasko: it is confusing in the code... We should add a comment here (this what you clarified to me)

auth/api/http/policies/transport.go
drasko: What about additng a query parameter instead of the body? Would it make sense? Because it will definitely be more instructive in the code.

drasko: What about additng a query parameter instead of the body? Would it make sense? Because it will definitely be more instructive in the code.

drasko: (share operation needs to be done also with query param, to stay consistent)

drasko (changes requested):

auth/api/http/policies/transport.go
dborovcanin: Since there is a list in the payload, we only can change this endpoint to remove only a single policy. I don't find removing policies one-by-one practical for the end-user, so I propose to keep it like this.

dborovcanin (approved):

drasko (approved):
LGTM

tofikmalik4545545-hue (approved):

PR #1493, opened 2021-11-03
NOISSUE - Update self register environment variable name

buraksekili:
### What does this do?

This PR updates the name of the environment variable responsible for allowing user creation by everybody or not. The previous name of it was `MF_ONLY_ADMIN_CREATES_USER`. Now, it is updated to `MF_USERS_ALLOW_SELF_REGISTER` which is a boolean variable. If `MF_USERS_ALLOW_SELF_REGISTER` is true, everybody can create a new user. Otherwise, only the admin can create a new user. Also, `MF_USERS_ALLOW_SELF_REGISTER` is `true` by default.

Furthermore, this PR updates the default admin creation process. The previous version was not checking the existence of the policy if `MF_USERS_ALLOW_SELF_REGISTER` is true. However, if `MF_USERS_ALLOW_SELF_REGISTER` is true, the policy has to be created. Therefore, this PR adds checking the existence of that policy when `MF_USERS_ALLOW_SELF_REGISTER` is true.

`-`
Update the environment variable.
No


drasko (approved):
LGTM

dborovcanin (approved):

PR #1505, opened 2021-11-25
NOISSUE - Apply policies to Channels

buraksekili:
### What does this do?

This PR adds policies for Channels entity encapsulating following operations; `view`, `update` and `delete`. The admin can do all of the operations on the all channels.

Updated Channels repository and tests.

### DIFF
--- a/things/postgres/channels.go
+++ b/things/postgres/channels.go
@@ -124,6 +124,7 @@ func (cr channelRepository) RetrieveAll(ctx context.Context, owner string, pm th
 	nq, name := getNameQuery(pm.Name)
 	oq := getOrderQuery(pm.Order)
 	dq := getDirQuery(pm.Dir)
+	ownerQuery := getOwnerQuery(pm.FetchSharedThings)
 	meta, mq, err := getMetadataQuery(pm.Metadata)
 	if err != nil {
 		return things.ChannelsPage{}, errors.Wrap(things.ErrSelectEntity, err)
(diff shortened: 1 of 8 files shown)


things/postgres/channels.go
```diff
@@ -137,13 +137,16 @@ func (cr channelRepository) RetrieveAll(ctx context.Context, owner string, pm th
...
 
 	if len(query) > 0 {
-		whereClause = fmt.Sprintf("AND %s", strings.Join(query, " AND "))
+		whereClause = fmt.Sprintf(" WHERE %s", strings.Join(query, " AND "))
 	}
 
-	q := fmt.Sprintf(`SELECT id, name, metadata FROM channels
-	      WHERE owner = :owner %s ORDER BY %s %s LIMIT :limit OFFSET :offset;`, whereClause, oq, dq)
+	q := fmt.Sprintf(`SELECT id, name, metadata FROM channels 
+	%s ORDER BY %s %s LIMIT :limit OFFSET :offset;`, whereClause, oq, dq)
```
drasko: Is this formatting OK? Looks like there is a trailing space at the end of the line, and line below is no indented (maybe this is OK, but previous was, and this probably makes it easier to read).

drasko (changes requested):

things/postgres/channels.go
buraksekili: Updated 👍

dborovcanin (approved):

things/service.go
```diff
@@ -386,16 +389,38 @@ func (ts *thingsService) CreateChannels(ctx context.Context, token string, chann
...
+	return chs, nil
+}
+
+func (ts *thingsService) createChannel(ctx context.Context, channel *Channel, identity *mainflux.UserIdentity) (Channel, error) {
+	chID, err := ts.idProvider.ID()
+	if err != nil {
+		return Channel{}, errors.Wrap(ErrCreateUUID, err)
+	}
+	channel.ID = chID
+	channel.Owner = identity.GetEmail()
```
mteodor, line 409: a question, if would be switching, at some point, to identity ID instead of the email would that be a problem for policies?

buraksekili: Actually, policies use ID instead of email. Here, `channel.Owner = identity.GetEmail()` email is used to store it on the database, not for policies.

drasko (approved):
LGTM

PR #1515, opened 2021-12-01
NOISSUE - UI Group fix

fbugarski:
Signed-off-by: Filip Bugarski <filipbugarski@gmail.com>

No
No


manuio (approved):
LGTM

ui/service.go
```diff
@@ -281,40 +282,51 @@ func (gs *uiService) ListChannels(ctx context.Context, token string) ([]byte, er
...
 	}
 
 	data := struct {
 		NavbarActive string
-		Groups       []auth.Group
+		Groups       []sdk.Group
 	}{
 		"groups",
-		grpsPage.Groups,
+		thsPage.Groups,
```
manuio: CI is failling, `thsPage.Groups` is of type `[]auth.Group` not `[]sdk.Group`

manuio (changes requested):

ui/service.go
fbugarski: It is connected with previous PR. It will be fine now, but i will change tshPage to grpsPage :)

cmd/ui/main.go
```diff
@@ -40,12 +40,14 @@ const (
 	defThingsAuthURL     = "localhost:8181"
 	defAuthURL           = "localhost:8189"
 	defThingsAuthTimeout = "1s"
+	defAuthToken         = "123"
```
drasko: There will not be such a thing, as it will not be used in single-user mode. This is maybe your development helper, but not needed to be set-up globlay later.

cmd/ui/main.go
```diff
@@ -40,12 +40,14 @@ const (
...
 
 	envLogLevel          = "MF_GUI_LOG_LEVEL"
 	envClientTLS         = "MF_GUI_CLIENT_TLS"
 	envCACerts           = "MF_GUI_CA_CERTS"
 	envPort              = "MF_GUI_PORT"
 	envRedirectURL       = "MF_GUI_REDIRECT_URL"
+	envAuthToken         = "MF_GUI_AUTH_TOKEN"
```
drasko: I do not see the need for this, UI will not be used in a single-user mode. You are using it just for development purposes.

cmd/ui/main.go
```diff
@@ -60,6 +62,7 @@ type config struct {
 	jaegerURL         string
 	thingsAuthURL     string
 	thingsAuthTimeout time.Duration
+	authToken         string
```
drasko: Nope ;)

cmd/ui/main.go
```diff
@@ -90,7 +93,7 @@ func main() {
 
 	tc := thingsapi.NewClient(conn, thingsTracer, cfg.thingsAuthTimeout)
 
-	svc := ui.New(tc, sdk)
+	svc := ui.New(tc, cfg.authToken, sdk)
```
drasko: Nope ;)

cmd/ui/main.go
```diff
@@ -146,6 +149,7 @@ func loadConfig() config {
 		caCerts:           mainflux.Env(envCACerts, defCACerts),
 		jaegerURL:         mainflux.Env(envJaegerURL, defJaegerURL),
 		thingsAuthURL:     mainflux.Env(envThingsAuthURL, defThingsAuthURL),
+		authToken:         mainflux.Env(envAuthToken, defAuthToken),
```
drasko: Nope ;)

ui/service.go
```diff
@@ -55,12 +55,14 @@ var _ Service = (*uiService)(nil)
 type uiService struct {
 	things mainflux.ThingsServiceClient
 	sdk    sdk.SDK
+	token  string
```
drasko: Nope ;)

ui/service.go
```diff
@@ -55,12 +55,14 @@ var _ Service = (*uiService)(nil)
...
 }
 
 // New instantiates the HTTP adapter implementation.
-func New(things mainflux.ThingsServiceClient, sdk sdk.SDK) Service {
+func New(things mainflux.ThingsServiceClient, token string, sdk sdk.SDK) Service {
```
drasko: Nope ;)

drasko (changes requested):

ui/service.go
```diff
@@ -325,32 +336,33 @@ func (gs *uiService) ListGroups(ctx context.Context, token string) ([]byte, erro
...
-	}
-	group, err := gs.sdk.Group(id, "123")
+	tpl := template.New("groups")
+	tpl = tpl.Funcs(template.FuncMap{
+		"toJSON": func(data map[string]interface{}) string {
+			ret, _ := json.Marshal(data)
+			return string(ret)
+		},
+	})
+	var err error
```
drasko: I do not like moving this here, I prefer it on the top of the fnction. Also, I do not like declaring `err` separately, I think this should be declared at the same time as `tpl`, as it was before.

Is there any special need to introduce `tpl := template.New("groups")`? Would notit be picked up via parsing globally in template dir?

drasko (changes requested):

ui/service.go
```diff
@@ -65,17 +68,24 @@ func New(things mainflux.ThingsServiceClient, sdk sdk.SDK) Service {
 	}
 }
 
-func (gs *uiService) Index(ctx context.Context, token string) ([]byte, error) {
-	tpl := template.New("index")
+func ParseTemplate(tmp string) (*template.Template, error) {
```
drasko: Is it used in other packages and has to be exported?

ui/service.go
```diff
@@ -95,33 +105,23 @@ func (gs *uiService) Index(ctx context.Context, token string) ([]byte, error) {
...
-	tpl = tpl.Funcs(template.FuncMap{
-		"toJSON": func(data map[string]interface{}) string {
-			ret, _ := json.Marshal(data)
-			return string(ret)
-		},
-	})
-	var err error
-
-	tpl, err = tpl.ParseGlob(templateDir + "/*")
+	tpl, err := ParseTemplate("/things.html")
```
drasko: No need for `/`, this should be handled in the `ParseTemplate()`. Just use file name, it will be simpler.

ui/service.go
```diff
@@ -65,17 +68,24 @@ func New(things mainflux.ThingsServiceClient, sdk sdk.SDK) Service {
 	}
 }
 
-func (gs *uiService) Index(ctx context.Context, token string) ([]byte, error) {
-	tpl := template.New("index")
+func ParseTemplate(tmp string) (*template.Template, error) {
+	tpl := template.New(tmp)
```
drasko: Do we need this at all, when we create a `tpl` by calling `ParseFiles()`?

ui/service.go
```diff
@@ -65,17 +68,24 @@ func New(things mainflux.ThingsServiceClient, sdk sdk.SDK) Service {
...
 	tpl = tpl.Funcs(template.FuncMap{
```
drasko, line 73: If we do not need this `New()` above, then move this `tpl.Funcs()` after `ParseFiles()` where `tpl` is created.

ui/service.go
```diff
@@ -18,6 +18,9 @@ import (
 
 const (
 	templateDir = "ui/web/template"
+	header      = templateDir + "/header.html"
```
drasko: I would propose other approach - instead of defining 3 different strings (`header`, `footer` and `navbar`), create a slice of  strings: `tmplFiles = []string{"header.html", "footer.html", "navbar.html"}`.

Then in the function `ParseTemplate()` add your file to this string array, prefix each array member with a `templateDir`, and pass the whole array to a `ParseFiles()` function - it is a [variadic function](https://pkg.go.dev/html/template#ParseFiles) and can accept array of strings.

ui/service.go
drasko: Additionaly, it would be good to make this function variadic, so that it can accept more files as an input (some service handlers would potentinally like to pass several template files associated with them).

drasko (changes requested):

ui/service.go
fbugarski: It must be called before the template is parsed.

ui/service.go
fbugarski: {"level":"warn","message":"Method index took 255.486µs to complete with error: open ui/web/templateindex.html: no such file or directory.","ts":"2021-12-09T13:24:28.490807676Z"}

ui/service.go
```diff
@@ -65,17 +67,27 @@ func New(things mainflux.ThingsServiceClient, sdk sdk.SDK) Service {
 	}
 }
 
+func parseTemplate(tmps ...string) (tpl *template.Template, err error) {
+	for _, tmp := range tmps {
+
```
drasko: Kill this white line

ui/service.go
```diff
@@ -65,17 +67,27 @@ func New(things mainflux.ThingsServiceClient, sdk sdk.SDK) Service {
...
+		tpl = template.New(tmp)
+		tpl = tpl.Funcs(template.FuncMap{
+			"toJSON": func(data map[string]interface{}) string {
+				ret, _ := json.Marshal(data)
+				return string(ret)
+			},
+		})
+		a := append(tmplFiles, templateDir+tmp)
```
drasko: Use `tmlp`, because `tmp` is often shortened for `temporary`. Also - first append just `a := append(tmplFiles, tmpl)`, then in foreach add `templateDir` prefix o all array members.

What does `a` stands for, btw?

ui/service.go
```diff
@@ -65,17 +67,27 @@ func New(things mainflux.ThingsServiceClient, sdk sdk.SDK) Service {
...
+		tpl, err = tpl.ParseFiles(a...)
```
drasko: This should not be in `range` loop, but later, in the end, when all files are well prepared in the array.

drasko (changes requested):

drasko (approved):
LGTM

PR #1517, opened 2021-12-02
MF-1506 - Move Groups back to domain services

### LINKED ISSUE #649: Introduce organizations (user groups) for full multi-tenancy
Organizations (or groups of users) should be introduced to enable full multi-tenancy.

Admin account equally should be introduced:
- Admin of the organization
- Admin of the whole Mainflux system (admin of admins)

Related to https://github.com/mainflux/mainflux/issues/397

### LINKED ISSUE #1506: Group-based Access Control
**ENHANCEMENT**

Bring Groups back to domain services and implement group-based policies. That is, Things groups belong to the Things service, and users groups belong to the Users service. There is absolutely no difference between them. This does not mean we'll repeat the code for groups in two services. Instead, we'll create an `internal` shared package(s) for group operation and use it in two different services. We can use an old implementation as a reference for the transport layer: https://github.com/mainflux/mainflux/tree/530f925c4d7e9c222cd04468751b1763f896519c.
So, the `internal` shared package should contain Groups transport, repository, and service layers, both specification and implementations. Things and Users services will extend respective layers and use the same implementation to avoid code duplication.

#### Policies
The access privileges are executed in the following format:
`S-A-O` - Subject-Action-Object
Where Subject is Client ID, Action is one of the supported actions (read, write, send, receive, list....), and Object is **Client ID** or **Group ID**.

The following table describes the access control rights:
image

There are three types of policies:

- M Policy for messages
- G Policy for Group rights
- C Policy for Clients that are group members 

M Policy represents the client's rights to send and receive messages in the group. Only group members with corresponding rights can publish or receive messages to/from the group.

G Policy represents the client's rights to modify the group itself. Only group members with correct rights can modify or update the group, or add/remove members to/from the group.

Finally, the C policy represents the rights the member has over other members of the group.

> Please note that this concept of multiple policy types does not have to map 1:1 with implementation. For example, a single policy with multiple rules can be used, as described [here](https://github.com/ultravioletrs/ultraviolet/issues/5#issuecomment-1168860794.) 

### Roles

Each client has a **role** that determines their rights on the "root" level. The "root" level access control is not related to groups and is used to provision clients and groups. There are two roles at the moment: **admin** and **general**. The admin client can access all the other clients and groups unconditionally. The general client can access only clients and groups created by it.

This is a must-have.

### LINKED ISSUE #1512: Move Groups API layer to shared internal package
**ENHANCEMENT**
As a part of #1509, groups' codebase needs to be shared internally between Users and Things. Groups codebase includes:

    -  **API layer**
    -  Repository layer (see #1513)
    -  Event sourcing (see #1510)

This is a must-have.

### LINKED ISSUE #1513: Move Groups repository layer to shared internal package
**ENHANCEMENT**
As a part of #1509, groups' codebase needs to be shared internally between Users and Things. Groups codebase includes:

    -  API layer (see #1512)
    -  **Repository layer**
    -  Event sourcing (see #1510)

This is a must-have.

dborovcanin:

### What does this do?
This pull request makes domain services (Users and Things) use the internal Groups package for entity groups manipulation.

This pull request resolves #1512, resolves #1513, and closes #1506, closes #649.

Groups are moved to domain services, which changes internal architecture when it comes to entity groups, also moving the corresponding API to the respective services.

Yes.

Yes.

### Notes
N/A


mteodor (approved):

auth/api/logging.go
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Mainflux
 // SPDX-License-Identifier: Apache-2.0
 
+//go:build !test
```
manuio, line 4: Is this normal?

dborovcanin:
We changed the approach to access control and we'll soon integrate a new solution with corresponding docs. Closing due to irrelevance.

PR #1523, opened 2021-12-09
MF-1469 - Indicate proper authentication scheme in Authorization header

### LINKED ISSUE #1469: Indicate proper authentication scheme in Authorization header
**FEATURE REQUEST**

    when authenticating api requests use `Authorization: Bearer <value>` not `Authorizaiton: <value>`

   should-have

bioinformatx:
### What does this do?
This pull request indicates the proper auth scheme in the HTTP Authorization header.

This pull request resolves #1469.

The user authentication scheme is changed to require a `Bearer ` prefix.
This is a breaking change since each user HTTP API call needs to have this prefix to be valid. 

Yes.

No, but documentation will be added in #1565.

### Notes
N/A


dborovcanin:
@stefankovacevic123 Please sign your commits, read [contribution guidelines](https://github.com/mainflux/mainflux/blob/master/CONTRIBUTING.md), and update your PR according to it.

drasko:
@stefankovacevic123 it is necessary to take in the account remarks from @dusanb94. Otherwise this PR will be closed due to the inactivity.

bioinformatx:
I am resolving the issue right now

drasko:
@stefankovacevic123 DCO issue should be fixed by following the instructions [here](https://github.com/mainflux/mainflux/pull/1523/checks?check_run_id=4541212126)

drasko:
Additionall, you need to change PR name, as @dusanb94 mentioned [here](https://github.com/mainflux/mainflux/pull/1523#issuecomment-989794650)

manuio (dismissed):
LGTM

dborovcanin (commented):
It looks like this PR does not indicate proper auth schema (#1469) but rather removes it: instead of using the `Bearer` prefix, we simply ignore it.  While this is OK for some specific use-cases (where devices are preconfigured to use the prefix with token-based authentication), it is not generic and does not actually take into account different authentication schemes. I propose that we elaborate on the possible solutions in the issue and decide where to go with this. Also, @stefankovacevic123 please fill out the issue template properly.

drasko:
@mteodor what is your opinion on this, as someone who opened an issue a s used `Bearer` token for the user DB read?

drasko:
Also, it would be good to hear @nmarcetic's opinion here.

dborovcanin:
I have nothing against this approach, but we need to decide if we are going to use any other Auth schema. We probably won't and in that sense, this implementation is a way to go.

OAuth enforces this "token_type token_value" syntax and I have never seen anyone using anything but "bearer" as a type, which is why we omitted it in the initial design. We can create out own schema (and I think it was actually the idea to make separate users and things toekns using different schemas) but that conceptionally does not match the idea of Ttken type: both users and things are "bearers" and it makes sense that both use "bearer" token.

mteodor:
my main reason for creating this issue was integration.
when you integrate with other services it makes sense that all services are aligned 
this is what a bearer token is
```
Bearer Authentication
Bearer authentication (also called token authentication) is an HTTP authentication scheme that involves security tokens called bearer tokens. The name “Bearer authentication” can be understood as “give access to the bearer of this token.” The bearer token is a cryptic string, usually generated by the server in response to a login request. The client must send this token in the Authorization header when making requests to protected resources:
```

and effectively we have only this type of api authentication



> @mteodor what is your opinion on this, as someone who opened an issue a s used `Bearer` token for the user DB read?

@drasko I dont understand this?

> instead of using the `Bearer` prefix, we simply ignore it

@dusanb94 we can enforce, but simply replacing it makes it not breaking existing applications that are built on top of our API

internal/httputil/auth.go
```diff
@@ -0,0 +1,17 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package httputil
+
+import (
+	"net/http"
+	"strings"
+)
+
+// FormatAuthString reads the value of request Authorization and removes the Bearer substring
```
drasko: Why do we do this? Why don't we just change API and demand that the new confention is to send `Bearer` in Autorisation header?

drasko (changes requested):

internal/httputil/auth.go
dborovcanin: I also agree. @mteodor I understand that this is for integration, and we should add it. However, since this is the standard, let's switch to it and require a prefix. That's all. At least when it comes to the API for users, we don't have to change the HTTP adapter for things in this PR. Also, this implies changes in SDK, tests, and docs.

drasko:
A [CBOR Web Token](https://datatracker.ietf.org/doc/html/rfc8392) standard should be examined in order to understand potential change of the Mainflux token into self-contained bearer token that can be a structure with a field `type`, telling if it is a User or a Thing token.

mteodor:
> Bearer



> A [CBOR Web Token](https://datatracker.ietf.org/doc/html/rfc8392) standard should be examined in order to understand potential change of the Mainflux token into self-contained bearer token that can be a structure with a field `type`, telling if it is a User or a Thing token.

This sounds like ok idea but I would go with separate issue for CBOR token

internal/httputil/auth.go
```diff
@@ -0,0 +1,20 @@
...
+	"net/http"
+	"strings"
+)
+
+// FormatAuthString reads the value of request Authorization and removes the Bearer substring or returns error if it does not exist
+func FormatAuthString(r *http.Request) (string, error) {
+	token := r.Header.Get("Authorization")
+	if !strings.Contains(token, "Bearer ") {
+		return token, errors.New("Token not containing a Bearer.")
+	}
```
dborovcanin: It should not containt `Bearer ` but start with it. Please extract `errors.New("Token not containing a Bearer.")` to a var. Error messages are not capitalized and it's not a full sentence.

internal/httputil/auth.go
```diff
@@ -0,0 +1,20 @@
...
+
+	return strings.ReplaceAll(token, "Bearer ", ""), nil
```
dborovcanin: Do not replace, only remove the prefix.

internal/httputil/auth.go
```diff
@@ -0,0 +1,20 @@
...
+package httputil
+
+import (
+	"errors"
+	"net/http"
+	"strings"
+)
+
+// FormatAuthString reads the value of request Authorization and removes the Bearer substring or returns error if it does not exist
+func FormatAuthString(r *http.Request) (string, error) {
```
dborovcanin: Maybe `ExtractAuthToken` makes for a little better naming.

go.mod
```diff
@@ -43,109 +43,5 @@ require (
 	golang.org/x/crypto v0.0.0-20220112180741-5e0467b6c7ce
 	golang.org/x/net v0.0.0-20220114011407-0dd24b26b47d
 	gonum.org/v1/gonum v0.9.3
-	google.golang.org/grpc v1.43.0
-)
-
-require (
```
dborovcanin: This does not look right. Please check your Go version.

dborovcanin (changes requested):
Please update PR, fix CI, and resolve conflicts.

dborovcanin:
@stefankovacevic123 Please update the branch and resolve conflicts and remarks so that we can proceed with this one.

internal/httputil/auth.go
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Mainflux
```
dborovcanin: Consider renaming the file name. Maybe `token.go` is more suitable.

internal/httputil/auth.go
```diff
@@ -0,0 +1,27 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/mainflux/mainflux/pkg/errors"
+)
+
+const (
+	// BearerPrefix represents the token prefix for Bearer authentication scheme.
+	BearerPrefix = "Bearer "
+)
```
dborovcanin:
```go
// BearerPrefix represents the token prefix for Bearer authentication scheme.
const BearerPrefix = "Bearer "
```

dborovcanin (changes requested):

auth/api/http/policies/transport.go
```diff
@@ -46,12 +46,18 @@ func decodePoliciesRequest(ctx context.Context, r *http.Request) (interface{}, e
...
 
-	var req policiesReq
+	t, err := httputil.ExtractAuthToken(r)
+	if err != nil {
+		return nil, err
+	}
+
+	req := policiesReq{
+		token: t,
+	}
```
dborovcanin, line 56: You can inline: `policiesReq{token: t}`.

dborovcanin (approved):

blokovi (approved):

drasko (approved):
LGTM

PR #1532, opened 2021-12-23
NOISSUE  - Add connect and Assign members to group - UI

fbugarski:

### DIFF
--- a/ui/api/endpoint.go
+++ b/ui/api/endpoint.go
@@ -260,16 +260,15 @@ func connectEndpoint(svc ui.Service) endpoint.Endpoint {
 	}
 }
 
-func disconnectEndpoint(svc ui.Service) endpoint.Endpoint {
+func disconnectThingEndpoint(svc ui.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
-		cr := request.(connectReq)
+		dcr := request.(disconnectThingReq)
 
-		if err := cr.validate(); err != nil {
+		if err := dcr.validate(); err != nil {
 			return nil, err
 		}
 
-		res, err := svc.Disconnect(ctx, cr.token, cr.ChannelIDs, cr.ThingIDs)
-
+		res, err := svc.Disconnect(ctx, dcr.token, []string{dcr.ChanID}, []string{dcr.ThingID})
 		if err != nil {
 			return nil, err
 		}
--- a/ui/api/endpoint.go
+++ b/ui/api/endpoint.go
@@ -232,7 +232,7 @@ func connectThingEndpoint(svc ui.Service) endpoint.Endpoint {
 			return nil, err
 		}
 
-		res, err := svc.Connect(ctx, cr.token, []string{cr.ChanID}, []string{cr.ThingID})
+		res, err := svc.ConnectChannelsByThing(ctx, cr.token, []string{cr.ChanID}, []string{cr.ThingID})
 		if err != nil {
 			return nil, err
 		}
@@ -242,7 +242,25 @@ func connectThingEndpoint(svc ui.Service) endpoint.Endpoint {
 	}
 }
 
-func connectEndpoint(svc ui.Service) endpoint.Endpoint {
+func connectChannelEndpoint(svc ui.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		cr := request.(connectChannelReq)
+
+		if err := cr.validate(); err != nil {
+			return nil, err
+		}
+
+		res, err := svc.ConnectThingsByChannel(ctx, cr.token, []string{cr.ChanID}, []string{cr.ThingID})
+		if err != nil {
+			return nil, err
+		}
+		return uiRes{
+			html: res,
+		}, err
+	}
+}
+
+func viewChannelsByThingEndpoint(svc ui.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(viewResourceReq)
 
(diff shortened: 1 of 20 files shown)


ui/service.go
```diff
@@ -336,6 +468,21 @@ func (gs *uiService) ViewGroup(ctx context.Context, token, id string) ([]byte, e
 	return btpl.Bytes(), nil
 }
 
+func (gs *uiService) Assign(ctx context.Context, token string, groupID, groupType string, memberIDs ...string) ([]byte, error) {
+	if err := gs.sdk.Assign(memberIDs, groupType, groupID, token); err != nil {
+		return []byte{}, err
+	}
+
```
drasko: Here you left an empty line, but in the function below you did not. Please choose one or another and stay consistent.

drasko (changes requested):
Very good, just one small remark regarding the style. PLease fix this and we can merge.

ui/api/endpoint.go
```diff
@@ -224,6 +224,96 @@ func removeChannelEndpoint(svc ui.Service) endpoint.Endpoint {
...
+		if err != nil {
+			return nil, err
+		}
+		return uiRes{
+			html: res,
+		}, err
+	}
+}
+
+func connectEndpoint(svc ui.Service) endpoint.Endpoint {
```
drasko: What's the difference of `connectEndpoint` and other 2 functions?

ui/api/transport.go
```diff
@@ -300,6 +466,65 @@ func decodeGroupUpdate(_ context.Context, r *http.Request) (interface{}, error)
...
+func parseSubtopic(subtopic string) (string, error) {
+	if subtopic == "" {
+		return subtopic, nil
+	}
+
+	subtopic, err := url.QueryUnescape(subtopic)
+	if err != nil {
+		return "", errMalformedSubtopic
+	}
+	subtopic = strings.Replace(subtopic, "/", ".", -1)
```
drasko: Why this replacement with `.`? Can't you split on `/`?

drasko (changes requested):

ui/api/endpoint.go
fbugarski: `connectEndpoint` is for listing all thing connections (button connections), `connectChannelEndpoint` is for listing all channel connections, and `connectThingEndpoint` is for connecting thing to channel and opposite.

ui/api/transport.go
fbugarski: UI doesn't support subtopic for now, i will remove this function.

ui/api/endpoint.go
drasko: Then the naming is not really good. Should be `listConnections` or something, as we use for Users and Things - AFAIK we are using `list-` prefix, right?

drasko: Accordingly, structs in `req` and `resp` should be renamed.

drasko: Also, why are we using `svc.ViewConnections()` here - should it not be `svc.ListConnections()` according to our standard, similar to [this](https://github.com/mainflux/mainflux/blob/master/things/api/things/http/endpoint.go#L173) (and same for `viewResourceReq`)

drasko (changes requested):

ui/api/transport.go
```diff
@@ -26,13 +28,25 @@ import (
 const (
 	contentType = "text/html"
 	staticDir   = "ui/web/static"
-	token       = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2MzkwODg1MTUsImlhdCI6MTYzOTA1MjUxNSwiaXNzIjoibWFpbmZsdXguYXV0aCIsInN1YiI6ImZscDFAZW1haWwuY29tIiwiaXNzdWVyX2lkIjoiYzkzY2FmYjMtYjNhNy00ZTdmLWE0NzAtMTVjMTRkOGVkMWUwIiwidHlwZSI6MH0.Ck9r2eJc2gieUdGy1Lt1BYLpYmf5rQT-KX1bckmwq28"
+	token       = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2NDE1MTkyNjEsImlhdCI6MTY0MTQ4MzI2MSwiaXNzIjoibWFpbmZsdXguYXV0aCIsInN1YiI6ImZscDFAZW1haWwuY29tIiwiaXNzdWVyX2lkIjoiYzkzY2FmYjMtYjNhNy00ZTdmLWE0NzAtMTVjMTRkOGVkMWUwIiwidHlwZSI6MH0.cqDOZdqiH9sXd1yuDwsv6-Mtb6_nVe_4c6cJK-iJ-Ig"
```
drasko, line 32: Leave a `TODO` comment above this `token` that it is a temporary token and will be removed once auth proxy is in place.

ui/service.go
```diff
@@ -271,6 +282,112 @@ func (gs *uiService) RemoveChannel(ctx context.Context, token, id string) ([]byt
...
+	var btpl bytes.Buffer
+	if err := tpl.ExecuteTemplate(&btpl, "connections", data); err != nil {
+		println(err.Error())
+	}
+	return btpl.Bytes(), nil
+}
+
+func (gs *uiService) ListChannelConnections(ctx context.Context, token, id string) ([]byte, error) {
+	tpl, err := parseTemplate("connectionsttc", "connectionsttc.html")
```
drasko: Please rename this file, because it is not self-documenting

drasko (changes requested):

ui/api/transport.go
drasko: Please add this `TODO` comment on a line in front of the token, as explained

drasko (approved):
LGTM

PR #1537, opened 2022-01-05
NOISSUE - Add API keys functions to CLI

fbugarski:
### What does this do?
Add API keys functions to CLI


drasko:
Please:
- Correct the PR description. If you do not have anything to put in the template, simply delete these parts, do not put `/`.
- Fix the CI, it is failing
- Update branch

cli/keys.go
```diff
@@ -0,0 +1,72 @@
...
+			if err := sdk.Revoke(args[0], args[1]); err != nil {
+				logError(err)
+				return
+			}
+
+			logOK()
+		},
+	},
+	cobra.Command{
+		Use:   "RetrieveKey",
```
drasko: Why this one is in capital letters, and other are not?

drasko: Also - short description do not correspond to te `Use` part

drasko (changes requested):

drasko:
@fbugarski please update branch and fix the CI

drasko:
@fbugarski is this ready for review? If yes - then please remove "Draft" tag.

pkg/sdk/go/keys.go
```diff
@@ -12,7 +12,7 @@ import (
 
 const keysEndpoint = "keys"
 
-func (sdk mfSDK) Issue(token string, k Key) (issueKeyRes, error) {
+func (sdk mfSDK) Issue(token string, k KeyReq) (issueKeyRes, error) {
```
drasko: Why is request called `KeyReq`, and the response `issueKeyRes`? Can we please align these two for consistence?

drasko (changes requested):

drasko (approved):
This looks good to me now, but @dusanb94 must approve as well.

cli/keys.go
```diff
@@ -0,0 +1,102 @@
...
+	"time"
+
+	mfxsdk "github.com/mainflux/mainflux/pkg/sdk/go"
+	"github.com/spf13/cobra"
+)
+
+var cmdAPIKeys = []cobra.Command{
+	{
+		Use:   "issue",
+		Short: "issue type duration <api_key_token>",
```
drasko: This needs to be rephrased better.

cli/keys.go
```diff
@@ -0,0 +1,102 @@
...
+				return
+			}
+
+			logOK()
+		},
+	},
+	{
+		Use:   "retrieve",
+		Short: "retrieve <key_id> <user_auth_token>",
+		Long:  `Retrieve API key with given id`,
```
drasko: `Retrieves` or `Retrieve`? If you use imperative here, use it for `Remove` also. Pick one and stay consistent.

drasko (changes requested):

pkg/sdk/go/sdk.go
```diff
@@ -116,6 +116,11 @@ type Member struct {
 	Type string
 }
 
+type KeyReq struct {
```
dborovcanin: Consider using Type and Duration directly - without wrapper `KeyReq` struct.

pkg/sdk/go/keys.go
```diff
@@ -6,43 +6,45 @@ import (
...
 
 	"github.com/mainflux/mainflux/pkg/errors"
 )
 
 const keysEndpoint = "keys"
 
-func (sdk mfSDK) Issue(token string, k Key) (issueKeyRes, error) {
-	data, err := json.Marshal(k)
+func (sdk mfSDK) Issue(token string, d time.Duration) (KeyRes, error) {
+	datareq := KeyReq{Type: 2, Duration: d}
```
drasko: Magic number `2` - define all constants at the top of the file and add appropriate comments.

drasko (changes requested):

pkg/sdk/go/responses.go
```diff
@@ -61,30 +61,29 @@ type MembersPage struct {
...
-func (res issueKeyRes) Empty() bool {
+func (res KeyRes) Empty() bool {
 	return res.Value == ""
 }
 
 type retrieveKeyRes struct {
 	ID        string     `json:"id,omitempty"`
 	IssuerID  string     `json:"issuer_id,omitempty"`
 	Subject   string     `json:"subject,omitempty"`
-	Type      uint32     `json:"type,omitempty"`
```
manuio, line 87: Is this ok?

pkg/sdk/go/keys.go
```diff
@@ -6,43 +6,58 @@ import (
...
 
 	"github.com/mainflux/mainflux/pkg/errors"
 )
 
+type keyReq struct {
+	Type     uint32        `json:"type,omitempty"`
+	Duration time.Duration `json:"duration,omitempty"`
+}
+
 const keysEndpoint = "keys"
```
manuio, line 19: You can probably define this in the const block below

cli/keys.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// NewKeysCmd returns keys command.
+func NewKeysCmd() *cobra.Command {
+	cmd := cobra.Command{
+		Use:   "keys",
+		Short: "Keys management",
+		Long:  `Keys management: issue, revoke, or retrieve API key.`,
+		Run: func(cmd *cobra.Command, args []string) {
+			logUsage("keys [issue | revoke | retrieve]")
```
manuio, line 84: In other commands this is defined as `cmd.Short`

manuio (changes requested):

pkg/sdk/go/keys.go
drasko: @manuio I am not sure, because block belov is `iota` enumeration.

manuio: oh yes, right. REsolving this one

cli/keys.go
fbugarski: In functions NewGroupsCmd, NewThingsCmd, NewChannlesCmd it is not defined as cmd.Short..
For example : https://github.com/mainflux/mainflux/blob/42dd813521e11b5ec2cac07509b07042ef35d5b7/cli/groups.go#L212
I followed other functions as examples to be consistent.

pkg/sdk/go/keys.go
```diff
@@ -6,43 +6,58 @@ import (
...
+const (
+	// LoginKey is temporary User key received on successfull login.
+	LoginKey uint32 = iota
+	// RecoveryKey represents a key for resseting password.
+	RecoveryKey
+	// APIKey enables the one to act on behalf of the user.
+	APIKey
+)
```
dborovcanin, line 28: Can you use these constants from Auth service? Please check if that's OK from dependencies perspective (we don't want to pull all the Auth service and its dependencies), but I think that it should be.

dborovcanin (approved):

pkg/sdk/go/responses.go
fbugarski: @manuio I removed type because you can use only type 2 to Issue API Key.

cli/keys.go
manuio: Right, it's consistent. But in fact this wasn't properly done in a first place. Firstly, we should use `cmd.Short` since this can be seen with `--help` flag while now it will be only displayed if you type a wrong command. Secondly, we should replace `cmd.Short` by `cmd.Use` following this Cobra dicctionnary: 
```
	// Use is the one-line usage message.
	// Recommended syntax is as follow:
	//   [ ] identifies an optional argument. Arguments that are not enclosed in brackets are required.
	//   ... indicates that you can specify multiple values for the previous argument.
	//   |   indicates mutually exclusive information. You can use the argument to the left of the separator or the
	//       argument to the right of the separator. You cannot use both arguments in a single use of the command.
	//   { } delimits a set of mutually exclusive arguments when one of the arguments is required. If the arguments are
	//       optional, they are enclosed in brackets ([ ]).
	// Example: add [-F file | -D dir]... [-f format] profile
	Use string
```

ILike that when we use the `--help` flag the generated doc would be complete.

If you want I can do it in a separate PR.  Or maybe you can take this one?

pkg/sdk/go/responses.go
manuio: Then, why `keyReq{Type: APIKey, Duration: d}` have a type?

cli/keys.go
dborovcanin: I would say it's better to do it in a separate PR so that we keep this one focused on API keys.

fbugarski: Lets do it in separate PR and we should incloude this issue in that PR : https://github.com/mainflux/mainflux/issues/1551.

manuio: @dusanb94 @fbugarski Ok, I can take both PRs.

manuio (approved):
LGTM

dborovcanin (approved):

drasko (approved):
LGTM

PR #1541, opened 2022-01-10
MF-1308 - Use IETF Health Check standard

### LINKED ISSUE #1308: Use IETF Health Check standard
**FEATURE REQUEST**

Mainflux microservices have a current `/version` API endpoint to return liveness info.

However, it would be better to adhere to an [IETF RFC: Health Check Response Format for HTTP APIs](https://tools.ietf.org/html/draft-inadarei-api-health-check-05).

There is a [go implementation](https://github.com/nelkinda/health-go) that can be used for this purpose.

nice-to-have

manuio:

Resolves #1308 

Resolved in UI repo here: https://github.com/mainflux/ui/pull/233
Resolved in devops repo here: https://github.com/mainflux/devops/pull/97
Resolved in docs repo here: https://github.com/mainflux/docs/pull/119

Since we wanted to have information about the build time and the corresponding commit hash of the version we added `build_time` and `commit` fields. However these fields are non standard and we should decide what to do about in this issue: https://github.com/mainflux/mainflux/issues/1549


consumers/writers/api/transport.go
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Mainflux
 // SPDX-License-Identifier: Apache-2.0
 
+//go:build !test
```
drasko: Why do we have both these lines. Are they necessary?

drasko:
I like this implementation. For now it looks maybe redundant, as version was functioning well, but I think in the future we will add some of the fields. For example, I think that `uptime` would be useful. Also, I find it more logical that we test status of the service via `/health` endpoint, rather than via `/version` endpoint.

On the other hand, I would like to hear from @nmarcetic and @dusanb94 does this change add the value, or maybe complicates things without particular benefit. @mteodor and @blokovi also - what do you think?

drasko (changes requested):
We should update Swagger YAML files as well

consumers/writers/api/transport.go
manuio: I don't know why, it was done by the editor. I removed this one and all others.

manuio:
> I like this implementation. For now it looks maybe redundant, as version was functioning well, but I think in the future we will add some of the fields. For example, I think that `uptime` would be useful. Also, I find it more logical that we test status of the service via `/health` endpoint, rather than via `/version` endpoint.
> 
> On the other hand, I would like to hear from @nmarcetic and @dusanb94 does this change add the value, or maybe complicates things without particular benefit. @mteodor and @blokovi also - what do you think?

@drasko I checked the code of the repo that you proposed and I prefered to use it as example than add it in the vendor. We were almost using nothing from there and the standard is not so complicated to follow.

blokovi (dismissed):
Just for reference, we should see if this can be used in [Dockerfile](https://github.com/mainflux/mainflux/blob/master/docker/Dockerfile) for HEALTCHECK instruction:
https://docs.docker.com/engine/reference/builder/#healthcheck

We already use this instruction in VerneMQ [Dockerfile](https://github.com/mainflux/mainflux/blob/master/docker/vernemq/Dockerfile#L50)

blokovi (dismissed):

drasko:
> Just for reference, we should see if this can be used in [Dockerfile](https://github.com/mainflux/mainflux/blob/master/docker/Dockerfile) for HEALTCHECK instruction: https://docs.docker.com/engine/reference/builder/#healthcheck
> 
> We already use this instruction in VerneMQ [Dockerfile](https://github.com/mainflux/mainflux/blob/master/docker/vernemq/Dockerfile#L50)

I find this very interesting. Do we have to change our Dockerfile for this?

What is K8s looking for in the response - only status code?

blokovi:
> I find this very interesting. Do we have to change our Dockerfile for this?
> 
> What is K8s looking for in the response - only status code?

@drasko 

Yes, change in Dockerfile is needed. 

That Docker healthcheck has not been integrated in the Kubernetes, there we can use [Liveness and readiness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)

http/api/logging.go
```diff
@@ -1,8 +1,6 @@
 // Copyright (c) Mainflux
 // SPDX-License-Identifier: Apache-2.0
 
-// +build !test
-
```
dborovcanin: Please don't remove test tags.

dborovcanin (changes requested):

http/api/logging.go
manuio: Ok but can you explain me why this tags are not everywhere? For what this is needed?

manuio: Replaced by the new one `//go:build`

pkg/sdk/go/health_test.go
```diff
@@ -24,14 +24,14 @@ func TestVersion(t *testing.T) {
 		empty bool
 		err   error
 	}{
-		"get version": {
+		"get things service version": {
```
dborovcanin: Please check all the response fields, not only Version.

dborovcanin (changes requested):

http/api/logging.go
dborovcanin: > Ok but can you explain me why this tags are not everywhere? For what this is needed?

It is used to ignore logging and metrics for testing (coverage included) - it is used here: https://github.com/mainflux/mainflux/blob/6483969927adb71736b8d4ed7553f8c2988973ce/scripts/ci.sh#L98

dborovcanin (dismissed):

health.go
```diff
@@ -0,0 +1,51 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Add(contentType, contentTypeJSON)
+		if r.Method != http.MethodGet && r.Method != http.MethodHead {
+			w.WriteHeader(http.StatusMethodNotAllowed)
+			return
+		}
+
+		res := HealthInfo{
+			Status:      svcStatus,
+			Description: service + " service",
```
mteodor: `" service"` as constant?

health.go
```diff
@@ -0,0 +1,51 @@
...
+			Version:     version,
```
mteodor: it would be good to have timestamp of build so that when debugging deployment by querying the health you can confirm that right version is deployed

health.go
```diff
@@ -0,0 +1,51 @@
...
+
+package mainflux
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+const (
+	version         string = "0.12.1"
```
mteodor: is there a way that version is automatically set by release process
something like explained here

manuio: You forgot the link :)

health.go
manuio: It should be covered by @blokovi proposition. There is an issue about: https://github.com/mainflux/mainflux/issues/1543

health.go
drasko: @mteodor explained where?

health.go
drasko: @mteodor there is a way to put timestamp and Git hash during the compilation phase. We had this before. I think it would make sense to add it back.

health.go
mteodor: https://polyverse.com/blog/how-to-embed-versioning-information-in-go-applications-f76e2579b572/

Makefile
```diff
@@ -11,8 +11,13 @@ DOCKERS_DEV = $(addprefix docker_dev_,$(SERVICES))
 CGO_ENABLED ?= 0
 GOARCH ?= amd64
 
+VERSION = 0.12.1
+TIME = $(shell date +%F_%T)
+
 define compile_service
-	CGO_ENABLED=$(CGO_ENABLED) GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) go build -mod=vendor -ldflags "-s -w" -o ${BUILD_DIR}/mainflux-$(1) cmd/$(1)/main.go
+	CGO_ENABLED=$(CGO_ENABLED) GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) \
+	go build -mod=vendor -ldflags "-s -w -X 'github.com/mainflux/mainflux.Version=$(VERSION)' -X 'github.com/mainflux/mainflux.BuildTime=$(TIME)'" \
```
drasko: When we are there, we could add Git hash as well (probably Git would need to beinstalled in Dockerfile for build as well).

Also, I am not sure that building for a certain version should go this way. I would rather expect that developer first tags  the release. The when the desired version is selected in ENV var, that version is checked-out and built. When `VERSION` is left empty, then master HEAD is built. Maybe in that case we can leave default version `"master"`, to give an information to the user that built was produced from master, and not from a tagged version.

@manuio @dusanb94 what do you think about that?

drasko (changes requested):

Makefile
dborovcanin: I think that Git hash is a good idea. The version really depends on how we decide to proceed. If we'll have a commit hash, it makes sense to bound Version to Git tag or release.

manuio: @drasko @dusanb94 I used `docker build --build-args` to pass version and commit using git commands directly from the Makefile. It avoids any installation of git in the Dockerfile.

blokovi (dismissed):

api/auth.yml
```diff
@@ -542,6 +553,31 @@ components:
           uniqueItems: true
           items:
             type: string
+    HealthInfo:
```
dborovcanin: @manuio Is it possible to extract `HealthInfo` to a separate file and share it between services?

manuio: @dusanb94 done. The only problem is if for example you want to verify yaml format with the online swagger editor, the example will be empty.

manuio: We are also loosing specific service description...

health.go
```diff
@@ -0,0 +1,74 @@
...
+package mainflux
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+const (
+	contentType     = "Content-Type"
+	contentTypeJSON = "application/json"
```
arvindh123: Content type can be 
application/health+json
as mentioned in  https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check-05#section-3
https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check-05#section-7

manuio: done

arvindh123:
In this PR we have only the version  of service and do not have any other health check status or metrics  like uptime, CPU usage , other component status.  

Example Output : https://tools.ietf.org/id/draft-inadarei-api-health-check-04.html#rfc.section.5

It seems to  be similar to the version endpoint

manuio:
> In this PR we have only the version of service and do not have any other health check status or metrics like uptime, CPU usage , other component status.
> 
> Example Output : https://tools.ietf.org/id/draft-inadarei-api-health-check-04.html#rfc.section.5
> 
> It seems to be similar to the version endpoint

@arvindh123 Indeed, for now we only have 3 standard fields: `status`, `version`, and `description`. `build_time` and `commit` are custom fields. I didn't want to add it in this PR but we can maybe open another issue to talk about this `check` field. However, as previously discussed maybe this can be solved with docker.

Makefile
```diff
@@ -23,6 +28,9 @@ define make_docker
 		--build-arg SVC=$(svc) \
 		--build-arg GOARCH=$(GOARCH) \
 		--build-arg GOARM=$(GOARM) \
+		--build-arg VERSION=$(shell git describe --abbrev=0 --tags) \
```
drasko: Please check if you need to change our Dockerfile as well, as it is `alpine go`, maybe you need to install Git into it (i.e. try to build Docker images, see if compilation will work).

drasko (changes requested):

Makefile
manuio: @drasko Compilation work, I already tested. There is no need to install git in the Dockerfile since you pass this values as arguments.

Makefile
```diff
@@ -12,7 +12,12 @@ CGO_ENABLED ?= 0
 GOARCH ?= amd64
 
 define compile_service
-	CGO_ENABLED=$(CGO_ENABLED) GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) go build -mod=vendor -ldflags "-s -w" -o ${BUILD_DIR}/mainflux-$(1) cmd/$(1)/main.go
+	CGO_ENABLED=$(CGO_ENABLED) GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) \
+	go build -mod=vendor -ldflags "-s -w \
+	-X 'github.com/mainflux/mainflux.BuildTime=$(TIME)' \
+	-X 'github.com/mainflux/mainflux.Version=$(VERSION)' \
+	-X 'github.com/mainflux/mainflux.Commit=$(COMMIT)'" \
```
dborovcanin, line 22: By default, these (`TIME`, `VERSION`, `COMMIT`) will be empty. Please set them like in `make_dockers`.

dborovcanin (changes requested):

Makefile
manuio: I we do so we have to install git in the Dockerfile. Maybe it's more understandable to do it like that:
```
	go build -mod=vendor -ldflags "-s -w \
	-X 'github.com/mainflux/mainflux.BuildTime=$(6)' \
	-X 'github.com/mainflux/mainflux.Version=$(4)' \
	-X 'github.com/mainflux/mainflux.Commit=$(5)'" \
	-o ${BUILD_DIR}/mainflux-$(1) cmd/$(1)/main.go
```

dborovcanin: No need for that. You can set it as variables in the Makefile like so:
```make
VERSION ?= $(shell git describe --abbrev=0 --tags) 
COMMIT ?= $(shell git rev-parse HEAD) 
TIME ?= $(shell date +%F_%T)
```
The `?=` will take care to set only if variables are empty, which will be the case for the local build, and won't be the case for Docker build.

drasko: But I think we need those filled in the Docker build as well, otherwise `/health` will not work correctly for our Docker images, right?

dborovcanin: Also, maybe it's better to have for TIME something like `date --rfc-3339=seconds` rather than a custom format.

manuio: @dusanb94 Let me try what you proposed. It's probably the way to go.
@drasko It is filled in the docker build. Here:
```
		--build-arg VERSION=$(VERSION) \
		--build-arg COMMIT=$(COMMIT) \
		--build-arg TIME=$(TIME) \
```
and here:
```
ARG VERSION
ARG COMMIT
ARG TIME
```

dborovcanin (dismissed):

api/schemas/HealthInfo.yml
```diff
@@ -0,0 +1,23 @@
...
+  status:
+    type: string
+    description: Service status.
+    enum:
+      - pass
+  version:
+    type: string
+    description: Service version.
+    example: 0.0.1
+  commit:
```
drasko, line 12: We cannot leave `commit`. We must adhere to standard. If you have fields that are missing, we probably must work with standardization body and argument why we need those, so that they can rectify in the next version pf the draft. Otherwise please find the closest field and put commit information there (probably `releaseId`).

drasko (changes requested):

api/schemas/HealthInfo.yml
manuio: ok, where I put `build_time`?

drasko: That's a good question. Let me check.

drasko: @manuio I have sent an e-mail to the author. Maybe we'll have to open an issue [here](https://github.com/inadarei/rfc-healthcheck/issues) as well. Let's give it a day or two, and if we have no answers, I will merge the current PR.

drasko: I have opened an issue as well: https://github.com/inadarei/rfc-healthcheck/issues/76

mteodor (dismissed):

health.go
```diff
@@ -0,0 +1,74 @@
...
+
+// HealthInfo contains version endpoint response.
+type HealthInfo struct {
+	// Status contains service status.
+	Status string `json:"status"`
+
+	// Version contains current service version.
+	Version string `json:"version"`
+
+	// Commit represents the service build time.
```
drasko: This is not correct. It is a Git commit hash.

drasko (changes requested):

health.go
manuio: right

drasko: Please fix

drasko (changes requested):

drasko (approved):
LGTM

drasko:
@manuio @dusanb94 I have merged this PR, I think it is very useful and that we have it. Once @inadarei or someone else reacts on https://github.com/inadarei/rfc-healthcheck/issues/76 we'll see how to progress with https://github.com/mainflux/mainflux/issues/1549

PR #1540, opened 2022-01-09
MF-1263 - Move repeating errors to the separate package

manuio:
Signed-off-by: Manuel Imperiale <manuel.imperiale@gmail.com>

Resolves #1263

PR #1544, opened 2022-01-12
MF-1348 - Add transport errors logging

### LINKED ISSUE #1237: Inconsistent error sources in services and unaddressed errors in transport.go files

Currently, we send, via `transport.go`, using `switch` case, errors and error codes destined to HTTP clients. We do it in various services. Errors come from different sources such as database, redis cache, etc. Before reaching the `switch` case in transport.go, errors pass by `service.go`

The first problem is that not all of the errors that rech the `transport.go` are addressed by the switch case. The second problem is that we have errors defined in different packages coming from the different sources and transport should know in advance about these packages and sources to handle errors correctly. To sum it up, we have a double problem: firstly, inconsistent error sources, secondly unaddressed errors.

Also, as a by-product, swagger, i.e. openapi specs are difficult to write so that they agree with error handling in `transport.go`.

Medium. We need this so we can have a good infrastructure to build usable HTTP clients. Also, good openapi files can automate doc generation and API code generation.

### LINKED ISSUE #1348: Add go-kit transport level logging
**ENHANCEMENT**

Currently Mainflux go-kit implementation of microservices do not contain transport-level logging, meaning that many of the HTTP requests that do not make it to the logging middleware (they do not pass validation on the transport level, so are not propagated to the next layer) will not be logged at all.

Following [the official doc](https://gokit.io/examples/stringsvc.html#transport-logging), add the transport level logging in all microservices.

Must-have

manuio:

Resolves #1348 
Resolves #1237


internal/httputil/logging.go
```diff
@@ -0,0 +1,25 @@
...
+
+// Middleware is an ErrorEncoder middleware
+type Middleware func(kithttp.ErrorEncoder) kithttp.ErrorEncoder
+
+// LoggingErrorEncoder is a go-kit error encoder logging decorator.
+func LoggingErrorEncoder(logger logger.Logger) Middleware {
+	return func(encode kithttp.ErrorEncoder) kithttp.ErrorEncoder {
+		return func(ctx context.Context, err error, w http.ResponseWriter) {
+			logger.Error(err.Error())
+			defer encode(ctx, err, w)
```
dborovcanin: No need to defer since this is the last operation in the function.

cmd/influxdb-reader/main.go
```diff
@@ -28,6 +28,8 @@ import (
 )
 
 const (
+	svcName = "influxdb-reader"
+
```
dborovcanin: Please remove an empty line.

logger/mock.go
```diff
@@ -0,0 +1,29 @@
...
+package logger
+
+var _ Logger = (*loggerMock)(nil)
+
+type loggerMock struct{}
+
+// NewMock returns wrapped go kit logger mock.
+func NewMock() (Logger, error) {
+	return &loggerMock{}, nil
+}
```
dborovcanin: No need to return an error here, since it's always `nil`. It will also simplify using the mock (instead of `logger, _ := logger.NewMock()` simply call `logger.NewMock()`)

dborovcanin (changes requested):
A very important contribution. Please fix minor remarks I had and it's ready.

internal/httputil/logging.go
```diff
@@ -0,0 +1,25 @@
...
+
+// LoggingErrorEncoder is a go-kit error encoder logging decorator.
+func LoggingErrorEncoder(logger logger.Logger) Middleware {
+	return func(encode kithttp.ErrorEncoder) kithttp.ErrorEncoder {
+		return func(ctx context.Context, err error, w http.ResponseWriter) {
+			logger.Error(err.Error())
+			encode(ctx, err, w)
+		}
+	}
+}
```
dborovcanin: This can be replaced with something like:
```go

// Copyright (c) Mainflux

package httputil

import (
	"context"
	"net/http"

	kithttp "github.com/go-kit/kit/transport/http"
)

// LoggingErrorEncoder is a go-kit error encoder logging decorator.
func LoggingErrorEncoder(logger logger.Logger, enc kithttp.ErrorEncoder) kithttp.ErrorEncoder {
	return func(ctx context.Context, err error, w http.ResponseWriter) {
		logger.Error(err.Error())
		enc(ctx, err, w)
	}
}
```
and later used like:
```go
kithttp.ServerErrorEncoder(httputil.LoggingErrorEncoder(logger, encodeError))
```

dborovcanin (changes requested):
Does this mean that in case the error is not on transport level, we will have errors logged twice - by the service and by the transport? Obviously, transport errors won't be logged twice as requests never reached the service.

drasko:
@manuio can you please proceed with this one, taking into accounts @dusanb94's remarks. I think we are close to merge it.

manuio:
@drasko Can we first merge this one to avoid conflicts? https://github.com/mainflux/mainflux/pull/1540 
We had a talk with @dusanb94 and the idea would be to return a specific error for `validation` that we will check in `LoggingErrorEncoder` to log or not the error.

mteodor (dismissed):

internal/httputil/logging.go
drasko: @manuio is this resolved?

drasko:
@manuio before continuing with this one, we'd like to do a small experiment with removing logging from service layer, and adding it to transport layer only, in order to avoid logging two times.

@dusanb94 will do this experiment and inform you how to proceede forward.

manuio:
@drasko ok! indeed, that would be a solution but more  HTTP API oriented. Isn't going to be a problem with gRPC logging?

drasko:
> @drasko ok! indeed, that would a solution but more HTTP API oriented. Isn't going to be a problem with gRPC logging?

Hmmm... maybe, but I am not sure. @dusanb94 will do a small PoC, taking into account gRPC API as well, and tell us.

dborovcanin:
> @drasko ok! indeed, that would be a solution but more HTTP API oriented. Isn't going to be a problem with gRPC logging?

@manuio gRPC should also be covered by this because we're relying on Go Kit. You can take a quick look at the PoC I wrote (for HTTP only at the moment), but it needs improvements in design: https://github.com/mainflux/mainflux/compare/master...dusanb94:transport-logging?expand=1

drasko:
@manuio can you please fix conflicts and update the branch. I think we are near to merging this one.

@dusanb94 no more particular remarks?

manuio:
@drasko After to talk with @dusanb94 we think that the best would be to have specific errors for validation such as `ErrEmptyToken`, `ErrEmptyID`, `ErrNameSize`, "ErrInvalidOffset", etc. The encodeError decorator would contain a switch case and only log this validation errors.  

The previous solution was to wrap validation errors with a single `ErrValidation` and use it in encodeError to log or not. 

That's why the solution with specific errors looks more fine grained and less redundant, if you agree I'll proceed.

dborovcanin:
> 

@manuio As we already discussed, this looks good to me. I only wondered if these errors should be on package level (different for each service), or global. However, since we have a lot of them repeating, probably global errors in `/internal/httputil/transport.go` are fine.

drasko:
@manuio @dusanb94 OK for me also - I agree with what @dusanb94 wrote. Please proceed in that direction.

auth/api/grpc/server.go
```diff
@@ -236,13 +237,21 @@ func encodeError(err error) error {
...
+	case errors.Contains(err, errors.ErrMalformedEntity),
+		err == apiutil.ErrInvalidAuthKey,
+		err == apiutil.ErrMissingID,
+		err == apiutil.ErrMissingMemberType,
+		err == apiutil.ErrMissingPolicySub,
+		err == apiutil.ErrMissingPolicyObj,
+		err == apiutil.ErrMissingPolicyAct:
+		return status.Error(codes.InvalidArgument, err.Error())
+	case errors.Contains(err, errors.ErrAuthentication),
+		err == apiutil.ErrMissingEmail):
```
dborovcanin: This looks like a syntax error.

auth/api/grpc/server.go
```diff
@@ -236,13 +237,21 @@ func encodeError(err error) error {
...
 		return status.Error(codes.Unauthenticated, err.Error())
 	case errors.Contains(err, errors.ErrAuthorization):
 		return status.Error(codes.PermissionDenied, err.Error())
-	case errors.Contains(err, auth.ErrKeyExpired):
+	case errors.Contains(err, auth.ErrKeyExpired),
+		err == apiutil.ErrMissingToken:
```
dborovcanin: Group cases by the resulting status - there are 2 cases that result in 
`return status.Error(codes.Unauthenticated, err.Error())`.

provision/api/requests.go
```diff
@@ -1,6 +1,8 @@
 package api
 
-import "github.com/mainflux/mainflux/pkg/errors"
+import (
+	"github.com/mainflux/mainflux/internal/apiutil"
+)
```
dborovcanin: `import "github.com/mainflux/mainflux/internal/apiutil"`

things/api/auth/grpc/requests.go
```diff
@@ -3,16 +3,22 @@
 
 package grpc
 
-import "github.com/mainflux/mainflux/pkg/errors"
+import (
+	"github.com/mainflux/mainflux/internal/apiutil"
+)
```
dborovcanin: `import "github.com/mainflux/mainflux/internal/apiutil"`

things/api/things/http/endpoint_test.go
```diff
@@ -1087,9 +1090,9 @@ func TestSearchThings(t *testing.T) {
 		{
 			desc:   "search things with zero limit",
 			auth:   token,
-			status: http.StatusOK,
+			status: http.StatusBadRequest,
```
dborovcanin, line 1093: Zero limit should fall back to the default. That is, this PR should not change the API.

things/api/things/http/endpoint_test.go
```diff
@@ -1907,9 +1910,9 @@ func TestListChannels(t *testing.T) {
 		{
 			desc:   "get a list of channels with zero limit and offset 1",
 			auth:   token,
-			status: http.StatusOK,
+			status: http.StatusBadRequest,
```
dborovcanin, line 1913: The same here.

things/api/things/http/requests.go
```diff
@@ -32,23 +32,23 @@ type createThingReq struct {
...
 
 func (req createThingReq) validate() error {
 	if req.token == "" {
-		return errors.ErrAuthentication
+		return apiutil.ErrMissingToken
 	}
 
-	if req.ID != "" && validateUUID(req.ID) != nil {
-		return errors.ErrMalformedEntity
+	if err := validateUUID(req.ID); req.ID != "" && err != nil {
```
dborovcanin: The same as:
```go
if err := validateUUID(req.ID); err != nil {
```
because `validateUUID` internally calls `uuid.FromString` that checks if the string is empty.

dborovcanin (changes requested):

things/api/things/http/endpoint_test.go
manuio: I did it because for all other services zero limit results in a BadRequest error.

things/api/things/http/requests.go
manuio: Yes but the method would return an error when you are creating a Thing without ID.

things/api/things/http/endpoint_test.go
dborovcanin: Since 0 limit is only the case when **actual limit param is sent and its value is 0**, this is a valid fix.

dborovcanin (dismissed):

readers/api/endpoint_test.go
```diff
@@ -67,7 +69,7 @@ func (tr testRequest) make() (*http.Response, error) {
 		return nil, err
 	}
 	if tr.token != "" {
-		req.Header.Set("Authorization", httputil.BearerPrefix+tr.token)
+		req.Header.Set("Authorization", apiutil.BearerPrefix+tr.token)
```
mteodor: what about things token?

readers/api/transport.go
```diff
@@ -74,79 +75,74 @@ func MakeHandler(svc readers.MessageRepository, tc mainflux.ThingsServiceClient,
...
-	t, err := httputil.ExtractAuthToken(r)
+	to, err := apiutil.ReadFloatQuery(r, toKey, 0)
 	if err != nil {
 		return nil, err
 	}
 
 	req := listMessagesReq{
 		chanID: bone.GetValue(r, "chanID"),
-		token:  t,
+		token:  apiutil.ExtractBearerToken(r),
```
mteodor: what about thing key request?

things/api/things/http/requests.go
```diff
@@ -175,20 +186,20 @@ type createChannelsReq struct {
...
 
 	if len(req.Channels) <= 0 {
-		return errors.ErrMalformedEntity
+		return apiutil.ErrEmptyList
 	}
 
 	for _, channel := range req.Channels {
-		if channel.ID != "" && validateUUID(channel.ID) != nil {
-			return errors.ErrMalformedEntity
+		if err := validateUUID(channel.ID); channel.ID != "" && err != nil {
```
mteodor: if `id == ""` wouldnt validate UUID fail anyways?

manuio: Yes, but we want to check only if an ID was sent.

manuio:
@drasko In this PR we are adding transport error logging but I had to change the logic with Token extraction to don't return errors in decode layers. For that reason I did another branch and continued in another PR with Bearer and Thing schemes. Can we review and merge this one ASAP and let @dusanb94  review the merged PR? I can add any remark that he have in the new PR.

auth/api/http/keys/endpoint_test.go
```diff
@@ -120,11 +122,11 @@ func TestIssue(t *testing.T) {
 			status: http.StatusCreated,
 		},
 		{
-			desc:   "issue recovery key",
+			desc:   "issue recovery key with wrong api type",
```
dborovcanin: What's `wrong api type`?

internal/apiutil/errors.go
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package apiutil
+
+import (
+	"github.com/mainflux/mainflux/pkg/errors"
+)
```
dborovcanin:
```go
import "github.com/mainflux/mainflux/pkg/errors"
```
Also, a nice comment in this file explaining what these errors represent would be good.

dborovcanin (changes requested):

things/api/things/http/requests.go
dborovcanin: I would also prefer having this separated, it's more readable:
```go
if channel.ID != "" {
    if err:=validateUUID(channel.ID); err != nil {
      return err
}
...
```

internal/apiutil/errors.go
```diff
@@ -5,7 +5,9 @@ package apiutil
 
 import "github.com/mainflux/mainflux/pkg/errors"
 
-// Errors used to distingate and log on API request validation
+// Errors defined in this file are used by the LoggingErrorEncoder decorator
+// to distingate and log API request validation errors and avoid that service
```
dborovcanin: To distinguish?

dborovcanin: @manuio Please fix this typo and we're ready to merge.

manuio: ok!

blokovi (approved):

dborovcanin (approved):

PR #1555, opened 2022-01-27
NOISSUE - Fix auth members list response

manuio:

### What does this do?
Fix the JSON tags of members response in auth service, add missing comments and remove unused structures

### DIFF
--- a/auth/api/http/groups/endpoint.go
+++ b/auth/api/http/groups/endpoint.go
@@ -358,11 +358,15 @@ func buildUsersResponse(mp auth.MemberPage) memberPageRes {
 			Limit:  mp.Limit,
 			Name:   mp.Name,
 		},
-		Members: []interface{}{},
+		Members: []memberRes{},
 	}
 
-	for _, m := range mp.Members {
-		res.Members = append(res.Members, m)
+	for _, member := range mp.Members {
+		memb := memberRes{
+			ID:   member.ID,
+			Type: member.Type,
+		}
+		res.Members = append(res.Members, memb)
 	}
 
 	return res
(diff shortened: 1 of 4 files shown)


auth/api/http/groups/endpoint.go
```diff
@@ -358,11 +358,15 @@ func buildUsersResponse(mp auth.MemberPage) memberPageRes {
...
-		Members: []interface{}{},
+		Members: []memberRes{},
 	}
 
-	for _, m := range mp.Members {
-		res.Members = append(res.Members, m)
+	for _, member := range mp.Members {
+		memb := memberRes{
+			ID:   member.ID,
+			Type: member.Type,
```
dborovcanin: Do we really need type here? It looks like it should be a part of `memberPageRes` because all the members are of the same type.

manuio: done

dborovcanin: Do we even need `memberRes` since we only use ID? Does it make sense to return only a list of strings that are IDs?

auth/api/http/groups/endpoint.go
```diff
@@ -350,19 +350,20 @@ func buildGroupsResponse(gp auth.GroupPage) groupPageRes {
...
 			Name:   mp.Name,
 		},
-		Members: []interface{}{},
+		Type:    groupType,
+		Members: []string{},
 	}
 
-	for _, m := range mp.Members {
-		res.Members = append(res.Members, m)
+	for _, member := range mp.Members {
```
drasko: Keep `m` here, it was better.

drasko (changes requested):

drasko:
@manuio can you please update the branch

dborovcanin (approved):

dborovcanin (approved):

drasko (approved):
LGTM

PR #1558, opened 2022-02-01
MF-1551 - Fix Cobra usage commands and clean unnecessary struct types

### LINKED ISSUE #1551: Fix Cobra usage commands and clean unnecessary struct types
**ENHANCEMENT**
   - code refactor

nice-to-have

Clean unnecessary struct types in the initialization of slices

For example:
https://github.com/mainflux/mainflux/blob/bcc8cf71191efc8af64e595d292a7ef9a416b0ae/cli/groups.go#L14
https://github.com/mainflux/mainflux/blob/bcc8cf71191efc8af64e595d292a7ef9a416b0ae/cli/bootstrap.go#L39

manuio:

### What does this do?
Fix Cobra usage commands and clean unnecessary struct types

Resolves #1551


cli/bootstrap.go
```diff
@@ -11,13 +11,13 @@ import (
 )
 
 var cmdBootstrap = []cobra.Command{
-	cobra.Command{
-		Use:   "add",
-		Short: "add <JSON_config> <user_auth_token>",
+	{
+		Use:   "add {JSON_config} {user_auth_token}",
```
drasko: Why are you using curly braces `{}` instead of [Linux-doc convention](https://web.mit.edu/rhel-doc/4/RH-DOCS/rhel-rg-en-4/s1-intro-conventions.html) for replaceable which is `<>`? Is this something demanded by Cobra?

Additionally, in Linux manuals "Brackets (`[]`) surround optional arguments, vertical bars (`|`) separate choices". I would like that we follow this convention also.

drasko (changes requested):

cli/bootstrap.go
manuio: Because of Cobra recommendation: https://pkg.go.dev/github.com/spf13/cobra#Command
What do you mean about `[]`? Isn't already like that?

drasko: This is related to a set (several) arguments, out of which at least one is mandatory. This notation is taken from [IBM](https://www.ibm.com/docs/en/iis/8.5?topic=home-reading-command-line-syntax). In the [Microsfot notation](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/command-line-syntax-key) the role of `<>` replaceable is better explained.

Point here is that we are really using replaceable - i.e. applying the actual value of `user_auth_token` - and not putting the literal item (i.e. string "user_auth_token"), selected from the set of possible items.

At least this is how I understand. @dusanb94 @mteodor what do you think?

Interesting discussion can be seen here: https://github.com/clap-rs/clap/discussions/2947 as well, where [Oracle convention](https://docs.oracle.com/cd/B10501_01/server.920/a90842/preface.htm#971044) is mentioned, and by it "Braces enclose two or more items, one of which is required."

manuio: TBH I also prefer `<>` and as you said `{}` isn't really correct if the value can change. I guess that the correct one would be `{<user_token>}` but is not really readable.  If others don't mind I'm going to revert.

I'd also like to know your opinion about descriptions (Short and Long). I don't find them really consistent.

blokovi (dismissed):

cli/bootstrap.go
dborovcanin: I agree with @drasko. As long we can follow rules, we should use the standard convention.
As for long and short - let's use it as intended. Short for quick explanation, long for the same explanation enriched with more details.

dborovcanin (approved):

blokovi (approved):

drasko (approved):
LGTM

PR #1559, opened 2022-02-03
MF-1240 - Return to service transport layer only service errors

### LINKED ISSUE #1240: Return to service transport layer only service errors
Currently, we send to transport.go errors from different sources (db, cache, etc.). However, not all of these errors are addressed by the switch case in the transport.go. So we have a double problem: firstly, inconsistent error sources, secondly unaddressed errors.

Also, we need to make swagger spec in agreement with new error handling, plus, we need to migrate swaggers to openapi 3. 

This relates also to issue #1261, so the single PR should tackle both issues at once.

manuio:

### What does this do?
Refactor encodeError method in services transport layer to ensure that service and db errors are properly encoded in the HTTP response. 

Resolves #1240 

yes


blokovi (dismissed):

auth/api/http/groups/transport.go
```diff
@@ -327,19 +326,22 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 		w.WriteHeader(http.StatusForbidden)
 	case errors.Contains(err, auth.ErrMemberAlreadyAssigned):
 		w.WriteHeader(http.StatusConflict)
-	case errors.Contains(err, io.EOF):
-		w.WriteHeader(http.StatusBadRequest)
-	case errors.Contains(err, io.ErrUnexpectedEOF):
-		w.WriteHeader(http.StatusBadRequest)
```
dborovcanin, line 333: Please doublecheck. These errors are not returned by the service layer but can happen during the decoding of the request body.

certs/api/transport.go
```diff
@@ -136,23 +135,27 @@ func decodeRevokeCerts(_ context.Context, r *http.Request) (interface{}, error)
...
+		errors.Contains(err, errors.ErrRemoveEntity):
+		w.WriteHeader(http.StatusInternalServerError)
+
 	default:
-		switch err.(type) {
-		case *json.SyntaxError:
-			w.WriteHeader(http.StatusBadRequest)
-		case *json.UnmarshalTypeError:
-			w.WriteHeader(http.StatusBadRequest)
-		default:
```
dborovcanin, line 155: Same here, this can happen on the transport layer, not the service.

http/api/endpoint_test.go
```diff
@@ -61,10 +61,10 @@ func (tr testRequest) make() (*http.Response, error) {
 func TestPublish(t *testing.T) {
 	chanID := "1"
 	contentType := "application/senml+json"
-	token := "auth_token"
-	invalidToken := "invalid_token"
+	thinKey := "thing_key"
```
dborovcanin: Fix typo.

http/mocks/things.go
```diff
@@ -31,13 +28,6 @@ func NewThingsClient(data map[string]string) mainflux.ThingsServiceClient {
 func (tc thingsClient) CanAccessByKey(ctx context.Context, req *mainflux.AccessByKeyReq, opts ...grpc.CallOption) (*mainflux.ThingID, error) {
 	key := req.GetToken()
 
-	// Since there is no appropriate way to simulate internal server error,
-	// we had to use this obscure approach. ErrorToken simulates gRPC
-	// call which returns internal server error.
-	if key == ServiceErrToken {
-		return nil, status.Error(codes.Internal, "internal server error")
-	}
-
```
dborovcanin: I'd like to keep both empty and invalid key cases.

provision/api/transport.go
```diff
@@ -89,24 +89,23 @@ func decodeMappingRequest(_ context.Context, r *http.Request) (interface{}, erro
...
+	case errors.Contains(err, errors.ErrAuthentication):
 		w.WriteHeader(http.StatusUnauthorized)
+
 	default:
-		switch err.(type) {
-		case *json.SyntaxError:
-			w.WriteHeader(http.StatusBadRequest)
-		case *json.UnmarshalTypeError:
-			w.WriteHeader(http.StatusBadRequest)
-		default:
```
dborovcanin, line 109: The same remark: can these errors happen on the transport layer? Check all the other places this section is removed.

dborovcanin (changes requested):

certs/api/transport.go
manuio: This errors are related to json.Unmarshal. You can check here https://pkg.go.dev/encoding/json#SyntaxError and here https://pkg.go.dev/encoding/json#UnmarshalTypeError

We are always returning `ErrMalformedEntity` wrapped with the json package error if an error occurs during Unmarshal. This check would be only useful if we return the json error without any wrap.

auth/api/http/groups/transport.go
manuio: This errors are related to protobuf Unmarshal. If there is any error returned we should Wrap it with `ErrMalformedEntity`.  If it occurs in a consumer handler then tthe error is simply never returned.

http/api/endpoint_test.go
manuio: thanks

http/mocks/things.go
manuio: ok

dborovcanin (approved):

blokovi (approved):

bootstrap/api/requests.go
```diff
@@ -8,6 +8,8 @@ import (
 	"github.com/mainflux/mainflux/pkg/errors"
 )
 
+const maxLimitSize = 100
```
drasko, line 11: No need for this "Size" I think. `maxLimit` is sufficient. Is there some particular reason you introduced this `Size` suffix?

bootstrap/api/requests_test.go
```diff
@@ -198,15 +198,15 @@ func TestListReqValidation(t *testing.T) {
 			desc:   "too large limit",
 			key:    "key",
 			offset: 0,
-			limit:  maxLimit + 1,
+			limit:  maxLimitSize + 1,
```
drasko, line 201: Same here.

drasko (changes requested):

bootstrap/api/requests.go
manuio: @drasko Consistency reasons. Also think about `maxNameSize` or any other. Are we going to use `maxName`?

drasko: Uh... I think it is not exactly the same, but I agree, I would probably keep `maxLimitSize` then. @dusanb94 what do you say?

dborovcanin: Keeping it makes sense, since Limit is actually name of the param (though limitSize sounds obnoxious).

drasko (approved):
LGTM

PR #1561, opened 2022-02-04
NOISSUE- Refactor MQTT subscriber

tzzed:
Signed-off-by: tzzed <zerouali.t@gmail.com>

### What does this do?
This PR is a correction of the Suscriber interface checker.
None


pkg/messaging/mqtt/publisher.go
```diff
@@ -8,11 +8,10 @@ import (
 	"time"
 
 	mqtt "github.com/eclipse/paho.mqtt.golang"
+
 	"github.com/mainflux/mainflux/pkg/messaging"
 )
 
-var _ messaging.Publisher = (*publisher)(nil)
```
manuio, line 14: Please, don't remove the interface check.

pkg/messaging/mqtt/subscriber.go
```diff
@@ -9,13 +9,12 @@ import (
 
 	mqtt "github.com/eclipse/paho.mqtt.golang"
 	"github.com/gogo/protobuf/proto"
+
 	log "github.com/mainflux/mainflux/logger"
 	"github.com/mainflux/mainflux/pkg/errors"
 	"github.com/mainflux/mainflux/pkg/messaging"
 )
 
-var _ messaging.Publisher = (*publisher)(nil)
```
manuio, line 17: same here

manuio (changes requested):

pkg/messaging/mqtt/publisher.go
tzzed: It is useless because it is checked by the function `NewPublisher(address string, timeout time.Duration) (messaging.Publisher, error)`

pkg/messaging/mqtt/subscriber.go
tzzed: Here is why I did the correction. As you can see the checked verifies `messaging.Publisher` but we are in subscriber file. `NewSubscriber` does the check.

pkg/messaging/mqtt/publisher.go
dborovcanin: I agree that it is not necessary because of the factory function. However, it is in sync with the rest of the codebase, it is self-documenting, and it is an explicit type assertion, rather than an implicit conversion that happens in `NewPublisher`, so I would also prefer if we keep it. The same goes for `Subscriber`, but with the correct interface. Other than this, the PR looks good and ready for merge.

dborovcanin (changes requested):

pkg/messaging/mqtt/publisher.go
tzzed: done.

dborovcanin (approved):

manuio (approved):
LGTM

PR #1562, opened 2022-02-07
MF - 1416 - Queue Abstraction for Mainflux & RabbitMQ Support

### LINKED ISSUE #1416: Queue Abstraction for Mainflux & RabbitMQ Support
**ENHANCEMENT**

**1. Describe the enhancement you are requesting. Enhancements include:**
This enhancement describes a solution where the NATS queue in Mainflux is replaced with a RabbitMQ queue. Mainflux design is flexible enough and allows the use of any queue or database with minimal development effort. Being able to select or plug in the desired/required message queue will be a differentiator in the future for the platform.

The solution requires a new implementation of the [pubsub ](https://github.com/mainflux/mainflux/blob/0fe9f55bae8d0887d5ea2d7be9fb483c6b0bfafc/pkg/messaging/pubsub.go#L26) interface. Example implementations exist in the code base for NATS and MQTT. 

NATS usage, through references to mainflux/mainflux/pkg/messaging/nats, is hardcoded in several services. Calls to nats.NewPublisher and nats.NewPubSub exist throughout the system. It is recommended that a messaging configuration interface be created and used to select a message broker at system startup. 

The following steps are required to implement this enhancement: 

**Configuration**

- Create an Interface for Messaging Configuration parameters. The configuration should be loaded at system startup and be available for publisher/subscriber factory methods described below. 
- Implement the Messaging Configuration with the existing NATS configuration.
messaging:
	system: nats
	url: ....
	etc.
- Implement the RabbitMQ configuration
messaging:
	system: rabbitmq
	url: ....
       etc. 
- Configuration should be extensible to include other message brokers in the future. 

**RabbitMQ**

- Implement the pubsub interface. 
- Create a Publisher factory method using the factory pattern. This method will use the configuration interface described above to create the appropriate message queue. 
- Use the publisher factory method to replace calls to the **nats.NewPublisher** method
- Create a PubSub factory method using the factory pattern on the same factory class mentioned above. This method will use the configuration described above to create the appropriate message queue. 
- Use the PubSub factory method to replace calls to the **nats.NewPubSub** method to generate a message broker. This method is used across the system by readers and consumers of messages. 
- Create RabbitMQ unit tests.

**2. Indicate the importance of this enhancement to you (must-have, should-have, nice-to-have).**
Must-have.

rodneyosodo:
### What does this do?
Changes the messaging broker from NATS to RabittMQ

Resolves #1416 

Not yet. Will include after approval

Not yet

### Notes
N/A


drasko:
@0x6f736f646f  CI is failing. I restarted it, but there seems to be a problem with rabbitmq tests.

rodneyosodo:
Working to resolve the issue

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -77,7 +77,7 @@ func NewPubSub(url string, logger log.Logger) (PubSub, error) {
 		channel:       ch,
 		logger:        logger,
 		subscriptions: make(map[string]bool),
-		done:          make(chan error),
+		done: make(chan bool),
```
manuio: Looks like the indentation is worng here

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -176,7 +175,6 @@ func (ps *pubsub) handle(deliveries <-chan amqp.Delivery, done chan error, h mes
 		if err := h(msg); err != nil {
 			ps.logger.Warn(fmt.Sprintf("Failed to handle Mainflux message: %s", err))
 		}
-		d.Ack(false)
 	}
-	done <- nil
-}
+	<- done 
+}
```
manuio: new line at the end of the file please

manuio (changes requested):

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: Resolved

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: Resolved

pkg/messaging/rabbitmq/doc.go
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+// Package rabbitmq hold the implementation of the Publisher and PubSub
```
dborovcanin: ...holds the implementation....

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,182 @@
...
+	RoutingKey         = "application specific routing key for fancy topologies"
+	Exchange           = "mainflux"
+	ExchangeKind       = "fanout"
+	QueueDurability    = true
+	QueueDelete        = false
+	QueueExclusivity   = false
+	QueueWait          = false
+	ConsumerTag        = "mainflux-consumer"
+	Mandatory          = false
+	Immediate          = false
```
dborovcanin: Is there any reason these constants are exported?

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,182 @@
...
+	log "github.com/mainflux/mainflux/logger"
+	"github.com/mainflux/mainflux/pkg/messaging"
+	amqp "github.com/rabbitmq/amqp091-go"
+)
+
+// SubjectAllChannels represents subject to subscribe for all the channels.
+const (
+	ChansPrefix        = "channels"
+	SubjectAllChannels = "channels.>"
+	RoutingKey         = "application specific routing key for fancy topologies"
```
dborovcanin: Please change the value of the `routingKey`.

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		return err
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		ps.handle(msgs, handler)
+		wg.Done()
+	}()
```
dborovcanin: There is no `wg.Wait`. What's the purpose of this WaitGroup?

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,182 @@
...
+	Close()
+}
+
+type pubsub struct {
+	conn          *amqp.Connection
+	logger        log.Logger
+	queue         amqp.Queue
+	channel       *amqp.Channel
+	subscriptions map[string]bool
+	done          chan bool
```
dborovcanin: Is this used?

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		ps.handle(msgs, handler)
+		wg.Done()
+	}()
+
+	ps.subscriptions[topic] = true
```
dborovcanin: Should this map be protected for concurrent read/write?

dborovcanin (changes requested):

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: No reason. I am changing that

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: Yes

pkg/messaging/rabbitmq/doc.go
rodneyosodo: Done

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: Done

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: Done

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: Removed the waitgroup. The wait command was blocking

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: How?

dborovcanin: Usually by using a mutex.

rodneyosodo: Done

pkg/messaging/rabbitmq/pubsub.go
dborovcanin: What for, what's the purpose of the `done` channel?

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,183 @@
...
+		queue:         queue,
+		channel:       ch,
+		logger:        logger,
+		subscriptions: make(map[string]bool),
+		done:          make(chan error),
+	}
+	return ret, nil
+}
+
+func (ps *pubsub) Publish(topic string, msg messaging.Message) error {
```
dborovcanin: There is no need to reimplement this in `pubsub`, you can use the implementation from the `publisher.go`:
```go

type pubsub struct {
	publisher
	logger log.Logger
	queue  amqp.Queue
	subscriptions map[string]bool
	mutex sync.Mutex
}
```

pkg/messaging/rabbitmq/publisher.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/golang/protobuf/proto"
+	"github.com/mainflux/mainflux/pkg/messaging"
+	amqp "github.com/rabbitmq/amqp091-go"
+)
+
+var _ messaging.Publisher = (*publisher)(nil)
+
+type publisher struct {
+	connection *amqp.Connection
+	channel    *amqp.Channel
```
dborovcanin: Since this is rather a simple struct, consider using `conn` and `ch` instead.

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,183 @@
...
+	subject := fmt.Sprintf("%s.%s.%s", exchange, chansPrefix, topic)
+
+	if err := ps.channel.ExchangeDeclare(subject, exchangeKind, true, false, false, false, nil); err != nil {
+		return err
+	}
+
+	if err := ps.channel.QueueBind(ps.queue.Name, routingKey, subject, false, nil); err != nil {
+		return err
+	}
+	msgs, err := ps.channel.Consume(ps.queue.Name, "", true, false, false, false, nil)
```
dborovcanin: Is it OK to have an empty string for a consumer here?

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	go ps.handle(msgs, handler)
+	ps.subscriptions[topic] = true
+
+	return nil
+}
+
+func (ps *pubsub) Unsubscribe(topic string) error {
+	defer ps.channel.Cancel("", false)
```
dborovcanin: How does canceling for an empty consumer work? Did you test subscribing with multiple consumers and canceling one of them (add that in test cases, as well)?

pkg/messaging/rabbitmq/pubsub_test.go
```diff
@@ -0,0 +1,243 @@
...
+	channel     = "9b7b1b3f-b1b0-46a8-a717-b8213f9eda3b"
+	subtopic    = "engine"
+)
+
+var (
+	msgChan = make(chan messaging.Message)
+	data    = []byte("payload")
+)
+
+func TestPubsub(t *testing.T) {
```
dborovcanin: If you take publisher implementation from `publisher.go` that's already tested in `publisher_test.go`, you can have only `Subscriber` (with `Subscribe` and `Unsubscribe`) methods tested here.

pkg/messaging/rabbitmq/pubsub_test.go
```diff
@@ -0,0 +1,243 @@
...
+	pubsubcases := []struct {
```
dborovcanin: Fix typos: `sub topic` -> `subtopic` and `Unsusbcribe` -> `Unsubscribe`.

dborovcanin (changes requested):

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: When we finish processing messages from the delivery channel we send a nil value

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: Good idea

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: An empty string will cause the library to generate a unique identity for each consumer. Since we have no way to id the subscriber, I think that is the best fit for now.

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: I am getting this error `panic: runtime error: invalid memory address or nil pointer dereferenc`. I have resorted to another approach

pkg/messaging/rabbitmq/publisher.go
rodneyosodo: Done

rodneyosodo: Done

pkg/messaging/rabbitmq/pubsub_test.go
rodneyosodo: Done

pkg/messaging/rabbitmq/pubsub_test.go
rodneyosodo: Implemented

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: Initially, I thought it would generate the id for the exact consumer but after going through the code it doesn't as referenced [here](https://github.com/rabbitmq/amqp091-go/blob/84fae140a2b7e958129da81bd7143bd93e289520/consumers.go#L100)

rodneyosodo: If we do a `ps.ch.close()` we close the whole channel so we can't subscribe to it. The best solution to this is if we are able to generate UUID so as to id the consumers so we can cancel the respective consumers. The current implementation works fine.

pkg/messaging/nats/pubsub.go
```diff
@@ -36,7 +36,7 @@ type PubSub interface {
 }
 
 type pubsub struct {
-	conn          *broker.Conn
+	publisher
```
drasko, line 45: Needs variable name

drasko (changes requested):

pkg/messaging/nats/pubsub.go
```diff
@@ -42,7 +42,7 @@ type subscription struct {
 }
 
 type pubsub struct {
-	conn          *broker.Conn
+	publisher     publisher
```
dborovcanin: Instead of having a `publisher` field, consider using a composition of `publisher` like:
```go
type pubsub struct {
	publisher
	logger        log.Logger
	mu            sync.Mutex
	queue         string
	subscriptions map[string]map[string]subscription
}
```
This will enable using `ps.conn` rather than `ps.publisher.conn` later on.

dborovcanin (changes requested):

pkg/messaging/rabbitmq/publisher.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	err = pub.ch.Publish(
+		subject,
+		routingKey,
+		mandatory,
+		immediate,
+		amqp.Publishing{
+			Headers:     amqp.Table{},
+			ContentType: "text/plain",
```
dborovcanin: Is this type correct? Since our message is binary, it looks like `application/octet-stream` is more suitable for us.

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,192 @@
...
+	amqp "github.com/rabbitmq/amqp091-go"
+)
+
+const (
+	chansPrefix = "channels"
+	// SubjectAllChannels represents subject to subscribe for all the channels.
+	SubjectAllChannels = "channels.>"
+	routingKey         = "mainfluxkey"
+	exchange           = "mainflux"
+	exchangeKind       = "fanout"
```
dborovcanin: Did you consider using `topic` kind? In that case, we have a queue per channel and the subtopic could be the routing key. This will have an impact on both publisher and subscriber implementation.

dborovcanin (changes requested):

pkg/messaging/nats/pubsub.go
rodneyosodo: That is what we had initially but @drasko had suggested with name it.

I will stick with the earlier changes we had.

pkg/messaging/rabbitmq/publisher.go
rodneyosodo: Yes `application/octet-stream` is more suitable

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: Yes. That will be a good approach but it poses come challenges because when we do a subscribe the topic preprocessing is done by the client compared to how it is implemented on the publish method.

pkg/messaging/nats/pubsub_test.go
```diff
@@ -189,7 +189,7 @@ func TestPubsub(t *testing.T) {
 			desc:         "Susbcribe to a topic with empty id",
 			topic:        fmt.Sprintf("%s.%s", chansPrefix, topic),
 			topicID:      "",
-			errorMessage: errors.New("empty ID"),
+			errorMessage: errors.New("empty id"),
```
manuio: I'd create errors in  a package for nats and rabbitmq and use them in the code and tests

manuio (changes requested):

pkg/messaging/nats/pubsub.go
dborovcanin: @drasko I suggested using the composition for reasons explained [here](https://github.com/mainflux/mainflux/pull/1562#discussion_r868436254).

pkg/messaging/rabbitmq/pubsub.go
dborovcanin: I understand that we'll have to disassemble the topic we created before calling Publish/Subscribe methods, but it's worth trying it because topic manipulation is rather simple. Of course, queue manipulation is also something to take care of (creating a queue if it doesn't exist and removing it if there are no messages and subscribers left). I'm only afraid this topic processing is going to be too cryptic - using the first part of the topic as a queue (name), and the rest as a subtopic (routingKey).

rodneyosodo: From [this](https://www.cloudamqp.com/blog/part4-rabbitmq-13-common-errors.html) reference we need to have multiple queues. The current implementation I have pushed is that we have direct exchange rather than topic exchange as direct is faster than topic exchange. We will have a queue per subscription.
Another way to implement it is to have queues based on services such as a coap queue and an MQTT queue. But this makes us not have many queues.

rodneyosodo: ## Overview
RabbitMQ is somehow different to Nats and Kafka in that when a publisher is publishing a message on a topic they will publish the message to an exchange. The message will be published with a routing key. This routing key will direct the message to the different queues onto which the subscriber will consume the message. This is how the message flows
```
publisher -> exchange -> routingkey -> queue -> consumer -> subscriber


The publisher publishes the message to the broker.
The exchange receives the message being published and uses the routing key to route the message to a specific queue
The routing key is used to route the message to a specific queue
The queue temporarily stores messages for the consumer before consuming them
```

image


## Exchange
When the exchange is declared we declare it with a specific kind. This means an exchange can either route messages based on topic, directly to queues or using message headers among other methods. We will be interested in the three routing techniques. The number of bindings increases the performance of topic routing. Fanout exchanges are very fast because there is no routing to process, but this changes if they are bound to a large number of queues. Direct exchanges are faster than topic exchanges if you don't need the wild card. We picked direct method as it strikes a balance between performance and reliability.

## Possible implementations

### One exchange one queue
- If you only have one queue, you'll end up repeating the logic of deciding which code should process the message, when RabbitMQ can do it for you with routing.
- RabbitMQ prefers many queues over a single large one
- This option is ruled out

### One exchange multiple queues
- This is the most viable solution currently
- This is because we will have single consumers on one queue
- The advantage of a single topic exchange is that you get the benefits of both a Direct Exchange and a Fanout Exchange depending on the binding key used.
- Further, configuration changes are easy to achieve and can often be done without disrupting any system processing at all

### Multiple exchanges one queue
- This option is ruled out based on having a single queue

### Multiple exchanges multiple queues
- This is the most preferred way of doing it as the system is highly decoupled. 
- If we cross the boundary and have many exchanges and many queues we are at the risk of the processes taking up more memory which may be significant but this really depends
- If the `msg.channel` is not provided the message won't be consumed as we can subscribe to an empty topic

## Wayforward
- Currently, we have implemented Direct as the exchange type. So if a consumer needs to subscribe to multiple topics we would change this implementation to the exchange having topic as the exchange type
- We have not explored exchange-to-exchange architectures

For more reference check out [this article](https://spring.io/blog/2011/04/01/routing-topologies-for-performance-and-scalability-with-rabbitmq/)

pkg/messaging/rabbitmq/publisher.go
```diff
@@ -0,0 +1,84 @@
...
+	err = pub.ch.Publish(
+		exchangeName,
+		subject,
+		false,
+		false,
+		amqp.Publishing{
+			Headers:     amqp.Table{},
+			ContentType: "application/octet-stream",
+			AppId:       "mainflux-publisher",
+			Body:        []byte(data),
```
dborovcanin: No need to cast, data is already `[]byte`.

pkg/messaging/rabbitmq/pubsub_test.go
```diff
@@ -0,0 +1,179 @@
...
+		},
+		{
+			desc:         "Unsubscribe to an already unsubscribed topic with an ID",
+			topic:        fmt.Sprintf("%s.%s", chansPrefix, topic),
+			topicID:      "topicid1",
+			errorMessage: rabbitmq.ErrNotSubscribed,
+			pubsub:       false,
+		},
+		{
+			desc:         "Susbcribe to a topic with a subtopic with an ID",
```
dborovcanin: Fix typo `Susbcribe`. Check for other typos as well.

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,167 @@
...
+	messaging.PubSub
+	Close()
+}
+
+type subscription struct {
+	cancel func() error
+}
+type pubsub struct {
+	publisher
+	queue         string
```
dborovcanin: What's the purpose of this and is it even used?

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,167 @@
...
+	default:
+		s = make(map[string]subscription)
+		ps.subscriptions[topic] = s
+	}
+
+	queue, err := ps.ch.QueueDeclare(topic, true, true, true, false, nil)
+	if err != nil {
+		return err
+	}
+	if err := ps.ch.QueueBind(queue.Name, queue.Name, exchangeName, false, nil); err != nil {
```
dborovcanin: Is it safer to use topic instead of `queue.Name`? Is it possible that the broker will change the name passed in QueueDeclare?

dborovcanin (changes requested):

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: It is not being used

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: The broker doesn't change the passed name

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/mainflux/mainflux/pkg/messaging"
+	amqp "github.com/rabbitmq/amqp091-go"
+)
+
+const (
+	chansPrefix = "channels"
+	// SubjectAllChannels represents subject to subscribe for all the channels.
+	SubjectAllChannels = "channels.>"
+	exchangeName       = "mainflux-exchange"
+	exchangeKind       = "direct"
```
dborovcanin: Is this used?

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,166 @@
...
+	go ps.handle(msgs, handler)
+	s[id] = subscription{
+		cancel: handler.Cancel,
+	}
+
+	return nil
+}
+
+func (ps *pubsub) Unsubscribe(id, topic string) error {
+	defer ps.ch.Cancel(id, false)
```
dborovcanin: Should we really cancel unconditionally?

dborovcanin (changes requested):

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: exchangeKind no. The others yes.

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: From the docs, this is what it says about the Cancel method

> Only use this method to cleanly stop receiving deliveries from the server and cleanly shut down the consumer chan identified by this tag.  Using this method and waiting for the remaining messages to flush from the consumer chan will ensure all messages received on the network will be delivered to the receiver of your consumer chan.

I believe this is what we want to achieve

dborovcanin: We are, but we should first handle the internal logic, then try to cancel and handle the potential errors. What happens if we have the same consumer subscribed to multiple topics? Does it cancel them all (because it's the same consumer ID)?
For example, a nice solution would be to put this in the `cancel` method of `subscriber` when you subscribe, so that you don't even call it explicitly when unsubscribing. Something like:
```go
	s[id] = subscription{
		cancel: func() error {
			if err := ps.ch.Cancel(id, false); err != nil {
				return err
			}
			return handler.Cancel()
		},
	}
```

rodneyosodo:
RabbitMQ expects that all client IDs are unique.
```
The consumer is identified by a unique string and scoped for all consumers on this channel.
```
So when we cancel the consumer we have a unique id that will only cancel one consumer. For our case, since we are passing in the unique ID, we need to generate a unique ID for every consumer or to create multiple channels per topic. Creating numerous channels is discouraged as by doing so it consumes system resources and makes it more difficult to configure firewalls

dborovcanin:
> RabbitMQ expects that all client IDs are unique.
> 
> ```
> The consumer is identified by a unique string and scoped for all consumers on this channel.
> ```
> 
> So when we cancel the consumer we have a unique id that will only cancel one consumer. For our case, since we are passing in the unique ID, we need to generate a unique ID for every consumer or to create multiple channels per topic. Creating numerous channels is discouraged as by doing so it consumes system resources and makes it more difficult to configure firewalls

OK. I let you choose between generating an ID and using a combination of client ID and topic.

rodneyosodo:
> > RabbitMQ expects that all client IDs are unique.
> > ```
> > The consumer is identified by a unique string and scoped for all consumers on this channel.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So when we cancel the consumer we have a unique id that will only cancel one consumer. For our case, since we are passing in the unique ID, we need to generate a unique ID for every consumer or to create multiple channels per topic. Creating numerous channels is discouraged as by doing so it consumes system resources and makes it more difficult to configure firewalls
> 
> OK. I let you choose between generating an ID and using a combination of client ID and topic.

Which one do you suggest we use? I had started with using a combination of client ID and topic

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -101,14 +101,16 @@ func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler)
...
+	clientID := fmt.Sprintf("%s-%s", topic, id)
+	msgs, err := ps.ch.Consume(topic, clientID, true, false, false, false, nil)
 	if err != nil {
 		return err
 	}
 	go ps.handle(msgs, handler)
 	s[id] = subscription{
 		cancel: func() error {
-			if err := ps.ch.Cancel(id, false); err != nil {
+			clientID := fmt.Sprintf("%s-%s", topic, id)
```
dborovcanin: Reuse `clientID` from L104.

pkg/messaging/nats/pubsub_test.go
```diff
@@ -27,9 +27,9 @@ var (
 )
 
 func TestPubsub(t *testing.T) {
-	err := pubsub.Subscribe(topicID, fmt.Sprintf("%s.%s", chansPrefix, topic), handler{})
+	err := pubsub.Subscribe(clientID, fmt.Sprintf("%s.%s", chansPrefix, topic), handler{})
```
dborovcanin, line 30: You can still keep `TestPublish` and `TestSubscribe` with separate cases (but in this single file, that's OK).

pkg/messaging/nats/pubsub_test.go
```diff
@@ -83,134 +83,134 @@ func TestPubsub(t *testing.T) {
...
+			desc:         "Subscribe to an already subscribed topic with an ID",
 			topic:        fmt.Sprintf("%s.%s", chansPrefix, topic),
-			topicID:      "topicid1",
-			errorMessage: errors.New("already subscribed to topic"),
+			clientID:     "clientid1",
+			errorMessage: nats.ErrAlreadySubscribed,
 			pubsub:       true,
 		},
 		{
 			desc:         "Unsubscribe to a topic with an ID",
```
dborovcanin: typo `Unsubscribe **from**...`. Same for all unsubscribe cases.

dborovcanin (changes requested):

pkg/messaging/nats/pubsub_test.go
rodneyosodo: Here we are subscribing so that we can receive the published messages

dborovcanin (approved):

drasko (approved):
LGTM

PR #1563, opened 2022-02-09
NOISSUE - Separate Keto hosts for read and write

blokovi:
### What does this do?
Separate Keto hosts for read and write

Related to https://github.com/mainflux/devops/issues/102
Since installing keto on Kubernetes with official helm charts creates two separate services for read and write, Auth service must connect to both of them with different hosts.

Updated README


cmd/auth/main.go
```diff
@@ -192,16 +196,16 @@ func initJaeger(svcName, url string, logger logger.Logger) (opentracing.Tracer,
 	return tracer, closer
 }
 
-func initKeto(hostAddress, readPort, writePort string, logger logger.Logger) (readerConnection, writerConnection *grpc.ClientConn) {
-	checkConn, err := grpc.Dial(fmt.Sprintf("%s:%s", hostAddress, readPort), grpc.WithInsecure())
+func initKeto(hostReadAddress, readPort, hostWriteAddress, writePort string, logger logger.Logger) (readerConnection, writerConnection *grpc.ClientConn) {
+	checkConn, err := grpc.Dial(fmt.Sprintf("%s:%s", hostReadAddress, readPort), grpc.WithInsecure())
```
dborovcanin: Consider renaming to `readConn`.

blokovi: @dusanb94 renamed, thanks

dborovcanin (approved):

auth/README.md
```diff
@@ -74,6 +74,10 @@ default values.
 | MF_AUTH_SECRET                | String used for signing tokens                                           | auth           |
 | MF_AUTH_LOGIN_TOKEN_DURATION  | The login token expiration period                                        | 10h            |
 | MF_JAEGER_URL                 | Jaeger server URL                                                        | localhost:6831 |
+| MF_KETO_READ_REMOTE_HOST      | Keto Read Host                                                           | mainflux-keto  |
```
drasko, line 77: Is it OK that they have identical host/container names? These are two different containers if I understand well.

drasko (changes requested):

auth/README.md
blokovi: Same container.
In docker deployment keto expose read and write on the same host with different ports.

blokovi: This PR doesn't change anything in keto deployment. It just enables Auth service to have separate URLs for read and write connection to Keto. That wasn't  possible because Auth used the same environment variable for both read and write connection

drasko: Can it be 2 different hosts? If not, then use one host with 2 ports, i.e. add just one variable: `MF_KETO_REMOTE_HOST` and two env vars for ports: `MF_KETO_READ_REMOTE_PORT` and `MF_KETO_WRITE_REMOTE_PORT` (additionally - do we need word "REMOTE" in these and do we use it in other var names?).

drasko (changes requested):

auth/README.md
blokovi: @drasko  What you suggest is that nothing to change, that's how it is now.

But we need Auth service to be capable to connect to two different keto hosts, not just the same host and different ports.
That is because in k8s setup keto read and write endpoints are on different hosts.
So, we can have like in this PR four envars (for read host, read port, write host, write port) or maybe two envars (each in `host:port` format). In any way we need possibility to configure differently both host and port part of connection urls

drasko: OK, thanks

drasko (approved):
LGTM

PR #1564, opened 2022-02-12
NOISSUE - Update dependencies

dborovcanin:

### What does this do?
This pull request updates dependencies. It also fixes the warning in policies tests.

There is no such issue.

There are no such changes.

N/A

N/A

### Notes
N/A


auth/api/http/policies/transport.go
```diff
@@ -96,10 +96,11 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 	default:
 		w.WriteHeader(http.StatusInternalServerError)
 	}
-	errorVal, ok := err.(errors.Error)
-	if ok {
-		if err := json.NewEncoder(w).Encode(errorRes{Err: errorVal.Msg()}); err != nil {
-			w.Header().Set("Content-Type", contentType)
+
+	errVal, ok := err.(errors.Error)
+	if ok && errVal.Msg() != "" {
```
drasko: Maybe inline this initialization, if not used otherwise

auth/api/http/policies/responses.go
```diff
@@ -39,7 +39,7 @@ func (res deletePoliciesRes) Headers() map[string]string {
 }
 
 func (res deletePoliciesRes) Empty() bool {
-	return false
+	return true
```
drasko, line 42: Why is this changed?

drasko (changes requested):

auth/api/http/policies/responses.go
dborovcanin: Since `deleteResponse` has no body, `Empty` should return `false`.

drasko: You introduced the change where it returns `true`. I am confused now - what should the return value be - `false` as it is today, or `true` as you propose.

dborovcanin: My bad, it is empty, so it should return true.

drasko: OK, good - resolving this one.

drasko (approved):
LGTM

PR #1570, opened 2022-03-03
NOISSUE - RabbitMQ build and deployment

### LINKED ISSUE #1610: Use go build tags for different brokers
**FEATURE REQUEST**

To prevent adding potentially huge dependencies in vendors, consider using different Go build tags for different supported internal message brokers.

rodneyosodo:
### What does this do?
It aggregates the different message broker deployments starting with rabbitmq

Resolves #1610

We are modularising the backend broker from nats to either nats or rabbitmq or Kafka

No

Yes

### Notes
N/A


docker/docker-compose.yml
```diff
@@ -305,11 +321,12 @@ services:
     depends_on:
       - things
       - nats
+      - rabbitmq
```
dborovcanin: Adding both Nats and RabbitMQ as dependencies is suboptimal. The one should choose between them, rather than always having them running together. Consider using multiple compose files (one for broker and the other for the Mainflux services), updating Makefile, or checking if there are options to do it with a single compose file.

dborovcanin (changes requested):
New `messaging/broker` and another wrapper look unnecessary.

rodneyosodo:
> New `messaging/broker` and another wrapper look unnecessary.

It aggregates the currently two brokers nats and rabbitmq based on the broker url provided

rodneyosodo:
To be continued after fixing message broker wrapper

cmd/mqtt/main.go
```diff
@@ -62,9 +62,9 @@ const (
 	defThingsAuthTimeout = "1s"
 	envThingsAuthURL     = "MF_THINGS_AUTH_GRPC_URL"
 	envThingsAuthTimeout = "MF_THINGS_AUTH_GRPC_TIMEOUT"
-	// Nats
-	defNatsURL = "nats://localhost:4222"
-	envNatsURL = "MF_NATS_URL"
+	// Message broker
```
manuio: Do we need this and other comments? Looks like it's the only service where it's done.

Makefile
```diff
@@ -111,4 +111,11 @@ rundev:
...
 run:
-	docker-compose -f docker/docker-compose.yml up
+ifeq ("$(MF_BROKER_TYPE)", "rabbitmq")
+	sed -i "s,file: broker/.*.yml,file: broker/rabbitmq.yml," docker/docker-compose.yml
+else ifeq ("$(MF_BROKER_TYPE)", "nats")
+	sed -i "s,file: broker/.*.yml,file: broker/nats.yml," docker/docker-compose.yml
+else
+	echo "Invalid broker type"; exit 1
+endif
+	docker-compose -f docker/docker-compose.yml up -d
```
drasko: Remove `-d`

drasko (changes requested):

Makefile
drasko: @0x6f736f646f please take care of these remarks

pkg/messaging/broker/pubsub.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"github.com/mainflux/mainflux"
+	log "github.com/mainflux/mainflux/logger"
+
+	"github.com/mainflux/mainflux/pkg/messaging/nats"
+	"github.com/mainflux/mainflux/pkg/messaging/rabbitmq"
+)
+
+const (
+	// chansPrefix = "channels"
```
drasko: Dead code?

pkg/messaging/broker/pubsub.go
```diff
@@ -0,0 +1,42 @@
...
+	// SubjectAllChannels represents subject to subscribe for all the channels.
+	SubjectAllChannels = "channels.>"
+)
+
+// PubSub type
+type PubSub nats.PubSub
+
+// NewPubSub This aggregates the NewPubSub function for all message brokers
+func NewPubSub(url, queue string, logger log.Logger) (nats.PubSub, error) {
+	brokerSelection := mainflux.Env(envBrokerType, defBrokerType)
```
drasko: I would prefer calling this one `brokerType` and not `brokerSelection`

pkg/messaging/rabbitmq/publisher.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+	subject := fmt.Sprintf("%s.%s", chansPrefix, topic)
+	if msg.Subtopic != "" {
+		subject = fmt.Sprintf("%s.%s", subject, msg.Subtopic)
+	}
+	err = pub.ch.Publish(
+		exchangeName,
+		subject,
+		false,
```
drasko: @dusanb94 maybe fields specific to different brokers should be grouped in `Options` substructure?

drasko (changes requested):

pkg/messaging/broker/pubsub.go
```diff
@@ -0,0 +1,42 @@
...
+
+	// SubjectAllChannels represents subject to subscribe for all the channels.
+	SubjectAllChannels = "channels.>"
+)
+
+// PubSub type
+type PubSub nats.PubSub
+
+// NewPubSub This aggregates the NewPubSub function for all message brokers
+func NewPubSub(url, queue string, logger log.Logger) (nats.PubSub, error) {
```
dborovcanin: You can join these files (`publisher.go` and `pubsub.go`) to a single `pkg/messaging/brokers/brokers.go`.

pkg/messaging/broker/publisher.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+var (
+	errEmptyBrokerType = errors.New("empty broker type")
+	defBrokerType      = "nats"
+)
+
+// NewPublisher This aggregates the NewPublisher function for all message brokers
+func NewPublisher(url string) (nats.Publisher, error) {
+	brokerSelection := mainflux.Env(envBrokerType, defBrokerType)
```
dborovcanin: It's probably better to receive broker type as a parameter and use `mainflux.Env` only in `main.go`. The reason is sneaky configuration bugs that can arise when loading config from env this way (encapsulated in the method itself).

pkg/messaging/broker/pubsub.go
```diff
@@ -0,0 +1,42 @@
...
+	brokerSelection := mainflux.Env(envBrokerType, defBrokerType)
+	if brokerSelection == "nats" {
```
dborovcanin: Use `switch-case` instead `if-else`.

pkg/messaging/broker/publisher.go
```diff
@@ -0,0 +1,41 @@
...
+	if brokerSelection == "nats" {
```
dborovcanin: Use `switch-case` instead of `if-then-else`.

dborovcanin (changes requested):

pkg/messaging/rabbitmq/publisher.go
dborovcanin: What exactly would be this structure? A message field? Or like the struct you pass when creating a new Pub(Sub) instance, like configuration?

dborovcanin (commented):
Once #1562 is merged, this PR will need to rebase. Also, docs and deployment need to be updated: https://github.com/mainflux/docs and https://github.com/mainflux/devops.

pkg/messaging/broker/brokers.go
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package broker
```
dborovcanin: Use plural - `brokers`.

pkg/messaging/broker/brokers.go
```diff
@@ -0,0 +1,66 @@
...
+
+	log "github.com/mainflux/mainflux/logger"
+	"github.com/mainflux/mainflux/pkg/messaging/nats"
+	"github.com/mainflux/mainflux/pkg/messaging/rabbitmq"
+)
+
+const (
+	// SubjectAllChannels represents subject to subscribe for all the channels.
+	SubjectAllChannels = "channels.>"
+)
```
dborovcanin:
```go
// SubjectAllChannels represents subject to subscribe for all the channels.
const SubjectAllChannels = "channels.>"
```

pkg/messaging/broker/brokers.go
```diff
@@ -0,0 +1,66 @@
...
+
+var (
+	errEmptyBrokerType = errors.New("empty broker type")
+)
```
dborovcanin:
```go
var errEmptyBrokerType = errors.New("empty broker type")
```

pkg/messaging/broker/brokers.go
```diff
@@ -0,0 +1,66 @@
...
+
+// PubSub type
```
dborovcanin: Create wrapper interfaces here like:
```go

// Publisher interface enriched with connection closing.
type Publisher interface {
	messaging.Publisher
	Close() error
}

// PubSub interface enriched with connection closing.
type PubSub interface {
	messaging.PubSub
	Close() error
}

// NewPublisher This aggregates the NewPublisher function for all message brokers
func NewPublisher(brokerType, url string) (Publisher, error) {
	// brokerType := mainflux.Env(envBrokerType, defBrokerType)
	switch brokerType {
	case "nats":
		pb, err := nats.NewPublisher(url)
		if err != nil {
			return nil, err
		}
		return pb, nil
	case "rabbitmq":
		pb, err := rabbitmq.NewPublisher(url)
		if err != nil {
			return nil, err
		}
		return pb, nil
	default:
		return nil, errEmptyBrokerType
	}
}

// NewPubSub This aggregates the NewPubSub function for all message brokers
func NewPubSub(brokerType, url, queue string, logger log.Logger) (PubSub, error) {
	// brokerType := mainflux.Env(envBrokerType, defBrokerType)
	switch brokerType {
	case "nats":
		pb, err := nats.NewPubSub(url, queue, logger)
		if err != nil {
			return nil, err
		}
		return pb, nil
	case "rabbitmq":
		pb, err := rabbitmq.NewPubSub(url, queue, logger)
		if err != nil {
			return nil, err
		}
		return pb, nil
	default:
		return nil, errEmptyBrokerType
	}
}

```
A small modification on broker implementations will be needed. Also, case `MQTT` is missing in this switch.

dborovcanin (changes requested):

pkg/messaging/broker/pubsub.go
rodneyosodo: So with build tags, we need actually different files for Kafka, nats and rabbitmq

rodneyosodo: Sorry this is in the Kafka branch

pkg/messaging/broker/brokers.go
rodneyosodo: is MQTT used as an internal message broker or an adapter to provide an MQTT API for sending messages through the platform?

dborovcanin: It is primarily used for forwarding MQTT messages, but It can be used for both.

pkg/messaging/broker/brokers.go
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package brokers
```
dborovcanin: /pkg/messaging/**brokers**/brokers.go

dborovcanin (changes requested):
@0x6f736f646f Please be careful and at least do basic compile tests. This code won't compile due to a missing import and an error in package naming. When you fix that, you'll need to fix other broker implementations to match the `Close` method signature.

pkg/messaging/broker/brokers.go
rodneyosodo: Should we modify the current MQTT adapter to allow for it to be used as an option in the internal broker?

dborovcanin: No need to modify adapter, only the messages Pub/Sub.

Makefile
```diff
@@ -111,4 +117,11 @@ rundev:
...
 run:
-	docker-compose -f docker/docker-compose.yml up
+ifeq ("$(MF_BROKER_TYPE)", "rabbitmq")
+	sed -i "s,file: brokers/.*.yml,file: brokers/rabbitmq.yml," docker/docker-compose.yml
+else ifeq ("$(MF_BROKER_TYPE)", "nats")
+	sed -i "s,file: brokers/.*.yml,file: brokers/nats.yml," docker/docker-compose.yml
+else
+	echo "Invalid broker type"; exit 1
+endif
+	docker-compose -f docker/docker-compose.yml up
```
dborovcanin: Add a new line at the end of the Makefile.

pkg/messaging/brokers/brokers_nats.go
```diff
@@ -0,0 +1,37 @@
+//go:build nats
+// +build nats
```
dborovcanin: Since Nats is the default, consider changing logic to:
```go
//go:build !rabbitmq
// +build !rabbitmq
```
This way we do not have to provide a `nats` tag to build with nats - it will work without tags and we have a more straightforward default build.

docker/.env
```diff
@@ -6,8 +6,20 @@ MF_NGINX_SSL_PORT=443
 MF_NGINX_MQTT_PORT=1883
 MF_NGINX_MQTTS_PORT=8883
 
-## NATS
-MF_NATS_URL=nats://nats:4222
+# Message Broker
+MF_BROKER_TYPE=nats
+MF_BROKER_URL=nats://broker:4222
```
dborovcanin: The broker URL needs to change depending on the broker type - here, it's hardcoded to use NATS.

dborovcanin (changes requested):

pkg/messaging/brokers/brokers_nats.go
```diff
@@ -0,0 +1,37 @@
...
+
```
dborovcanin, line 3: Consider adding `init` function to log for which broker the binary is built (https://go.dev/doc/effective_go#init).

pkg/messaging/brokers/brokers_rabbitmq.go
```diff
@@ -0,0 +1,36 @@
+//go:build rabbitmq
+// +build rabbitmq
+
+// Copyright (c) Mainflux
```
dborovcanin, line 4: The same comment for `init` function.

dborovcanin (changes requested):

pkg/messaging/brokers/brokers_nats.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/mainflux/mainflux/pkg/messaging/nats"
+)
+
+// SubjectAllChannels represents subject to subscribe for all the channels.
+const SubjectAllChannels = "channels.>"
+
+type (
+	Publisher nats.Publisher
+	PubSub    nats.PubSub
+)
```
dborovcanin: Let's simplify remove these wrappers and update the base interface with `Close`:
```go
// Publisher specifies message publishing API.
type Publisher interface {
	// Publishes message to the stream.
	Publish(topic string, msg Message) error

	// Close gracefully closes message publisher's connection.
	Close() error
}

// MessageHandler represents Message handler for Subscriber.
type MessageHandler interface {
	// Handle handles messages passed by underlying implementation.
	Handle(msg Message) error

	// Cancel is used for cleanup during unsubscribing and it's optional.
	Cancel() error
}

// Subscriber specifies message subscription API.
type Subscriber interface {
	// Subscribe subscribes to the message stream and consumes messages.
	Subscribe(id, topic string, handler MessageHandler) error

	// Unsubscribe unsubscribes from the message stream and
	// stops consuming messages.
	Unsubscribe(id, topic string) error

	// Close gracefully closes message subscriber's connection.
	Close() error
}
```
and use `messaging.{Publisher,Subscriber,PubSub}` everwhere instead of wrapped versions (`brokers.PubSub`, `nats.PubSub`, `mqtt.PubSub`...).

dborovcanin (changes requested):

docker/brokers/nats.yml
```diff
@@ -0,0 +1,8 @@
+services:
+  broker:
+    image: nats:2.2.4-alpine
+    command: "-c /etc/nats/nats.conf"
+    volumes:
+      - ./../nats/:/etc/nats
+    ports:
+      - ${MF_NATS_PORT}:${MF_NATS_PORT}
```
dborovcanin: Please add an empty line.

docker/brokers/rabbitmq.yml
```diff
@@ -0,0 +1,11 @@
...
+  broker:
+    image: rabbitmq:3.7-management
+    environment:
+      RABBITMQ_ERLANG_COOKIE: ${MF_RABBITMQ_COOKIE}
+      RABBITMQ_DEFAULT_USER: ${MF_RABBITMQ_USER}
+      RABBITMQ_DEFAULT_PASS: ${MF_RABBITMQ_PASS}
+      RABBITMQ_DEFAULT_VHOST: ${MF_RABBITMQ_VHOST}
+    ports:
+      - ${MF_RABBITMQ_PORT}:${MF_RABBITMQ_PORT}
+      - ${MF_RABBITMQ_HTTP_PORT}:${MF_RABBITMQ_HTTP_PORT}
```
dborovcanin: Please add an empty line.

pkg/messaging/pubsub.go
```diff
@@ -26,6 +29,9 @@ type Subscriber interface {
 	// Unsubscribe unsubscribes from the message stream and
 	// stops consuming messages.
 	Unsubscribe(id, topic string) error
+
+	// Close gracefully closes message subscriber's connection.
+	Close()
```
dborovcanin: Close should return an error. In case the underlying implementation is not returning it, return nil.

dborovcanin (changes requested):

pkg/messaging/brokers/brokers_rabbitmq.go
```diff
@@ -0,0 +1,36 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package brokers
+
+import (
```
dborovcanin, line 9: This code won't compile due to missing import.

dborovcanin: Consider adding this line `MF_BROKER_TYPE=rabbitmq make http coap mqtt opcua lora` to `/scripts/ci.sh` to add compile check to CI.

docker/brokers/rabbitmq.yml
```diff
@@ -0,0 +1,11 @@
+services:
+  broker:
+    image: rabbitmq:3.7-management
```
dborovcanin: Use more recent image such as `rabbitmq:3.9.20-management-alpine`.

pkg/messaging/rabbitmq/publisher.go
```diff
@@ -18,15 +18,8 @@ type publisher struct {
...
-// Publisher wraps messaging Publisher exposing
-// Close() method for RabbitMQ connection.
-type Publisher interface {
-	messaging.Publisher
-	Close()
-}
-
 // NewPublisher returns RabbitMQ message Publisher.
-func NewPublisher(url string) (Publisher, error) {
+func NewPublisher(url string) (messaging.Publisher, error) {
```
dborovcanin, line 22: Remove `"amqp://%s"` and let the user completely define the URL, just like we do with NATS.

docker/.env
dborovcanin: This is still not addressed.

dborovcanin (changes requested):

scripts/ci.sh
```diff
@@ -65,6 +65,8 @@ setup_mf() {
         fi
     done
     make -j$NPROC
+    echo "Compile check for rabbitmq..."
+    MF_BROKER_TYPE=rabbitmq make http
```
dborovcanin, line 68: Put this before ` make -j $NPROC`.

scripts/ci.sh
```diff
@@ -81,6 +83,7 @@ setup() {
     setup_lint
 }
 
+
```
dborovcanin: Remove added empty line.

Makefile
```diff
@@ -111,4 +117,12 @@ rundev:
 	cd scripts && ./run.sh
 
 run:
-	docker-compose -f docker/docker-compose.yml up
+ifeq ("$(MF_BROKER_TYPE)", "rabbitmq")
+	sed -i "s,file: brokers/.*.yml,file: brokers/rabbitmq.yml," docker/docker-compose.yml
+	export MF_BROKER_URL=amqp://$(MF_RABBITMQ_USER):$(MF_RABBITMQ_PASS)@broker:$(MF_RABBITMQ_PORT)$(MF_RABBITMQ_VHOST) && docker-compose -f docker/docker-compose.yml up
```
dborovcanin: This will not work unless the user has locally set `MF_RABBITMQ_*` env vars set because it will be replaced with empty values. I would suggest doing something like this instead:
```make
	sed -i "s,file: brokers/.*.yml,file: brokers/rabbitmq.yml," docker/docker-compose.yml
	sed -i "s,MF_BROKER_URL: .*,MF_BROKER_URL: $$\{MF_RABBITMQ_URL\}," docker/docker-compose.yml
```

dborovcanin (changes requested):

Makefile
rodneyosodo: I thought we needed to build the broker URL for instance when using rabbitmq

Makefile
dborovcanin: This one is still not resolved.

Makefile
```diff
@@ -111,4 +117,13 @@ rundev:
 	cd scripts && ./run.sh
 
 run:
-	docker-compose -f docker/docker-compose.yml up
+ifeq ("$(MF_BROKER_TYPE)", "rabbitmq")
+	sed -i "s,file: brokers/.*.yml,file: brokers/rabbitmq.yml," docker/docker-compose.yml
+	sed -i 's#{MF_BROKER_URL}#MF_RABBITMQ_URL#' docker/docker-compose.yml
```
dborovcanin: This approach is not flexible enough and will work only when you first replace `${MF_BROKER_URL}` with `${MF_RABBITMQ_URL}`. If you want to run it with any other broker later, it won't work. Since you [changed](https://github.com/mainflux/mainflux/pull/1570/files#diff-423deb13b7c401b1a7f41ee91c77f722e11d2f317d6a66b546524e8a04cc8b03R289) `docker-compose.yml` to use `${MF_NATS_URL}`, it will not work at all. I prefer using `MF_BROKER_URL: ${MF_NATS_URL}` over `MF_BROKER_URL: ${MF_BROKER_URL}` in `docker-compose.yml`, but Makefile `sed` command needs an update.

dborovcanin (changes requested):

docker/.env
```diff
@@ -6,8 +6,21 @@ MF_NGINX_SSL_PORT=443
...
+MF_NATS_URL=nats://broker:4222
+
+## RabbitMQ
+MF_RABBITMQ_PORT=5672
+MF_RABBITMQ_HTTP_PORT=15672
+MF_RABBITMQ_USER=mainflux
+MF_RABBITMQ_PASS=mainflux
+MF_RABBITMQ_COOKIE=mainflux
+MF_RABBITMQ_VHOST=/
+MF_RABBITMQ_URL=amqp://mainflux:mainflux@broker:5672/
```
dborovcanin: You can reuse previous values to construct the URL: `MF_RABBITMQ_URL=amqp://$MF_RABBITMQ_USER:$MF_RABBITMQ_PASS@broker:$MF_RABBITMQ_PORT$MF_MF_RABBITMQ_VHOST`

scripts/ci.sh
```diff
@@ -65,6 +65,8 @@ setup_mf() {
         fi
     done
     make -j$NPROC
+    echo "Compile check for rabbitmq..."
+    MF_BROKER_TYPE=rabbitmq make -j $NPROC http
```
dborovcanin: Put this before `make -j $NPROC`: https://github.com/mainflux/mainflux/pull/1570#discussion_r901527315

dborovcanin (changes requested):
We're almost done.

cmd/mqtt/main.go
rodneyosodo: This was initially done but I am changing it.

rodneyosodo:
> Once #1562 is merged, this PR will need to rebase. Also, docs and deployment need to be updated: https://github.com/mainflux/docs and https://github.com/mainflux/devops.

Once this is merged we will update the docs and devops repos

dborovcanin (approved):

drasko (approved):
LGTM

PR #1572, opened 2022-03-07
NOISSUE - Fix CoAP adapter

dborovcanin:

### What does this do?
This pull request fixes the CoAP adapter in observe mode.
Since NATS PubSub implementation is limited to one subscriber per topic, and CoAP will have multiple subscribers with different tokens per single topic, NATS PubSub wrapper won't work for the CoAP adapter.

There are no such issues.

There are no breaking changes. Observe mod now takes into account the Observe option.

No.

No, since now the adapter acts according to the protocol spec.

### Notes
N/A

### DIFF
--- a/coap/api/transport.go
+++ b/coap/api/transport.go
@@ -31,7 +31,7 @@ const (
 	authQuery = "auth"
 )
 
-var channelPartRegExp = regexp.MustCompile(`/channels/([\w\-]+)/messages(/[^?]*)?(\?.*)?$`)
+var channelPartRegExp = regexp.MustCompile(`^/channels/([\w\-]+)/messages(/[^?]*)?(\?.*)?$`)
 
 var (
 	errMalformedSubtopic = errors.New("malformed subtopic")
@@ -105,7 +105,7 @@ func handler(w mux.ResponseWriter, m *mux.Message) {
 		case errors.Contains(err, errors.ErrAuthorization),
 			errors.Contains(err, errors.ErrAuthentication):
 			resp.Code = codes.Unauthorized
-		case errors.Contains(err, coap.ErrUnsubscribe):
+		default:
 			resp.Code = codes.InternalServerError
 		}
 		sendResp(w, &resp)
@@ -135,7 +135,7 @@ func decodeMessage(msg *mux.Message) (messaging.Message, error) {
 		return messaging.Message{}, err
 	}
 	channelParts := channelPartRegExp.FindStringSubmatch(path)
-	if len(channelParts) < 2 {
+	if len(channelParts) < 3 {
 		return messaging.Message{}, errMalformedSubtopic
 	}
 
@@ -145,7 +145,7 @@ func decodeMessage(msg *mux.Message) (messaging.Message, error) {
 	}
 	ret := messaging.Message{
 		Protocol: protocol,
-		Channel:  parseID(path),
+		Channel:  channelParts[1],
 		Subtopic: st,
 		Payload:  []byte{},
 		Created:  time.Now().UnixNano(),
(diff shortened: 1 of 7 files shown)


coap/api/transport.go
```diff
@@ -154,8 +155,8 @@ func decodeMessage(msg *mux.Message) (messaging.Message, error) {
 
 func parseID(path string) string {
 	vars := strings.Split(path, "/")
-	if len(vars) > 1 {
-		return vars[1]
+	if len(vars) > 2 {
```
drasko: It would be good to leave a small comment here, explaining where the number `2` comes from so that the reader can understand quickly.

drasko (changes requested):

rodneyosodo (commented):
LGTM.

So for the coap adapter, we will not be using the mainflux nats abstraction?

coap/api/transport.go
```diff
@@ -136,7 +145,7 @@ func decodeMessage(msg *mux.Message) (messaging.Message, error) {
 	}
 	ret := messaging.Message{
 		Protocol: protocol,
-		Channel:  parseID(path),
+		Channel:  channelParts[1],
```
manuio, line 154: It's not rlated to your modifications but there is a length check for `channelParts  < 2` but it should be `< 3` right?

manuio (changes requested):

coap/api/transport.go
dborovcanin: That's correct, good catch.

dborovcanin:
> LGTM.
> 
> So for the coap adapter, we will not be using the mainflux nats abstraction?

@0x6f736f646f Unfortunately, we can't use it. The reason is that we can have multiple subscriptions per topic, but for different clients, which is not allowed by our PubSub. The other option is to create PubSub per client which is not a good idea anyways.

manuio (dismissed):
LGTM

coap/api/transport.go
```diff
@@ -70,63 +73,69 @@ func handler(w mux.ResponseWriter, m *mux.Message) {
...
+	if msg.Options == nil {
+		return messaging.Message{}, errBadOptions
+	}
 	path, err := msg.Options.Path()
 	if err != nil {
 		return messaging.Message{}, err
 	}
 	channelParts := channelPartRegExp.FindStringSubmatch(path)
-	if len(channelParts) < 2 {
+	if len(channelParts) < 3 {
```
drasko: This magic number deserves a comment with an example how channel looks like and why we have 3 parts

drasko (changes requested):

coap/api/transport.go
```diff
@@ -70,73 +79,79 @@ func handler(w mux.ResponseWriter, m *mux.Message) {
...
-		logger.Warn("Nil options")
-		resp.Code = codes.BadOption
-		return
-	}
+
 	msg, err := decodeMessage(m)
 	if err != nil {
 		logger.Warn(fmt.Sprintf("Error decoding message: %s", err))
 		resp.Code = codes.BadRequest
+		sendResp(w, &resp)
```
mteodor, line 87: should we check if resp is nil ?

docker/docker-compose.yml
```diff
@@ -105,6 +105,8 @@ services:
       - ./nats/:/etc/nats
     networks:
       - mainflux-base-net
+    ports:
```
mteodor, line 108: why this?

dborovcanin: This is not related to this particular PR, but it is nice to have NATS port opened so that you can run services locally for development purposes, and still be able to connect to the other services running in Docker.

coap/api/transport.go
```diff
@@ -70,73 +79,79 @@ func handler(w mux.ResponseWriter, m *mux.Message) {
...
+	obs, err := m.Options.Observe()
+	if err != nil {
+		logger.Warn(fmt.Sprintf("Error reading observe option: %s", err))
+		return errBadOptions
 	}
+	if obs == 0 {
+		c := coap.NewClient(c, m.Token, logger)
+		return service.Subscribe(context.Background(), key, msg.Channel, msg.Subtopic, c)
+	}
+	return service.Unsubscribe(context.Background(), key, msg.Channel, msg.Subtopic, m.Token.String())
```
mteodor, line 132: why `m.Token.String()`?

coap/api/transport.go
```diff
@@ -70,73 +79,79 @@ func handler(w mux.ResponseWriter, m *mux.Message) {
...
+}
+
+func handleGet(m *mux.Message, c mux.Client, msg messaging.Message, key string) error {
+	var obs uint32
+	obs, err := m.Options.Observe()
+	if err != nil {
+		logger.Warn(fmt.Sprintf("Error reading observe option: %s", err))
+		return errBadOptions
 	}
+	if obs == 0 {
```
mteodor: why `0`? is it supposed to be `COAP_OBSERVE_CANCEL` and `COAP_OBSERVE_ESTABLISH`?
anyways we should make constant with meaningful name

coap/api/transport.go
dborovcanin: In the CoAP lib we use, Token is implemented as a byte slice, and, since Token is used to identify the session, a string is simpler to use (that's why they provided this `String` method).

coap/api/transport.go
dborovcanin: According to the spec, zero value of `Observe` option sent by the client means `Subscribe` while any other value indicates `Unsubscribe`.

mteodor (approved):

manuio (approved):
LGTM

drasko (approved):
LGTM

PR #1579, opened 2022-03-15
NOISSUE - Kafka support

### LINKED ISSUE #990: Add Kafka writer
**FEATURE REQUEST**

May Mainflux users have already legacy systems (apps) running on an enterprise bus which is often Kafka.

It would be high effort for them to switch to NATS, i.e. change their legacy apps to subscribe to NATS provided by Mainflux, or use Mainflux API to fetch the data.

For this purpose we can provide a NATS -> Kafka adapter/writer. It would be a small piece of software, very similar to existing writers (just that it will not write in some DB, but Kafka). It would just bluntly take messages from NATS and throw them into Kafka.

Must-have

rodneyosodo:
### What does this do?
Adds Kafka support. Since we are introducing a multi-broker system, we need to add support for Kafka. This allows us to use Kafka as a message broker for our system. Kafka has been chosen because it is the most popular legacy message broker. This allows us to support legacy systems that use Kafka.

Resolves #990 

N/A

Yes

https://github.com/mainflux/docs/pull/157

### Notes
The publisher implementation is based on the `segmentio/kafka-go` library. Publishing messages is well supported by the library, but subscribing to topics is not. The library does not provide a way to subscribe to all topics, but only to a specific topic. This is a problem because the Mainflux platform uses a topic per channel, and the number of channels is not known in advance. The solution is to use the Zookeeper library to get a list of all topics and then subscribe to each of them. The list of topics is obtained by connecting to the Zookeeper server and reading the list of topics from the `/brokers/topics` node. The first message received from the topic can be lost if subscription happens closely followed by publishing. After the subscription, we guarantee that all messages will be received.


pkg/messaging/broker/pubsub_kafka.go
```diff
@@ -0,0 +1,30 @@
...
+package broker
+
+import (
+	log "github.com/mainflux/mainflux/logger"
+
+	"github.com/mainflux/mainflux/pkg/messaging/kafka"
+)
+
+const (
+	// chansPrefix = "channels"
```
manuio: Remove commented code

pkg/messaging/broker/pubsub_nats.go
```diff
@@ -0,0 +1,29 @@
...
+
+package broker
+
+import (
+	log "github.com/mainflux/mainflux/logger"
+	"github.com/mainflux/mainflux/pkg/messaging/nats"
+)
+
+const (
+	// chansPrefix = "channels"
```
manuio: Same here

pkg/messaging/broker/pubsub_rabbitmq.go
```diff
@@ -0,0 +1,29 @@
...
+
+package broker
+
+import (
+	log "github.com/mainflux/mainflux/logger"
+	"github.com/mainflux/mainflux/pkg/messaging/rabbitmq"
+)
+
+const (
+	// chansPrefix = "channels"
```
manuio: and here

manuio (changes requested):

Makefile
```diff
@@ -16,7 +16,7 @@ TIME ?= $(shell date +%F_%T)
 
 define compile_service
 	CGO_ENABLED=$(CGO_ENABLED) GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) \
-	go build -mod=vendor -ldflags "-s -w \
+	go build -mod=vendor -tags $(MF_BROKER_TYPE) -ldflags "-s -w \
```
dborovcanin: Add a default value to `MF_BROKER_TYPE`, otherwise, the build will if it's empty.

Makefile
```diff
@@ -111,4 +111,4 @@ rundev:
 	cd scripts && ./run.sh
 
 run:
-	docker-compose -f docker/docker-compose.yml up
+	docker-compose -f docker/docker-compose.yml up -d
```
drasko: Remove `-d`

pkg/messaging/nats/pubsub.go
```diff
@@ -15,10 +15,11 @@ import (
...
 )
 
-const chansPrefix = "channels"
-
-// SubjectAllChannels represents subject to subscribe for all the channels.
-const SubjectAllChannels = "channels.>"
+const (
+	chansPrefix = "channels"
+	// SubjectAllChannels represents subject to subscribe for all the channels.
+	SubjectAllChannels = "channels.>"
```
drasko: I would construct this one as `chansPrefix + ".>"`

twins/README.md
```diff
@@ -22,8 +22,8 @@ default values.
 | MF_TWINS_DB                | Database name                                                        | mainflux              |
 | MF_TWINS_DB_HOST           | Database host address                                                | localhost             |
 | MF_TWINS_DB_PORT           | Database host port                                                   | 27017                 |
-| MF_THINGS_STANDALONE_EMAIL | User email for standalone mode (no gRPC communication with users)       |                |
-| MF_THINGS_STANDALONE_TOKEN | User token for standalone mode that should be passed in auth header     |                |
+| MF_THINGS_STANDALONE_EMAIL | User email for standalone mode (no gRPC communication with users)    |                       |
```
drasko: Remove tabs, use spaces

drasko (changes requested):

pkg/messaging/rabbitmq/publisher.go
```diff
@@ -65,7 +68,7 @@ func (pub *publisher) Publish(topic string, msg messaging.Message) error {
 		immediate,
 		amqp.Publishing{
 			Headers:     amqp.Table{},
-			ContentType: "text/plain",
+			ContentType: "application/octet-stream",
```
manuio: You can use a const here

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -34,6 +34,7 @@ var (
 	errAlreadySubscribed = errors.New("already subscribed to topic")
 	errNotSubscribed     = errors.New("not subscribed")
 	errEmptyTopic        = errors.New("empty topic")
+	errEmptyID           = errors.New("empty ID")
```
manuio: If I remember well we are not using upper-case in logs

pkg/messaging/rabbitmq/publisher.go
rodneyosodo: This is linked with RabbitMQ branch [here](https://github.com/mainflux/mainflux/pull/1570)

pkg/messaging/nats/pubsub.go
rodneyosodo: That is what is implemented currently across all brokers. Should we change?

pkg/messaging/rabbitmq/pubsub.go
rodneyosodo: This is also linked with the RabbitMQ branch let me update it [here](https://github.com/mainflux/mainflux/pull/1570)

docker/broker/rabbitmq.yml
```diff
@@ -0,0 +1,11 @@
...
+  broker:
+    image: rabbitmq:3.7-management
+    environment:
+      RABBITMQ_ERLANG_COOKIE: ${MF_RABBITMQ_COOKIE}
+      RABBITMQ_DEFAULT_USER: ${MF_RABBITMQ_USER}
+      RABBITMQ_DEFAULT_PASS: ${MF_RABBITMQ_PASS}
+      RABBITMQ_DEFAULT_VHOST: ${MF_RABBITMQ_VHOST}
+    ports:
+      - ${MF_RABBITMQ_PORT}:${MF_RABBITMQ_PORT}
+      - ${MF_RABBITMQ_HTTP_PORT}:${MF_RABBITMQ_HTTP_PORT}
```
dborovcanin: Please add an empty line.

pkg/messaging/README.md
```diff
@@ -2,8 +2,8 @@
 
 `messaging` package defines `Publisher`, `Subscriber` and an aggregate `Pubsub` interface. 
 
-`Subscriber` interface defines methods used to subscribe to a message broker such as MQTT or NATS or RabbitMQ. 
+`Subscriber` interface defines methods used to subscribe to a message broker such as MQTT or NATS or RabbitMQ or Kafka. 
 
-`Publisher` interface defines methods used to publish messages to a message broker such as MQTT or NATS or RabbitMQ.
+`Publisher` interface defines methods used to publish messages to a message broker such as MQTT or NATS or RabbitMQ or Kafka.
```
dborovcanin: `Publisher` interface defines methods used to publish messages to a message broker.
Currently, supported brokers are MQTT, NATS, RabbitMQ, and Kafka.

pkg/messaging/kafka/publisher.go
```diff
@@ -0,0 +1,73 @@
...
+	})
+	defer writer.Close()
+
+	kafkaMsg := kafka.Message{
+		Value: data,
+		Topic: subject,
+	}
+	if err := writer.WriteMessages(context.TODO(), kafkaMsg); err != nil {
+		// Sometime it take time for leader to be elected. If that is so, we retry to publish message
+		if strings.Contains(fmt.Sprint(err), "[5] Leader Not Available:") {
```
dborovcanin: Extract this string literal to a constant if there is no other way to decide the error type. Use `err.Error()` instead of `fmt.Sprint(err)`.

dborovcanin: Looks like there is already an existing exported error for this in `vendor/github.com/segmentio/kafka-go/error.goL166` so you can reuse it.

pkg/messaging/kafka/publisher.go
```diff
@@ -0,0 +1,73 @@
...
+			time.Sleep(2 * time.Second)
```
dborovcanin: Consider adding an exponential backoff.

pkg/messaging/kafka/pubsub.go
```diff
@@ -0,0 +1,149 @@
...
+	if id == "" {
+		return ErrEmptyID
+	}
+	if topic == "" {
+		return ErrEmptyTopic
+	}
+	ps.mu.Lock()
+	defer ps.mu.Unlock()
+	// Check topic
+	s, ok := ps.subscriptions[topic]
```
dborovcanin: You can use `subs` here and later `s := subs[id]` or `sub := subs[id]` instead of `reader := subs[id]`.

pkg/messaging/kafka/pubsub_test.go
```diff
@@ -0,0 +1,221 @@
...
+	// Test Subscribe and Unsubscribe
+	subcases := []struct {
+		desc         string
+		topic        string
+		topicID      string
+		errorMessage error
+		pubsub       bool //true for subscribe and false for unsubscribe
+	}{
+		{
+			desc:         "Susbcribe to a topic with an ID",
```
dborovcanin: Fix typos `susbcribe`.

dborovcanin (changes requested):

dborovcanin:
@0x6f736f646f Do not forget to update `Makefile` and `docker-compose.yml` to enable run with `MF_BROKER_TYPE=kafka make run`.

Makefile
```diff
@@ -123,6 +123,9 @@ ifeq ("$(MF_BROKER_TYPE)", "rabbitmq")
 else ifeq ("$(MF_BROKER_TYPE)", "nats")
 	sed -i "s,file: brokers/.*.yml,file: brokers/nats.yml," docker/docker-compose.yml
 	sed -i "s,MF_BROKER_URL: .*,MF_BROKER_URL: $$\{MF_NATS_URL\}," docker/docker-compose.yml
+else ifeq ("$(MF_BROKER_TYPE)", "kafka")
+	sed -i "s,file: brokers/.*.yml,file: brokers/kafka.yml," docker/docker-compose.yml
+	sed -i "s,MF_BROKER_URL: .*,MF_BROKER_URL: $$\{MF_KAFKA_URL\}," docker/docker-compose.yml
```
dborovcanin: There is no `MF_KAFKA_URL` in `.env` file.

dborovcanin (changes requested):
Please fix CI.

pkg/messaging/kafka/publisher.go
```diff
@@ -0,0 +1,91 @@
...
+		if delay > backoffCeiling {
+			delay = backoffCeiling
+		}
+
+		time.Sleep(time.Duration(delay) * time.Second)
+		attempts += 1
+
+		// Sometime it take time for leader to be elected. If that is so, we retry to publish message
+		err := writer.WriteMessages(context.TODO(), msg)
+		fmt.Println(err.Error())
```
dborovcanin: Remove.

dborovcanin (changes requested):

pkg/messaging/kafka/publisher.go
rodneyosodo: I get a `runtime error: invalid memory address or nil pointer dereference [recovered]` sometimes when it tries and checks if the error is `unexpected EOF`

pkg/messaging/kafka/publisher.go
```diff
@@ -0,0 +1,78 @@
...
+		Async:   true,
+	})
+	defer writer.Close()
+
+	kafkaMsg := kafka.Message{
+		Value: data,
+		Topic: subject,
+	}
+	for _, backoff := range backoffSchedule {
+		err := writer.WriteMessages(context.TODO(), kafkaMsg)
```
dborovcanin: Since the write is set to be async, the error is unlikely to happen. It's OK to keep it async, but check what are the ways to log the possible error(s). Take a look at `ErrorLogger` field of writer config.

pkg/messaging/kafka/publisher.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	data, err := proto.Marshal(&msg)
+	if err != nil {
+		return err
+	}
+	subject := fmt.Sprintf("%s.%s", chansPrefix, topic)
+	if msg.Subtopic != "" {
+		subject = fmt.Sprintf("%s.%s", subject, msg.Subtopic)
+	}
+	writer := kafka.NewWriter(kafka.WriterConfig{
```
dborovcanin: Check writer config options and try to find the best one for our use case. Looks like there are many options, backoff included (so you don't have to implement it here).

dborovcanin (changes requested):

vendor/github.com/segmentio/kafka-go/writer.go
```diff
@@ -0,0 +1,1272 @@
...
+			// operation?
+			//
+			// * If all goroutines that added message to this batch have stopped
+			//   waiting for it, should we abort?
+			//
+			// * If the writer has been closed? It reduces the durability
+			//   guarantees to abort, but may be better to avoid long wait times
+			//   on close.
+			//
+			delay := backoff(attempt, 100*time.Millisecond, 1*time.Second)
```
rodneyosodo: Here

pkg/messaging/kafka/publisher.go
rodneyosodo: They have implemented it but with no configurable parameters for the writer [here](https://github.com/mainflux/mainflux/pull/1579/files#r905379708)

dborovcanin: There is a `MaxAttempts` field that can be used for that. The lib is a little bit complex, but you can try and debug the code to check how it works.

rodneyosodo: It backs off up to 1s then stops

```
kafka writer: writing 1 messages to channels.topic.engine (partition: 0)
kafka writer: error writing messages to channels.topic.engine (partition 0): kafka.(*Client).Produce: dial tcp 127.0.0.1:9092: connect: connection refused
kafka writer: backing off 100ms writing 1 messages to channels.topic.engine (partition: 0)
kafka writer: writing 1 messages to channels.topic.engine (partition: 0)
kafka writer: error writing messages to channels.topic.engine (partition 0): kafka.(*Client).Produce: dial tcp 127.0.0.1:9092: connect: connection refused
kafka writer: backing off 400ms writing 1 messages to channels.topic.engine (partition: 0)
kafka writer: writing 1 messages to channels.topic.engine (partition: 0)
kafka writer: error writing messages to channels.topic.engine (partition 0): kafka.(*Client).Produce: dial tcp 127.0.0.1:9092: connect: connection refused
kafka writer: backing off 900ms writing 1 messages to channels.topic.engine (partition: 0)
kafka writer: writing 1 messages to channels.topic.engine (partition: 0)
kafka writer: error writing messages to channels.topic.engine (partition 0): kafka.(*Client).Produce: dial tcp 127.0.0.1:9092: connect: connection refused
kafka writer: backing off 1s writing 1 messages to channels.topic.engine (partition: 0)
```

dborovcanin: That should be OK, 1s should be more than enough. In case of, say, 10 attempts - it's up to 10s. We assume that Kafka is ready to accept messages before we start and backoff is only for cases where something malfunctioning or Kafka is overloaded. How did you test this, did you wait some time after compose is up, and did you check Kafka logs to be sure it's up and ready to accept connections before you started testing?

rodneyosodo: When testing it on the platform all other services throw this error message `{"level":"error","message":"Failed to connect to postgres: dial tcp 172.19.0.6:5432: connect: connection refused","ts":"2022-06-24T13:56:42.923294665Z"}`  but after that it connects. Publishing from one adapter and receiving in another works fine.

pkg/messaging/kafka/publisher.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/mainflux/mainflux/pkg/messaging"
+	"github.com/segmentio/kafka-go"
+)
+
+var (
+	_ messaging.Publisher = (*publisher)(nil)
+)
+
+type publisher struct {
+	conn *kafka.Conn
```
dborovcanin: This field is not used.

pkg/messaging/kafka/publisher.go
```diff
@@ -0,0 +1,65 @@
...
+	url  string
+}
+
+// NewPublisher returns Kafka message Publisher.
+func NewPublisher(url string) (messaging.Publisher, error) {
+	conn, err := kafka.Dial("tcp", url)
+	if err != nil {
+		return nil, err
+	}
```
dborovcanin: It's OK to check the connection, but there is no point in keeping this in the `conn` field that is not used.

pkg/messaging/kafka/publisher.go
```diff
@@ -0,0 +1,65 @@
...
+	data, err := proto.Marshal(&msg)
+	if err != nil {
+		return err
+	}
+	subject := fmt.Sprintf("%s.%s", chansPrefix, topic)
+	if msg.Subtopic != "" {
+		subject = fmt.Sprintf("%s.%s", subject, msg.Subtopic)
+	}
+	writer := kafka.NewWriter(kafka.WriterConfig{
```
dborovcanin: It looks like `kafka.NewWriter(kafka.WriterConfig)` is deprecated and the preferred way is to use the Wrteir struct directly.

dborovcanin (changes requested):
CI is failing.

rodneyosodo:
The `LeaderNotAvailable` is retriable according to [this](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol) docs. A possible workaround would be to retry and if not possibly fail and manage the fails. The [library](https://github.com/segmentio/kafka-go) we are using retries up to 1s then fails. As demonstrated earlier, if you retry for up to at least 10 seconds the writer will write the message. The current work around it is to create topics using the Kafka command inside the broker as referenced [here](https://github.com/Shopify/sarama/issues/272#issuecomment-261834761). 
Our problem is mainly on the [CI](https://semaphoreci.com/mainflux/mainflux/branches/pull-request-1579/builds/61) as when mainflux is deployed you can be able to publish a message from one adapter and receive it using another. The interim workaround is to create the topics first before testing the code.

Another problem is that on mainflux when we publish message for the first time it throws the same error `mainflux-http   | {"level":"warn","message":"Method publish to channel 242bcd02-fb24-4c8c-b56d-de97ad56de34 took 49.085996ms to complete with error: [5] Leader Not Available: the cluster is in the middle of a leadership election and there is currently no leader for this partition and hence it is unavailable for writes.","ts":"2022-06-27T14:01:40.789712934Z"}`

This is an error with regard to Kafka as it is being experienced with other library users such as [Sarama](https://github.com/Shopify/sarama/issues/272).

My recommendation would be to increase the retrying on public method upto 10s then manage the error after that

dborovcanin:
> The `LeaderNotAvailable` is retriable according to [this](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol) docs


Can you point out the exact section for this?

> Our problem is mainly on the [CI](https://semaphoreci.com/mainflux/mainflux/branches/pull-request-1579/builds/61) as when mainflux is deployed you can be able to publish a message from one adapter and receive it using another. The interim workaround is to create the topics first before testing the code.

Looks like the problem is actually that first-time writing always arises this error, not due to Kafka not being ready but because we do not have a topic created. I guess Kafka creates a topic internally when we try to publish, and the next attempt will pass because on the first - Kafka took care of topic for us. Please investigate if it works like that and how to correctly handle topics lifetime (we do not remove them when we unsubscribe, and I assume there are other ways to configure Kafka to handle it for us).

rodneyosodo:
> Can you point out the exact section for this?

You can check it from [here](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol/#AGuideToTheKafkaProtocol-ErrorCodes)

> Please investigate if it works like

Yes it works like that.
I have tried two approaches:

1. To create the topic as reference [here](https://github.com/segmentio/kafka-go#to-create-topics) but only when `KAFKA_AUTO_CREATE_TOPICS_ENABLE='false'`

     This doesn't work as expected as 
```go
topicConfigs := []kafka.TopicConfig{
	{
		Topic:             subject,
		NumPartitions:     1,
		ReplicationFactor: 1,
	},
}
err = pub.conn.CreateTopics(topicConfigs...)
if err != nil {
	return err
}
```
   after initially creating the kafka connection object.

3. To create the topic while publishing as reference [here](https://github.com/segmentio/kafka-go#missing-topic-creation-before-publication) 
    This one worked fine.

rodneyosodo:
## Topic management
Kafka topics are the categories used to organize messages. Each topic has a name that is unique across the entire Kafka cluster.

### Creation

- We are manually creating the topics before publishing to it. 
- If we try and create an already created topic the topic it is idempotent since the values haven't changed

### Deletion
- Since we are creating the topics ourselves we have the privilege to delete them ourselves. This is not an elegant solution but a feasible solution to our case.
- Before we can delete any topic we would want to make sure the deletion of topics is enabled in our cluster. Set `delete.topic.enable=true`.
- We have to make sure all consumers have stopped consuming the data from the topic we want to delete, otherwise, kafka will throw errors.
- If there is a chance that a consumer is up and running, the topic will get auto-created if the cluster-wide property `auto.create.topics.enable=true`. Not bad per se, but it will use a default number of partitions (1) and a replication factor (1), which might not be what we wanted.

We are currently deleting topics when we close the publisher interface

### Publishing message
- Configured retries to be at least 1000 since we care about data delivery to clients [Ref](https://newrelic.com/blog/best-practices/kafka-best-practices)
- Configured producer for to wait for acknowledgment  [Ref](https://newrelic.com/blog/best-practices/kafka-best-practices)
- Stuck to random partitioning when writing to topics [Ref](https://www.pepperdata.com/blog/kafka-optimization-four-best-practices/)

pkg/messaging/kafka/publisher.go
```diff
@@ -0,0 +1,109 @@
...
+	data, err := proto.Marshal(&msg)
+	if err != nil {
+		return err
+	}
+	subject := fmt.Sprintf("%s.%s", chansPrefix, topic)
+	if msg.Subtopic != "" {
+		subject = fmt.Sprintf("%s.%s", subject, msg.Subtopic)
+	}
+
+	conn, err := kafka.Dial("tcp", pub.url)
```
dborovcanin: Revert to having a connection as a `publisher` field and close it when you close the publisher, so we don't open and close the connection all the time - we keep it open.

pkg/messaging/kafka/publisher.go
```diff
@@ -0,0 +1,109 @@
...
+		return err
+	}
+
+	writer := &kafka.Writer{
+		Addr:         kafka.TCP(pub.url),
+		Async:        true,
+		MaxAttempts:  100,
+		RequiredAcks: kafka.RequireAll,
+	}
+	defer writer.Close()
```
dborovcanin: Is there a need to create a new writer always or you can always use the same one - as a field of the publisher?

pkg/messaging/kafka/publisher.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	pub.mu.Lock()
+	defer pub.mu.Unlock()
+	if !pub.topicInTopics(subject) {
+		pub.topics = append(pub.topics, subject)
+	}
+	return nil
+}
+
+func (pub *publisher) Close() error {
```
dborovcanin, line 100: The publisher is closed only at the end of its lifetime - removing topics that way does not sound particularly useful since Publisher is meant to be long-lasting.

pkg/messaging/kafka/pubsub.go
```diff
@@ -0,0 +1,159 @@
...
+		// Check topic ID
+		if _, ok := s[id]; ok {
+			return ErrAlreadySubscribed
+		}
+	default:
+		s = make(map[string]subscription)
+		ps.subscriptions[topic] = s
+	}
+
+	reader := kafka.NewReader(kafka.ReaderConfig{
```
dborovcanin: What happens to the reader when there is no topic existing?

pkg/messaging/kafka/pubsub.go
```diff
@@ -0,0 +1,159 @@
...
+		return err
+	}
+	delete(subs, id)
+	if len(subs) == 0 {
+		delete(ps.subscriptions, topic)
+	}
+	return nil
+}
+
+func (ps *pubsub) Close() error {
```
dborovcanin, line 141: Simplify with something like this:
```suggestion
func (ps *pubsub) Close() error {
	ps.mu.Lock()
	defer ps.mu.Unlock()
	for _, subs := range ps.subscriptions {
		for _, s := range subs {
			if s.cancel != nil {
				if err := s.close(); err != nil {
					return err
				}
			}
		}
	}
	return nil
}
```
where `s.close` is something like this:
```go
func (s subscription) close() error {
	if err := s.cancel(); err != nil {
		return err
	}
	return s.Close()
}
```

dborovcanin (changes requested):

pkg/messaging/kafka/pubsub.go
rodneyosodo: If `auto.create.topics.enable=true` in your Kafka configuration file it will automatically create that topic.

pkg/messaging/kafka/publisher.go
rodneyosodo: What about cleaning up?

dborovcanin: I am not sure. Is there a way to configure Kafka to delete the topic if no client has pushed to it in a while (say 1d) or no client is subscribed? If we leave topics intact, how does that affect performance?

rodneyosodo: From what I have reviewed I have not seen a way to automatically delete the topic after a specific period. That option is only done with message using [delete.retention.ms](https://kafka.apache.org/documentation/#topicconfigs_delete.retention.ms). Also, the number of topics isn't all that important in and of itself, but each Kafka topic is partitioned, and the total number of partitions may have an effect on performance [[1]](https://www.instaclustr.com/blog/the-power-of-kafka-partitions-how-to-get-the-most-out-of-your-kafka-cluster/).

drasko (dismissed):
LGTM

users/postgres/users.go
```diff
@@ -115,7 +115,7 @@ func (ur userRepository) RetrieveByEmail(ctx context.Context, email string) (use
 }
 
 func (ur userRepository) RetrieveByID(ctx context.Context, id string) (users.User, error) {
-	q := `SELECT email, password, metadata FROM users WHERE id = $1`
+	q := `SELECT email, password, metadata FROM users WHERE state = 'active' AND id = $1`
```
arvindh123: This part of code is covered in this PR ? Or it should be done in separate PR , 
How we retrieve the non active user id ?

rodneyosodo: No. Let me revert back

rodneyosodo:
[From](https://github.com/segmentio/kafka-go/issues/283) this it appears that in the segmentio library you can list all topics with a bit of work around. But the problem is when the MQTT forwarder is subscribing to the broker, it uses the wildcard to subscribe to all topics but they are no topics in the kakfa broker.

rodneyosodo:
Regarding Kafka support for wildcard, I have tested using [confluent-kafka-python](https://github.com/confluentinc/confluent-kafka-python) and it supports. I think [the issue is the library we are using](https://github.com/segmentio/kafka-go/pull/663#issuecomment-980583030) I have tried to pass in a regex on the topic and it throws this error

```bash
2023/02/07 14:24:37 Failed to join group groupID2: [17] Invalid Topic: a request which attempted to access an invalid topic (e.g. one which has an illegal name), or if an attempt was made to write to an internal topic (such as the consumer offsets topic)
2023/02/07 14:24:37 Leaving group groupID2, member main@r0x6f736f646f (github.com/segmentio/kafka-go)-77b03394-27c9-4846-89c2-fe1fb51f7034
```

I have tried with another library https://github.com/confluentinc/confluent-kafka-go and it works. I want to write the implementation on this kafka PR and update after testing it on mainflux

dborovcanin:
Let's address the subscriber problem and research our options in a separate PR and make this PR only a `Publisher` implementation. Using Subscriber will simply panic. Don't forget to document it.

dborovcanin:
This PR might be a better fit for https://github.com/absmach/mg-contrib since Kafka does not support all the features MG messaging requires, but may also be useful for integration with systems that use Kafka and do not need those features.

dborovcanin:
As already commented [here](https://github.com/absmach/magistrala/pull/1579#issuecomment-2228807609), this may be a better fit for the `mg-contrib` repository. I'm closing this PR.

PR #1581, opened 2022-03-18
MF-1580 - Influxdb Writer changes format of update-time to string

### LINKED ISSUE #1580: Influxdb Writer changes format of update-time to string
**BUG REPORT**

I am using influxdb to store telemetry data, and while sending update-time as float data type to mainflux mqtt-broker, I noticed that the influxdb-writer service converts the update-time's data type from float to string. The writer shouldn't convert the datatype of incoming data and keep it as it is.

3. What are the expected results?
I expect update-time to be a floating-point number.

4. What are the received results?
Update time is stored as a string.

5. What are the steps to reproduce the issue?
Run latest Docker composition, run InfluxDB Writer, provision and connect thing and channel, send message to any of adapters.

6. In what environment did you encounter the issue?
Latest Docker compose; tested with multiple systems.

7. Additional information you deem important:
N/A

Hasan98-git:
### What does this do?
This will stop influxdb-writer service from changing the datatype of update-time variable to string

Resolves #1580

It won't break current functionality

No

No

### DIFF
--- a/consumers/writers/influxdb/fields.go
+++ b/consumers/writers/influxdb/fields.go
@@ -9,7 +9,7 @@ import (
 type fields map[string]interface{}
 
 func senmlFields(msg senml.Message) fields {
-	updateTime := strconv.FormatFloat(msg.UpdateTime, 'f', -1, 64)
+	// updateTime := strconv.FormatFloat(msg.UpdateTime, 'f', -1, 64)
 	ret := fields{
 		"protocol":   msg.Protocol,
 		"unit":       msg.Unit,
--- a/consumers/writers/influxdb/fields.go
+++ b/consumers/writers/influxdb/fields.go
@@ -1,19 +1,16 @@
 package influxdb
 
 import (
-	"strconv"
-
 	"github.com/mainflux/mainflux/pkg/transformers/senml"
 )
 
 type fields map[string]interface{}
 
 func senmlFields(msg senml.Message) fields {
-	// updateTime := strconv.FormatFloat(msg.UpdateTime, 'f', -1, 64)
 	ret := fields{
 		"protocol":   msg.Protocol,
 		"unit":       msg.Unit,
-		"updateTime": updateTime,
+		"updateTime": msg.UpdateTime,
 	}
 
 	switch {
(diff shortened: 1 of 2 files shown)


consumers/writers/influxdb/fields.go
```diff
@@ -9,7 +9,7 @@ import (
 type fields map[string]interface{}
 
 func senmlFields(msg senml.Message) fields {
-	updateTime := strconv.FormatFloat(msg.UpdateTime, 'f', -1, 64)
+	// updateTime := strconv.FormatFloat(msg.UpdateTime, 'f', -1, 64)
```
manuio: You can simply remove it

manuio (dismissed):

consumers/writers/influxdb/fields.go
Hasan98-git: Okay, but I have checked that the build fails, I thing i will need to remove the strconv package also and replace retur
n value of updateTime with msg.UpdateTime

dborovcanin:
@Hasan98-git Please update tests accordingly.

Hasan98-git:
@dusanb94 can you guide on how and where to update the tests?

dborovcanin:
> @dusanb94 can you guide on how and where to update the tests?

Sure, @Hasan98-git. You can exec tests locally via `go test --race -v ./... -count=1` (you have to have Docker service running). You'll notice that error is in the InfluxDB reader. You can run the same command to form `readers/influxdb` to check what's the probelem.

dborovcanin:
Hello, @Hasan98-git. Tests fix is a little complex and out of the scope of this PR. We should probably switch to Flux and InfluxDB v2 anyways, but that's also not in the scope of this PR. Would you allow me to push test fixes (it's actually Reader update) to this branch to prevent breaking the `master` branch? I want to keep your contribution, but I'd also like to keep the latest codebase working.

dborovcanin (approved):

manuio (approved):
LGTM

Hasan98-git:
@dusanb94 I have worked on the test but I wasn't able to figure out why it was failing, the issue is that after removing the strconv functionality from consumers/writers/influxdb/fields.go the writer writes the vlaue of updateTime as a zero e-g in the test (readers/influxdb/messages_test.go) the value of updateTime was set to 1234 but the writer writes it as 0 in the db due to which the test fails, can you help?

Hasan98-git:
@dusanb94 Hi, yeah that will be great, thanks for the help.

PR #1589, opened 2022-04-01
NOISSUE - Get user by ID and get all users by SDK

rodneyosodo:

### What does this do?
It underlines the clear distinction between get user by id and get all users if you are admin.

Resolvers #1618

It adds get all users to the mainflux SDK and also get user by id. The current implementation for cli fails to get user by id as show here

```sh
mainflux-cli users get $USERTOKEN

{}

```

The new implementation presents
```sh
./mainflux-cli users get all $USERTOKEN  
[
  {
    "email": "admin@example.com",
    "id": "3803babf-7bf3-458b-a337-6905b575a6c6"
  },
  {
    "email": "pedantic_elion@email.com",
    "id": "be1c2ef2-6712-41d5-b64d-d22c0a5d3803"
  },
  {
    "email": "reverent_murdock@email.com",
    "id": "124422b7-883e-460c-bebc-12ea04d4934b"
  }
]

./mainflux-cli users get 124422b7-883e-460c-bebc-12ea04d4934b $USERTOKEN

{
  "email": "reverent_murdock@email.com",
  "id": "124422b7-883e-460c-bebc-12ea04d4934b"
}

```

Yes

No

### Notes
N/A


manuio (dismissed):
LGTM

dborovcanin (changes requested):
Please provide pagination options, as well.
Also, please don't forget to update the tests.

rodneyosodo:
Where should I provide the pagination option
- If it is on the cli, I have provided
- If it is on the SDK, I haven't as the user's endpoint doesn't provide a means of querying it

pkg/sdk/go/users_test.go
```diff
@@ -128,6 +128,165 @@ func TestCreateUser(t *testing.T) {
...
+		desc     string
+		token    string
+		offset   uint64
+		limit    uint64
+		err      error
+		response []sdk.User
+		name     string
+	}{
+		{
+			desc:     "get a list users",
```
manuio: `get a list of users`

pkg/sdk/go/users_test.go
```diff
@@ -128,6 +128,165 @@ func TestCreateUser(t *testing.T) {
...
+			token:    token,
+			offset:   0,
+			limit:    5,
```
manuio: Define a constant for `limit`.

pkg/sdk/go/users_test.go
```diff
@@ -128,6 +128,165 @@ func TestCreateUser(t *testing.T) {
...
+			err:      nil,
+			response: users[0:5],
```
manuio: Use the const here: `users[0:limit]`

manuio (changes requested):

provision/service.go
```diff
@@ -34,6 +34,13 @@ var (
 	ErrFailedBootstrap          = errors.New("failed to create bootstrap config")
 	ErrFailedBootstrapValidate  = errors.New("failed to validate bootstrap config creation")
 	ErrGatewayUpdate            = errors.New("failed to updated gateway metadata")
+
+	// Limit query parameter
+	Limit uint = 10
```
mteodor: this doesnt need to be exported, right?

pkg/sdk/go/users_test.go
rodneyosodo: The limit changes. As we test we can choose a limit of 5 or 110 for testing

provision/service.go
rodneyosodo: I can add it without defining it when calling the users function in the SDK i.e 
`if _, err := ps.sdk.Users(token, 0, 10, ""); err != nil {}`

pkg/sdk/go/users_test.go
manuio: The limit is related to the size of the slice and it's reapeated many times. Creating a const you avoid any bug or repetition and it's self explanatory.

provision/service.go
rodneyosodo: Aah. I get what you are referring to

rodneyosodo: Done

pkg/sdk/go/users_test.go
rodneyosodo: Done

pkg/sdk/go/users_test.go
rodneyosodo: Done

pkg/sdk/go/channels_test.go
```diff
@@ -386,7 +386,7 @@ func TestChannelsByThing(t *testing.T) {
 			desc:     "get a list of channels by thing with limit greater than max",
 			thing:    tid,
 			token:    token,
-			offset:   0,
+			offset:   Offset,
```
manuio: This can be probably defined without to export, i.e `offset` and `limit`

pkg/sdk/go/sdk.go
```diff
@@ -130,8 +130,11 @@ type SDK interface {
 	// CreateUser registers mainflux user.
 	CreateUser(token string, user User) (string, error)
 
-	// User returns user object.
-	User(token string) (User, error)
+	// User returns user object by id.
+	User(token, id string) (User, error)
+
+	// Users returns list of users.
+	Users(token string, offset, limit uint64, name string) (UsersPage, error)
```
manuio: You can also add the `metadata` filter, check here: https://github.com/mainflux/mainflux/blob/master/users/postgres/users.go#L173. Please, also check about the `name` filter. From what I see what we were using it's the `email`. Check here: https://github.com/mainflux/mainflux/blob/master/users/api/transport.go#L145 The name is missing, you can probably add it.
BTW, tests are probably working because name is empty, add also tests to filter by name and metadata.

manuio (changes requested):

pkg/sdk/go/channels_test.go
rodneyosodo: I have exported it as the test is similar for things and also channels

manuio: I'd define consts per test, no need to export. Eventually use different values for each test.

rodneyosodo: Done

pkg/sdk/go/sdk.go
rodneyosodo: Fantastic. Let me implement that

rodneyosodo: This also applies to things. I will add it in the same pull request

manuio: great! BTW, I recommend you to use a structure for all filters and pass it everywhere as argument. Similar to `PageMetadata` here: https://github.com/mainflux/mainflux/blob/master/things/service.go#L50
It's simpler at many levels.

cli/channels.go
```diff
@@ -45,9 +45,14 @@ var cmdChannels = []cobra.Command{
 				logUsage(cmd.Use)
 				return
 			}
-
+			filteredInfo := mfxsdk.Filter{
```
manuio: I'd use `pageFilters` instead of `filteredInfo`

pkg/sdk/go/channels.go
```diff
@@ -76,8 +76,13 @@ func (sdk mfSDK) CreateChannels(chs []Channel, token string) ([]Channel, error)
 	return ccr.Channels, nil
 }
 
-func (sdk mfSDK) Channels(token string, offset, limit uint64, name string) (ChannelsPage, error) {
-	url := fmt.Sprintf("%s/%s?offset=%d&limit=%d&name=%s", sdk.thingsURL, channelsEndpoint, offset, limit, name)
+func (sdk mfSDK) Channels(token string, filter Filter) (ChannelsPage, error) {
+	endpoint, err := sdk.parseFilteredValues(filter)
```
manuio: `endpoint` naming is confusing here. You are creating the query parameters part of the URL. Probably the best is to change the method to accept `sdk.thingsURL` and `channelsEndpoint`. For example `sdk.createURL(sdk.thingsURL, channelsEndpoint,, filters)`

manuio (changes requested):

cli/channels.go
rodneyosodo: Done

pkg/sdk/go/channels.go
rodneyosodo: Done

rodneyosodo:
https://github.com/mainflux/mainflux/pull/1594#issuecomment-1112040258

Users pagemetadata has name field instead of email field

rodneyosodo:
- Used `pageMetadata` as referenced [here](https://github.com/mainflux/mainflux/pull/1594#issuecomment-1112040258)
- Change Users `Name` to `Email` field

pkg/sdk/go/sdk.go
```diff
@@ -85,6 +89,11 @@ type User struct {
 	Password string                 `json:"password,omitempty"`
 	Metadata map[string]interface{} `json:"metadata,omitempty"`
 }
+type UserPage users.PageMetadata
+type UserPageMetadata struct {
+	Metadata map[string]interface{} `json:"metadata,omitempty"`
+	UserPage
+}
```
dborovcanin: Don't import Users service as a dependency to SDK. We are trying to keep SDK as simple and independent as possible, even at cost of copying structs from other services. Also, there is nothing wrong with using a single metadata page that joins all the metadata fields for all the entities in the SDK; unused fields will be simply ignored.

dborovcanin (changes requested):

pkg/sdk/go/sdk.go
```diff
@@ -349,3 +367,33 @@ func (sdk mfSDK) sendThingRequest(req *http.Request, key, contentType string) (*
 
 	return sdk.client.Do(req)
 }
+
+func (sdk mfSDK) parseUserFilteredValues(baseURL, endpoint string, filter UserPageMetadata) (string, error) {
```
manuio: This can be improved by Marshaling `UserPageMetadata` in a `map[string]interface` and then adding query parameters to the URL using a for loop and a switch case. Something like that: 
```
        var queryParams := ""
	for key, val := range pageMeta {
		switch k {
		case
			"offset",
			"limit",
			"name",
			"email":
			queryParams = fmt.Sprintf(`%s&%s=%s `, queryParams, key, val)
		case "metadata":
                       ...
		}
	}
	return fmt.Sprintf("%s/%s?%s, baseURL, endpoint, queryParams)
```

You have to also check if it's or not the first query parameter to handle properly `?` and `&`.

manuio (changes requested):

pkg/sdk/go/sdk.go
rodneyosodo: I am reverting to using one PageMetadata which contains all the information for paging. For users, we will be interested in email while for things and channels we will be interested in the name
```go
type PageMetadata struct {
	Total    uint64
	Offset   uint64
	Limit    uint64
	Email    string
	Name     string
	Metadata map[string]interface{}
}
```

pkg/sdk/go/sdk.go
```diff
@@ -130,8 +140,11 @@ type SDK interface {
 	// CreateUser registers mainflux user.
 	CreateUser(token string, user User) (string, error)
 
-	// User returns user object.
-	User(token string) (User, error)
+	// User returns user object by id.
+	User(token, id string) (User, error)
+
+	// Users returns list of users.
+	Users(token string, filter PageMetadata) (UsersPage, error)
```
manuio: I'd use `pm PageMetadata`

pkg/sdk/go/sdk.go
```diff
@@ -149,7 +162,7 @@ type SDK interface {
 	CreateThings(things []Thing, token string) ([]Thing, error)
 
 	// Things returns page of things.
-	Things(token string, offset, limit uint64, name string) (ThingsPage, error)
+	Things(token string, filter PageMetadata) (ThingsPage, error)
```
manuio: same here

pkg/sdk/go/sdk.go
```diff
@@ -210,7 +223,7 @@ type SDK interface {
 	CreateChannels(channels []Channel, token string) ([]Channel, error)
 
 	// Channels returns page of channels.
-	Channels(token string, offset, limit uint64, name string) (ChannelsPage, error)
+	Channels(token string, filter PageMetadata) (ChannelsPage, error)
```
manuio: and here

pkg/sdk/go/sdk.go
```diff
@@ -349,3 +362,29 @@ func (sdk mfSDK) sendThingRequest(req *http.Request, key, contentType string) (*
 
 	return sdk.client.Do(req)
 }
+
+func (sdk mfSDK) parseFilteredValues(baseURL, endpoint string, filter PageMetadata) (string, error) {
```
manuio: Can you find a better name for this function?

manuio (changes requested):

manuio:
@0x6f736f646f can you check why tests are failling please?

cli/channels.go
```diff
@@ -45,9 +45,15 @@ var cmdChannels = []cobra.Command{
 				logUsage(cmd.Use)
 				return
 			}
+			pageFilters := mfxsdk.PageMetadata{
```
manuio: `pageFilters` could be `pageMetadata`

pkg/sdk/go/users_test.go
```diff
@@ -281,15 +281,15 @@ func TestUsers(t *testing.T) {
 			token:  token,
 			offset: offset,
 			limit:  0,
-			err:    nil,
+			err:    createError(sdk.ErrFailedFetch, http.StatusInternalServerError),
```
manuio: Errors related to request limit are `http.StatusBadRequest` and `ErrFailedFetch` is probably too generic as error.

manuio (changes requested):

pkg/sdk/go/users_test.go
rodneyosodo: That is what is generated by the user service application. We would probably need to change that too.

manuio: Indeed, `ErrLimitSize` is missing here: https://github.com/mainflux/mainflux/blob/master/users/api/transport.go#L288 It should be added for the case returning `w.WriteHeader(http.StatusBadRequest)`. Can you check if there are more errors missing?
About `ErrFailedFetch` maybe no need to change it since it's an SDK error.

drasko (dismissed):
LGTM

drasko:
@dusanb94 please do a review of this one, so we can move forward.

cli/channels.go
```diff
@@ -45,9 +45,15 @@ var cmdChannels = []cobra.Command{
 				logUsage(cmd.Use)
 				return
 			}
+			pageMetadata := mfxsdk.PageMetadata{
+				Name:     "",
```
manuio, line 57: Name is missing here

cli/things.go
```diff
@@ -45,9 +45,14 @@ var cmdThings = []cobra.Command{
 				logUsage(cmd.Use)
 				return
 			}
-
+			pageMetadata := mfxsdk.PageMetadata{
+				Name:     "",
```
manuio, line 56: and here

cli/users.go
```diff
@@ -38,16 +38,30 @@ var cmdUsers = []cobra.Command{
...
 		Run: func(cmd *cobra.Command, args []string) {
-			if len(args) != 1 {
+			if len(args) != 2 {
 				logUsage(cmd.Use)
 				return
 			}
-
-			u, err := sdk.User(args[0])
+			pageMetadata := mfxsdk.PageMetadata{
+				Email:    "",
```
manuio, line 57: Email is missing

pkg/sdk/go/sdk.go
```diff
@@ -349,3 +362,29 @@ func (sdk mfSDK) sendThingRequest(req *http.Request, key, contentType string) (*
 
 	return sdk.client.Do(req)
 }
+
+func (sdk mfSDK) parseQueryParameters(baseURL, endpoint string, pm PageMetadata) (string, error) {
+	var resp map[string]interface{}
+	var queryParams string = ""
```
manuio: `var queryParams string` would be more consistent

manuio (changes requested):

cli/users.go
rodneyosodo: Yes. Because currently it is used to get all users. Our cli doesn't have the capability to get a specific user based on the email

cli/things.go
rodneyosodo: Yes. Similar conversation [here](https://github.com/mainflux/mainflux/pull/1589#discussion_r881846268)

cli/channels.go
rodneyosodo: Yes. Similar conversation [here](https://github.com/mainflux/mainflux/pull/1589#discussion_r881846268)

manuio: Let's open an issue :)

rodneyosodo: Okay. No problem. I will work on another PR to add the changes

rodneyosodo: Implementing with this spec in mind
```sh
mainflux-cli users get [all | email <email_address> | metadata <metadata_json_string> | <user_id> ] <user_auth_token>
```

cli/users.go
drasko: What is this "list all users with e-mail addresses"? Why do we provide e-mail addresses if we are listing all users?

rodneyosodo: No, we aren't doing that. That is why the email field is empty for listing all users. There is a separate command to list a user by providing their email address as outlined [here](https://github.com/mainflux/mainflux/pull/1589#discussion_r885735257).

provision/service.go
```diff
@@ -34,6 +34,12 @@ var (
 	ErrFailedBootstrap          = errors.New("failed to create bootstrap config")
 	ErrFailedBootstrapValidate  = errors.New("failed to validate bootstrap config creation")
 	ErrGatewayUpdate            = errors.New("failed to updated gateway metadata")
+
+	// Limit query parameter
+	limit uint = 10
+	// Offset query parameter
+	offset uint = 0
+	// Name query parameter
```
dborovcanin: What's this comment for?

pkg/sdk/go/users_test.go
```diff
@@ -181,3 +399,129 @@ func TestCreateToken(t *testing.T) {
...
+	}
+
+	mainfluxSDK := sdk.NewSDK(sdkConf)
+	user := sdk.User{Email: "user@example.com", Password: "password"}
+
+	mockAuthzDB := map[string][]mocks.SubjectSet{}
+	mockAuthzDB[user.Email] = append(mockAuthzDB[user.Email], mocks.SubjectSet{Object: "authorities", Relation: "member"})
+	auth := mocks.NewAuthService(map[string]string{user.Email: user.Email}, mockAuthzDB)
+
+	tkn, _ := auth.Issue(context.Background(), &mainflux.IssueReq{Id: user.ID, Email: user.Email, Type: 0})
```
dborovcanin: Instead of literal 0, use key type constants from `auth/keys.go`.

pkg/sdk/go/users_test.go
```diff
@@ -181,3 +399,129 @@ func TestCreateToken(t *testing.T) {
...
```
dborovcanin: Instead of literal 0, use key type constants from `auth/keys.go`.

pkg/sdk/go/users_test.go
```diff
@@ -128,6 +132,220 @@ func TestCreateUser(t *testing.T) {
...
+		offset   uint64
+		limit    uint64
+		err      error
+		response []sdk.User
+		email    string
+		metadata map[string]interface{}
+	}{
+		{
+			desc:   "get a list users",
+			token:  token,
```
dborovcanin: `get a lit of users` or simply `list users`.

pkg/sdk/go/users_test.go
```diff
@@ -128,6 +132,220 @@ func TestCreateUser(t *testing.T) {
...
+	}
+
+	mainfluxSDK := sdk.NewSDK(sdkConf)
+	user := sdk.User{Email: "user@example.com", Password: "password"}
+
+	mockAuthzDB := map[string][]mocks.SubjectSet{}
+	mockAuthzDB[user.Email] = append(mockAuthzDB[user.Email], mocks.SubjectSet{Object: "authorities", Relation: "member"})
+	auth := mocks.NewAuthService(map[string]string{user.Email: user.Email}, mockAuthzDB)
+
+	tkn, _ := auth.Issue(context.Background(), &mainflux.IssueReq{Id: user.ID, Email: user.Email, Type: 0})
```
dborovcanin: Instead of literal 0, use key type constants from `auth/keys.go`.

pkg/sdk/go/users_test.go
```diff
@@ -128,6 +132,220 @@ func TestCreateUser(t *testing.T) {
...
```
dborovcanin: Instead of literal 0, use key type constants from `auth/keys.go`.

pkg/sdk/go/sdk.go
```diff
@@ -353,3 +358,29 @@ func (sdk mfSDK) sendThingRequest(req *http.Request, key, contentType string) (*
...
+		case "total", "offset", "limit", "name", "email":
+			queryParams = fmt.Sprintf("%s%s=%s&", queryParams, key, fmt.Sprint(val))
+		case "metadata":
+			metadataJSON, err := json.Marshal(pm.Metadata)
+			if err != nil {
+				return "", err
+			}
+			queryParams = fmt.Sprintf("%s%s=%s&", queryParams, key, string(metadataJSON))
+		}
+	}
```
dborovcanin: You can attach this to a utility method of PageMetadata that will return URL-encoded query params for the given Page metadata, something like:
```go
func (sdk mfSDK) withQueryParams(baseURL, endpoint string, pm PageMetadata) (string, error) {
    q, err := pm.Query()
    if err != nil {
        return "", err
    }
	return fmt.Sprintf("%s/%s?%s", baseURL, endpoint, q), nil
}
```
where `pm.Query` converts page metadata to query params.

dborovcanin (changes requested):

pkg/sdk/go/sdk.go
```diff
@@ -353,3 +358,37 @@ func (sdk mfSDK) sendThingRequest(req *http.Request, key, contentType string) (*
...
+
+func (sdk mfSDK) withQueryParams(baseURL, endpoint string, pm PageMetadata) (string, error) {
+	q, err := pm.Query()
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%s/%s?%s", baseURL, endpoint, q), nil
+}
+
+func (pm PageMetadata) Query() (string, error) {
```
dborovcanin: Please double-check this implementation. There is a `Values` type from `net/url` that you can use for this purpose, rather than manually constructing a query. It is supposed to provide you a better way of handling query and escaping it correctly. Consider something like:
```go
func (pm PageMetadata) Query() (string, error) {
    var resp map[string]interface{}
    jsonMarshal, err := json.Marshal(pm)
    if err != nil {
    	return "", err
    }
    err = json.Unmarshal(jsonMarshal, &resp)
    if err != nil {
    	return "", err
    }
    q := url.Values{}
    for key, val := range resp {
    	var v string
    	ok := false
    	if key != "metadata" {
    		v, ok = val.(string)
    		if !ok {
    			return "", ErrMetadata
    		}
    	}
    	switch key {
    	case "name", "email":
    		q.Add(key, v)
    	case "total", "offset", "limit":
    		if _, err := strconv.ParseUint(v, 10, 64); err != nil {
    			return "", err
    		}
    		q.Add(key, v)
    	case "metadata":
    		md, err := json.Marshal(pm.Metadata)
    		if err != nil {
    			return "", err
    		}
    		q.Add(key, string(md))
    	}
    }
    return q.Encode(), nil
}
```

dborovcanin (changes requested):

pkg/sdk/go/sdk.go
rodneyosodo: Since in pagemetadata only email and name are string, this raises an error
```
  v, ok = val.(string)
  if !ok {
   return "", error
  }
```
I have changed it to just use `fmt.Sprint(val)` for all values apart from pageMetadata

dborovcanin: `Since in pagemetadata only email and name are string, this raises an error` Makes sense. Since we do not get many benefits from using this marshaling to JSON and back just to remove unused fields while also losing type information, I would propose sticking to the original PageMetadata and simple `if`:
```go
func (pm PageMetadata) query() (string, error) {
	q := url.Values{}
	if pm.Level != 0 {
		q.Add("level", strconv.FormatUint(pm.Level, 10))
	}
	if pm.Total != 0 {
		q.Add("total", strconv.FormatUint(pm.Total, 10))
	}
	if pm.Offset != 0 {
		q.Add("offset", strconv.FormatUint(pm.Offset, 10))
	}
	if pm.Limit != 0 {
		q.Add("limit", strconv.FormatUint(pm.Limit, 10))
	}
	if pm.Level != 0 {
		q.Add("level", strconv.FormatUint(pm.Level, 10))
	}
	if pm.Email != "" {
		q.Add("email", pm.Email)
	}
	if pm.Name != "" {
		q.Add("name", pm.Name)
	}
	if pm.Type != "" {
		q.Add("type", pm.Type)
	}
	if pm.Metadata != nil {
		md, err := json.Marshal(pm.Metadata)
		if err != nil {
			return "", err
		}
		q.Add("metadata", string(md))
	}
	return q.Encode(), nil
}
```

rodneyosodo: We don't retain a value of zero for uint fields. For instance, if we pass in a limit of 0 it should return an `ErrFailedFetch`. I have changed it to allow that.

dborovcanin: @0x6f736f646f Great, even simpler. Only keep this method unexported, we use it only locally.

pkg/sdk/go/users_test.go
```diff
@@ -181,3 +400,129 @@ func TestCreateToken(t *testing.T) {
...
+		{
+			desc:  "update email for user",
+			user:  sdk.User{ID: userID, Email: "user2@example.com", Password: "password"},
+			token: usertoken,
+			err:   nil,
+		},
+		{
+			desc:  "update email for non existing user",
+			user:  sdk.User{ID: "0", Email: "user2@example.com", Password: "password"},
+			token: wrongValue,
```
dborovcanin: This case should have a valid token so that it passes authorization.

pkg/sdk/go/users_test.go
```diff
@@ -128,6 +133,220 @@ func TestCreateUser(t *testing.T) {
...
+	for _, tc := range cases {
+		filter := sdk.PageMetadata{
+			Email:    tc.email,
+			Total:    uint64(200),
+			Offset:   uint64(tc.offset),
+			Limit:    uint64(tc.limit),
+			Metadata: tc.metadata,
+		}
+		_, err := mainfluxSDK.Users(tc.token, filter)
+		assert.Equal(t, tc.err, err, fmt.Sprintf("%s: expected error %s, got %s", tc.desc, tc.err, err))
```
dborovcanin, line 319: Do not only test errors but also test responses, similar to `TestUser`.

dborovcanin (changes requested):
Just update tests we're ready.

pkg/sdk/go/users_test.go
rodneyosodo: For this, we were not checking the response because it doesn't return in the order they were created. For instance, you want to query 5 users, it will return random 5 users i.e 
```
Email:"test-65@example.com"
Email:"test-67@example.com"
Email:"test-75@example.com"
Email:"test-93@example.com"
Email:"test-94@example.com"
```

Instead of users 1 to 5

pkg/sdk/go/users_test.go
rodneyosodo: This is returning nil. I don't think it is correct. Should it return, `ErrFailedFetch`?

pkg/sdk/go/users_test.go
```diff
@@ -128,6 +133,220 @@ func TestCreateUser(t *testing.T) {
...
+			desc:     "get a list of users with same email address and no metadata",
+			token:    token,
+			offset:   0,
+			limit:    100,
+			err:      nil,
+			email:    "mainflux.com",
+			metadata: make(map[string]interface{}),
+		},
+		{
+			desc:   "get a list of users with same email address and metadata",
```
manuio, line 298: ` same email and metadata`

pkg/sdk/go/users_test.go
```diff
@@ -128,6 +133,220 @@ func TestCreateUser(t *testing.T) {
...
+		{
+			desc:   "get a list of users with limit greater than max",
+			token:  token,
+			offset: offset,
+			limit:  110,
+			err:    createError(sdk.ErrFailedFetch, http.StatusBadRequest),
+			email:  email,
+		},
+		{
+			desc:     "get a list of users with same email address and no metadata",
```
manuio: `same email and metadata`

pkg/sdk/go/users_test.go
```diff
@@ -128,6 +133,220 @@ func TestCreateUser(t *testing.T) {
...
+			offset: 0,
+			limit:  100,
+			err:    nil,
+			email:  "",
+			metadata: map[string]interface{}{
+				"name": "demo5",
+			},
+		},
+		{
+			desc:     "get a list of users with same no metadata and email address",
```
manuio: Fix `same no metadata`

pkg/sdk/go/users_test.go
```diff
@@ -128,6 +133,220 @@ func TestCreateUser(t *testing.T) {
...
+			token:    token,
+			offset:   0,
+			limit:    100,
+			err:      nil,
+			email:    "mainflux.com",
```
manuio: This is not an email how test can pass?

pkg/sdk/go/users_test.go
```diff
@@ -181,3 +400,129 @@ func TestCreateToken(t *testing.T) {
...
+	}{
+		{
+			desc:  "update email for user",
+			user:  sdk.User{ID: userID, Email: "user2@example.com", Password: "password"},
+			token: usertoken,
+			err:   nil,
+		},
+		{
+			desc:  "update email for non existing user",
+			user:  sdk.User{ID: "0", Email: "user2@example.com", Password: "password"},
```
manuio: Why ID is `0`?

manuio (changes requested):

pkg/sdk/go/users_test.go
rodneyosodo: Since it is checking for non-existing user. The id doesn't matter as it is the token that is used to id the specific user.

pkg/sdk/go/users_test.go
rodneyosodo: because we were not checking the response

pkg/sdk/go/users_test.go
manuio: oh yes, sorry :)

pkg/sdk/go/users_test.go
dborovcanin: You can use `assert.ElementsMatch()` in that case.

pkg/sdk/go/users_test.go
dborovcanin: It should, which means that the test case was not correct. Since it doesn't work when you fix the case, it also means that either repository mock or service is not working.

dborovcanin (changes requested):

pkg/sdk/go/users_test.go
rodneyosodo: With `cli` it is able to update while with the test it fails. I will look into the mock and check what is the problem

rodneyosodo: If the token is valid, it will return the useridentity which has a valid email address. The valid email address will pass our mock tests. If we pass an invalid token it will fail to pass as you are `Unauthorized`. This means when we are updating the user it doesn't matter the value of userID or userEmail but rather the token to get the actual user as it uses the Email and ID from the token.

dborovcanin: This effectively means that the user can update only its own metadata. If that's the case, please update all the tests (and service and mocks if needed) to match the actual endpoint and what it does.

pkg/sdk/go/users_test.go
dborovcanin: Please don't forget this one.

dborovcanin (changes requested):

dborovcanin (dismissed):

drasko (dismissed):
LGTM

cli/channels.go
```diff
@@ -37,17 +37,28 @@ var cmdChannels = []cobra.Command{
 		},
 	},
 	{
-		Use:   "get [all | <channel_id>] <user_auth_token>",
+		Use:   "get [all | name <channel_name> | metadata <metadata_json_string> | <channel_id>] <user_auth_token>",
```
manuio: I'd use `name` and `metadata` as flags like we did for other filters like `limit` and `offset`

manuio (changes requested):

pkg/sdk/go/things_test.go
```diff
@@ -320,26 +324,36 @@ func TestThings(t *testing.T) {
...
 			err:      nil,
 			response: []sdk.Thing{},
+			metadata: make(map[string]interface{}),
 		},
 	}
 	for _, tc := range cases {
-		page, err := mainfluxSDK.Things(tc.token, tc.offset, tc.limit, tc.name)
+		filter := sdk.PageMetadata{
+			Name:     tc.name,
+			Total:    uint64(200),
```
manuio: Create a const for `200`.

pkg/sdk/go/things_test.go
```diff
@@ -320,26 +324,36 @@ func TestThings(t *testing.T) {
...
+			Offset:   uint64(tc.offset),
```
manuio, line 351: Is this cast needed?

cli/channels.go
dborovcanin: Like `get --name <name_valu>`? This makes sense.

manuio (changes requested):

cli/channels.go
manuio: Yes `--name` and `--metadata`

pkg/sdk/go/things_test.go
rodneyosodo: Yes

cli/channels.go
```diff
@@ -37,24 +37,27 @@ var cmdChannels = []cobra.Command{
 		},
 	},
 	{
-		Use:   "get [all | name <channel_name> | metadata <metadata_json_string> | <channel_id>] <user_auth_token>",
+		Use:   "get [all | <channel_id>] <user_auth_token>",
 		Short: "Get channel",
 		Long: `Get all channels, group by id, group by name or group by metadata.
```
manuio: I'd simplify it: `Get all channels or get channel by id. Channels can be filtered by name or metadata` Keep in mind that with `--help` flag you should see the available flags (filters)

cli/things.go
```diff
@@ -37,23 +37,26 @@ var cmdThings = []cobra.Command{
 		},
 	},
 	{
-		Use:   "get [all | name <thing_name> | metadata <metadata_json_string> | <thing_id>] <user_auth_token>",
+		Use:   "get [all | <thing_id>] <user_auth_token>",
 		Short: "Get things",
 		Long: `Get all things, group by id, group by name or group by metadata.
```
manuio: same here

cli/users.go
```diff
@@ -38,23 +38,26 @@ var cmdUsers = []cobra.Command{
 		},
 	},
 	{
-		Use:   "get [all | email <email_address> | metadata <metadata_json_string> | <user_id> ] <user_auth_token>",
+		Use:   "get [all | <user_id> ] <user_auth_token>",
 		Short: "Get users",
 		Long: `Get all users, group by id, group by email or group by metadata.
```
manuio: same here

manuio (changes requested):

dborovcanin (approved):

drasko (approved):
LGTM

PR #1597, opened 2022-04-26
NOISSUE - Fix Nginx entrypoint script

dborovcanin:

### What does this do?
This pull request fixes the Nginx entrypoint script location mapping.
Also, dependencies are updated.

There is no such issue.

There are no such changes.

N/A

N/A

### Notes
N/A


manuio (dismissed):
LGTM

manuio (dismissed):
LGTM

Makefile
```diff
@@ -111,4 +111,4 @@ rundev:
 	cd scripts && ./run.sh
 
 run:
-	docker-compose -f docker/docker-compose.yml up
+	docker-compose -f docker/docker-compose.yml up -d
```
drasko: Why are we running as a daemon now?

drasko (changes requested):

Makefile
dborovcanin: Because there is no point in blocking the terminal as there are too many services and too many logs to track them from the single output.

manuio (approved):
LGTM

drasko (approved):
LGTM

PR #1598, opened 2022-04-27
MF-1588 - Update Subscriber interface

### LINKED ISSUE #1588: Update Subscriber interface
**ENHANCEMENT**
Currently, the Mainflux Subscriber interface is limited per topic: you can't have multiple subscribers to a single topic. This is OK for Writers but turns out to be limiting for internal use in protocol adapters (see #1572, #1583). Consider adding subscriber ID alongside with topic for Subscribe interface to overcome this limit and use internal implementation instead of implementing custom logic per adapter.

This is a should-have enhancement.

rodneyosodo:

### What does this do?
This PR updated the subscriber interface. It does so by adding an id field to the subscriber. This is to help us have multiple subscriptions per topic instead of only one.

Resolves #1588.

It adds an id field to the subscriber interface methods.

Yes.
No.

### Notes
N/A


pkg/messaging/nats/pubsub.go
```diff
@@ -81,13 +87,22 @@ func (ps *pubsub) Publish(topic string, msg messaging.Message) error {
...
-	if _, ok := ps.subscriptions[topic]; ok {
+	// If there are no subscription, create an empty map assigned to the topic.
+	s, ok := ps.subscriptions[topic]
+	if !ok {
+		s = make(map[string]subscription)
+	}
+	// If subscription exists return already subscribed.
+	if _, ok := s[id]; ok {
 		return errAlreadySubscribed
 	}
```
dborovcanin: Consider raplacing:
```go
switch ok {
    case true:
        if _, ok := s[id]; ok {
            return errAlreadySubscribed
	}
    default:
        s = make(map[string]subscription)
        ps.subscriptions[topic] = s

}
```
This way you check for `errAlreadySubscribed` only if the first map exist. The same effect as:
```go
if ok {
    if _, ok := s[id]; ok {
        return errAlreadySubscribed
    }
} else {
    s = make(map[string]subscription)
        ps.subscriptions[topic] = s
}
```
Please note that here `ps.subscriptions[topic] = s` is sufficient and there is no need to assign it on L129 (or anywhere else) since maps in Go are reference types (passed around as pointers).

pkg/messaging/nats/pubsub.go
```diff
@@ -97,34 +112,59 @@ func (ps *pubsub) Subscribe(topic string, handler messaging.MessageHandler) erro
...
+		}
+		if err := current.Unsubscribe(); err != nil {
+			return err
+		}
+	}
+	delete(s, id)
+	if len(s) == 0 {
+		delete(ps.subscriptions, topic)
+	} else {
+		ps.subscriptions[topic] = s
```
dborovcanin: This is surplus.

pkg/messaging/pubsub.go
```diff
@@ -10,16 +10,22 @@ type Publisher interface {
 }
 
 // MessageHandler represents Message handler for Subscriber.
-type MessageHandler func(msg Message) error
+type MessageHandler interface {
+	// Handle handles messages passed by underlying implementation.
+	Handle(msg Message) error
+
+	// Cancel is used for cleanup after unsubscribing and it's optional.
```
dborovcanin: Technically, it is not used **after** unsubscribing, but during the unsubscribing process. It is used **before** calling internal unsubscribe, so maybe rephrase to something like:

>  Cancel is used for cleanup during unsubscribing and it's optional.

cmd/coap/main.go
```diff
@@ -77,13 +77,14 @@ func main() {
 
 	tc := thingsapi.NewClient(conn, thingsTracer, cfg.thingsAuthTimeout)
 
-	nc, err := broker.Connect(cfg.natsURL)
+	nps, err := nats.NewPubSub(cfg.natsURL, "coap", logger)
```
dborovcanin: The queue needs to be empty here. Otherwise, out of all clients subscribed to a single subject (topic), only one from the queue will receive the message.

pkg/messaging/nats/pubsub.go
```diff
@@ -81,13 +87,22 @@ func (ps *pubsub) Publish(topic string, msg messaging.Message) error {
 	return nil
 }
 
-func (ps *pubsub) Subscribe(topic string, handler messaging.MessageHandler) error {
+func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) error {
+	if id == "" {
+		return errEmptyID
+	}
```
dborovcanin, line 93: This is not mandatory. An empty string ID is still a unique ID. However, I'd like to keep it anyways to enforce non-empty strings. Also, for other protocols such as MQTT - client IDs are actually required.

pkg/messaging/mqtt/subscriber.go
```diff
@@ -18,14 +18,19 @@ import (
...
 var _ messaging.Subscriber = (*subscriber)(nil)
 
 type subscriber struct {
-	client  mqtt.Client
-	timeout time.Duration
-	logger  log.Logger
+	client        mqtt.Client
+	timeout       time.Duration
+	logger        log.Logger
+	subscriptions map[string]map[string]mqtt.Token
```
dborovcanin: Unlike the CoAP token that identifies a session, `matt.Token` is, per documentation:
> Token defines the interface for the tokens used to indicate when actions have completed.

That is, the token is an artificial implementation of async operations within the MQTT library we use. However, if we want to have a correct implementation, we'll need to have an MQTT client (form the Paho lib) per ID (instead of mqtt.Token) - because MQTT client ID is unique per MQTT 3.1.1 spec.
Something like: `map[string]mqtt.Client` Where keys are client IDs, and values are MQTT Clients. Those MQTT clients will take care of its subscriptions per topic.

dborovcanin (changes requested):

pkg/messaging/nats/pubsub.go
```diff
@@ -97,34 +115,58 @@ func (ps *pubsub) Subscribe(topic string, handler messaging.MessageHandler) erro
...
+			}
+		default:
+			return errNotSubscribed
+		}
+	default:
 		return errNotSubscribed
 	}
-
-	if err := sub.Unsubscribe(); err != nil {
-		return err
```
dborovcanin: Simplify with:
```go
subs, ok := ps.subscriptions[topic]
if !ok {
	return errNotSubscribed
}
sub, ok := subs[id]
if !ok {
	return errNotSubscribed
}
if sub.cancel != nil {
	if err := sub.cancel(); err != nil {
		return err
	}
}
if err := sub.Unsubscribe(); err != nil {
	return err
}
```

pkg/messaging/mqtt/subscriber.go
```diff
@@ -18,55 +18,99 @@ import (
...
+	}
+	// Check topic
+	s, ok := sub.subscriptions[topic]
+	switch ok {
+	case true:
+		// Check client ID
+		if _, ok := s[id]; ok {
+			return errAlreadySubscribed
+		}
+	default:
```
dborovcanin, line 73: There is no need to create a new client for each topic, but only for each client ID. We have to store topics for each client anyway (because the library does not provide us to access to all the subscribed topics for a single client) so that we know when to close the client (when we empty all the topics the client is subscribed to), we can do something like this:
```go

type subscription struct {
	client mqtt.Client
	topics []string
}

// NewSubscriber returns a new MQTT message subscriber.
func NewSubscriber(address string, timeout time.Duration, logger log.Logger) (messaging.Subscriber, error) {
	ret := subscriber{
		address:       address,
		timeout:       timeout,
		logger:        logger,
		subscriptions: make(map[string]subscription),
	}
	return ret, nil
}
```

dborovcanin (changes requested):

pkg/messaging/mqtt/subscriber.go
rodneyosodo: Have a look at the implementation. With topics being a map, we would have a simpler implementation and we wouldn't need to write the `contains` and `delete` functions

pkg/messaging/mqtt/subscriber.go
```diff
@@ -16,21 +16,58 @@ import (
...
+}
+
+// contains checks if a topic is present
+func (sub subscription) contains(topic string) bool {
+	for _, v := range sub.topics {
+		if v == topic {
+			return true
+		}
+	}
+	return false
```
dborovcanin: You can replace this implementation with simple:
```go
return sub.indexOf(topic) != -1
```

pkg/messaging/mqtt/subscriber.go
```diff
@@ -16,21 +16,58 @@ import (
...
+	sub.topics[len(sub.topics)-1] = ""
+	return true
+}
+
 type subscriber struct {
 	address       string
 	timeout       time.Duration
 	logger        log.Logger
-	subscriptions map[string]map[string]mqtt.Client
+	subscriptions map[string]subscription
```
dborovcanin, line 40: This map needs to be protected for concurrent use.

pkg/messaging/mqtt/subscriber.go
```diff
@@ -51,22 +88,26 @@ func (sub subscriber) Subscribe(id, topic string, handler messaging.MessageHandl
...
 	}
-	opts := mqtt.NewClientOptions().SetUsername(username).AddBroker(sub.address)
-	client := mqtt.NewClient(opts)
-	token := client.Connect()
-	if token.Error() != nil {
-		return token.Error()
+	// Check topic
+	ok = s.contains(topic)
+	if ok {
+		return errAlreadySubscribed
```
dborovcanin: If the previous block (`if !ok {....}`) executes, this condition will be satisfied and we'll return an error. We actually want to continue with subscribing.

pkg/messaging/mqtt/subscriber.go
```diff
@@ -83,19 +124,18 @@ func (sub subscriber) Unsubscribe(id, topic string) error {
...
 	case true:
-		// Check client ID
-		if _, ok := s[id]; !ok {
+		// Check topic
+		if ok := s.contains(topic); !ok {
 			return errNotSubscribed
 		}
 	default:
 		return errNotSubscribed
 	}
```
dborovcanin: You can replace this entire `switch` with an `if`:
```go
if !ok || (ok && !s.contains(topic)) {
    return errNotSubscribed
}
```

pkg/messaging/mqtt/subscriber.go
```diff
@@ -16,21 +16,58 @@ import (
...
+}
+
+// Deletes a topic from the slice
+func (sub subscription) delete(topic string) bool {
+	index := sub.indexOf(topic)
+	if index == -1 {
+		return false
+	}
+	copy(sub.topics[index:], sub.topics[index+1:])
+	sub.topics[len(sub.topics)-1] = ""
```
dborovcanin: Probably the simplest way to handle slice growth is to actually copy it. Something like:
```go
topics := make([]string, len(sub.topics)-1)
copy(topics[:index], sub.topics[:index])
copy(topics[index:], sub.topics[index+1:])
sub.topics = topics
```

dborovcanin (changes requested):

pkg/messaging/mqtt/subscriber.go
```diff
@@ -16,57 +17,142 @@ import (
...
+	return true
+}
```
dborovcanin: Move unexported `subscription` struct methods to the bottom of the file.

pkg/messaging/mqtt/subscriber.go
dborovcanin: Please double-check this implementation - since you're only copying, it looks like it does not actually free any space. You have an example here: https://github.com/mainflux/mainflux/pull/1598/files#r863116725.
It is especially dangerous because this check will, eventually, never be satisfied: https://github.com/mainflux/mainflux/pull/1598/files#diff-3740ddbdbe88f161342d4cb97fb90d933e7ac87450cb30dcb246002912230e95R153.

pkg/messaging/mqtt/subscriber.go
```diff
@@ -16,57 +17,142 @@ import (
...
+	s, ok := sub.subscriptions[id]
+	switch ok {
+	case true:
+		// Check topic
+		if ok = s.contains(topic); ok {
+			return errAlreadySubscribed
+		}
+		s.topics = append(s.topics, topic)
+	default:
+		opts := mqtt.NewClientOptions().SetUsername(username).AddBroker(sub.address)
```
dborovcanin: You forgot to set the client ID.

cmd/twins/main.go
```diff
@@ -301,13 +311,17 @@ func newService(ps messaging.PubSub, chanID string, users mainflux.AuthServiceCl
...
-	return svc
+type handlerFunc func(msg messaging.Message) error
+
+func (h handlerFunc) Handle(msg messaging.Message) error {
+	return h(msg)
+}
+
+func (h handlerFunc) Cancel() error {
+	return nil
 }
```
dborovcanin: Move this to the bottom of the file.

dborovcanin (changes requested):
We're very close.

dborovcanin (approved):

drasko (approved):
LGTM

PR #1605, opened 2022-05-10
Login with externaly issued token

mteodor:

This accepts authorization token issued outside the mainflux.
example flow in UI
user tries to access UI, UI is behind auth proxy, gets redirected to authentication server
after successful authentication auth proxy sends token with each request 
UI first tries to access `tokens` endpoint , checks for `Authorization` header and if it is there extracts user data from it and creates user if it is not missing, if successful just returns token unchanged

yes


drasko:
Closing this one, as functionality will be added via implementation of this issue: https://github.com/mainflux/mainflux/issues/1887

PR #1606, opened 2022-05-10
NOISSUE - Make application/json content-type valid in http-adapter

manuio:

### What does this do?
Make application/json content-type valid in http-adapter

N/A
yes
no


mteodor (approved):

mteodor (dismissed):

drasko:
Why is this PR needed? Because of generic JSON writer?

@dusanb94 please comment on this PR, this looks strange to me. It is impractical to enter one by one content type, IMHO Mainflux is for now enabled only with SenML protocol knowledge.

manuio:
> Why is this PR needed? Because of generic JSON writer?
> 
> @dusanb94 please comment on this PR, this looks strange to me. It is impractical to enter one by one content type, IMHO Mainflux is for now enabled only with SenML protocol knowledge.

@drasko Yes it'sbecause of the generic JSON writer. Currently the http-adapter only accept `application/senml+json`If you want to send a payload with pure JSON you have to use the same content-type in the header and it's not really consistent. This PR only make both types valid. BTW, I should maybe add `application/senml+cbor`. Or should I remove content-type check?

dborovcanin (dismissed):

dborovcanin:
> > Why is this PR needed? Because of generic JSON writer?
> > @dusanb94 please comment on this PR, this looks strange to me. It is impractical to enter one by one content type, IMHO Mainflux is for now enabled only with SenML protocol knowledge.
> 
> @drasko Yes it'sbecause of the generic JSON writer. Currently the http-adapter only accept `application/senml+json`If you want to send a payload with pure JSON you have to use the same content-type in the header and it's not really consistent. This PR only make both types valid. BTW, I should maybe add `application/senml+cbor`. Or should I remove content-type check?

TBH, adding CBOR CT also makes sense, since we support, well, all three.

http/api/endpoint_test.go
```diff
@@ -83,36 +87,50 @@ func TestPublish(t *testing.T) {
...
+			msg:         msgJSON,
+			contentType: contentTypeJSON,
 			key:         thingKey,
 			status:      http.StatusAccepted,
 		},
 		"publish message with empty key": {
 			chanID:      chanID,
 			msg:         msg,
-			contentType: contentType,
+			contentType: contentTypeSenML,
```
drasko: Why only SenML, what about the other 3 CTs we support?

http/api/transport.go
```diff
@@ -176,7 +180,7 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 	}
 
 	if errorVal, ok := err.(errors.Error); ok {
-		w.Header().Set("Content-Type", contentType)
+		w.Header().Set("Content-Type", contentTypeJSON)
```
drasko: Why only JSON, what about other 3 CTs we support?

http/api/endpoint_test.go
```diff
@@ -63,10 +63,14 @@ func (tr testRequest) make() (*http.Response, error) {
 
 func TestPublish(t *testing.T) {
 	chanID := "1"
-	contentType := "application/senml+json"
+	contentTypeSenML := "application/senml+json"
+	contentTypeCBOR := "application/senml+cbor"
```
drasko: Looks like naming is not consistent. If we call upper SenML (should be maybe `SenMLJSON`), this one is also SenML (just that it is CBOR encoded). So in the upper we use semantic representation (SenML) and in the second we use encoding/serialization (although semantic structures are still the same).

drasko (changes requested):

http/api/endpoint_test.go
manuio: There are already tests for CTs validation and it's pointless to test 3 CTs for `empty key`

http/api/transport.go
manuio: Because it's the CT of the error response

http/api/transport.go
```diff
@@ -96,7 +98,9 @@ func parseSubtopic(subtopic string) (string, error) {
 }
 
 func decodeRequest(ctx context.Context, r *http.Request) (interface{}, error) {
-	if !strings.Contains(r.Header.Get("Content-Type"), contentType) {
+	if !strings.Contains(r.Header.Get("Content-Type"), contentTypeSenMLJSON) &&
+		!strings.Contains(r.Header.Get("Content-Type"), contentTypeSenMLCBOR) &&
+		!strings.Contains(r.Header.Get("Content-Type"), contentTypeJSON) {
```
dborovcanin: Extract to a variable `ct := r.Header.Get("Content-Type")`. Also, why don't we compare with `==` rather than using `strings.Contains`?

http/api/transport.go
```diff
@@ -29,8 +29,10 @@ import (
 )
 
 const (
-	protocol    = "http"
-	contentType = "application/senml+json"
+	protocol         = "http"
+	contentTypeSenMLJSON = "application/senml+json"
+	contentTypeSenMLCBOR  = "application/senml+cbor"
+	contentTypeJSON  = "application/json"
```
dborovcanin: Consider shortening to `ctSenmlJSON`, `ctSenmlCBOR`, `ctJSON` (I know it's `SenML` and not `Senml`, but this looks more natural; and since it's a local variable used only once, `ct` instead of `contentType` sounds acceptable).

pkg/sdk/go/message.go
```diff
@@ -21,7 +21,7 @@ func (sdk mfSDK) SendMessage(chanName, msg, key string) error {
 		subtopicPart = fmt.Sprintf("/%s", strings.Replace(chanNameParts[1], ".", "/", -1))
 	}
 
-	url := fmt.Sprintf("%s/channels/%s/messages%s", sdk.httpAdapterURL, chanID, subtopicPart)
+	url := fmt.Sprintf("%s/channels/%s/messages/json%s", sdk.httpAdapterURL, chanID, subtopicPart)
```
dborovcanin: Why did you add this subtopic?

dborovcanin (changes requested):
Please remove the binary file from `cmd/cli`.

dborovcanin (dismissed):

dborovcanin:
@manuio Please remove the binary from `/cmd/cli`.

drasko (approved):
LGTM

dborovcanin (approved):

PR #1620, opened 2022-06-22
MF1621 - Logical user removal

### LINKED ISSUE #1621: Add logical user removal for Mainflux users.
**FEATURE REQUEST**

   the existing issue, otherwise proceed to step 2.

Add logical user removal for Mainflux users.

This means adding a flag for the user to determine whether it's active or not.
The flag is binary (true or false for two user states) but in the future, we might add different states.

The deactivated user can not log in.
When listing users, Admin can have a query parameter provided filter only inactive, only active, or all the users. By default, only active users are listed.

Should have

_In the future, all the policies attached to user should be removed as well, and reverting/activating user should be supported._

rodneyosodo:

### What does this do?
Remove the user logically. I have added a new field called active so when we remove the user we set the value to be false. Listing users, you are also given the opportunity to list removed used or active users

Resolves #1621

N/A

Not yet

### Notes
N/A


users/tracing/users.go
```diff
@@ -75,12 +75,20 @@ func (urm userRepositoryMiddleware) UpdatePassword(ctx context.Context, email, p
 	return urm.repo.UpdatePassword(ctx, email, password)
 }
 
-func (urm userRepositoryMiddleware) RetrieveAll(ctx context.Context, offset, limit uint64, ids []string, email string, um users.Metadata) (users.UserPage, error) {
+func (urm userRepositoryMiddleware) RetrieveAll(ctx context.Context, active bool, offset, limit uint64, ids []string, email string, um users.Metadata) (users.UserPage, error) {
```
dborovcanin: I'm afraid bool is not enough to represent all the states: `active`, `inactive`, `all`.

users/users.go
```diff
@@ -65,10 +66,13 @@ type UserRepository interface {
...
 
 	// RetrieveAll retrieves all users for given array of userIDs.
-	RetrieveAll(ctx context.Context, offset, limit uint64, userIDs []string, email string, m Metadata) (UserPage, error)
+	RetrieveAll(ctx context.Context, active bool, offset, limit uint64, userIDs []string, email string, m Metadata) (UserPage, error)
 
 	// UpdatePassword updates password for user with given email
 	UpdatePassword(ctx context.Context, email, password string) error
+
+	// Remove logically removes the user identified with the provided ID
+	Remove(ctx context.Context, u User) error
```
dborovcanin: Maybe `Deactivate` is more suitable here.

dborovcanin (changes requested):

pkg/sdk/go/sdk.go
```diff
@@ -153,6 +153,9 @@ type SDK interface {
 	// UpdatePassword updates user password.
 	UpdatePassword(oldPass, newPass, token string) error
 
+	// DeleteUser removes existing thing.
```
manuio: fix `existing thing`

pkg/sdk/go/sdk.go
```diff
@@ -153,6 +153,9 @@ type SDK interface {
...
+	DeleteUser(id, token string) error
```
manuio: Consider calling this method `DisableUser`

manuio (changes requested):

pkg/sdk/go/sdk.go
rodneyosodo: I will change to `Deactivate` and `DeactivateUser` as indicated [here](https://github.com/mainflux/mainflux/pull/1620#discussion_r903986811)

users/tracing/users.go
rodneyosodo: Changed to using string

users/users.go
```diff
@@ -39,6 +39,7 @@ type User struct {
 	Email    string
 	Password string
 	Metadata Metadata
+	Active   string
```
dborovcanin: I would suggest using `State` instead of `Active` here, with constraints (allowed values, case sensitivity) checked in the code. I would keep plain VARCHAR in the DB due to simplicity and readability (over enumeration and integer types).
Replace `active` with `state` at other places as well.

users/postgres/init.go
```diff
@@ -78,6 +78,13 @@ func migrateDB(db *sqlx.DB) error {
 					`ALTER TABLE IF EXISTS users ADD PRIMARY KEY (id)`,
 				},
 			},
+			{
+				Id: "users_5",
+				Up: []string{
+					`ALTER TABLE IF EXISTS users ADD COLUMN IF NOT EXISTS
+					 active CHAR(10) NOT NULL DEFAULT 'active'`,
```
dborovcanin: Check if CHAR is the most suitable type here.

dborovcanin (changes requested):

users/postgres/users.go
```diff
@@ -38,7 +38,10 @@ func NewUserRepo(db Database) users.UserRepository {
 }
 
 func (ur userRepository) Save(ctx context.Context, user users.User) (string, error) {
-	q := `INSERT INTO users (email, password, id, metadata) VALUES (:email, :password, :id, :metadata) RETURNING id`
+	if user.State == "" {
+		user.State = "active"
+	}
```
dborovcanin: I'd prefer not to set default here but on the service layer with the rest of the service logic.

users/postgres/init.go
```diff
@@ -78,6 +78,13 @@ func migrateDB(db *sqlx.DB) error {
 					`ALTER TABLE IF EXISTS users ADD PRIMARY KEY (id)`,
 				},
 			},
+			{
+				Id: "users_5",
+				Up: []string{
+					`ALTER TABLE IF EXISTS users ADD COLUMN IF NOT EXISTS
+					 state VARCHAR(10) NOT NULL DEFAULT 'active'`,
```
dborovcanin: Add a little more flexible size limit (64 should be plenty enough)

cli/users.go
```diff
@@ -58,6 +58,7 @@ var cmdUsers = []cobra.Command{
 				Offset:   uint64(Offset),
 				Limit:    uint64(Limit),
 				Metadata: metadata,
+				State:    State,
```
dborovcanin: Consider renaming `state` with `status`.

internal/apiutil/errors.go
```diff
@@ -30,6 +30,9 @@ var (
 	// ErrEmailSize indicates that email size exceeds the max.
 	ErrEmailSize = errors.New("invalid email size")
 
+	// ErrActiveState indicates an invalid activity state.
+	ErrActiveState = errors.New("invalid activity state")
```
dborovcanin: Just `ErrInvalidStatus` here.

dborovcanin (changes requested):
Consider enriching users' tests with a new user with an `inactive` status to see if tests pass.

users/mocks/users.go
```diff
@@ -110,6 +110,21 @@ func (urm *userRepositoryMock) RetrieveAll(ctx context.Context, state string, of
 		return up, nil
 	}
 
+	if state == "active" || state == "inactive" {
+
```
dborovcanin: Please remove an empty line.

dborovcanin: There should be no need to sort users, but you can use `assert.ElementsMatch`.

users/service_test.go
```diff
@@ -390,3 +390,100 @@ func TestSendPasswordReset(t *testing.T) {
 
 	}
 }
+
+func TestDeactivateUser(t *testing.T) {
```
dborovcanin: Add a case for logging in as a deactivated user.

pkg/sdk/go/sdk.go
```diff
@@ -153,6 +154,9 @@ type SDK interface {
 	// UpdatePassword updates user password.
 	UpdatePassword(oldPass, newPass, token string) error
 
+	// DeleteUser deactivates the user.
+	DeactivateUser(id, token string) error
```
dborovcanin: Consider renaming to something more generic like `ChangeUserStatus`.

dborovcanin (changes requested):

users/mocks/users.go
rodneyosodo: The `sortUsers` is needed when getting querying a list of users from allusers

dborovcanin (changes requested):
Rename `state` to `status`. It is a little less ambiguous.

cli/users.go
```diff
@@ -140,6 +141,24 @@ var cmdUsers = []cobra.Command{
 				return
 			}
 
+			logOK()
+		},
+	},
+	{
+		Use:   "deactivate <user_id> <user_auth_token>",
```
manuio: I'd use `disable` instead of `deactivate`

users/api/metrics.go
```diff
@@ -120,11 +120,20 @@ func (ms *metricsMiddleware) SendPasswordReset(ctx context.Context, host, email,
...
 		ms.latency.With("method", "list_members").Observe(time.Since(begin).Seconds())
 	}(time.Now())
 
-	return ms.svc.ListMembers(ctx, token, groupID, offset, limit, gm)
+	return ms.svc.ListMembers(ctx, token, groupID, state, offset, limit, gm)
+}
+
+func (ms *metricsMiddleware) ChangeUserStatus(ctx context.Context, token string, id string) (err error) {
+	defer func(begin time.Time) {
+		ms.counter.With("method", "remove_user").Add(1)
```
manuio: This should be `change_user_status` but maybe better to find another naming. Why not `DisableUser` and `EnableUser`?

users/api/transport.go
```diff
@@ -152,8 +160,13 @@ func decodeListUsers(_ context.Context, r *http.Request) (interface{}, error) {
 		return nil, err
 	}
 
+	s, err := apiutil.ReadStringQuery(r, stateKey, "active")
```
manuio: create a const for `active`

users/api/transport.go
```diff
@@ -258,9 +271,14 @@ func decodeListMembersRequest(_ context.Context, r *http.Request) (interface{},
 	if err != nil {
 		return nil, err
 	}
+	s, err := apiutil.ReadStringQuery(r, stateKey, "active")
```
manuio: use it here

users/api/transport.go
```diff
@@ -110,6 +111,13 @@ func MakeHandler(svc users.Service, tracer opentracing.Tracer, logger logger.Log
 		opts...,
 	))
 
+	mux.Delete("/users/:userID", kithttp.NewServer(
```
manuio: `"/users/:id"`

users/api/transport.go
```diff
@@ -258,9 +271,14 @@ func decodeListMembersRequest(_ context.Context, r *http.Request) (interface{},
...
 	}
+	s, err := apiutil.ReadStringQuery(r, stateKey, "active")
+	if err != nil {
+		return nil, err
+	}
 
 	req := listMemberGroupReq{
 		token:    apiutil.ExtractBearerToken(r),
+		state:    s,
 		groupID:  bone.GetValue(r, "groupId"),
```
manuio: You could also fix this and use `id`

users/api/transport.go
```diff
@@ -269,6 +287,15 @@ func decodeListMembersRequest(_ context.Context, r *http.Request) (interface{},
 	return req, nil
 }
 
+func decodeChangeUserStatus(_ context.Context, r *http.Request) (interface{}, error) {
+	req := changeUserStatusReq{
+		token:  apiutil.ExtractBearerToken(r),
+		userID: bone.GetValue(r, "userID"),
```
manuio: `id`

users/mocks/users.go
```diff
@@ -110,6 +110,20 @@ func (urm *userRepositoryMock) RetrieveAll(ctx context.Context, offset, limit ui
 		return up, nil
 	}
 
+	if state == "active" || state == "inactive" {
```
manuio: Probably better to create consts for `active` `inactive`

users/service.go
```diff
@@ -53,7 +53,7 @@ type Service interface {
 	ViewProfile(ctx context.Context, token string) (User, error)
 
 	// ListUsers retrieves users list for a valid admin token.
-	ListUsers(ctx context.Context, token string, offset, limit uint64, email string, meta Metadata) (UserPage, error)
+	ListUsers(ctx context.Context, token string, state string, offset, limit uint64, email string, meta Metadata) (UserPage, error)
```
manuio: `state string, offset, limit uint64, email string, meta Metadat` should be passed as a `PageMetadata` structure

users/service.go
```diff
@@ -73,7 +73,10 @@ type Service interface {
 	SendPasswordReset(ctx context.Context, host, email, token string) error
 
 	// ListMembers retrieves everything that is assigned to a group identified by groupID.
-	ListMembers(ctx context.Context, token, groupID string, offset, limit uint64, meta Metadata) (UserPage, error)
+	ListMembers(ctx context.Context, token, groupID string, state string, offset, limit uint64, meta Metadata) (UserPage, error)
```
manuio: Same here

manuio (changes requested):

users/api/metrics.go
dborovcanin: @0x6f736f646f Since we will have only these two states, and it's unlikely that will change in foreseeable future, the @manuio proposal sounds OK. It's a cleaner service interface and we can utilize the same repository implementation.

cli/users.go
drasko: If the user's state is `active`, and not `enabled`, I'd leave `deactivate`.

manuio: I'd change `active` by `enabled`. `active` could refer to the user activity in the platform.

cli/users.go
```diff
@@ -140,6 +141,42 @@ var cmdUsers = []cobra.Command{
 				return
 			}
 
+			logOK()
+		},
+	},
+	{
+		Use:   "enable <user_id> <user_auth_token>",
+		Short: "Change user status to active",
```
manuio: `Change user status to enabled` ?

cli/users.go
```diff
@@ -140,6 +141,42 @@ var cmdUsers = []cobra.Command{
...
+				logError(err)
+				return
+			}
+
+			logOK()
+		},
+	},
+	{
+		Use:   "disable <user_id> <user_auth_token>",
+		Short: "Change user status to inactive",
```
manuio: same here

internal/apiutil/errors.go
```diff
@@ -30,6 +30,9 @@ var (
 	// ErrEmailSize indicates that email size exceeds the max.
 	ErrEmailSize = errors.New("invalid email size")
 
+	// ErrInvalidStatus indicates an invalid activity status.
+	ErrInvalidStatus = errors.New("invalid activity status")
```
manuio: Maybe better `invalid user account status`?

users/api/endpoint.go
```diff
@@ -118,7 +118,7 @@ func listUsersEndpoint(svc users.Service) endpoint.Endpoint {
 		if err := req.validate(); err != nil {
 			return users.UserPage{}, err
 		}
-		up, err := svc.ListUsers(ctx, req.token, req.offset, req.limit, req.email, req.metadata)
+		up, err := svc.ListUsers(ctx, req.token, req.state, req.offset, req.limit, req.email, req.metadata)
```
manuio: I'd use ` req.status`?

users/api/logging.go
```diff
@@ -79,7 +79,7 @@ func (lm *loggingMiddleware) ViewProfile(ctx context.Context, token string) (u u
 	return lm.svc.ViewProfile(ctx, token)
 }
 
-func (lm *loggingMiddleware) ListUsers(ctx context.Context, token string, offset, limit uint64, email string, um users.Metadata) (e users.UserPage, err error) {
+func (lm *loggingMiddleware) ListUsers(ctx context.Context, token string, state string, offset, limit uint64, email string, um users.Metadata) (e users.UserPage, err error) {
```
manuio: `status string`

users/api/logging.go
```diff
@@ -157,7 +157,7 @@ func (lm *loggingMiddleware) SendPasswordReset(ctx context.Context, host, email,
 	return lm.svc.SendPasswordReset(ctx, host, email, token)
 }
 
-func (lm *loggingMiddleware) ListMembers(ctx context.Context, token, groupID string, offset, limit uint64, m users.Metadata) (mp users.UserPage, err error) {
+func (lm *loggingMiddleware) ListMembers(ctx context.Context, token, groupID string, state string, offset, limit uint64, m users.Metadata) (mp users.UserPage, err error) {
```
manuio: `status`

users/api/requests.go
```diff
@@ -139,20 +143,38 @@ func (req passwChangeReq) validate() error {
 
 type listMemberGroupReq struct {
 	token    string
+	state    string
```
manuio: `status`

manuio (changes requested):

users/api/requests.go
```diff
@@ -44,6 +44,7 @@ func (req viewUserReq) validate() error {
 
 type listUsersReq struct {
 	token    string
+	state    string
```
manuio: `status`

users/api/requests.go
```diff
@@ -62,6 +63,9 @@ func (req listUsersReq) validate() error {
 	if len(req.email) > maxEmailSize {
 		return apiutil.ErrEmailSize
 	}
+	if req.state != "all" && req.state != "active" && req.state != "inactive" {
```
manuio: Create consts for these status.  Maybe even change them by `enabled and `disabled`?

manuio (changes requested):

users/service.go
```diff
@@ -16,8 +16,9 @@ const (
 	memberRelationKey = "member"
 	authoritiesObjKey = "authorities"
 	usersObjKey       = "users"
-	active            = "active"
-	inactive          = "inactive"
+	EnableStatusKey   = "enabled"
```
manuio: `EnabledStatusKey`

users/service.go
```diff
@@ -16,8 +16,9 @@ const (
...
+	DisableStatusKey  = "disabled"
```
manuio: `DisabledSatusKey`

manuio (changes requested):

cli/users.go
drasko: Please leave "active"

Users cannot be enabled, they are not devices. Users are active/inactive and/or deactivated. Please keep this terminology, not enabled/disabled.

users/service.go
drasko: Users are active/inactive, not enabled.

users/service.go
```diff
@@ -31,6 +33,12 @@ var (
 
 	// ErrPasswordFormat indicates weak password.
 	ErrPasswordFormat = errors.New("password does not meet the requirements")
+
+	// ErrAlreadyEnabledUser indicates the user is enabled.
+	ErrAlreadyEnabledUser = errors.New("the user is already enabled")
```
drasko, line 40: Please do not use `enabled` terminology here, stick to `activated`. @dborovcanin please make sure the terminology is consistent, as I can see that there is some confusion which might lead to incosistence.

drasko (changes requested):

cli/users.go
manuio: Why a user can't be disabled? For example, Azure is using this therminology here: https://docs.microsoft.com/en-us/azure/active-directory/authentication/howto-mfa-userstates#azure-ad-multi-factor-authentication-user-states and there are many others...

Inactive can refer to a user not using the platform for a long time, right? In any case I agree to chose a naming and to be consistent in all the code.

users/service.go
manuio: @drasko I proposed this naming 3 days ago and @dborovcanin agreed. If you prefer active/inactive then @0x6f736f646f shoudl rename all

drasko: Active/inactive/deactivated.

users/tracing/users.go
mteodor: wouldnt enum be more suitable

users/api/endpoint.go
```diff
@@ -188,6 +188,20 @@ func listMembersEndpoint(svc users.Service) endpoint.Endpoint {
 	}
 }
 
+func removeUserEndpoint(svc users.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(removeUserReq)
+		if err := req.validate(); err != nil {
+			return nil, err
+		}
+		err := svc.RemoveUser(ctx, req.token, req.userID)
```
mteodor: you can inline error check

users/service.go
```diff
@@ -328,7 +332,21 @@ func (svc usersService) ListMembers(ctx context.Context, token, groupID string,
...
+
+func (svc usersService) RemoveUser(ctx context.Context, token, id string) error {
+	_, err := svc.identify(ctx, token)
+	if err != nil {
+		return err
+	}
+
+	dbUser, err := svc.users.RetrieveByID(ctx, id)
+	if err != nil {
+		return errors.Wrap(errors.ErrAuthentication, err)
```
mteodor: is this ok to be authentication error?

users/users.go
```diff
@@ -65,10 +66,13 @@ type UserRepository interface {
...
 
 	// RetrieveAll retrieves all users for given array of userIDs.
-	RetrieveAll(ctx context.Context, offset, limit uint64, userIDs []string, email string, m Metadata) (UserPage, error)
+	RetrieveAll(ctx context.Context, status string, offset, limit uint64, userIDs []string, email string, m Metadata) (UserPage, error)
 
 	// UpdatePassword updates password for user with given email
 	UpdatePassword(ctx context.Context, email, password string) error
+
+	// ChangeStatus changes users status to enabled or disabled
+	ChangeStatus(ctx context.Context, u User, status string) error
```
dborovcanin: While this is practically user update, no need to pass the entire user - the ID is sufficient.

users/tracing/users.go
```diff
@@ -75,12 +75,20 @@ func (urm userRepositoryMiddleware) UpdatePassword(ctx context.Context, email, p
 	return urm.repo.UpdatePassword(ctx, email, password)
 }
 
-func (urm userRepositoryMiddleware) RetrieveAll(ctx context.Context, offset, limit uint64, ids []string, email string, um users.Metadata) (users.UserPage, error) {
+func (urm userRepositoryMiddleware) RetrieveAll(ctx context.Context, state string, offset, limit uint64, ids []string, email string, um users.Metadata) (users.UserPage, error) {
```
dborovcanin: state -> status

users/api/requests.go
```diff
@@ -62,6 +63,9 @@ func (req listUsersReq) validate() error {
 	if len(req.email) > maxEmailSize {
 		return apiutil.ErrEmailSize
 	}
+	if req.status != users.AllStatusKey && req.status != users.EnabledStatusKey && req.status != users.DisabledStatusKey {
+		return apiutil.ErrInvalidStatus
+	}
```
manuio: Can you please split this cases:
```
	if req.status != users.AllStatusKey &&
	    req.status != users.EnabledStatusKey &&
	    req.status != users.DisabledStatusKey {
		return apiutil.ErrInvalidStatus
	}
```

users/api/requests.go
```diff
@@ -139,20 +143,38 @@ func (req passwChangeReq) validate() error {
...
 func (req listMemberGroupReq) validate() error {
 	if req.token == "" {
 		return apiutil.ErrBearerToken
 	}
 
-	if req.groupID == "" {
+	if req.id == "" {
 		return apiutil.ErrMissingID
 	}
+	if req.status != users.AllStatusKey && req.status != users.EnabledStatusKey && req.status != users.DisabledStatusKey {
```
manuio: same here

manuio (changes requested):

users/service.go
```diff
@@ -322,13 +344,45 @@ func (svc usersService) ListMembers(ctx context.Context, token, groupID string,
...
+				Limit:  pm.Limit,
 			},
 		}, nil
 	}
 
-	return svc.users.RetrieveAll(ctx, offset, limit, userIDs, "", m)
+	return svc.users.RetrieveAll(ctx, pm.Status, pm.Offset, pm.Limit, userIDs, pm.Email, pm.Metadata)
+}
+
+func (svc usersService) EnableUser(ctx context.Context, token, id string) error {
```
dborovcanin, line 362: You can wrap both implementations of `EnableUser` and `DisableUser` with:
```go
func (svc usersService) changeStatus(ctx context.Context, token, id, status string) error {
	if _, err := svc.identify(ctx, token); err != nil {
		return err
	}

	dbUser, err := svc.users.RetrieveByID(ctx, id)
	if err != nil {
		return errors.Wrap(errors.ErrNotFound, err)
	}
	if dbUser.Status == status {
		return errStatus
	}

	return svc.users.ChangeStatus(ctx, dbUser, status)
}
```
and use it like:
```go
func (svc usersService) EnableUser(ctx context.Context, token, id string) error {
	if _, err := svc.changeStatus(ctx, token, id, EnabledStatusKey); err != nil {
		return errors.Wrap(ErrAlreadyEnabledUser, err)
	}
	return nil
}
```

dborovcanin (changes requested):

dborovcanin (dismissed):

users/api/transport.go
```diff
@@ -110,6 +111,20 @@ func MakeHandler(svc users.Service, tracer opentracing.Tracer, logger logger.Log
...
+		encodeResponse,
+		opts...,
+	))
+
+	mux.Post("/users/:id/disable", kithttp.NewServer(
+		kitot.TraceServer(tracer, "disable_user")(disableUserEndpoint(svc)),
+		decodeChangeUserStatus,
+		encodeResponse,
+		opts...,
+	))
```
manuio, line 126: Do we need 2 endpoints to change the status? You could fusion them or even use update_user one, right?

manuio, line 126: Do we need 2 endpoints to change the status? You could fusion them or even use update_user one, right? 

Don't forget to add the API doc here: https://github.com/mainflux/mainflux/blob/master/api/openapi/users.yml

manuio (changes requested):

users/api/transport.go
dborovcanin: We can, but I would say that this is OK at the moment. It's like an endpoint to enable and disable users - without exposing how it's done internally. If we add more status types, we definitely won't add an endpoint for each of them.

rodneyosodo: I have updated the api documentation

api/openapi/users.yml
```diff
@@ -215,6 +216,42 @@ paths:
           description: Missing or invalid content type.
         '500':
           $ref: "#/components/responses/ServiceError"
+  /users/{userID}/enable:
```
manuio: `userId`

api/openapi/users.yml
```diff
@@ -215,6 +216,42 @@ paths:
...
+    post:
+      summary: Enables a user account
+      description: |
+        Enables a disabled user account given userId.
+      tags:
+        - users
+      responses:
```
manuio, line 228: add section: 
```
      parameters:
        - $ref: "#/components/parameters/UserId"
```

api/openapi/users.yml
```diff
@@ -215,6 +216,42 @@ paths:
...
+          description: User enabled.
+        '400':
+          description: Failed due to malformed JSON.
+        '404':
+          description: Failed due to non existing user.
+        '401':
+          description: Missing or invalid access token provided.
+        '500':
+         $ref: "#/components/responses/ServiceError"
+  /users/{userID}/disable:
```
manuio: same here

api/openapi/users.yml
```diff
@@ -215,6 +216,42 @@ paths:
...
+    post:
+      summary: Disables a user account
+      description: |
+        Disables a user account given userId.
+      tags:
+        - users
+      responses:
```
manuio, line 248: and here. BTW, you can verify with: https://editor.swagger.io/

manuio (changes requested):

api/openapi/users.yml
rodneyosodo: I didn't know about this. Thanks

api/openapi/users.yml
```diff
@@ -215,6 +216,46 @@ paths:
...
+          description: Failed due to non existing user.
+        '401':
+          description: Missing or invalid access token provided.
+        '500':
+         $ref: "#/components/responses/ServiceError"
+  /users/{userId}/disable:
+    post:
+      summary: Disables a user account
+      description: |
+        Disables a user account given userId.
```
manuio: `Disables a user account for a given user ID.`

manuio (changes requested):

api/openapi/users.yml
```diff
@@ -353,7 +394,14 @@ components:
         default: 0
         minimum: 0
       required: false
-
+    Status:
+      name: status
+      description: The status of users to retrieve during retrieval.
```
manuio: Sounds redundant. Maybe simply `User account status`?

manuio (changes requested):

api/openapi/users.yml
```diff
@@ -215,6 +216,46 @@ paths:
...
+         $ref: "#/components/responses/ServiceError"
+  /users/{userId}/disable:
+    post:
+      summary: Disables a user account
+      description: |
+        Disables a user account given userId.
+      tags:
+        - users
+      parameters:
+        - $ref: "#/components/parameters/UserID"
```
manuio: If I remember well the recommendation is to use pure PascalCase

rodneyosodo: Yes

api/openapi/users.yml
```diff
@@ -220,11 +220,11 @@ paths:
     post:
       summary: Enables a user account
       description: |
-        Enables a disabled user account given userId.
+        Enables a disabled user account given user ID.
```
manuio: Do you mean `giving` or `for given`?

rodneyosodo: `for a given`

api/openapi/users.yml
```diff
@@ -215,6 +216,46 @@ paths:
           description: Missing or invalid content type.
         '500':
           $ref: "#/components/responses/ServiceError"
+  /users/{userId}/enable:
+    post:
+      summary: Enables a user account
+      description: |
+        Enables a disabled user account  for a given user ID.
```
manuio: There is an extra white space here `account  for`

api/openapi/users.yml
```diff
@@ -353,7 +394,14 @@ components:
...
         minimum: 0
       required: false
-
+    Status:
+      name: status
+      description: User account status.
+      in: query
+      schema:
+        type: string
+        default: active
```
manuio: `        default: enabled`

manuio (changes requested):

api/openapi/users.yml
rodneyosodo: Updated

manuio (dismissed):
LGTM

manuio:
@drasko Can we merge this one?

drasko:
@manuio I'll review

drasko:
This must wait for @dborovcanin's approval

manuio:
@drasko He had approved before my last remark

users/postgres/init.go
```diff
@@ -78,6 +78,13 @@ func migrateDB(db *sqlx.DB) error {
 					`ALTER TABLE IF EXISTS users ADD PRIMARY KEY (id)`,
 				},
 			},
+			{
+				Id: "users_5",
+				Up: []string{
+					`ALTER TABLE IF EXISTS users ADD COLUMN IF NOT EXISTS
+					status VARCHAR(64) NOT NULL DEFAULT 'enabled'`,
```
dborovcanin: Consider using SMALLINT and enumeration for status. While it is more cryptic than VARCHAR, it's also unlikely that we'll have any other status anytime soon. Unlike simple boolean, it also leaves us the space for adding potential new statuses without changing the schema in the future.

rodneyosodo: `SMALLINT` and `ENUM` are both small and fast. `SMALLINT` have the advantage of changing to multiple values in the future without schema change but this needs to be defined somewhere so that someone can know the values for each int and are therefore more cryptic than `ENUM`.  `ENUM` has an additional advantage of being safe whereas in `VARCHAR` or `SMALLINT` you have to specify the check function. I propose the use of `ENUM` since I doubt that the values will change in the near future.

dborovcanin (changes requested):
Since we now rely on DB enum, please also test repository behavior in case of inserting an invalid value for User status.

dborovcanin (approved):

drasko (approved):
LGTM

PR #1622, opened 2022-06-24
MF-1197 - Add MQTT adapter tests

### LINKED ISSUE #1197: Add MQTT adapter tests
**FEATURE REQUEST**

MQTT adapter needs unit tests.

This is a must-have.

zzokki81:
### What does this do?
Added MQTT unit tests

Resolves #1197 

MQTT unit tests

No


mqtt/handler_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+func TestHandler_AuthConnect(t *testing.T) {
+
+	cases := []struct {
+		desc    string
+		err     error
+		session *session.Client
+	}{
+		{
+			desc:    "should return error if client is nil",
```
manuio: The description should explain the test that you are doing not the error that is supposed to return. In this case it can be `Handle message without active session`.

Same for all other description

mqtt/mocks/client.go
```diff
@@ -0,0 +1,28 @@
...
+func NewClient() auth.Client {
+	return MockClient{}
+}
+
+func (cli MockClient) Authorize(ctx context.Context, chanID, thingID string) error {
+	return nil
+}
+
+func (cli MockClient) Identify(ctx context.Context, thingKey string) (string, error) {
+
```
manuio: kill this blank line

manuio (changes requested):

mqtt/handler_test.go
```diff
@@ -0,0 +1,162 @@
...
+	rdb "github.com/go-redis/redis/v8"
+	"github.com/mainflux/mainflux/logger"
+	"github.com/mainflux/mainflux/mqtt/mocks"
+	"github.com/mainflux/mainflux/mqtt/redis"
+	"github.com/mainflux/mainflux/pkg/messaging"
+	"github.com/ory/dockertest/v3"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestHandler_AuthConnect(t *testing.T) {
```
manuio: Rename the method `TestAuthConnect`

mqtt/handler_test.go
```diff
@@ -0,0 +1,162 @@
...
+		},
+	}
+
+	for _, tc := range cases {
+		err := setupTest().AuthConnect(tc.session)
+		assert.True(t, errors.Contains(err, tc.err), fmt.Sprintf("%s: expected %s got %s\n", tc.desc, tc.err, err))
+	}
+}
+
+func TestHandler_AuthPublish(t *testing.T) {
```
manuio: Rename `TestAuthPublish`

manuio (changes requested):

mqtt/handler_test.go
```diff
@@ -0,0 +1,181 @@
...
+		{
+			desc:   "Handle subscribe authorization on client correctly",
+			client: &sessionClient,
+			err:    nil,
+			topic:  &topics,
+		},
+	}
+
+	for _, tc := range cases {
+		err := setupTest().AuthSubscribe(tc.client, tc.topic)
```
manuio: Create the session handler outside the for loop: `handler := createHandler()`

mqtt/handler_test.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	cases := []struct {
+		desc   string
+		client *session.Client
+		err    error
+		topic  *[]string
+	}{
+		{
+			desc:   "Handle subscribe authorization on client without active session",
```
manuio: Avoid repeting `Handle subscribe authorization on client` for all descriptions. For example here `Subscribe without active session` would be enough

manuio (changes requested):

mqtt/handler_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+import (
+	"fmt"
+	"github.com/mainflux/mainflux/pkg/uuid"
+	"github.com/stretchr/testify/require"
+	"log"
+	"testing"
+
+	"github.com/mainflux/mainflux/pkg/errors"
+
```
manuio: remove this blank line

mqtt/handler_test.go
```diff
@@ -0,0 +1,189 @@
...
+			desc:    "connect without active session",
+			err:     errInvalidConnect,
+			session: nil,
+		},
+		{
+			desc: "connect when id is invalid",
+			err:  errors.New("thing identify error"),
+			session: &session.Client{
+				ID:       "123",
+				Username: "testUsername",
```
manuio: create a global const and use it in all tests

manuio (changes requested):

mqtt/handler_test.go
```diff
@@ -0,0 +1,189 @@
...
+		err     error
+		session *session.Client
+	}{
+		{
+			desc:    "connect without active session",
+			err:     errInvalidConnect,
+			session: nil,
+		},
+		{
+			desc: "connect when id is invalid",
```
manuio: `connect with invalid id`

mqtt/handler_test.go
```diff
@@ -0,0 +1,189 @@
...
+			err:  errors.New("thing identify error"),
```
manuio: Remove this error and use one from the package. If there is no error in the package maybe you can change a bit the code and wrap the returned error with a new one.

mqtt/handler_test.go
```diff
@@ -0,0 +1,189 @@
...
+			session: &session.Client{
+				ID:       "123",
+				Username: "testUsername",
+				Password: []byte(""),
+			},
+		},
+		{
+			desc: "connect when username is invalid",
```
manuio: `connect with valid password and invalid username`

mqtt/handler_test.go
```diff
@@ -0,0 +1,189 @@
...
+			err:  errors.ErrAuthentication,
+			session: &session.Client{
+				ID:       "123",
+				Username: "testUsername",
+				Password: []byte("password"),
+			},
+		},
+		{
+			desc: "connect with right username and password",
```
manuio: `connect with valid username and password"`

manuio (changes requested):

mqtt/handler_test.go
```diff
@@ -2,21 +2,26 @@ package mqtt
...
 	"github.com/mainflux/mainflux/mqtt/redis"
+	"github.com/mainflux/mainflux/pkg/errors"
 	"github.com/mainflux/mainflux/pkg/messaging"
+	"github.com/mainflux/mainflux/pkg/uuid"
 	"github.com/mainflux/mproxy/pkg/session"
 	"github.com/ory/dockertest/v3"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"log"
+	"testing"
```
manuio: Move this 2 packets to the top

mqtt/handler_test.go
```diff
@@ -2,21 +2,26 @@ package mqtt
...
+)
+
+const (
+	id         = "123"
+	username   = "testUsername"
+	usernameOK = "ok"
```
manuio: Why not to use a single user for the tests?

manuio (changes requested):

mqtt/handler_test.go
```diff
@@ -0,0 +1,239 @@
...
+	for _, tc := range cases {
+		handler.Publish(tc.client, &tc.topic, &tc.payload)
+		assert.Contains(t, buf.String(), tc.expected)
+	}
+}
+
+func newHandler() session.Handler {
+	logger, err := logger.New(&buf, "debug")
+	if err != nil {
+		panic(err)
```
manuio: Replace it by something like `log.Fatalf("Could not create logger: %s", err))`

manuio (changes requested):

mqtt/handler_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/mainflux/mainflux/pkg/uuid"
+	"github.com/mainflux/mproxy/pkg/session"
+	"github.com/ory/dockertest/v3"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	id              = "123"
+	username        = "username"
```
manuio: Since we are using thingID as MQTT username it would be better to use an uuid here:
```
	thingID  = "thingID"
```

mqtt/handler_test.go
```diff
@@ -0,0 +1,239 @@
...
+	invalidUsername = "invalidUsername"
```
manuio: `invalidID = "invalidID"`

mqtt/handler_test.go
```diff
@@ -0,0 +1,239 @@
...
+	for _, tc := range cases {
+		err := handler.AuthConnect(tc.session)
+		assert.True(t, errors.Contains(err, tc.err), fmt.Sprintf("%s: expected %s got %s\n", tc.desc, tc.err, err))
+	}
+}
+
+func TestAuthPublish(t *testing.T) {
+	handler := newHandler()
+	var topic string
+	var payload = []byte("payload")
```
manuio: Use a valid payload. For example `[]byte("[{"n":"test-name", "v": 1.2}]")`

mqtt/handler_test.go
```diff
@@ -0,0 +1,239 @@
...
+		desc    string
+		client  *session.Client
+		err     error
+		topic   *string
+		payload []byte
+	}{
+		{
+			desc:    "publish without active session",
+			client:  nil,
+			err:     errNilClient,
```
manuio: It's not realted to this PR but can you change this error name please? `errInactiveSession` for example

mqtt/handler_test.go
```diff
@@ -0,0 +1,239 @@
...
+		log.Fatalf("Could not start container: %s", err)
+	}
+
+	redisClient := rdb.NewClient(&rdb.Options{
+		Addr:     fmt.Sprintf("localhost:%s", container.GetPort("6379/tcp")),
+		Password: "",
+		DB:       0,
+	})
+
+	eventStore := redis.NewEventStore(redisClient, "")
```
manuio: No need to create a real even store. Use a mock instead. The only method to implement is the one reaturning the client ID: `ID()`

manuio (changes requested):

mqtt/handler_test.go
```diff
@@ -0,0 +1,373 @@
...
+		err     error
+		session *session.Client
+	}{
+		{
+			desc:    "connect without active session",
+			err:     errMissingClient,
+			session: nil,
+		},
+		{
+			desc: "connect with valid id",
```
manuio: `connect with invalid password"` 

Maybe you can add another test with empty client ID?

mqtt/handler_test.go
```diff
@@ -0,0 +1,373 @@
...
+
+	cases := []struct {
+		desc    string
+		client  *session.Client
+		err     error
+		topic   *string
+		payload []byte
+	}{
+		{
+			desc:    "publish without active session",
```
manuio: `publish with inactive client`

mqtt/handler_test.go
```diff
@@ -0,0 +1,373 @@
...
+		},
+		{
+			desc:    "publish without topic",
+			client:  &sessionClient,
+			err:     errMissingTopicPub,
+			topic:   nil,
+			payload: payload,
+		},
+		{
+			desc:    "publish with active session, valid topic and payload",
```
manuio: `publish with malformed topic`

manuio (changes requested):

manuio (changes requested):

mqtt/handler_test.go
```diff
@@ -0,0 +1,381 @@
...
+		desc     string
+		client   *session.Client
+		topic    string
+		payload  []byte
+		expected string
+	}{
+		{
+			desc:     "publish without active session",
+			client:   nil,
+			topic:    "topic",
```
manuio: Use a valid topic here to be sure that the error comes from the inactive session

mqtt/handler_test.go
```diff
@@ -0,0 +1,381 @@
...
+			payload:  []byte("payload"),
+			expected: errMissingClient.Error(),
+		},
+		{
+			desc:     "publish with invalid channel parts",
```
manuio: This should be `publish with invalid topic`

mqtt/handler_test.go
```diff
@@ -0,0 +1,381 @@
...
+			client:   &sessionClient,
+			topic:    "topic",
+			payload:  []byte("payload"),
+			expected: "published with client_id " + clientID + " to the topic: " + "topic",
```
manuio: Create a variable for this error

mqtt/handler_test.go
```diff
@@ -0,0 +1,381 @@
...
+		desc     string
+		client   *session.Client
+		topic    []string
+		expected string
+	}{
+		{
+			desc:     "subscribe without active session",
+			client:   nil,
+			topic:    topics,
+			expected: "failed to subscribe: " + errMissingClient.Error(),
```
manuio: Create an error for `failed to subscribe:`

manuio (changes requested):

mqtt/handler.go
```diff
@@ -69,7 +77,7 @@ func (h *handler) AuthConnect(c *session.Client) error {
 	}
 
 	if err := h.es.Connect(c.Username); err != nil {
-		h.logger.Warn("Failed to publish connect event: " + err.Error())
+		h.logger.Warn("failed to publish connect event: " + err.Error())
```
manuio: I'd change this `Warn` by `Error` and create a const for the message

manuio (changes requested):

mqtt/handler.go
```diff
@@ -153,38 +161,38 @@ func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
...
 	if c == nil {
-		h.logger.Error("Nil client disconnect")
+		h.logger.Error(errors.Wrap(errFailedDisconnectMsg, errClientNotInitialized).Error())
 		return
 	}
-	h.logger.Info("Disconnect - Client with ID: " + c.ID + " and username " + c.Username + " disconnected")
+	h.logger.Info(fmt.Sprintf(infoDisconnectMsg, c.ID, c.Username))
 	if err := h.es.Disconnect(c.Username); err != nil {
-		h.logger.Warn("Failed to publish disconnect event: " + err.Error())
+		h.logger.Warn("failed to publish disconnect event: " + err.Error())
```
manuio: Same here. `logger.Error` and create a const for the message. Also, test it

manuio (changes requested):

mqtt/handler_test.go
```diff
@@ -0,0 +1,392 @@
...
+			err:     errClientNotInitialized,
+			session: nil,
+		},
+		{
+			desc: "connect without clientID",
+			err:  nil,
+			session: &session.Client{
+				ID:       "",
+				Username: thingID,
+				Password: []byte("password"),
```
manuio: Use `password` const here

mqtt/handler_test.go
```diff
@@ -0,0 +1,392 @@
...
+		session *session.Client
+	}{
+		{
+			desc:    "connect without active session",
+			err:     errClientNotInitialized,
+			session: nil,
+		},
+		{
+			desc: "connect without clientID",
+			err:  nil,
```
manuio: Is it normal to don't return error if client ID is empty?

manuio (changes requested):

mqtt/handler_test.go
```diff
@@ -0,0 +1,392 @@
...
+
+	for _, tc := range cases {
+		err := handler.AuthConnect(tc.session)
+		assert.True(t, errors.Contains(err, tc.err), fmt.Sprintf("%s: expected %s got %s\n", tc.desc, tc.err, err))
+	}
+}
+
+func TestAuthPublish(t *testing.T) {
+	handler := newHandler()
+	var invalidTopic = "invalidTopic"
```
manuio: This var is used in several tests, create a const instead.

manuio (changes requested):

mqtt/handler_test.go
```diff
@@ -0,0 +1,406 @@
...
+	var topic = "channels/" + chID + "/messages"
+
+	cases := []struct {
+		desc    string
+		client  *session.Client
+		err     error
+		topic   *string
+		payload []byte
+	}{
+		{
```
manuio, line 109: `publish successfully` test is missing

mqtt/handler_test.go
```diff
@@ -0,0 +1,406 @@
...
+		client   *session.Client
+		expected string
+	}{
+		{
+			desc:     "connect without active session",
+			client:   nil,
+			expected: fmt.Sprint(errors.Wrap(mqtt.ErrFailedConnect, mqtt.ErrClientNotInitialized).Error()),
+		},
+		{
+			desc:     "connect with valid session",
```
manuio: `connect with active session` would be more consistent with previous message

mqtt/handler_test.go
```diff
@@ -0,0 +1,406 @@
...
+	buf.Reset()
+	cases := []struct {
+		desc     string
+		client   *session.Client
+		expected string
+	}{
+		{
+			desc:     "connect without active session",
+			client:   nil,
+			expected: fmt.Sprint(errors.Wrap(mqtt.ErrFailedConnect, mqtt.ErrClientNotInitialized).Error()),
```
manuio: use `err: errors.Wrap(mqtt.ErrFailedConnect, mqtt.ErrClientNotInitialized)` here and extract the string latter

mqtt/handler_test.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+}
+
+func TestConnect(t *testing.T) {
+	handler := newHandler()
+	buf.Reset()
+	cases := []struct {
+		desc     string
+		client   *session.Client
+		expected string
```
manuio: `err`

mqtt/handler_test.go
```diff
@@ -0,0 +1,406 @@
...
+
+	for _, tc := range cases {
+		handler.Connect(tc.client)
+		assert.Contains(t, buf.String(), tc.expected)
+	}
+}
+
+func TestPublish(t *testing.T) {
+	handler := newHandler()
+	buf.Reset()
```
manuio: I'd rename `buf` `logBuffer`

manuio (changes requested):

manuio (dismissed):
LGTM

mqtt/handler_test.go
```diff
@@ -0,0 +1,421 @@
...
+	authClient := mocks.NewClient(map[string]string{password: thingID}, map[string]interface{}{chanID: thingID})
+
+	var redisClient *redis.Client
+
+	pool, err := dockertest.NewPool("")
+	if err != nil {
+		log.Fatalf("Could not connect to docker: %s", err)
+	}
+
+	container, err := pool.Run("redis", "5.0-alpine", nil)
```
dborovcanin: Spinning a new Docker container for each `newHandler` call is not a good idea. Why do we even use Redis and not mock for events?

dborovcanin (changes requested):
Testing using log messages is usually not a good idea. While it makes sense for some methods, we definitely should not spin Redis for each `newHandler` call, and we should use an error message to compare (see error wrapping), not hardcoded log constants.

manuio (dismissed):
LGTM

mqtt/handler.go
```diff
@@ -23,14 +24,28 @@ var _ session.Handler = (*handler)(nil)
...
+	LogInfoPublished                   = "published with client_id %s to the topic %s"
+	LogErrFailedConnect                = "failed to connect: "
+	LogErrFailedSubscribe              = "failed to subscribe: "
+	LogErrFailedUnsubscribe            = "failed to unsubscribe: "
+	LogErrFailedPublish                = "failed to publish: "
+	LogErrFailedDisconnect             = "failed to disconnect: "
+	LogErrFailedPublishDisconnectEvent = "failed to publish disconnect event: "
+	logErrFailedParseSubtopic          = "failed to parse subtopic: "
+	LogErrFailedPublishConnectEvent    = "failed to publish connect event: "
+	LogErrFailedPublishToMsgBroker     = "failed to publish to mainflux message broker: "
```
dborovcanin, line 40: Put log messages to a separate `const` block.

mqtt/handler.go
```diff
@@ -23,14 +24,28 @@ var _ session.Handler = (*handler)(nil)
...
+	errMalformedSubtopic               = errors.New("malformed subtopic")
```
dborovcanin: Is this error covered by tests?

mqtt/handler_test.go
```diff
@@ -0,0 +1,391 @@
...
+	invalidChanID  = "1"
+	clientID       = "clientID"
+	invalidThingID = "invalidThingID"
+	password       = "password"
+)
+
+var (
+	invalidTopic  = "invalidTopic"
+	idProvider    = uuid.NewMock()
+	logBuffer     = bytes.Buffer{}
```
dborovcanin, line 37: Add a comment, something like:
Test log messages for cases the handler does not provide a return value.

dborovcanin (changes requested):
This obscure testing of logs pointed out potential flows in mProxy. Also, we should definitely decouple mutual dependency between Mainflux and mProxy (mProxy is using Mainflux errors, while Mainflux is using mProxy in MQTT adapter).

manuio (dismissed):
LGTM

mqtt/handler.go
```diff
@@ -20,17 +21,33 @@ import (
 
 var _ session.Handler = (*handler)(nil)
 
-const protocol = "mqtt"
+const (
+	protocol                           = "mqtt"
```
dborovcanin: Keep `protocol` as a separate constant, since this is a block for logging constants.

dborovcanin (changes requested):
While checking the log message is a quirky way to test - for some methods that don't return errors but log them - that's the only way. I left only one cosmetical remark and we can merge.

manuio (approved):
LGTM

dborovcanin (approved):

drasko (approved):
LGTM

drasko:
Merged! Thanks @zzokki81. This was a nice solution with log analysis, but we should improve in the future by adding return values to the handlers.

dborovcanin:
The main improvement would be adding return errors where it's missing, and using pubsub mocks to check if pubsub methods are called for successful flow (rather than checking for info logs). Also, moving the logger from the handler implementation to the logging wrapper/middleware is a good idea.

drasko:
Issue to track: https://github.com/mainflux/mproxy/issues/31

PR #1624, opened 2022-06-28
NOISSUE - Fix make run for OS X and docker-compose hierarchy

manuio:

### What does this do?
Fix `make run` and the nats conf path for docker-compose


manuio:
One way to make the GNU version of the SED to work on the Mac OS X, is to directly install the gnu-sed along with the default names which will assure you that you won't have to run different commands on both the operating systems. 

Closing this one.

PR #1625, opened 2022-07-04
MF-1623 - Bring back WebSocket Adapter

### LINKED ISSUE #1623: Bring back WebSocket Adapter
The initial commit is bringing back the ws directory, without any changes

AryanGodara:
Signed-off-by: AryanGodara [aryangodara03@gmail.com](mailto:aryangodara03@gmail.com)

### What does this do?
This pull request brings back the WebSocket adapter

Resolves #1623

Nil

Yes

Yes

### Notes
N/A


ws/adapter_test.go
```diff
@@ -0,0 +1,120 @@
...
+		{
+			desc: "publish valid message",
+			msg:  msg,
+			err:  nil,
+		},
+		{
+			desc: "publish empty message",
+			msg:  mainflux.Message{},
+			err:  ws.ErrFailedMessagePublish,
+		},
```
manuio, line 85: Add more tests. With invalid ID, invalid Channel, etc etc

ws/adapter_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	cases := []struct {
+		desc     string
+		chanID   string
+		subtopic string
+		channel  *ws.Channel
+		err      error
+	}{
+		{
+			desc:    "subscription to valid channel",
```
manuio: `subscribe to valid channel`

ws/adapter_test.go
```diff
@@ -0,0 +1,120 @@
...
+			chanID:  chanID,
+			channel: channel,
+			err:     nil,
+		},
+		{
+			desc:    "subscription to channel that should fail",
+			chanID:  "0",
+			channel: channel,
+			err:     ws.ErrFailedSubscription,
+		},
```
manuio, line 163: `subscribe to invalid channel`

manuio (changes requested):

ws/adapter_test.go
AryanGodara: I'm still learning about tests, so I'll get to this after fixing other errors.
Thank you for the help!

AryanGodara:
After commie `cd4160e`
I have made the changes like implementing client.go with adapter.go
Correcting the "authKey" in Subscribe(), Unsubscribe() and Publish()
Also, wrote the code for Handle() and Cancel(), after taking help from @0x6f736f646f 

@dborovcanin and @0x6f736f646f please take a look.
For now, the code is compiling and ws handshake is completing.
(It was working earlier also, but took few hours to get everything to work again, after implementing the above changes)

AryanGodara:
It seems to be working now, 
Manual Test: 

```
$make run 
$ ./mainflux-cli provision test (To get readymade thing and channel, that are connected (saves time on doing manually)
$ go run cmd/main.go (server starts listening)
$ go run cmd/ws_client.go (ws connection successful)

$ ./coap-cli-linux get channels/CHANNELID/messages --auth THINGKEY -o (Start listening on coap, to check if there is error with only ws-adapter, or in general error with sending messages)

$ curl -s -S -i -X POST -H "Content-Type: application/senml+json" -H "Authorization: Thing THINGKEY" https://localhost/http/channels/CHANNELID/messages -d '[{"random":"message}]'

OUTPUT:- received the message on both cli and ws terminal windows (tried multiple times)

File: ws/api/transport.go [hunk 4]
```diff
@@ -0,0 +1,254 @@
...
+		}
+
+		if len(elem) > 1 && (strings.Contains(elem, "*") || strings.Contains(elem, ">")) {
+			return "", errMalformedSubtopic
+		}
+
+		filteredElems = append(filteredElems, elem)
+	}
+
+	subtopic = strings.Join(filteredElems, ",")
```
dborovcanin: Topics are joined by `.`, not `,`.

ws/api/transport.go
```diff
@@ -0,0 +1,254 @@
...
+	logger            log.Logger
+	channelPartRegExp = regexp.MustCompile(`^/channels/([\w\-]+)/messages(/[^?]*)?(\?.*)?$`)
+)
+
+type subscription struct {
+	pubID    string
+	chanID   string
+	subtopic string
+	conn     *websocket.Conn
+	// client   *ws.Connclient
```
dborovcanin: Remove.

ws/api/transport.go
```diff
@@ -0,0 +1,254 @@
...
+)
+
+var (
+	errUnauthorizedAccess = errors.New("missing or invalid credentials provided")
+	errMalformedSubtopic  = errors.New("malformed subtopic")
+)
+
+var (
+	upgrader = websocket.Upgrader{
+		ReadBufferSize:  1024,
```
dborovcanin: Extract `1024` to constant.

ws/api/transport.go
```diff
@@ -0,0 +1,254 @@
...
+	mux := bone.New()
+	mux.GetFunc("/channels/:id/messages", handshake(svc))
+	mux.GetFunc("/channels/:id/messages/*", handshake(svc))
+	mux.GetFunc("/version", mainflux.Health(protocol))
+	mux.Handle("/metrics", promhttp.Handler())
+
+	return mux
+}
+
+func handshake(svc ws.Service) http.HandlerFunc {
```
dborovcanin: This function is overly large. Split it to simplify.

manuio:
@AryanGodara Please, update your branch and remove draft tag when the PR is ready

ws/client.go
```diff
@@ -0,0 +1,94 @@
...
+	"time"
+
+	"github.com/gorilla/websocket"
+	"github.com/mainflux/mainflux/logger"
+	"github.com/mainflux/mainflux/pkg/messaging"
+)
+
+// Client wraps WS client.
+type Client interface {
+	Token() string
```
dborovcanin: `Token` is not used.

ws/api/transport.go
```diff
@@ -0,0 +1,254 @@
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		sub, err := authorize(r)
```
dborovcanin: Move authorization to the service layer.

ws/api/transport.go
```diff
@@ -0,0 +1,254 @@
...
+		ReadBufferSize:  1024,
+		WriteBufferSize: 1024,
+		CheckOrigin:     func(r *http.Request) bool { return true },
+	}
+	auth              mainflux.ThingsServiceClient
+	logger            log.Logger
+	channelPartRegExp = regexp.MustCompile(`^/channels/([\w\-]+)/messages(/[^?]*)?(\?.*)?$`)
+)
+
+type subscription struct {
```
dborovcanin: You have `subscription`, `ws.Client` and `ws.Channel`. Try to unite them to a single `Client`.

ws/api/transport.go
```diff
@@ -0,0 +1,254 @@
...
+			}
+		}
+
+		sub.channel = ws.NewChannel()
+
+		wg := new(sync.WaitGroup)
+		wg.Add(2)
+
+		go sub.listen(wg)
+		go sub.broadcast(wg, svc)
```
dborovcanin: Decouple subscriber and service. You can use the channel for that (`broadcast` sends to it, and the service reads and does `Push`).

dborovcanin (changes requested):
Overall - this is very good. There are some minor bugs, but you need to try and simplify and improve the code structure, but we're very close.

ws/api/transport.go
AryanGodara: Combined ConnClient{} and Channel{} structs in ws/client.go

ws/api/transport.go
AryanGodara: subcription{} struct has `*websocket.Conn` field, which is needed to read and write messages using ws protocol. And holds `pubID`, `chanID` and `subtopic`, I need to use a struct to hold these, and they're all required in listen() and broadcast methods.

ws/adapter.go
```diff
@@ -0,0 +1,135 @@
...
+	ErrFailedConnection = errors.New("failed to connect to message broker")
+)
+
+// Service specifies web socket service API.
+type Service interface {
+	// Publish Message
+	Publish(ctx context.Context, token string, msg messaging.Message) error
+
+	// Authorize message is used to authenticate the thingKey and chanID
+	Authorize(ctx context.Context, thingKey, chanID string) (*mainflux.ThingID, error)
```
dborovcanin: This is internal to the adapter implementation and should not be exposed.

ws/client.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	if err := c.client.WriteMessage(1, m.Payload); err != nil {
+		c.logger.Error(fmt.Sprintf("Error sending message: %s", err))
+	}
+
+	return c.client.Close()
+}
+
+func (c *Connclient) Handle(msg messaging.Message) error {
+	return c.client.WriteMessage(websocket.TextMessage, msg.Payload)
```
dborovcanin: We need to prevent getting clients' own messages over WS. In order to do that, we need to know the client ID here, either by setting it up in the Client structure or passing it as an argument (which will mess up the implementation of the `messaging.MessageHandler` interface).

ws/docker-compose-config.yml
```diff
@@ -0,0 +1,18 @@
...
+    MF_WS_ADAPTER_LOG_LEVEL: ${MF_WS_ADAPTER_LOG_LEVEL}
+    MF_WS_ADAPTER_PORT: ${MF_WS_ADAPTER_PORT}
+    MF_BROKER_URL: ${MF_NATS_URL}
+    MF_JAEGER_URL: ${MF_JAEGER_URL}
+    MF_THINGS_AUTH_GRPC_URL: ${MF_THINGS_AUTH_GRPC_URL}
+    MF_THINGS_AUTH_GRPC_TIMEOUT: ${MF_THINGS_AUTH_GRPC_TIMEOUT}
+  ports:
+    - ${MF_WS_ADAPTER_PORT}:${MF_WS_ADAPTER_PORT}
+  networks:
+    - mainflux-base-net
```
dborovcanin: Add an empty line.

ws/adapter.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Subscribes to a channel with specified id.
+	Subscribe(ctx context.Context, thingKey, chanID, subtopic string, channel Client) error
```
dborovcanin: Rename `channel Client`.

dborovcanin (changes requested):

ws/client.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/gorilla/websocket"
+	"github.com/mainflux/mainflux/pkg/messaging"
+)
+
+type Client interface {
+	Handle(m messaging.Message) error
+	Cancel() error
+}
```
dborovcanin: This interface is not used.

ws/docker-compose-config.yml
```diff
@@ -0,0 +1,19 @@
+ws-adapter:
```
dborovcanin: Remove this file in favor of `docker/docker-compose.yml`

ws/api/requests.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Mainflux
+// SPDX-Licence-Identifier: Apache-2.0
+
+package api
+
+import "github.com/gorilla/websocket"
+
+type publishReq struct {
+	thingKey string // pubID = thingKey (delete this comment later)
```
dborovcanin: Later is now, remove the comment. :)

ws/api/requests.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Mainflux
+// SPDX-Licence-Identifier: Apache-2.0
+
+package api
+
+import "github.com/gorilla/websocket"
+
+type publishReq struct {
```
dborovcanin: Rename to `connReq`.

ws/api/endpoints.go
```diff
@@ -0,0 +1,178 @@
...
+			return "", errMalformedSubtopic
+		}
+
+		filteredElems = append(filteredElems, elem)
+	}
+
+	subtopic = strings.Join(filteredElems, ".")
+
+	return subtopic, nil
+}
```
dborovcanin: Use a single function to parse subtopic. You can use https://github.com/mainflux/mainflux/blob/6bd745b658ff1e04c40394a732c731fa99882c32/http/api/transport.go#L71 as a refenerce.

ws/api/endpoints.go
```diff
@@ -0,0 +1,178 @@
...
+	req := publishReq{
+		thingKey: authKey,
+		chanID:   chanID,
+	}
+
+	return req, nil
+}
+
+func (req publishReq) broadcast(msgs chan<- messaging.Message) {
+
```
dborovcanin: Remove the empty line.

ws/api/endpoints.go
```diff
@@ -0,0 +1,178 @@
...
+	req := publishReq{
+		thingKey: authKey,
+		chanID:   chanID,
+	}
+
+	return req, nil
+}
+
+func (req publishReq) broadcast(msgs chan<- messaging.Message) {
```
dborovcanin: Consider using different naming for this func. While we practically do broadcast using a message broker, we actually only use a single receiver (message broker).

ws/api/logging.go
```diff
@@ -0,0 +1,92 @@
...
+// 		if err != nil {
+// 			lm.logger.Warn(fmt.Sprintf("%s with error: %s", message, err))
+// 			return
+// 		}
+// 		lm.logger.Info(fmt.Sprintf("%s without errors.", message))
+// 	}(time.Now())
+
+// 	return lm.svc.authorize(ctx, thingKey, chanID)
+// }
+
```
dborovcanin: Remove commented code.

ws/api/logging.go
```diff
@@ -0,0 +1,92 @@
...
+	return lm.svc.Subscribe(ctx, thingKey, chanID, subtopic, c)
+}
+
+func (lm *loggingMiddleware) Unsubscribe(ctx context.Context, thingKey, chanID, subtopic string) (err error) {
+	defer func(begin time.Time) {
+		destChannel := chanID
+		if subtopic != "" {
+			destChannel = fmt.Sprintf("%s.%s", destChannel, subtopic)
+		}
+		message := fmt.Sprintf("Method unsusubscribe to channel %s took %s to complete", destChannel, time.Since(begin))
```
dborovcanin: ...unsubscribe from...

ws/api/metrics.go
```diff
@@ -0,0 +1,68 @@
...
+
+// func (mm *metricsMiddleware) Authorize(ctx context.Context, thingKey, chanID string) (*mainflux.ThingID, error) {
+// 	defer func(begin time.Time) {
+// 		mm.counter.With("method", "authorize").Add(1)
+// 		mm.latency.With("method", "authorize").Observe(time.Since(begin).Seconds())
+// 	}(time.Now())
+
+// 	return mm.svc.authorize(ctx, thingKey, chanID)
+// }
+
```
dborovcanin: Remove.

scripts/run.sh
```diff
@@ -50,6 +50,11 @@ MF_THINGS_LOG_LEVEL=info MF_THINGS_HTTP_PORT=8182 MF_THINGS_AUTH_GRPC_PORT=8183
 ###
 MF_HTTP_ADAPTER_LOG_LEVEL=info MF_HTTP_ADAPTER_PORT=8185 MF_THINGS_AUTH_GRPC_URL=localhost:8183 $BUILD_DIR/mainflux-http &
 
+###
```
dborovcanin, line 53: Also need to add service to the Makefile.

health.go
```diff
@@ -51,8 +52,10 @@ type HealthInfo struct {
 // Health exposes an HTTP handler for retrieving service health.
 func Health(service string) http.HandlerFunc {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Println("Do we reach Healh()?")
```
dborovcanin: Yes, we do. 🙂 Remove.

docker/ssl/authorization.js
```diff
@@ -138,6 +138,18 @@ function setKey(r) {
...
+    //     for (a in r.args) {
+    //         if (a == 'authorization' && r.args[a] === clientKey) {
+    //             return clientKey
+    //         }
+    //     }
+
+    //     r.error('Authorization param does not match certificate')
+    //     return '';
+    // }
+
```
dborovcanin: Uncomment.

docker/nginx/nginx-x509.conf
```diff
@@ -121,6 +121,14 @@ http {
...
             proxy_pass http://mqtt_ws_cluster;
         }
+
+        # Proxy pass to mainflux-ws-adapter
+        # location /ws/ {
+        #     include snippets/verify-ssl-client.conf;
+        #     include snippets/proxy-headers.conf;
+        #     include snippets/ws-upgrade.conf;
+        #     proxy_pass http://ws-adapter:${MF_WS_ADAPTER_PORT}/;
+        # }
```
dborovcanin: Uncomment.

docker/nginx/nginx-key.conf
```diff
@@ -110,6 +110,13 @@ http {
             include snippets/ws-upgrade.conf;
             proxy_pass http://mqtt_ws_cluster;
         }
+
+        # Proxy pass to mainflux-ws-adapter
+        # location /ws/ {
+        #     include snippets/proxy-headers.conf;
+        #     include snippets/ws-upgrade.conf;
+        #     proxy_pass http://ws-adapter:${MF_WS_ADAPTER_PORT}/;
+        # }
```
dborovcanin: Uncomment.

docker/docker-compose.yml
```diff
@@ -346,3 +346,22 @@ services:
...
+  #     MF_WS_ADAPTER_LOG_LEVEL: ${MF_WS_ADAPTER_LOG_LEVEL}
+  #     MF_WS_ADAPTER_PORT: ${MF_WS_ADAPTER_PORT}
+  #     MF_BROKER_URL: ${MF_NATS_URL}
+  #     MF_JAEGER_URL: ${MF_JAEGER_URL}
+  #     MF_THINGS_AUTH_GRPC_URL: ${MF_THINGS_AUTH_GRPC_URL}
+  #     MF_THINGS_AUTH_GRPC_TIMEOUT: ${MF_THINGS_AUTH_GRPC_TIMEOUT}
+  #   ports:
+  #     - ${MF_WS_ADAPTER_PORT}:${MF_WS_ADAPTER_PORT}
+  #   networks:
+  #     - mainflux-base-net
```
dborovcanin: Uncomment.

cmd/ws/main.go
```diff
@@ -0,0 +1,227 @@
...
+		logger.Error(fmt.Sprintf("Failed to init Jaeger client: %s", err))
+		os.Exit(1)
+	}
+
+	return tracer, closer
+}
+
+func newService(tc mainflux.ThingsServiceClient, nps messaging.PubSub, logger logger.Logger) adapter.Service {
+	svc := adapter.New(tc, nps)
+
```
dborovcanin: Remove the empty line.

cmd/ws/main.go
```diff
@@ -0,0 +1,227 @@
...
+	svc = api.LoggingMiddleware(svc, logger)
+
```
dborovcanin: This one as well.

cmd/ws/main.go
```diff
@@ -0,0 +1,227 @@
...
+			Help:      "Total duration of requests in microseconds",
+		}, []string{"method"}),
+	)
+
+	return svc
+}
+
+func startWSServer(ctx context.Context, cfg config, svc adapter.Service, l logger.Logger) error {
+	p := fmt.Sprintf(":%s", cfg.port)
+
```
dborovcanin: Remove the empty line.

cmd/ws/main.go
```diff
@@ -0,0 +1,227 @@
...
+	errCh := make(chan error, 2)
+
```
dborovcanin: And this one.

dborovcanin (changes requested):

ws/docker-compose-config.yml
AryanGodara: This was just to store until review😅. I'll put this in `docker/docker-compose.yml` now👍.

ws/api/endpoints.go
AryanGodara: changed this to `listen`, as this one is only 
listening for messages from ws-client using conn.readmessage
`publish` function publishes to the message broker

ws/api/endpoints.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/go-zoo/bone"
+	"github.com/gorilla/websocket"
+	"github.com/mainflux/mainflux/pkg/errors"
+	"github.com/mainflux/mainflux/pkg/messaging"
+	"github.com/mainflux/mainflux/ws"
+)
+
+var (
+	channelPartRegExp = regexp.MustCompile(`^/channels/([\w\-]+)/messages(/[^?]*)?(\?.*)?$`)
+)
```
dborovcanin: Inline (remove parentheses).

ws/api/endpoints.go
```diff
@@ -0,0 +1,173 @@
...
+		if err := svc.Subscribe(context.Background(), req.thingKey, req.chanID, req.subtopic, client); err != nil {
+			logger.Warn(fmt.Sprintf("Failed to subscribe to broker: %s", err.Error()))
+			if err == ws.ErrFailedConnection {
+				req.conn.Close()
+				return
+			}
+		}
+
+		logger.Debug(fmt.Sprintf("Successfully upgraded communication to WS on channel %s", req.chanID))
+
```
dborovcanin: Remove the empty line.

ws/api/endpoints.go
```diff
@@ -0,0 +1,173 @@
...
+		msgs := make(chan messaging.Message)
+
+		// Listen for messages received from the chan messages, and publish them to broker
+		go publish(svc, msgs, req.thingKey)
+
```
dborovcanin: Same.

ws/client.go
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-Licence-Identifier: Apache-2.0
+
+package ws
+
+import (
+	"github.com/gorilla/websocket"
+	"github.com/mainflux/mainflux/pkg/messaging"
+)
+
+type Connclient struct {
```
dborovcanin: You can rename it to `Client` and use `id` instead of `pubID`.

ws/api/endpoints.go
```diff
@@ -0,0 +1,173 @@
...
+		conn, err := upgrader.Upgrade(w, r, nil)
+		if err != nil {
+			logger.Warn(fmt.Sprintf("Failed to upgrade connection to websocket: %s", err.Error()))
+			return
+		}
+
+		// Configure the closing of the ws connection
+		conn.SetCloseHandler(func(code int, text string) error {
+			return svc.Unsubscribe(context.Background(), req.thingKey, req.chanID, req.subtopic)
+		})
```
dborovcanin: As per `SetCloseHandler` documentation:

>  Applications should only set a close handler when the application must perform some action before sending a close message back to the peer.

Can you think of other ways of unsubscribing (maybe when you close the `msgs` channel)?

ws/api/requests.go
```diff
@@ -0,0 +1,13 @@
...
+
+package api
+
+import "github.com/gorilla/websocket"
+
+type connReq struct {
+	thingKey string
+	chanID   string
+	subtopic string
+	conn     *websocket.Conn
```
dborovcanin, line 12: I would suggest using this like a pure DTO and not adding any logic here. You can leave this logic to the Client (because the client operates on the adapter level).

dborovcanin (changes requested):

ws/adapter_test.go
```diff
@@ -0,0 +1,256 @@
...
+func NewThingsClient() mainflux.ThingsServiceClient {
+	return mocks.NewThingsClient(map[string]string{thingKey: chanID})
+}
+
+// func TestMain(m *testing.M) {
+
+// 	TestPublish(&testing.T{})
+// 	TestSubscribe(&testing.T{})
+// 	TestUnsubscribe(&testing.T{})
+// }
```
dborovcanin: Remove unused code/

ws/api/endpoint_test.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+const (
+	chanID   = "1"
+	id       = "1"
+	thingKey = "thing_key"
+	protocol = "ws"
+	// invalidKey = "invalid_key"
+	// stringMsg  = `[{"n":"current","t":-1,"v":1.6}]`
+	// msgJSON    = `{"field1":"val1","field2","val2"}`
```
dborovcanin: Remove.

ws/api/endpoint_test.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	// subtopic, err := parseSubTopic(subtopic)
+	// if err != nil {
+	// 	if header {
+	// 		return fmt.Sprintf("%s/channels/%s/messages", u, chanID), err
+	// 	}
+	// 	return fmt.Sprintf("%s/channels/%s/messages?authorization=%s", u, chanID, thingKey), err
+	// }
+
```
dborovcanin: Remove.

ws/api/endpoint_test.go
```diff
@@ -0,0 +1,144 @@
...
+	// if thingKey == mocks.ServiceErrToken {
+	// 	res.StatusCode = http.StatusServiceUnavailable
+	// } else if thingKey == "invalid" {
+	// 	res.StatusCode = http.StatusForbidden
+	// }
+
+	// if err != nil {
+	// 	res.StatusCode = http.StatusBadRequest
+	// }
```
dborovcanin: Remove.

ws/api/endpoint_test.go
```diff
@@ -0,0 +1,144 @@
...
+		{"connect to invalid channel id", "", "", true, thingKey, http.StatusBadRequest, []byte{}},
+		{"connect with empty thingKey", id, "", true, "", http.StatusForbidden, []byte{}},
+		{"connect with invalid thingKey", id, "", true, "invalid", http.StatusForbidden, []byte{}},
+		{"connect unable to authorize", id, "", true, mocks.ServiceErrToken, http.StatusServiceUnavailable, []byte{}},
+		{"connect and send message with thingKey as query parameter", id, "", false, thingKey, http.StatusSwitchingProtocols, msg},
+		{"connect and send message that cannot be published", id, "", true, thingKey, http.StatusSwitchingProtocols, []byte{}},
+		{"connect and send message to subtopic", id, "subtopic", true, thingKey, http.StatusSwitchingProtocols, msg},
+		{"connect and send message to subtopic with invalid name", id, "sub/a*b/topic", true, thingKey, http.StatusBadRequest, msg},
+		{"connect and send message to nested subtopic", id, "subtopic/nested", true, thingKey, http.StatusSwitchingProtocols, msg},
+		{"connect and send message to all subtopics", id, ">", true, thingKey, http.StatusSwitchingProtocols, msg},
```
dborovcanin: Use the same formatting as in adapter_test.go.

ws/api/endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+		conn, err := upgrader.Upgrade(w, r, nil)
+		if err != nil {
+			logger.Warn(fmt.Sprintf("Failed to upgrade connection to websocket: %s", err.Error()))
+			return
+		}
+
+		req.conn = conn
+		client := ws.NewClient(conn, "")
+
+		// Subscribe using the adapterservice
```
dborovcanin: No need for this comment.

ws/api/endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+				return
+			case ws.ErrFailedConnection:
+				w.WriteHeader(http.StatusServiceUnavailable)
+				req.conn.Close()
+				return
+			default:
+				w.WriteHeader(http.StatusUnauthorized)
+				req.conn.Close()
+				return
+			}
```
dborovcanin: Extract this to `encodeError`.

ws/api/endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+	if authKey == "" {
+		authKeys := bone.GetQuery(r, "authorization")
+		if len(authKeys) == 0 {
+			logger.Debug("Missing authorization key.")
+			return connReq{}, errUnauthorizedAccess
+		}
+		authKey = authKeys[0]
+	}
+
+	//todo: Check this later
```
dborovcanin: Did you check? :)

ws/api/endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+	if authKey == "invalid" {
+		return connReq{}, errUnauthorizedAccess
+	}
+	if authKey == "unavailable" {
+		return connReq{}, ws.ErrEmptyID
+	}
```
dborovcanin: What's this?

ws/api/endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+
+	req := connReq{
+		thingKey: authKey,
+		chanID:   chanID,
+	}
+
+	// Starting parsing subtopic from here
+	channelParts := channelPartRegExp.FindStringSubmatch(r.RequestURI)
+	if len(channelParts) < 2 { // Open conversation for this on gh
```
dborovcanin: Did you open a conversation?

ws/api/endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+	if chanID == "0" || len(chanID) == 0 {
+		return connReq{}, errors.ErrMalformedEntity
+	}
+
+	req := connReq{
+		thingKey: authKey,
+		chanID:   chanID,
+	}
+
+	// Starting parsing subtopic from here
```
dborovcanin: This comment is redundant.

ws/mocks/things.go
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Mainflux
```
dborovcanin: If this is exactly the same as the HTTP adapter mock, you can use that one. Later, we should move mocks to the respective services, rather than copying them.

dborovcanin (changes requested):

ws/mocks/things.go
AryanGodara: Yes, it's the same; I'll import the HTTP one and delete this file.

ws/api/endpoints.go
AryanGodara: Please take a look at the `encodeErro()` function, 
 I will switch to your implementation of `encodeError()`, if current one isn't correct

api/openapi/websocket.yml
```diff
@@ -0,0 +1,41 @@
+openapi: 3.0.1
```
dborovcanin, line 1: You can try using AsyncAPI for Websockets like [here](https://github.com/mainflux/mainflux/tree/master/api/asyncapi).

ws/adapter.go
```diff
@@ -0,0 +1,148 @@
...
+	ErrUnauthorizedAccess = errors.New("missing or invalid credentials provided")
+
+	// ErrEmptyTopic and ErrEmptyID indicate absence of channelID or thingKey in the request
+	ErrEmptyTopic = errors.New("empty topic")
+	ErrEmptyID    = errors.New("empty id")
+)
+
+// Service specifies web socket service API.
+type Service interface {
+	// Publish Message
```
dborovcanin, line 50: Update the comment.

ws/adapter.go
```diff
@@ -0,0 +1,148 @@
...
+	if len(chanID) == 0 || len(thingKey) == 0 {
+		return ErrUnauthorizedAccess
+	}
+
+	thid, err := svc.Authorize(ctx, thingKey, chanID)
+	if err != nil {
+		return ErrUnauthorizedAccess
+	}
+
+	// c.id = thid.GetValue()
```
dborovcanin: Remove commented code.

ws/api/endpoints.go
```diff
@@ -0,0 +1,199 @@
...
+			case errors.ErrMalformedEntity:
+				w.WriteHeader(http.StatusBadRequest)
+				return
+			default:
+				logger.Warn(fmt.Sprintf("Failed to authorize: %s", err.Error()))
+				w.WriteHeader(http.StatusServiceUnavailable)
+				return
+			}
+		}
+
```
dborovcanin: Handle error separately (`encodeError`).

ws/adapter.go
```diff
@@ -0,0 +1,148 @@
...
+	// Publish Message
+	Publish(ctx context.Context, thingKey string, msg messaging.Message) error
+
+	// Subscribes to a channel with specified id.
+	Subscribe(ctx context.Context, thingKey, chanID, subtopic string, client messaging.MessageHandler) error
+
+	// Unsubscribe method is used to stop observing resource.
+	Unsubscribe(ctx context.Context, thingKey, chanID, subtopic string) error
+
+	Authorize(context.Context, string, string) (*mainflux.ThingID, error)
```
dborovcanin: I don't like the idea of exposing the `Authorize` method. The adapter should take care of it silently, without exposing it for general use.

ws/mocks/pubsub.go
```diff
@@ -0,0 +1,25 @@
...
+	return nil
+}
+
+func (pubsub mockPubSub) Unsubscribe(string, string) error {
+	return nil
+}
+
+func (pb mockPubSub) Close() error {
+	return nil
+}
```
dborovcanin: You need to provide a cases where these methods actually fail so you can test your adapter to cover those.

dborovcanin (changes requested):

ws/adapter.go
AryanGodara: This was because I couldn't use it without exposing. (After the change to start using MessageHandler interface in `Subscribe()` function above this).
I don't currently know how to fix this problem, I'll try this tomorrow.

ws/api/endpoints.go
```diff
@@ -0,0 +1,223 @@
...
+// 	case ws.ErrAlreadySubscribed:
+// 		req.conn.Close()
+// 	case ws.ErrFailedConnection:
+// 		w.WriteHeader(http.StatusServiceUnavailable)
+// 		req.conn.Close()
+// 	default:
+// 		w.WriteHeader(http.StatusUnauthorized)
+// 		req.conn.Close()
+// 	}
+// }
```
dborovcanin: Remove commented code.

ws/mocks/pubsub.go
dborovcanin: This does not seem to be addressed, but manually resolved.

ws/api/endpoints.go
```diff
@@ -0,0 +1,223 @@
...
+		w.WriteHeader(http.StatusForbidden)
+	case errMalformedSubtopic:
+		w.WriteHeader(http.StatusBadRequest)
+	case errors.ErrMalformedEntity:
+		w.WriteHeader(http.StatusBadRequest)
+	default:
+		logger.Warn(fmt.Sprintf("Failed to authorize: %s", err.Error()))
+		w.WriteHeader(http.StatusServiceUnavailable)
+	}
+}
```
dborovcanin: Why do we need two methods?

ws/api/endpoints.go
```diff
@@ -0,0 +1,223 @@
...
+
+		if err := svc.Subscribe(context.Background(), req.thingKey, req.chanID, req.subtopic, client); err != nil {
+			fmt.Println("Error in subscribe")
+			logger.Warn(fmt.Sprintf("Failed to subscribe to broker: %s", err.Error()))
+			if err == ws.ErrEmptyID || err == ws.ErrEmptyTopic {
+				w.WriteHeader(http.StatusBadRequest)
+			} else {
+				encodeError(&req, w, err)
+				return
+			}
```
dborovcanin: Avoid using `else`. Also, avoid handling responses (`w.Write` and `w.WriteHeader`) in multiple locations in the code.

dborovcanin (changes requested):

ws/adapter.go
```diff
@@ -0,0 +1,146 @@
...
+		default:
+			return ErrFailedSubscription
+		}
+	}
+
+	return nil
+}
+
+func (svc *adapterService) Unsubscribe(ctx context.Context, thingKey, chanID, subtopic string) error {
+	if len(chanID) == 0 || len(thingKey) == 0 {
```
dborovcanin:
```go
if chanID == "" || thingKey == "" {...}
```

ws/adapter.go
```diff
@@ -0,0 +1,146 @@
...
+	c.id = thid.GetValue()
+
+	subject := fmt.Sprintf("%s.%s", "channels", chanID)
+	if subtopic != "" {
+		subject = fmt.Sprintf("%s.%s", subject, subtopic)
+	}
+
+	if err := svc.pubsub.Subscribe(thid.GetValue(), subject, c); err != nil {
+		switch err {
+		case ErrAlreadySubscribed:
```
dborovcanin: Once #1633 is merged, this case will be no more.

ws/client.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func NewClient(c *websocket.Conn, thingKey string) *Client {
+	return &Client{
+		conn: c,
+		id:   thingKey,
+	}
+}
+
+func (c *Client) Publish(svc Service, logger logger.Logger, thingKey, chanID, subtopic string, msgs <-chan []byte) {
```
dborovcanin: Rename to something like `Do` or `Process` or `Handle`.

ws/api/endpoints.go
```diff
@@ -0,0 +1,212 @@
...
+		if err != nil {
+			encodeError(&req, w, err)
+			return
+		}
+		conn, err := upgrader.Upgrade(w, r, nil)
+		if err != nil {
+			logger.Warn(fmt.Sprintf("Failed to upgrade connection to websocket: %s", err.Error()))
+			return
+		}
+		req.conn = conn
```
dborovcanin, line 36: Please double-check what happens with the WS connection once the header or response is written to the response writer (in the `encodeError`). I'd assume that connections are closed and released, but you need to do some profiling and see.

ws/api/endpoints.go
```diff
@@ -0,0 +1,212 @@
...
+		client := ws.NewClient(conn, "")
+
+		if err := svc.Subscribe(context.Background(), req.thingKey, req.chanID, req.subtopic, client); err != nil {
+			fmt.Println("Error in subscribe")
+			logger.Warn(fmt.Sprintf("Failed to subscribe to broker: %s", err.Error()))
+			encodeError(&req, w, err)
+		}
+
+		// In case previous subscription wasn't closed properly
```
dborovcanin: If `Subscribe` fails, clean up and exit, don't do anything else.

ws/api/endpoints.go
```diff
@@ -0,0 +1,212 @@
...
+				return
+			}
+			client = ws.NewClient(conn, client.GetID())
+		}
+
+		logger.Debug(fmt.Sprintf("Successfully upgraded communication to WS on channel %s", req.chanID))
+		msgs := make(chan []byte)
+
+		// Listen for messages received from the chan messages, and publish them to broker
+		go client.Publish(svc, logger, req.thingKey, req.chanID, req.subtopic, msgs)
```
dborovcanin: Since the adapter is already using Client in the Subscribe, it would be nice to decouple them here (so that client does not use the adapter in Publish. You can do something similar to `listen` (even at the cost of exposing messaging package to the API layer).

dborovcanin (changes requested):

ws/adapter.go
```diff
@@ -0,0 +1,146 @@
...
+	// ErrFailedConnection indicates that service couldn't connect to message broker.
+	ErrFailedConnection = errors.New("failed to connect to message broker")
+
+	// ErrInvalidConnection indicates that client couldn't subscribe to message broker
+	ErrInvalidConnection = errors.New("nats: invalid connection")
+
+	// ErrAlreadySubscribed indicates that client couldn't subscribe, as it was already subscribed
+	ErrAlreadySubscribed = errors.New("already subscribed to topic")
+
+	// ErrUnauthorizedAccesss indicates that client provided missing or invalid credentials
```
dborovcanin: Fix typo.

ws/api/endpoints.go
AryanGodara: Since, now, encodeError is only called here for decodeRequest(r), only if there is an error.
The function will call return, before a ws-connection is ever created (before this `conn, err := upgrader.Upgrade(w, r, nil)`)

And in the case that subscribe() fails, we no longer call encodeError, just do `req.conn.close()` and return

ws/adapter.go
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) Mainflux
+// SPDX-Licence-Identifier: Apache-2.0
```
dborovcanin: Fix header: `License`. Check all the other typos of the same type.

ws/client.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (c *Client) Cancel() error {
+	if c.conn == nil {
+		return nil
+	}
+	return c.conn.Close()
+}
+
+func (c *Client) Handle(msg messaging.Message) error {
+	if msg.GetPublisher() == c.id {
```
dborovcanin, line 36: Add a short explanation comment.

ws/client.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/gorilla/websocket"
+	"github.com/mainflux/mainflux/pkg/messaging"
+)
+
+type Client struct {
+	conn *websocket.Conn
+	id   string
+}
+
+func NewClient(c *websocket.Conn, thingKey string) *Client {
```
dborovcanin: Since this method is never called with `thingKey` argument other than an empty string, you can remove it:
```go
func NewClient(c *websocket.Conn) *Client {...}
```

ws/api/endpoints.go
```diff
@@ -0,0 +1,231 @@
...
+		req.conn = conn
+		client := ws.NewClient(conn, "")
+
+		if err := svc.Subscribe(context.Background(), req.thingKey, req.chanID, req.subtopic, client); err != nil {
+			logger.Warn(fmt.Sprintf("Failed to subscribe to broker: %s", err.Error()))
+			req.conn.Close()
+			return
+		}
+
+		// In case previous subscription wasn't closed properly
```
dborovcanin: This section should not exist. `svc.Subscribe` or, rather, the underlying implementation of the `pubsub.Subscribe` handles such cases.

ws/adapter_test.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	for _, tt := range cases {
+		pubsub.SetFail(tt.fail)
+		if tt.pubsub == true {
+			t.Run(tt.name, func(t *testing.T) {
+				assert.Equal(t, tt.err, svc.Subscribe(context.Background(), tt.thingKey, tt.chanID, tt.subtopic, c))
+			})
+		} else {
+			t.Run(tt.name, func(t *testing.T) {
```
dborovcanin: No need to call `t.Run`, you can directly `assert.Equal`. Use `switch-case` instead of `if-else`.

ws/adapter_test.go
```diff
@@ -0,0 +1,244 @@
...
+	Protocol:  protocol,
+	Payload:   []byte(`[{"n":"current","t":-5,"v":1.2}]`),
+}
+
+func newService(cc mainflux.ThingsServiceClient) (ws.Service, mocks.MockPubSub) {
+	pubsub := mocks.NewPubSub()
+	return ws.New(cc, pubsub), pubsub
+}
+
+func NewThingsClient() mainflux.ThingsServiceClient {
```
dborovcanin: No need for this wrapper, call `httpmock.NewThingsClient(map[string]string{thingKey: chanID})` directly.

ws/api/endpoint_test.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func newHTTPServer(svc ws.Service) *httptest.Server {
+	logger := log.NewMock()
+	mux := api.MakeHandler(svc, logger)
+	return httptest.NewServer(mux)
+}
+
+func NewThingsClient() mainflux.ThingsServiceClient {
+	return httpmock.NewThingsClient(map[string]string{thingKey: chanID})
```
dborovcanin: No need for this wrapper, call `httpmock.NewThingsClient(map[string]string{thingKey: chanID})` directly.

dborovcanin (changes requested):

ws/adapter_test.go
AryanGodara: t.Run makes it a little easier to debug. Since all the testcases become individually available, in Testing section of vsCode.
But I'll make this change!

ws/adapter.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+var (
+	// ErrFailedMessagePublish indicates that message publishing failed.
+	ErrFailedMessagePublish = errors.New("failed to publish message")
+
+	// ErrFailedSubscription indicates that client couldn't subscribe to specified channel
+	ErrFailedSubscription = errors.New("failed to subscribe to a channel")
+
+	// ErrFailedSubscription indicates that client couldn't unsubscribe from specified channel
```
dborovcanin: // ErrFailedUnsubscribe

ws/adapter.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// ErrInvalidConnection indicates that client couldn't subscribe to message broker
+	ErrInvalidConnection = errors.New("nats: invalid connection")
+
+	// ErrUnauthorizedAccess indicates that client provided missing or invalid credentials
+	ErrUnauthorizedAccess = errors.New("missing or invalid credentials provided")
+
+	// ErrEmptyTopic and ErrEmptyID indicate absence of channelID or thingKey in the request
+	ErrEmptyTopic = errors.New("empty topic")
+	ErrEmptyID    = errors.New("empty id")
```
dborovcanin: Separate these comments (because of `godoc`).

ws/client.go
```diff
@@ -0,0 +1,41 @@
...
+	return &Client{
+		conn: c,
+		id:   "",
+	}
+}
+
+func (c *Client) GetID() string {
+	return c.id
+}
+
```
dborovcanin: Is this used?

ws/client.go
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package ws
+
+import (
+	"github.com/gorilla/websocket"
+	"github.com/mainflux/mainflux/pkg/messaging"
+)
+
+type Client struct {
```
dborovcanin, line 12: Add comments for all the exported names.

ws/adapter_test.go
```diff
@@ -0,0 +1,236 @@
...
+			err:      ws.ErrUnauthorizedAccess,
+		},
+	}
+
+	for _, tt := range cases {
+		assert.Equal(t, tt.err, svc.Publish(context.Background(), tt.thingKey, tt.msg))
+	}
+}
+
+func TestSubUnsub(t *testing.T) {
```
dborovcanin: Separate `Subscribe` and `Unsubscribe` tests.

dborovcanin (changes requested):

ws/api/endpoint_test.go
```diff
@@ -0,0 +1,187 @@
...
+			msg:      msg,
+		},
+	}
+
+	for _, tt := range cases {
+		conn, res, err := handshake(ts.URL, tt.chanID, tt.subtopic, tt.thingKey, tt.header)
+		assert.Equal(t, tt.status, res.StatusCode, fmt.Sprintf("expected status code '%d' got '%d'\n", tt.status, res.StatusCode))
+		if err != nil {
+			return
+		}
```
dborovcanin: This must be changed - it makes tests fail silently. You can add an expected error to the test case and do something like:
```go
assert.Equal(t, err, tc.err, fmt.Sprintf(....))
if tc.err != nil {
		err = conn.WriteMessage(websocket.TextMessage, tt.msg)
		assert....
}
```

dborovcanin (changes requested):

ws/api/endpoint_test.go
```diff
@@ -0,0 +1,189 @@
...
+			chanID:   id,
+			subtopic: ">",
+			header:   true,
+			thingKey: thingKey,
+			status:   http.StatusSwitchingProtocols,
+			msg:      msg,
+		},
+	}
+
+	for _, tt := range cases {
```
dborovcanin: What does `tt` stand for? Use `tc` (test case) for the sake of consistency with other tests.

ws/api/endpoint_test.go
```diff
@@ -0,0 +1,189 @@
...
+		conn, res, err := handshake(ts.URL, tt.chanID, tt.subtopic, tt.thingKey, tt.header)
+		assert.Equal(t, tt.status, res.StatusCode, fmt.Sprintf("expected status code '%d' got '%d'\n", tt.status, res.StatusCode))
+
+		if tt.status == http.StatusSwitchingProtocols {
+			assert.Nil(t, err, fmt.Sprintf("unexpected error %s\n", err))
+
+			err = conn.WriteMessage(websocket.TextMessage, tt.msg)
+			assert.Nil(t, err, fmt.Sprintf("unexpected error %s\n", err))
+		}
```
dborovcanin: In assertion messages use `tc.desc` so that you know which exactly case is failing.

ws/api/endpoint_test.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func TestHandshake(t *testing.T) {
+	thingsClient := httpmock.NewThingsClient(map[string]string{thingKey: chanID})
+	svc := newService(thingsClient)
+	ts := newHTTPServer(svc)
+	defer ts.Close()
+
+	cases := []struct {
+		name     string
```
dborovcanin: Rename to `desc` for consistency.

ws/api/endpoints.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+		svc.Publish(context.Background(), req.thingKey, m)
+	}
+	if err := svc.Unsubscribe(context.Background(), req.thingKey, req.chanID, req.subtopic); err != nil {
+		logger.Warn(fmt.Sprintf("Failed to subscribe to broker: %s", err.Error()))
+		req.conn.Close()
+	}
+}
+
+func encodeError(req *connReq, w http.ResponseWriter, err error) {
```
dborovcanin: Please double-check possible errors here, this smells like too many.

ws/client.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Handle handles the sending and receiving of messages via the broker
+func (c *Client) Handle(msg messaging.Message) error {
+	// To prevent publisher from receiving its own published message
+	if msg.GetPublisher() == c.id {
+		return nil
+	}
+
+	return c.conn.WriteMessage(websocket.TextMessage, msg.Payload)
+}
```
dborovcanin: This is not covered by tests.

dborovcanin (changes requested):
Almost there, only minor remarks are left, mostly for testing.

ws/client.go
AryanGodara: I can create a new test `TestPubSub` with a channel, that checks if message is received but that will be kind of repitition of the `pkg/messaging/pubsub` tests.
Should I add this test?

dborovcanin: I meant more like: check if `p.conn.WriteMessage` works correctly - focus on WebSockets here. Like: creating a connection, waiting for messages (plain WebsSokets), and in the test case checking if Handle will send you the message.

AryanGodara: I see, but since this looks like an integration test, should I create a new test file for this, or just add another function to ws/api/endpoint_test.go ?

I have a small ws-client program, I'll modify that code to test the handle function

ws/adapter.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	thid, err := svc.authorize(ctx, thingKey, chanID)
+	if err != nil {
+		return ErrUnauthorizedAccess
+	}
+
+	c.id = thid.GetValue()
+
+	subject := fmt.Sprintf("%s.%s", "channels", chanID)
```
dborovcanin: Extract to constant `chansPrefix = "channels"`.

ws/adapter.go
```diff
@@ -0,0 +1,146 @@
...
+	if chanID == "" || thingKey == "" {
+		return ErrUnauthorizedAccess
+	}
+
+	thid, err := svc.authorize(ctx, thingKey, chanID)
+	if err != nil {
+		return ErrUnauthorizedAccess
+	}
+
+	subject := fmt.Sprintf("%s.%s", "channels", chanID)
```
dborovcanin: Same here.

ws/api/endpoints.go
```diff
@@ -0,0 +1,186 @@
...
+				statusCode = http.StatusUnauthorized
+			case codes.PermissionDenied:
+				statusCode = http.StatusForbidden
+			case codes.Internal:
+				statusCode = http.StatusInternalServerError
+			default:
+				statusCode = http.StatusInternalServerError
+			}
+		}
+	}
```
dborovcanin: These cases are not covered by tests. Since gRPC calls errors are wrapped by `errors.ErrAuthorization` on the adapter (service) level - it looks like gRPC errors will never happen anyways - so we don't need these cases. Please double-check.

ws/api/endpoints.go
```diff
@@ -0,0 +1,186 @@
...
+			logger.Debug(fmt.Sprintf("Closing WS connection: %s", err.Error()))
+			close(msgs)
+			return
+		}
+
+		if err != nil {
+			logger.Warn(fmt.Sprintf("Failed to read message: %s", err.Error()))
+			close(msgs)
+			return
+		}
```
dborovcanin, line 133: These scenarios are not tested, which also results in L153-L156 never executing during tests. It's important to test API layer extensively because the entire protocol support happens here.

dborovcanin (changes requested):

ws/client.go
AryanGodara: I have to still add testcases list, and make some changes, but this is the approach I'm taking ->

Inside adapter_test.go, I've added a new function, to test the handler ->

If svc.Publish works -> connection.ReadMessage works -> c.conn.WriteMesssage inside `handle()` also works.
And I'll add test cases where we publish to ourselves (same chanID, thingKey, topic, etc.) To check if the extra condition inside handle also works.

Wanted to check if this is the right approach before progressing further

```
msgChan = make(chan []byte)
done    = make(chan interface{})

...
(Other vars and constants)
(TestPublish, TestSubscribe, TestUnsubscribe)
...

func TestWebsocketConn(t *testing.T) {
	thingsClient := httpmock.NewThingsClient(map[string]string{thingKey: chanID})
	svc, _ := newService(thingsClient)

	conn, _, err := websocket.DefaultDialer.Dial(socketUrl, nil)
	// defer conn.Close()
	assert.Nil(t, err, fmt.Sprintln("expected nil error when creating new ws connection"))
	c := ws.NewClient(conn)

	err = svc.Subscribe(context.Background(), thingKey, chanID, "", c)
	assert.Nil(t, err, fmt.Sprintln("expected nil error when creating new ws connection"))
	err = svc.Subscribe(context.Background(), thingKey, chanID, msg.Subtopic, c)
	assert.Nil(t, err, fmt.Sprintln("expected nil error when creating new ws connection"))

	go receiveHandler(t, conn)

	err = svc.Publish(context.Background(), thingKey, msg)
	assert.Nil(t, err, fmt.Sprintln("expected nil error when publishing"))
	data := <-msgChan
	receivedMsg := messaging.Message{}
	err = json.Unmarshal(data, &receivedMsg)
	assert.Nil(t, err, fmt.Sprintf("expected no error while unmarshalling: %s", err))
	assert.Equal(t, msg, receivedMsg, fmt.Sprintf("expected %+v, got %+v", msg, receivedMsg))

}

func receiveHandler(t *testing.T, connection *websocket.Conn) {
	defer close(done)

	for {
		_, msg, err := connection.ReadMessage()
		assert.Nil(t, err, fmt.Sprintf("unexpected error while reading ws message: %s", err))
		msgChan <- msg
	}
}

```

ws/api/endpoints.go
AryanGodara: These are tested manually, when I force close or suspend and close, either the terminal, or the ws-client.
And then connect to them again, and in both cases the closure is handled by the ws-adapter.
I'm able to normally connect back using the previously unexpectely closed client, and continue to send and receive messages.

ws/api/endpoints.go
```diff
@@ -0,0 +1,172 @@
...
+		conn, err := upgrader.Upgrade(w, r, nil)
+		if err != nil {
+			logger.Warn(fmt.Sprintf("Failed to upgrade connection to websocket: %s", err.Error()))
+			return
+		}
+		req.conn = conn
+		client := ws.NewClient(conn)
+
+		if err := svc.Subscribe(context.Background(), req.thingKey, req.chanID, req.subtopic, client); err != nil {
+			logger.Warn(fmt.Sprintf("Failed to subscribe to broker: %s", err.Error()))
```
dborovcanin: This log is not needed, `Subscribe` is wrapped by logging middleware.

ws/api/endpoints.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+		svc.Publish(context.Background(), req.thingKey, m)
+	}
+	if err := svc.Unsubscribe(context.Background(), req.thingKey, req.chanID, req.subtopic); err != nil {
+		logger.Warn(fmt.Sprintf("Failed to subscribe to broker: %s", err.Error()))
+		req.conn.Close()
+	}
+}
+
+func encodeError(req *connReq, w http.ResponseWriter, err error) {
```
dborovcanin: No need to pass the request as a pointer.

ws/api/endpoints.go
```diff
@@ -0,0 +1,172 @@
...
+			logger.Warn(fmt.Sprintf("Failed to read message: %s", err.Error()))
+			close(msgs)
+			return
+		}
+
+		msgs <- payload
+	}
+}
+
+func process(svc ws.Service, req *connReq, msgs <-chan []byte) {
```
dborovcanin: No need to pass the request as a pointer.

ws/client_test.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+		msgChan <- message
+	}
+}
+func TestHandle(t *testing.T) {
+	s := httptest.NewServer(http.HandlerFunc(handler))
+	defer s.Close()
+
+	// Convert http://127.0.0.1 to ws://127.0.0.
+	u := "ws" + strings.TrimPrefix(s.URL, "http")
```
dborovcanin: Use `strings.Replace(...)`.

ws/client_test.go
```diff
@@ -0,0 +1,93 @@
...
+		assert.Nil(t, err, fmt.Sprintf("expected nil error from handle, got: %s", err))
+
+		select {
+		case rec := <-msgChan:
+			receivedMsg = rec
+		case <-time.After(time.Duration(5) * time.Second):
+			receivedMsg = []byte{}
+		}
+		assert.Equal(t, tc.expectedPayload, receivedMsg, fmt.Sprintf("%s: expected %+v, got %+v", tc.desc, msg, receivedMsg))
+		// assert.Equal(t, 0, len(receivedMsg), fmt.Sprintf("expected empty message, got %+v", receivedMsg))
```
dborovcanin: Remove.

ws/adapter_test.go
```diff
@@ -0,0 +1,255 @@
...
+			err:      ws.ErrFailedSubscription,
+		},
+		{
+			desc:     "subscribe to channel with valid thingKey, chanID, subtopic",
+			thingKey: thingKey,
+			chanID:   chanID,
+			subtopic: subTopic,
+			pubsub:   true,
+			fail:     false,
+			err:      nil,
```
dborovcanin: Put a success case(s) first. The same goes for the other tests.

dborovcanin (changes requested):

ws/api/endpoints.go
```diff
@@ -149,12 +148,11 @@ func process(svc ws.Service, req *connReq, msgs <-chan []byte) {
 		svc.Publish(context.Background(), req.thingKey, m)
 	}
 	if err := svc.Unsubscribe(context.Background(), req.thingKey, req.chanID, req.subtopic); err != nil {
-		logger.Warn(fmt.Sprintf("Failed to subscribe to broker: %s", err.Error()))
 		req.conn.Close()
 	}
 }
 
-func encodeError(req *connReq, w http.ResponseWriter, err error) {
+func encodeError(req connReq, w http.ResponseWriter, err error) {
```
dborovcanin: Param `req` is not needed.

dborovcanin (changes requested):

dborovcanin (approved):

ws/client_test.go
```diff
@@ -0,0 +1,92 @@
...
+		publisher       string
+		expectedPayload []byte
+	}{
+		{
+			desc:            "handling with different id from ws.Client",
+			publisher:       msg.Publisher,
+			expectedPayload: msg.Payload,
+		},
+		{
+			desc:            "handling with same id as ws.Client",
```
dborovcanin: `handling with the same id as ws.Client (empty by default)`

ws/client_test.go
```diff
@@ -0,0 +1,92 @@
...
+		msg.Publisher = tc.publisher
+		err = c.Handle(msg)
+		assert.Nil(t, err, fmt.Sprintf("expected nil error from handle, got: %s", err))
+
+		select {
+		case rec := <-msgChan:
+			receivedMsg = rec
+		case <-time.After(time.Duration(5) * time.Second):
+			receivedMsg = []byte{}
+		}
```
dborovcanin: Instead of waiting, add a counter on the server side and a flag to indicate whether to wait for messages and check payload or not.

dborovcanin (changes requested):

dborovcanin (dismissed):

dborovcanin (approved):

drasko (approved):
LGTM

drasko:
@manuio this one seems to be ready. PLease review and approve if OK.

manuio (approved):
LGTM

PR #1633, opened 2022-07-26
MF-1630 - Replace old subscriptions with a new one instead of throwing an error

### LINKED ISSUE #1630: Replace old subscriptions with a new one instead of throwing an error
**FEATURE REQUEST**

No.

The implementation of the WS adapter in #1625 pointed out a different problem: What happens if the connection is closed and `Unsubscribe` fails? Since the connection is closed, there is no way for the client to call `Unsubscribe` anymore. And in the NATS (the default) implementation of the `PubSub`, trying to subscribe returns an error if the subscription already exists, leaving the client in the limbo state - it can't unsubscribe (the connection is broken) and there is no way to subscribe (the entry exists due to unsuccessful previous unsubscribe attempt).
The most straightforward workaround for this problem is to adopt the MQTT approach - Calling subscribe for the client that's already subscribed results in **unsubscribe the old entry** and **creating a new subscription** - effectively replacing the old subscription with the new one.

4. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).

This is a must-have because for protocols that support subscribing we can't rely on a connection to never break or calling `Unsubscribe` to never fail.

AryanGodara:
Signed-off-by: AryanGodara [aryangodara03@gmail.com](mailto:aryangodara03@gmail.com)

### What does this do?
Replaced old subscriptions instead the `pubsub.Subscribe()`, instead of returning an `alreadySubscribed` error

Resolves #1630

N/A

No

No

### Notes
N/A


manuio:
@AryanGodara Can you please check why CI is failling?

AryanGodara:
> @AryanGodara Can you please check why CI is failling?

Yes, I'm working on it.
the error is related to sync.mutex. I'm working on it

AryanGodara:
@manuio I have made some changes to pubsub_test.go, can you please check, whether they're appropriate (Whenever you have the time). Or if there is a better way to do this.
In brief, earlier Subscribe() call to already subscribed, simply returned an error. But now, it also deletes the previous subscription. So, I have to call `test to Subscribe` again, before calling the tests for unsubscribing. (As there is a test to call subscribe on already subscribe, between the two)

One other way, to avoid changing tests, is to unsubscribe from previous, then subscribe again, and return the ErrAlreadySubscribed.

I can change to above way of handling Subscriptions, if this is better.

pkg/messaging/nats/pubsub.go
```diff
@@ -104,6 +114,10 @@ func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler)
 		Subscription: sub,
 		cancel:       handler.Cancel,
 	}
+
+	if alreadySubscribed {
+		return ErrAlreadySubscribed
+	}
```
dborovcanin: You should not return an error, but only replace the old client with the new one.

dborovcanin (changes requested):

dborovcanin (changes requested):
@AryanGodara This needs to be done for all the supported message brokers (MQTT and RabbitMQ) so that all the implementations match as much as possible.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -80,9 +79,22 @@ func (ps pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) e
...
+			if err != nil {
+				return err
+			}
+			s = subscription{
+				client: client,
+				topics: []string{topic},
+			}
 		}
-		s.topics = append(s.topics, topic)
+		s.topics = append(s.topics, topic) // Check status of this line
```
dborovcanin: What does this comment mean?

dborovcanin (changes requested):
As you can see, code in different implementations is repeated - mostly the code that holds the subscriptions. You can consider how we can improve this and prevent code repetition.

pkg/messaging/mqtt/pubsub.go
AryanGodara: removing it, I did not understand this comment at first

pkg/messaging/nats/pubsub.go
```diff
@@ -77,7 +78,14 @@ func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler)
...
 		// Check topic ID
 		if _, ok := s[id]; ok {
-			return ErrAlreadySubscribed
+			// Unlocking, so that Unsubscribe() can access ps.subscriptions
+			ps.mu.Unlock()
+			if err := ps.Unsubscribe(id, topic); err != nil {
+				return err
+			}
+			ps.mu.Lock()
+			s = make(map[string]subscription)
```
dborovcanin: You need to replace only an entry in the `s[i]`, not the entire `ps.subscriptions[topic]`. That is, for the given topic (`ps.subscriptions[topic]`), you need to replace an existing subscription under the given ID (`ps.subscriptions[topic][id]`)

dborovcanin (changes requested):

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -79,7 +79,17 @@ func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler)
...
 		if _, ok := s[id]; ok {
-			return ErrAlreadySubscribed
+			// Unlocking, so Unsubscribe() can access ps.subscriptions
+			ps.mu.Unlock()
+			if err := ps.Unsubscribe(id, topic); err != nil {
+				return err
+			}
+			ps.mu.Lock()
+
+			if len(s) == 0 {
```
drasko: What is this? We are checking the length of a map within a loop i which we can go only if the element `id` exists (which should of course mean that map is not empty), and then in the code that is probably never reached we are setting subscriptions map to an uninitialized map `s`?

drasko (changes requested):

pkg/messaging/nats/pubsub.go
```diff
@@ -77,7 +78,16 @@ func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler)
 	case true:
 		// Check topic ID
 		if _, ok := s[id]; ok {
-			return ErrAlreadySubscribed
+			// Unlocking, so that Unsubscribe() can access ps.subscriptions
+			ps.mu.Unlock()
+			if err := ps.Unsubscribe(id, topic); err != nil {
+				return err
+			}
+			ps.mu.Lock()
```
dborovcanin, line 86: You need to be **very** careful with locks. Where do you unlock this one, for example? While there is `defer` at the beginning to save you, what happens if `ps.Unsubscribe` fails and you don't end you **don't** lock it? That's why we need test scenarios where `ps.Unsubcribe` return an error.

pkg/messaging/nats/pubsub.go
```diff
@@ -77,7 +78,16 @@ func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler)
...
+			if len(s) == 0 {
```
dborovcanin: This should work, but only because maps are passed around by pointer (so that `s` is updated by the previous Unsubscribe call), not because it's a good design. Test it.

pkg/messaging/rabbitmq/pubsub.go
dborovcanin: @drasko Because map can be emptied by calling unsubscribe at L84. However, this is dangerous and will probably work only because maps are passed around by pointer. See my previous comment.

dborovcanin (changes requested):

pkg/messaging/nats/pubsub.go
AryanGodara: In this case, I can add:
```
if err != nil {
  ps.mu.Lock()
  return err
```

I was assuming unsubscribe shoudn't fail, since if the subscription already exists, that means the values are correct, and unsubscribe will have no problem.
That was wrong of me to assume, I'll fix this ASAP.

For the ps.mu.Lock() below that, it's unlocked by the defer unlock in the beginning.
Earlier I was using multiple lock and unlock statements inside the switch blocks. But Drasko suggested that it looked untidy, and unreadable. So, we switched to this cleaner approach.
The tests were all passing, and I even tried it with my ws-adapter code.

dborovcanin: `I was assuming unsubscribe shouldn't fail, since if the subscription already exists, that means the values are correct, and unsubscribe will have no problem.` Exactly, that's not something you can control. You can solve it whichever way you want. Hint:  you can call `defer`  from anywhere, it does not have to be at the begging of the function (while that's the usual pattern).

pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -79,7 +80,19 @@ func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler)
 	case true:
 		// Check topic ID
 		if _, ok := s[id]; ok {
-			return ErrAlreadySubscribed
+			// Unlocking, so Unsubscribe() can access ps.subscriptions
+			ps.mu.Unlock()
+			if err := ps.Unsubscribe(id, topic); err != nil {
+				ps.mu.Lock()
```
dborovcanin: This lock happens only to prevent blocking in the deferred `unlock` call. Consider this:
```go
	ps.mu.Lock()
	// Check topic
	s, ok := ps.subscriptions[topic]
	switch ok {
	case true:
		// Check topic ID
		if _, ok := s[id]; ok {
			// Unlocking, so that Unsubscribe() can access ps.subscriptions
			ps.mu.Unlock()
			if err := ps.Unsubscribe(id, topic); err != nil {
				return err
			}

			ps.mu.Lock()
			s, ok = ps.subscriptions[topic]
			if !ok {
				s = make(map[string]subscription)
				ps.subscriptions[topic] = s
			}
		}
	default:
		s = make(map[string]subscription)
		ps.subscriptions[topic] = s
	}
	defer ps.mu.Unlock()
```
It looks more dangerous and complex, but should be correct and not cause any deadlocks - and you improve performance. Talking about performance - this approach holds `mu.Lock` while connecting to the message broker which is suboptimal - sometimes connecting to the broker includes opening a TCP connection, allocating system resources, even filesystem (for QoS2), and can take a lot of time. In the future, we definitely should think of possible improvements here.

dborovcanin (approved):

pkg/messaging/mqtt/pubsub.go
```diff
@@ -73,14 +72,28 @@ func (ps pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) e
...
+			ps.mu.Unlock()
+			if err := ps.Unsubscribe(id, topic); err != nil {
+				return err
+			}
+			ps.mu.Lock()
+			if len(ps.subscriptions) == 0 {
+				client, err := newClient(ps.address, id, ps.timeout)
+				if err != nil {
+					return err
+				}
```
dborovcanin: What if `newClient` call fails? You locked the mutex and exit - it's a deadlock.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub.go
```diff
@@ -81,12 +82,14 @@ func (ps pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) e
...
 			ps.mu.Unlock()
 			if err := ps.Unsubscribe(id, topic); err != nil {
+				ps.mu.Lock()
 				return err
 			}
 			ps.mu.Lock()
 			if len(ps.subscriptions) == 0 {
 				client, err := newClient(ps.address, id, ps.timeout)
 				if err != nil {
+					ps.mu.Unlock()
```
dborovcanin: No need to unlock here - you'll do it in the deferred call.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub.go
```diff
@@ -80,7 +79,24 @@ func (ps pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) e
 	case true:
 		// Check topic
 		if ok = s.contains(topic); ok {
-			return errAlreadySubscribed
+			// Unlocking, so that Unsubscribe() can access ps.subscriptions
+			ps.mu.Unlock()
+			if err := ps.Unsubscribe(id, topic); err != nil {
+				ps.mu.Lock()
```
dborovcanin: Add a comment on why you lock here.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -80,7 +80,23 @@ func (ps pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) e
...
 		// Check topic
 		if ok = s.contains(topic); ok {
-			return errAlreadySubscribed
+			// Unlocking, so that Unsubscribe() can access ps.subscriptions
+			ps.mu.Unlock()
+			if err := ps.Unsubscribe(id, topic); err != nil {
+				ps.mu.Lock() // Lock so that deferred unlock handle it
+				return err
+			}
+			ps.mu.Lock()
```
dborovcanin:
```suggestion
			err := ps.Unsubscribe(id, topic)
			ps.mu.Lock() // Lock so that deferred unlock handle it
			if err != nil {
				return err
			}
```

dborovcanin (dismissed):

pkg/messaging/nats/pubsub.go
```diff
@@ -69,20 +70,33 @@ func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler)
...
+			ps.mu.Unlock()
+			if err := ps.Unsubscribe(id, topic); err != nil {
+				return err
+			}
+
+			ps.mu.Lock()
+			s, ok = ps.subscriptions[topic]
+			if !ok {
+				s = make(map[string]subscription)
+				ps.subscriptions[topic] = s
```
drasko: Can this be contracted into:
```
ps.subscriptions[topic] = make(map[string]subscription)
```

If yes, do we need at all value `s` on line `88`? It looks like it is not used.

drasko (changes requested):

pkg/messaging/nats/pubsub.go
AryanGodara:
```
_, ok = ps.subscriptions[topic]
if !ok {
	ps.subscriptions[topic] = make(map[string]subscription)
}
```
Is this better @drasko ?

drasko: Yes, this looks better to me

drasko: Even more, I would embed check in `if` initialization - no need for two lines.

```go
if _, ok := ...; !ok {
...
}
```

AryanGodara: just a min

dborovcanin: This reassigning of the `s` is necessary here because `ps.subscriptions[topic]` can change as soon as the lock is released. I don't only think of `Unsubscribe` call that will certainly change it, but also possible `Subscribe` calls that wait for the lock to be released - we can't know.

drasko: @dborovcanin the way it is implemented now (in the newest change that @AryanGodara added) looks very nice to me.

Can you please review and approve if everything is OK now.

pkg/messaging/nats/pubsub.go
```diff
@@ -69,20 +70,31 @@ func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler)
...
+			ps.mu.Lock()
+			if _, ok = ps.subscriptions[topic]; !ok {
+				ps.subscriptions[topic] = make(map[string]subscription)
+			}
 		}
 	default:
 		s = make(map[string]subscription)
 		ps.subscriptions[topic] = s
 	}
+	defer ps.mu.Unlock()
```
drasko: I am not sure if this `defer` should be put here or under previous `Lock()`, where it was before. Both are valid, and even if you are putting Unlock-Lock after, it is not confusing, because we know that second `Lock()` is here just to come back to previous state.

I let @dborovcanin decide what place is better for this `defer()`.

drasko (commented):
This starts to look very good to me now. I can approve this one practically, I'll just wait for a few comments from @dborovcanin and we are ready to merge, I think.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -80,7 +80,24 @@ func (ps pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) e
...
 		if ok = s.contains(topic); ok {
-			return errAlreadySubscribed
+			// Unlocking, so that Unsubscribe() can access ps.subscriptions
+			ps.mu.Unlock()
+			err := ps.Unsubscribe(id, topic)
+			ps.mu.Lock() // Lock so that deferred unlock handle it
+			if err != nil {
+				return err
+			}
+			ps.mu.Lock()
```
dborovcanin: Remove. If there is no error - you will end up with to subsequent `mu.Lock()` calls that will certainly cause dedalock.

pkg/messaging/nats/pubsub.go
dborovcanin: This place is better. While it looks a little more dangerous, it's actually better because you do not have to call an artificial `lock` so that an earlier `defer` unlocks it.

dborovcanin (changes requested):
Once the remark in MQTT PubSub is resolved, we're ready.

pkg/messaging/nats/pubsub.go
```diff
@@ -69,20 +70,29 @@ func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler)
...
+				return err
+			}
+
+			ps.mu.Lock()
+			s = ps.subscriptions[topic]
 		}
-	default:
+	}
+	defer ps.mu.Unlock()
+	if s == nil {
```
drasko, line 92: Add comment above that `s` can be changed while `ps.mu` was unlocked

drasko (changes requested):

pkg/messaging/nats/pubsub.go
```diff
@@ -84,6 +84,7 @@ func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler)
 			}
 
 			ps.mu.Lock()
+			// value of s can changed while ps.mu is unlocked
```
drasko: `can be changed` - please fix typo

drasko (changes requested):

drasko (approved):
LGTM

PR #1643, opened 2022-08-15
MF-1640 - Create tests for mqtt pubsub

### LINKED ISSUE #1640: Create tests for mqtt pubsub
**FEATURE REQUEST**

No

Tests are missing for the mqtt implementation of pubsub in pkg/messaging/mqtt

Should have

AryanGodara:
Signed-off-by: aryan <aryangodara03@gmail.com>

### What does this do?
Adds tests for the mqtt implementation of pkg/messaging

Resolves #1640 

Nil

Yes

N/A

### Notes
N/A


pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,286 @@
...
+
+var (
+	msgChan = make(chan messaging.Message)
+	data    = []byte("payload")
+)
+
+func TestPublisher(t *testing.T) {
+	err := pubsub.Subscribe(clientID, fmt.Sprintf("%s.%s", chansPrefix, topic), handler{})
+	require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+	err = pubsub.Subscribe(clientID, fmt.Sprintf("%s.%s.%s.", chansPrefix, topic, subtopic), handler{})
```
drasko: Why are you subscribing 2 times in a row? Are you testing something? If yes, explain what you test in a comment (while other tests below have their `desc`, here it is not so evident).

drasko: Additionally - why are you subscribing at all in something that is called `TestPublisher`?

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,286 @@
...
+		err := publisher.Publish(topic, expectedMsg)
+		require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+
+		receivedMsg := <-msgChan
+		assert.Equal(t, expectedMsg, receivedMsg, fmt.Sprintf("%s: expected %+v got %+v\n", tc.desc, expectedMsg, receivedMsg))
+	}
+}
+
+func TestPubSub(t *testing.T) {
+	// Test Subscribe and Unsubscribe
```
drasko: Are you testing publishing capabilities also of `pubsub`, or only subscription?

drasko (changes requested):
Not bad at all. I think you are close, lets clarify these few remarks I left.

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: pubsub only tests the subscribing and unsubscribing capability

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: We need to subscribe to the topic/subtopic. Then we start publishing messages on that topic/subtopic.

pkg/messaging/mqtt/setup_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	handleInterrupt(pool, container)
+
+	address := fmt.Sprintf("%s:%s", "localhost", container.GetPort("1883"))
+	if err := pool.Retry(func() error {
+		publisher, err = mqtt.NewPublisher(address, 30*time.Second)
+		return err
+	}); err != nil {
+		log.Fatalf("Could not connect to docker: %s", err)
+	}
```
AryanGodara: @drasko 
Can you please review this part?
It seems correct from my side, but the tests are failing to start: `could not connect to docker: reached retry deadline`

pkg/messaging/mqtt/pubsub_test.go
drasko: @dborovcanin why is this called `TestPubSub()`, and not `TestSubscriber()`?

drasko (changes requested):

pkg/messaging/mqtt/pubsub.go
```diff
@@ -46,31 +47,34 @@ type pubsub struct {
 	subscriptions map[string]subscription
 }
 
+// func (ps *pubsub)
+
```
dborovcanin: Remove.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -46,31 +47,34 @@ type pubsub struct {
...
-	ret := pubsub{
+	ret := &pubsub{
 		publisher: publisher{
 			client:  client,
 			timeout: timeout,
 		},
 		address:       url,
 		timeout:       timeout,
 		logger:        logger,
+		mu:            &sync.RWMutex{},
```
dborovcanin: No need to use a pointer here.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -215,6 +219,6 @@ func (sub subscription) delete(topic string) bool {
 	topics := make([]string, len(sub.topics)-1)
 	copy(topics[:index], sub.topics[:index])
 	copy(topics[index:], sub.topics[index+1:])
-	sub.topics = topics
+	// sub.topics = topics
```
dborovcanin: Fix.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,287 @@
...
+	msgChan = make(chan messaging.Message)
+	data    = []byte("payload")
+)
+
+func TestPublisher(t *testing.T) {
+	// Subscribing with topic, and with subtopic, so that we can publish messages
+	err := pubsub.Subscribe(clientID, fmt.Sprintf("%s.%s", chansPrefix, topic), handler{false})
+	require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+	err = pubsub.Subscribe(clientID, fmt.Sprintf("%s.%s.%s.", chansPrefix, topic, subtopic), handler{false})
+	require.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
```
dborovcanin: I would prefer to use pure MQTT client for subscribing (`client.Subscribe`) directly instead of our Subscriber to isolate Pubslih testing from Subscriber. That is, we'll prevent errors in our Subscriber implementation to fail the Publish test. The same goes for our Subscriber - use native MQTT client to publish to prevent coupling between two. The same should be applied to all the other tests but in the separate PR.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,287 @@
...
+				assert.Equal(t, err, pc.errorMessage)
+			}
+		} else {
+			err := pubsub.Unsubscribe(pc.clientID, pc.topic)
+			if pc.errorMessage == nil {
+				require.Nil(t, err, fmt.Sprintf("%s got unexpected error: %s", pc.desc, err))
+			} else {
+				assert.Equal(t, err, pc.errorMessage)
+			}
+		}
```
dborovcanin: This looks too complex, try to simplify.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub_test.go
dborovcanin: This being said, it would be nice to have one test that actually does test this compatibility between Publisher and Subscriber implementations so we're sure that messages are sent and received in a compatible way.

AryanGodara: I will implement this asap!

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: This is the way it's implemented for nats and rabbitmq also.
I'll simplify this, and also make the change in those two files as well.

pkg/messaging/mqtt/pubsub.go
AryanGodara: @dborovcanin This is a problem, that I'm still trying to fix.
If I don't initialize sync.RWMutex{} here, then I get the `nil memory reference error` for some reason. 
This doesn't happen for rabbitmq or nats. This is the only fix I found for the problem, I'm still working on it

AryanGodara:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4764be]

goroutine 139 [running]:
testing.tRunner.func1.2({0x93f400, 0xdd7bd0})
        /usr/lib/go-1.18/src/testing/testing.go:1389 +0x24e
testing.tRunner.func1()
        /usr/lib/go-1.18/src/testing/testing.go:1392 +0x39f
panic({0x93f400, 0xdd7bd0})
        /usr/lib/go-1.18/src/runtime/panic.go:838 +0x207
sync.(*RWMutex).Lock(0x0?)
        /usr/lib/go-1.18/src/sync/rwmutex.go:139 +0x1e
github.com/mainflux/mainflux/pkg/messaging/mqtt.(*pubsub).Subscribe(0xc00007c2d0, {0x9e6f4d, 0x24}, {0xc000029d30, 0xe}, {0xaaf9c8, 0xc000029d28})
        /home/aryan/mainflux/pkg/messaging/mqtt/pubsub.go:77 +0xbc
github.com/mainflux/mainflux/pkg/messaging/mqtt_test.TestPublisher(0xc0002b8ea0?)
        /home/aryan/mainflux/pkg/messaging/mqtt/pubsub_test.go:31 +0x10a
testing.tRunner(0xc0002b8ea0, 0xa1acb8)
        /usr/lib/go-1.18/src/testing/testing.go:1439 +0x102
created by testing.(*T).Run
        /usr/lib/go-1.18/src/testing/testing.go:1486 +0x35f
FAIL    github.com/mainflux/mainflux/pkg/messaging/mqtt 12.346s
FAIL
```

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: I have made some changes, one commented out, and other not commented out.
Please tell if either of them is better than the current version

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: I have implemented this.
Then, changed the TestPubSub to TestSubscriber

Then, created a new function -> TestPubSub, which needs to be simplified, but it basically does what TestPublisher did earlier, ie., use pubsub.Subscribe, so it's basically testing both Publisher and Subscriber together.

pkg/messaging/mqtt/pubsub.go
dborovcanin: If you don't use a pointer - you don't have nil. :)

pkg/messaging/mqtt/pubsub_test.go
dborovcanin: `TestSubscriber` mixes testing `Subscribe` and `Unsubscribe` methods. TBH, I don't mind that because it's a part of the same functionality (even inseparable) but it's a little cleaner to have them separated. The complexity of executing test cases is an indication that we should split them. For `Subscribe` use the same principle as for the `Publish` test - use the plain client to publish messages you expect on the Subscriber side. For `Unsubscribe` - you obviously need to call `Subscribe` (you can do that before iterating through the cases, though), there is nothing we can do about it.

pkg/messaging/mqtt/pubsub_test.go
dborovcanin: @AryanGodara `TestSubscriber` mixes testing `Subscribe` and `Unsubscribe` methods. TBH, I don't mind that because it's a part of the same functionality (even inseparable) but it's a little cleaner to have them separated. The complexity of executing test cases is an indication that we should split them. For `Subscribe` use the same principle as for the `Publish` test - use the plain client to publish messages you expect on the Subscriber side. For `Unsubscribe` - you obviously need to call `Subscribe` (you can do that before iterating through the cases, though), there is nothing we can do about it.

dborovcanin:
Please fix CI.

pkg/messaging/mqtt/pubsub_test.go
dborovcanin: > This is the way it's implemented for nats and rabbitmq also. I'll simplify this, and also make the change in those two files as well.

Let's not use this PR for fixing other tests. You can file a separate PR for that.

pkg/messaging/mqtt/setup_test.go
```diff
@@ -0,0 +1,121 @@
...
+	logger, err = logg.New(os.Stdout, "error")
+	if err != nil {
+		log.Fatalf(err.Error())
+	}
+	if err := pool.Retry(func() error {
+		pubsub, err = mqtt_pubsub.NewPubSub(address, "", 30*time.Second, logger)
+		return err
+	}); err != nil {
+		log.Fatalf("Could not connect to docker: %s", err)
+	}
```
dborovcanin: No need for separate `Publisher` and `PubSub` since you know the underlying implementation of `Publisher` is shared. So use only `NewPubSub` to test both.

pkg/messaging/mqtt/pubsub_test.go
dborovcanin: This entire test is confusing. We need to test: _Publish, Subscribe, Unsubscribe, and PubSub_ integration between `publish` and `subscribe`. `Subscribe` and `Unsubscribe` are coupled since you can't test `Unsubscribe` without `Subscribe`, but that's fine.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub.go
```diff
@@ -147,32 +152,42 @@ func (ps pubsub) Unsubscribe(id, topic string) error {
...
+// 		var msg messaging.Message
+// 		if err := proto.Unmarshal(m.Payload(), &msg); err != nil {
+// 			ps.logger.Warn(fmt.Sprintf("Failed to unmarshal received message: %s", err))
+// 			return
+// 		}
+// 		if err := h.Handle(msg); err != nil {
+// 			ps.logger.Warn(fmt.Sprintf("Failed to handle Mainflux message: %s", err))
+// 		}
+// 	}
+// }
```
dborovcanin: Remove commented code.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -191,6 +206,21 @@ func newClient(address, id string, timeout time.Duration) (mqtt.Client, error) {
...
 
+//? Prototype, removed the proto.Unmarshall()
+func (ps *pubsub) mqttHandler(h messaging.MessageHandler) mqtt.MessageHandler {
+	return func(c mqtt.Client, m mqtt.Message) {
+		var msg messaging.Message
+		msg.Payload = m.Payload()
+		// if err := json.Unmarshal(m.Payload(), &msg); err != nil {
+		// 	ps.logger.Warn(fmt.Sprintf("Failed to unmarshal received message: %s", err))
+		// 	return
+		// }
```
dborovcanin: Remove commented code.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -191,6 +206,21 @@ func newClient(address, id string, timeout time.Duration) (mqtt.Client, error) {
 	return client, nil
 }
 
+//? Prototype, removed the proto.Unmarshall()
```
dborovcanin: What does this mean?

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,607 @@
...
+	msgChan = make(chan messaging.Message)
+	data    = []byte("payload")
+)
+
+// Tests the MQTT Broker being used for the tests
+func TestMQTTBroker(t *testing.T) {
+	// Subscribing with topic, and with subtopic, so that we can publish messages
+	client, err := newClient(address, "clientID", 30*time.Second)
+	defer client.Unsubscribe()
+	defer client.Disconnect(5)
```
dborovcanin: Defer in tests is not particularly useful. You can use `t.Cleanup` instead.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,607 @@
...
+	token.Wait()
+	require.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+}
+
+// Tests only the Publisher
+func TestPublisher(t *testing.T) {
+	// Subscribing with topic, and with subtopic, so that we can publish messages
+	client, err := newClient(address, "clientID1", 30*time.Second)
+	defer client.Unsubscribe()
+	defer client.Disconnect(5)
```
dborovcanin: Same here.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,607 @@
...
+	require.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+}
+
+// Tests only the Subscriber
+func TestSubscribe(t *testing.T) {
+	cases := []struct {
+		desc         string
+		topic        string
+		clientID     string
+		errorMessage error
```
dborovcanin: You can use error type, not error message.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,607 @@
...
+			assert.Equal(t, err, tc.errorMessage, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, err, tc.errorMessage))
+		}
+	}
+}
+
+// Tests only the unsubscriber
+func TestUnsubscribe(t *testing.T) {
+	client, err := newClient(address, "clientID2", 30*time.Second)
+	defer client.Unsubscribe()
+	defer client.Disconnect(5)
```
dborovcanin: Same, use `t.Cleanup`.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,607 @@
...
+			topic:        fmt.Sprintf("%s.%s", chansPrefix, topic),
+			clientID:     "",
+			errorMessage: mqtt_pubsub.ErrEmptyID,
+			pubsub:       false,
+			handler:      handler{false},
+		},
+	}
+
+	for _, pc := range cases {
+
```
dborovcanin: Remove the empty line.

pkg/messaging/mqtt/setup_test.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func mqttHandler(h messaging.MessageHandler) mqtt.MessageHandler {
+	return func(c mqtt.Client, m mqtt.Message) {
+		var msg messaging.Message
+		msg.Payload = m.Payload()
+		// if err := proto.Unmarshal(m.Payload(), &msg); err != nil {
+		// 	logger.Warn(fmt.Sprintf("Failed to unmarshal received message: %s", err))
+		// 	return
+		// }
```
dborovcanin: Remove unused code.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub.go
AryanGodara: There was an error while unmarshalling the message into messaging.Message
I think it was because the test only contained the payload, and not the entire message

pkg/messaging/mqtt/pubsub.go
```diff
@@ -21,13 +21,14 @@ const (
...
-	errConnect                = errors.New("failed to connect to MQTT broker")
-	errSubscribeTimeout       = errors.New("failed to subscribe due to timeout reached")
-	errUnsubscribeTimeout     = errors.New("failed to unsubscribe due to timeout reached")
-	errUnsubscribeDeleteTopic = errors.New("failed to unsubscribe due to deletion of topic")
-	errNotSubscribed          = errors.New("not subscribed")
-	errEmptyTopic             = errors.New("empty topic")
-	errEmptyID                = errors.New("empty ID")
+	ErrConnect                = errors.New("failed to connect to MQTT broker")
+	ErrSubscribeTimeout       = errors.New("failed to subscribe due to timeout reached")
+	ErrUnsubscribeTimeout     = errors.New("failed to unsubscribe due to timeout reached")
```
rodneyosodo: Is this a global variable or only used within the package?

pkg/messaging/mqtt/pubsub.go
```diff
@@ -87,7 +88,9 @@ func (ps pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) e
 			if err != nil {
 				return err
 			}
-			if len(ps.subscriptions) == 0 {
+			if len(ps.subscriptions[id].topics) == 0 {
+				// ps.subscriptions = make(map[string]subscription)
```
rodneyosodo: Remove commented code

pkg/messaging/mqtt/pubsub.go
```diff
@@ -21,13 +21,14 @@ const (
...
-	errEmptyTopic             = errors.New("empty topic")
-	errEmptyID                = errors.New("empty ID")
+	ErrConnect                = errors.New("failed to connect to MQTT broker")
+	ErrSubscribeTimeout       = errors.New("failed to subscribe due to timeout reached")
+	ErrUnsubscribeTimeout     = errors.New("failed to unsubscribe due to timeout reached")
+	ErrUnsubscribeDeleteTopic = errors.New("failed to unsubscribe due to deletion of topic")
+	ErrNotSubscribed          = errors.New("not subscribed")
+	ErrEmptyTopic             = errors.New("empty topic")
+	ErrEmptyID                = errors.New("empty ID")
+	ErrFailed                 = errors.New("failed")
```
rodneyosodo: What has failed?
I have seen an instance where you are using if after failing to handle mainflux message. Probably change it to something specific i.e
```go
	ErrFailedToHandleMessage  = errors.New("failed to handle mainflux message")

```

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,599 @@
...
+	token = client.Unsubscribe(topic)
+	token.Wait()
+	require.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+	token = client.Unsubscribe(fmt.Sprintf("%s.%s", topic, subtopic))
+	token.Wait()
+	require.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+}
+
+// Tests only the Publisher
+func TestPublisher(t *testing.T) {
```
rodneyosodo: Add a case where you publish with empty topic

pkg/messaging/mqtt/setup_test.go
```diff
@@ -0,0 +1,147 @@
...
+			}
+			return
+		}
+		if err := h.Handle(msg); err != nil {
+			logger.Warn(fmt.Sprintf("Failed to handle Mainflux message: %s", err))
+		}
+	}
+}
+
+func newClient(address, id string, timeout time.Duration) (mqtt.Client, error) {
```
rodneyosodo, line 93: Why are you using using here?
```go
func newClient(address, id string, timeout time.Duration) (mqtt.Client, error)
```
I thought when you instantiate publisher or pubsub it will automatically create a client and manage it for you.

rodneyosodo (changes requested):

pkg/messaging/mqtt/setup_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/ory/dockertest/v3"
+)
+
+var (
+	publisher messaging.Publisher
+	pubsub    messaging.PubSub
+
+	logger logg.Logger
+
+	address = "tcp://mqtt.eclipseprojects.io:1883"
```
rodneyosodo: Is this still needed?

rodneyosodo (changes requested):

pkg/messaging/mqtt/setup_test.go
AryanGodara: I am also testing publisher and subscribe-unsubscribe separately.
So, to test only `publish`, I use default mqtt.client.Subscribe(), instead of using the pubsub implementation.
Similarly for testing only subscribe, that's why I've created this newClient function. Same goes for mqttHandler in setup_test

pkg/messaging/mqtt/pubsub.go
AryanGodara: They are used inside mqtt_test package, and may also be exported as required (The error variables are global in the nats and rabbitmq implementations. So, this is also to make mqtt uniform with them).

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,599 @@
...
+	channel     = "9b7b1b3f-b1b0-46a8-a717-b8213f9eda3b"
+	subtopic    = "engine"
+)
+
+var (
+	msgChan = make(chan messaging.Message)
+	data    = []byte("payload")
+)
+
+// Tests the MQTT Broker being used for the tests
```
dborovcanin: Remove this test. We do not test broker or client lib. :)

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,599 @@
...
+
+	token = client.Unsubscribe(topic)
+	token.Wait()
+	require.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+	token = client.Unsubscribe(fmt.Sprintf("%s.%s", topic, subtopic))
+	token.Wait()
+	require.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+}
+
+// Tests only the Publisher
```
dborovcanin: These comments are unnecessary. The name `TestPublish` says enough.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,599 @@
...
+
+	token = client.Unsubscribe(topic)
+	token.Wait()
+	require.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+	token = client.Unsubscribe(fmt.Sprintf("%s.%s", topic, subtopic))
+	token.Wait()
+	require.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+}
+
+// Tests only the Subscriber
```
dborovcanin: Same.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,599 @@
...
+		switch tc.err {
+		case nil:
+			assert.Nil(t, err, "%s: got unexpected error: %s", tc.desc, err)
+		default:
+			assert.Equal(t, err, tc.err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, err, tc.err))
+		}
+	}
+}
+
+// Tests only the unsubscriber
```
dborovcanin: Same.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,599 @@
...
+			case nil:
+				t.Errorf("%s: got unexpected error: %s", tc.desc, err)
+			default:
+				assert.Equal(t, err.Error(), err.Error(), fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, err, tc.err))
+			}
+		}
+	}
+}
+
+// Tests both publisher and subscriber
```
dborovcanin: Same.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,599 @@
...
+			case nil:
+				assert.Nil(t, err, "%s: got unexpected error: %s", tc.desc, err.Error())
+			default:
+				assert.Equal(t, err, tc.err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, err, tc.err))
+			}
+		}
+	}
+}
+
+// Tests both Subscribe and Unsubscribe
```
dborovcanin: No need for this test. These cases are part of either `TestSubscribe` or, more likely, `TestUnsubscribe`.

pkg/messaging/mqtt/setup_test.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+const (
+	username = "mainflux-mqtt"
+	qos      = 2
+)
+
+func TestMain(m *testing.M) {
+
```
dborovcanin: Remove the empty line.

pkg/messaging/mqtt/setup_test.go
dborovcanin: @AryanGodara Please do not mark it as resolved yet but wait for @0x6f736f646f to address the answer and let him close the discussion if the answer is OK.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: The unsubscribe test uses client.subscribe, while testsubunsub uses pubsub to subscribe and unsubscribe, so it tests the complete cycle of sub and unsub.
I created this, since there is also a testpubsub, which tests pubsub.publish and pubsub.subscribe together

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,599 @@
...
+	data    = []byte("payload")
+)
+
+// Tests the MQTT Broker being used for the tests
+func TestMQTTBroker(t *testing.T) {
+	// Subscribing with topic, and with subtopic, so that we can publish messages
+	client, err := newClient(address, "clientID", 30*time.Second)
+	t.Cleanup(func() {
+		defer client.Unsubscribe()
+		defer client.Disconnect(5)
```
dborovcanin: No need to defer, `Cleanup` will do that for you.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,599 @@
...
+		},
+	}
+	for _, tc := range cases {
+		err := pubsub.Subscribe(tc.clientID, tc.topic, tc.handler)
+		switch tc.err {
+		case nil:
+			assert.Nil(t, err, "%s: got unexpected error: %s", tc.desc, err)
+		default:
+			assert.Equal(t, err, tc.err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, err, tc.err))
+		}
```
dborovcanin: Replace with:
```go
assert.Equal(t, err, tc.err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, err, tc.err))
```
Both cases fit with this since it's OK to compare nil error to nil.
Check all the other tests for the same.

pkg/messaging/mqtt/pubsub_test.go
dborovcanin: Makes sense. Please rename it to `TestSubscriber` since you are testing - the `Subscriber` interface. Add some explanation in the comment explaining how this differs from `TestSubscribe` and `TestUnsubscribe`.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub.go
```diff
@@ -96,6 +99,8 @@ func (ps pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) e
 					client: client,
 					topics: []string{topic},
 				}
+
+				ps.subscriptions[id] = s
 			}
 		}
 		s.topics = append(s.topics, topic)
```
dborovcanin: Since `pubsub` holds a map of `subscription` objects and not pointers to them, this won't work. It is changing topics of the copy of subscription and will never reflect in the `subscriptions` map. Please double-check this implementation.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub.go
AryanGodara: I have made changes to this, and checked by printing the contents of the map, and running test cases of subscribing and unsubscribing, and match the subscription map with the expected value.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -207,7 +223,7 @@ func (sub subscription) indexOf(element string) int {
 }
 
 // Deletes a topic from the slice
-func (sub subscription) delete(topic string) bool {
+func (sub *subscription) delete(topic string) bool {
```
dborovcanin: While you fixed the problem of modifying the subscription from this method by attaching it to the pointer rather than the struct, have in mind that this does not exactly solve the problem. Here:
```go
func (ps *pubsub) Unsubscribe(id, topic string) error {
	if id == "" {
		return ErrEmptyID
	}
	if topic == "" {
		return ErrEmptyTopic
	}
	ps.mu.Lock()
	defer ps.mu.Unlock()
	// Check client ID
	s, ok := ps.subscriptions[id]
	switch ok {
	case true:
		// Check topic
		if ok := s.contains(topic); !ok {
			return ErrNotSubscribed
		}
	default:
		return ErrNotSubscribed
	}
	token := s.client.Unsubscribe(topic)
	if token.Error() != nil {
		return token.Error()
	}

	ok = token.WaitTimeout(ps.timeout)
	if !ok {
		return ErrUnsubscribeTimeout
	}
	if ok := s.delete(topic); !ok {
		return ErrUnsubscribeDeleteTopic
	}
	if len(s.topics) == 0 {
		delete(ps.subscriptions, id)
	}
	return token.Error()
}
```
You call it on `s` where `s` is `s, ok := ps.subscriptions[id]` which means that you are effectively calling it on the **copy** of the value in the map and all the changes on `s` will not be reflected in `ps.subscriptions[id]`.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub.go
AryanGodara:
```go
if ok := s.delete(topic); !ok {
		return errUnsubscribeDeleteTopic
		return ErrUnsubscribeDeleteTopic
	}
	ps.subscriptions[id] = s

	if len(s.topics) == 0 {
		delete(ps.subscriptions, id)
	}
	return token.Error()
```

Is this okay, since it updates the map attached to pubsub, and later len(s.topics) is equivalent to len(ps.subscriptions)

dborovcanin: It is, but probably the simplest solution would be to use pointers in `subscriptions`  map like:
```go
type pubsub struct {
	publisher
	logger        log.Logger
	mu            *sync.RWMutex
	address       string
	timeout       time.Duration
	subscriptions map[string]*subscription
}
```

AryanGodara: @dborovcanin , I tested with converting map to pointer, and the complexity remains mostly the same. With more added confusion of maintaining pointers. (And sometimes I get nil pointer dereference panic also)

So, I feel like keeping the current format is the better approach (Since it only adds 2 lines to the program) (where we reassign `subscriptions[id] =s` , to update the original map)

pkg/messaging/mqtt/pubsub.go
```diff
@@ -87,18 +88,19 @@ func (ps pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) e
...
 					return err
 				}
 				s = subscription{
 					client: client,
-					topics: []string{topic},
+					topics: []string{},
 				}
 			}
 		}
 		s.topics = append(s.topics, topic)
```
dborovcanin: This will not work for the same reasons I mentioned before. `s` **is a copy of ps.subscriptions[id]** and will not be affected by `ps.Unsubscribe`. You can test by inspecting the `ps.subscriptions` value - there will be two "topic.engine" topics for the client with `clientid1`.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub.go
```diff
@@ -87,18 +88,20 @@ func (ps pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) e
 			if err != nil {
 				return err
 			}
-			if len(ps.subscriptions) == 0 {
+			s = ps.subscriptions[id]
+			if len(ps.subscriptions[id].topics) == 0 {
```
dborovcanin: Use `s.topics`.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -87,18 +88,20 @@ func (ps pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) e
 			if err != nil {
 				return err
 			}
-			if len(ps.subscriptions) == 0 {
+			s = ps.subscriptions[id]
```
dborovcanin: Add a short comment explaining why you reassign `s` here.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -191,6 +193,23 @@ func newClient(address, id string, timeout time.Duration) (mqtt.Client, error) {
 	return client, nil
 }
 
+func (ps *pubsub) mqttHandler(h messaging.MessageHandler) mqtt.MessageHandler {
+	return func(c mqtt.Client, m mqtt.Message) {
+		var msg messaging.Message
+		if err := proto.Unmarshal(m.Payload(), &msg); err != nil {
```
dborovcanin, line 201: If unmarshalling fails, the payload is likely either empty or messed up. Either way, no need to continue with `Handle`.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub.go
arvindh123: @AryanGodara @dborovcanin 
The nil pointer dereference error might be cause due to *sync.RWMutex, 
https://stackoverflow.com/questions/44215799/initializing-a-rwmutex-crashes-in-golang

We can try with pointer for subscription struct alone like:
```
type pubsub struct {
    publisher
    logger        log.Logger
    mu            sync.RWMutex
    address       string
    timeout       time.Duration
    subscriptions map[string]*subscription
}
```

arvindh123: I tried with above struct, 
And The error happens nil pointer panic happenes due unsubscription in subscribe function  		
```err := ps.Unsubscribe(id, topic)```
The Unscubscribe remove the topic and if the topics is 0 then it deletes the subscription sturct itself, 
We need to need to have some function like ```UnsubscribeForResubscribe``` which will not delete the topic or subscription struct


I tried the ```subscription``` in like : 
```go
func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) error {
	if id == "" {
		return ErrEmptyID
	}
	if topic == "" {
		return ErrEmptyTopic
	}
	ps.mu.Lock()
	defer ps.mu.Unlock()
	// Check client ID
	s, ok := ps.subscriptions[id]
	if !ok || s == nil {
		client, err := newClient(ps.address, id, ps.timeout)
		if err != nil {
			return err
		}
		ps.subscriptions[id] = &subscription{
			client: client,
			topics: []string{topic},
		}
	}
	s = ps.subscriptions[id]
	if ok = s.contains(topic); ok {
		// Unlocking, so that Unsubscribe() can access ps.subscriptions
		ps.mu.Unlock()
		fmt.Println(ps.subscriptions[id])

		err := ps.Unsubscribe(id, topic)

		ps.mu.Lock() // Lock so that deferred unlock handle it
		if err != nil {
			return err
		}
		if len(s.topics) == 0 {
			client, err := newClient(ps.address, id, ps.timeout)
			if err != nil {
				return err
			}
			s = &subscription{
				client: client,
				topics: []string{},
			}
		}
	}
	s.topics = append(s.topics, topic)
	token := s.client.Subscribe(topic, qos, ps.mqttHandler(handler))
	if token.Error() != nil {
		return token.Error()
	}
	if ok := token.WaitTimeout(ps.timeout); !ok {
		return ErrSubscribeTimeout
	}
	return token.Error()
}

```

And ```Unsubscription``` function which was suggested by @dborovcanin
```go

func (ps *pubsub) Unsubscribe(id, topic string) error {
	if id == "" {
		return ErrEmptyID
	}
	if topic == "" {
		return ErrEmptyTopic
	}
	ps.mu.Lock()
	defer ps.mu.Unlock()
	// Check client ID
	s, ok := ps.subscriptions[id]
	switch ok {
	case true:
		// Check topic
		if ok := s.contains(topic); !ok {
			return ErrNotSubscribed
		}
	default:
		return ErrNotSubscribed
	}
	token := s.client.Unsubscribe(topic)
	if token.Error() != nil {
		return token.Error()
	}

	ok = token.WaitTimeout(ps.timeout)
	if !ok {
		return ErrUnsubscribeTimeout
	}
	if ok := s.delete(topic); !ok {
		return ErrUnsubscribeDeleteTopic
	}
	if len(s.topics) == 0 {
		delete(ps.subscriptions, id)
	}
	return token.Error()
}

```

The modification part suggested in subscribe function 

```go
	if ok = s.contains(topic); ok {
		// Unlocking, so that Unsubscribe() can access ps.subscriptions
		ps.mu.Unlock()
		fmt.Println(ps.subscriptions[id])

		err := ps.UnsubscribeForResubscribe(id, topic)

		ps.mu.Lock() // Lock so that deferred unlock handle it
		if err != nil {
			return err
		}
		if len(s.topics) == 0 {
			client, err := newClient(ps.address, id, ps.timeout)
			if err != nil {
				return err
			}
			s = &subscription{
				client: client,
				topics: []string{},
			}
		}
	}
```

pkg/messaging/mqtt/pubsub.go
```diff
@@ -87,18 +88,21 @@ func (ps pubsub) Subscribe(id, topic string, handler messaging.MessageHandler) e
 			if err != nil {
 				return err
 			}
-			if len(ps.subscriptions) == 0 {
+			// ps.subscriptions[id] was updated after call to Unsubscribe(), but s is a copy, so it wasn't updated
```
dborovcanin: @AryanGodara @arvindh123 A combination of your ideas can be something like this:

1. First, this entire `switch` can be replaced with:
```go
	// If the client exists, check if it's subscribed to the topic and unsubscribe if needed.
	switch ok {
	case true:
		if ok := s.contains(topic); ok {
			if err := s.unsubscribe(topic, ps.timeout); err != nil {
				return err
			}
		}
	default:
		client, err := newClient(ps.address, id, ps.timeout)
		if err != nil {
			return err
		}
		s = subscription{
			client: client,
			topics: []string{},
		}
	}
        s.topics = append(s.topics, topic)
	ps.subscriptions[id] = s
...
```
2. Where unsubscribe looks like:
```go
func (s *subscription) unsubscribe(topic string, timeout time.Duration) error {
	token := s.client.Unsubscribe(topic)
	if token.Error() != nil {
		return token.Error()
	}

	if ok := token.WaitTimeout(timeout); !ok {
		return ErrUnsubscribeTimeout
	}
	if ok := s.delete(topic); !ok {
		return ErrUnsubscribeDeleteTopic
	}
	return token.Error()
}
```
3. Modify `Unsubscribe` to also utilize `s.unsubscribe(topic, timeout)` accordingly.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub.go
```diff
@@ -147,32 +152,30 @@ func (ps pubsub) Unsubscribe(id, topic string) error {
...
+	opts := mqtt.NewClientOptions().
+		SetUsername(username).
+		AddBroker(address).
+		SetClientID(id).
+		SetConnectionLostHandler(func(c mqtt.Client, err error) {
+			time.Sleep(500 * time.Millisecond)
+		}).
+		SetReconnectingHandler(func(c mqtt.Client, options *mqtt.ClientOptions) {
+			time.Sleep(500 * time.Millisecond)
+		})
```
dborovcanin: What's this?

pkg/messaging/mqtt/pubsub.go
```diff
@@ -135,10 +140,10 @@ func (ps pubsub) Unsubscribe(id, topic string) error {
 	case true:
 		// Check topic
 		if ok := s.contains(topic); !ok {
-			return errNotSubscribed
+			return ErrNotSubscribed
 		}
 	default:
-		return errNotSubscribed
+		return ErrNotSubscribed
```
dborovcanin, line 140: You can simplify this `switch` with short-circuit evaluation:
```go
s, ok := ps.subscriptions[id]
if !ok || !s.contains(topic) {
	return ErrNotSubscribed
}
```
Also, remove unnecessary comments (_Check client id_ and _Check topic_)

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub.go
AryanGodara: I implemented this when the mosquitto-broker was disconnecting sometimes, and causing tests to fail randomly.
This is just to give time for broker to reconnect, in case of failure while testing.
I found this on stack-overflow, and it fixed that bug of tests randomly failing. And 0.5 seconds is small enough that it doesn't cause tests to run for long durations.

pkg/messaging/mqtt/pubsub.go
AryanGodara: I've implemented this, and all tests are passing!

pkg/messaging/mqtt/pubsub.go
dborovcanin: Did you try and test without them? Do we need both since effectively they do the same thing?

dborovcanin: Looks like the problem is in `TestAll` test. Please check the logic behind it.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,584 @@
...
+			assert.Equal(t, err, tc.err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, tc.err, err))
+		default:
+			err := pubsub.Unsubscribe(tc.clientID, tc.topic)
+			assert.Equal(t, err, tc.err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, tc.err, err))
+		}
+	}
+}
+
+// Test Subscribe, Unsubscribe and Publish functions together (Test Entire PubSub interface)
+func TestAll(t *testing.T) {
```
dborovcanin: Use this test to test only PubSub - Publisher and Subscriber compatibility. No need for Unsubscribe - it's covered in the previous test.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,584 @@
...
+			handler:  handler{false},
+		},
+	}
+	for _, tc := range cases {
+		err := pubsub.Subscribe(tc.clientID, tc.topic, tc.handler)
+		assert.Equal(t, err, tc.err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, err, tc.err))
+	}
+}
+
+func TestPubSub(t *testing.T) {
```
dborovcanin, line 207: This test and `TestSubscribe` are testing Subscribe method - please join them.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub.go
AryanGodara: The problem is, when I subscribe to a topic using pubsub.Subscribe() then publish a message to that topic, using client.Publish() {where client is a basic mqtt-client from `"github.com/eclipse/paho.mqtt.golang"`} both don't work together.
That's why TestAll is failing.
I've removed that test, but still getting this error in TestSubscribe() now. I'm working on fixing this currently.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,390 @@
...
+	channel     = "9b7b1b3f-b1b0-46a8-a717-b8213f9eda3b"
+	subtopic    = "engine"
+)
+
+var (
+	msgChan = make(chan messaging.Message)
+	data    = []byte("payload")
+)
+
+//? Use client to subscribe, Publish() to send messages
```
dborovcanin: Fix comment (`//?`).

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,390 @@
...
+			assert.Equal(t, tc.payload, receivedMsg.Payload, fmt.Sprintf("%s: expected %+v got %+v\n", tc.desc, tc.payload, receivedMsg.Payload))
+		}
+	}
+
+	token = client.Unsubscribe(topic)
+	token.Wait()
+	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+	token = client.Unsubscribe(fmt.Sprintf("%s.%s", topic, subtopic))
+	token.Wait()
+	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
```
dborovcanin: Move this to `tc.Cleanup`, something like:
```go
t.Cleanup(func() {
    token := client.Unsubscribe(topic, fmt.Sprintf("%s.%s", topic, subtopic))
    token.WaitTimeout(3 * time.Second)
    assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
    client.Disconnect(10)
})
```

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,390 @@
...
+	token = client.Subscribe(fmt.Sprintf("%s.%s", topic, subtopic), qos, nil)
+	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+
+	// publish with empty topic
+	err = pubsub.Publish("", messaging.Message{Payload: data})
+	assert.Equal(t, err, mqtt_pubsub.ErrEmptyTopic, fmt.Sprintf("Publish with empty topic: expected: %s, got: %s", mqtt_pubsub.ErrEmptyTopic, err))
+
+	cases := []struct {
+		desc     string
+		topic    string
```
dborovcanin: You are not using this field.

pkg/messaging/mqtt/pubsub_test.go
dborovcanin: @AryanGodara Do not forget this case.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,390 @@
...
+			assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+			token.Wait()
+
+			receivedMsg := <-msgChan
+			assert.Equal(t, expectedMsg.Payload, receivedMsg.Payload, fmt.Sprintf("%s: expected %+v got %+v\n", tc.desc, expectedMsg, receivedMsg))
+		}
+	}
+}
+
+//? Should Test Publish() + Subscribe()
```
dborovcanin: What's with these question marks?

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,390 @@
...
+
+	token = client.Unsubscribe(topic)
+	token.Wait()
+	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+	token = client.Unsubscribe(fmt.Sprintf("%s.%s", topic, subtopic))
+	token.Wait()
+	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+}
+
+//todo: Add client.Publish() -> Token to check if Subscribe() actually worked.
```
dborovcanin: Comment?

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,390 @@
...
+			assert.Equal(t, err, tc.err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, tc.err, err))
+		default:
+			err := pubsub.Unsubscribe(tc.clientID, tc.topic)
+			assert.Equal(t, err, tc.err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, tc.err, err))
+		}
+	}
+}
+
+type handler struct {
+	fail      bool
```
dborovcanin, line 421: This field seems to be unused since it's always false in tests.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,390 @@
...
+				Publisher: tc.clientID,
+				Channel:   channel,
+				Subtopic:  subtopic,
+				Payload:   data,
+			}
+			err := pubsub.Publish(topic, expectedMsg)
+			assert.Nil(t, err, fmt.Sprintf("%s: got unexpected error: %s\n", tc.desc, err))
+
+			receivedMsg := <-msgChan
+			assert.Equal(t, expectedMsg.Payload, receivedMsg.Payload, fmt.Sprintf("%s: expected %+v got %+v\n", tc.desc, expectedMsg, receivedMsg))
```
dborovcanin: Add a comment explaining the logic behind this.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: This is to test failing handler. Useful in TestUnsubscribe. But mqtt broker's Unsubscrribe() function behaves differently than nats and rabbitmq.
I left it here, to make things uniform across all tests, and leave room to change this functionality later.

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: custom comments. To make sure I followed all the remarks😅. I'm removing them.

pkg/messaging/mqtt/pubsub_test.go
dborovcanin: No need to keep it just for the sake of consistency between tests. If it's useful for the `Unsubscribe` test, why don't we use it?

AryanGodara: I added the cancel field to subscription. Now, like we nats and rabbitmq. We have an extra step(function call) before unsubscribing mqtt client-topic.
I've also added the testcases for this.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -123,6 +125,12 @@ func (ps *pubsub) Unsubscribe(id, topic string) error {
 		return ErrNotSubscribed
 	}
 
+	if s.cancel != nil {
+		if err := s.cancel(); err != nil {
+			return err
+		}
+	}
```
dborovcanin: You can make this part of the `s.unsubscribe` method.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -25,28 +25,34 @@ var (
...
 
 	token := client.Subscribe(topic, qos, nil)
 	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
 	token = client.Subscribe(fmt.Sprintf("%s.%s", topic, subtopic), qos, nil)
 	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
 
+	t.Cleanup(func() {
+		token := client.Unsubscribe(topic)
+		token.Wait()
+		assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
```
dborovcanin: You can use a single call since `Unsubscribe` accepts multiple topics.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,428 @@
...
+			assert.Equal(t, tc.payload, receivedMsg.Payload, fmt.Sprintf("%s: expected %+v got %+v\n", tc.desc, tc.payload, receivedMsg.Payload))
+		}
+	}
+
+	token = client.Unsubscribe(topic)
+	token.Wait()
+	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+	token = client.Unsubscribe(fmt.Sprintf("%s.%s", topic, subtopic))
+	token.Wait()
+	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
```
dborovcanin: Remove these now that you added them to `tc.Cleanup`.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,428 @@
...
+			handler:  handler{false, ""},
+		},
+	}
+	for _, tc := range cases {
+		err = pubsub.Subscribe(tc.clientID, tc.topic, tc.handler)
+		assert.Equal(t, err, tc.err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, err, tc.err))
+
+		if tc.err == nil {
+			token := client.Subscribe(tc.topic, qos, nil)
+			assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
```
dborovcanin: Why do you subscribe here? You already subscribed in the `pubsub.Subscribe`.

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: I can't send messages from client without subscribing to that topic.
I tried client.Connect() first, but that is giving an mqtt error of "Not connected".

dborovcanin: @AryanGodara There are multiple problems with this test. First, let's make `msgChan` function-level (on the level of the test function), not a package-level variable. Second, remove `SetDefaultPublishHandler(mqttHandler(handler{false, id}))` from `newClient` - that's what cause this test to **pass**. Third, you will face a deadlock once you do that. I'll let you try and fix it on your own.

AryanGodara: msgChan is used inside the `Handle()` function, which is a method of `handler`. So, if I make it function-level.
I'll have to pass it to `mqttHandler() function`, which is not exported from package:rabbitmq.
So, this will make all testing really difficult.
I can, perhaps redefine it inside each function. 

`SetDefaultPublishHandler(mqttHandler(handler{false, id}))`: I'll do this part today. And figure out how to complete the test.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,426 @@
...
+			handler:  handler{false, "", &msgChan},
+		},
+	}
+	for _, tc := range cases {
+		err = pubsub.Subscribe(tc.clientID, tc.topic, tc.handler)
+		assert.Equal(t, err, tc.err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, err, tc.err))
+
+		if tc.err == nil {
+			token := client.Subscribe(tc.topic, qos, mqttHandler(handler{false, "clientID1", &msgChan}))
+			assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
```
dborovcanin: Why do we still have this - what's the purpose of it in this test and why does the test get stuck if we remove it (hint: because handling in `pubsub.Subscribe` fails, but you need to figure out why)?

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,426 @@
...
+			err := pubsub.Unsubscribe(tc.clientID, tc.topic)
+			assert.Equal(t, tc.err, err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, tc.err, err))
+		}
+	}
+}
+
+type handler struct {
+	fail      bool
+	publisher string
+	msgChan   *chan (messaging.Message)
```
dborovcanin: Why pointer? What are reference types in Go?

dborovcanin (changes requested):
Almost there. You need to understand why something happens when you make changes in code, rather than declaring that it works without getting into what actually changed.

pkg/messaging/mqtt/setup_test.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		log.Fatalf("Could not start container: %s", err)
+	}
+
+	handleInterrupt(m, pool, container)
+
+	address = fmt.Sprintf("%s:%s", "localhost", container.GetPort("1883/tcp"))
+	pool.MaxWait = 120 * time.Second
+
+	logger, err = logg.New(os.Stdout, "error")
```
dborovcanin: Use `logg.Debug.String()` as the log level (also rename `logg` alias) and run tests with `-v` flag. That will help you debug tests. 🙂

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: Because when we create a new client in the first line of the function, we are not setting any handler. And since we're no longer using default message handler, so we need the client.Subscribe() function to set the handler for the client.
One other function to set a message handler for the mqtt-client, without subscribing is client.AddRoute(). I used that but that also resulted in not receiving message on channel.
Then, I also added the client.Connect(), but that is returning error `not connected`.
That's why I stuck with the client.Subscribe() technique😅

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: Reference types are maps channels and slices.
But I thought that since default is pass by value for all types (in general), that's why I used a channel pointer to be safe.
Searching for answer on google, will update it to just chan(messaging.Message) if possible!

AryanGodara: Changed it to non-pointer!

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,427 @@
...
+				Publisher: "clientID1",
+				Channel:   channel,
+				Subtopic:  subtopic,
+				Payload:   data,
+			}
+			data, err := proto.Marshal(&expectedMsg)
+			assert.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+
+			token := client.Publish(tc.topic, qos, false, data)
+			token.WaitTimeout(500)
```
dborovcanin: Use `token.WaitTimeout(100*time.Millisecond)` so we know units. Also, 500ns is quite a short period, especially if the message size is significant.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -65,114 +67,107 @@ func NewPubSub(url, queue string, timeout time.Duration, logger log.Logger) (mes
...
+	opts := mqtt.NewClientOptions().
+		SetUsername(username).
+		AddBroker(address).
+		SetClientID(id).
+		SetConnectionLostHandler(func(c mqtt.Client, err error) {
+			time.Sleep(500 * time.Millisecond)
+		}).
+		SetReconnectingHandler(func(c mqtt.Client, options *mqtt.ClientOptions) {
+			time.Sleep(500 * time.Millisecond)
+		})
```
dborovcanin: Remove.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,427 @@
...
+			}
+
+			// publish message, and then receive it on message Channel
+			err := pubsub.Publish(topic, expectedMsg)
+			assert.Nil(t, err, fmt.Sprintf("%s: got unexpected error: %s\n", tc.desc, err))
+
+			receivedMsg := <-msgChan
+			assert.Equal(t, expectedMsg, receivedMsg, fmt.Sprintf("%s: expected %+v got %+v\n", tc.desc, expectedMsg, receivedMsg))
+		default:
+			assert.Equal(t, err, tc.err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, err, tc.err))
```
dborovcanin: Move this to L243, right after the `err := pubsub.Subscribe(tc.clientID, tc.topic, tc.handler)`. After that, use `if tc.err == nil {publish and so on....}` instead of a switch.
The same as in the previous test.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,427 @@
...
+
+func TestUnsubscribe(t *testing.T) {
+	msgChan := make(chan messaging.Message)
+
+	cases := []struct {
+		desc     string
+		topic    string
+		clientID string
+		err      error
+		pubsub   bool //true for subscribe and false for unsubscribe
```
dborovcanin: Rename to `subscribe` but keep the comment.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,427 @@
...
+			err := pubsub.Unsubscribe(tc.clientID, tc.topic)
+			assert.Equal(t, tc.err, err, fmt.Sprintf("%s: expected: %s, but got: %s", tc.desc, tc.err, err))
+		}
+	}
+}
+
+type handler struct {
+	fail      bool
+	publisher string
+	msgChan   chan (messaging.Message)
```
dborovcanin:
```
msgChan chan messaging.Message
```

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,426 @@
...
+	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+	token = client.Subscribe(fmt.Sprintf("%s.%s", topic, subtopic), qos, mqttHandler(handler{false, "clientID1", msgChan}))
+	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+
+	t.Cleanup(func() {
+		token := client.Unsubscribe(topic, fmt.Sprintf("%s.%s", topic, subtopic))
+		token.WaitTimeout(100 * time.Millisecond)
+		assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+
+		client.Disconnect(5)
```
dborovcanin: Increase to something like 100 (milliseconds).

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,426 @@
...
+			Subtopic:  tc.subtopic,
+			Payload:   tc.payload,
+		}
+		err := pubsub.Publish(topic, expectedMsg)
+		assert.Nil(t, err, fmt.Sprintf("%s: got unexpected error: %s\n", tc.desc, err))
+
+		receivedMsg := <-msgChan
+		if tc.payload == nil {
+			assert.Equal(t, 0, len(receivedMsg.Payload), fmt.Sprintf("%s: expected %+v got %+v\n", tc.desc, tc.payload, receivedMsg.Payload))
+		} else {
```
dborovcanin: Avoid using `else`.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,426 @@
...
+
+func TestPublisher(t *testing.T) {
+	msgChan := make(chan messaging.Message)
+
+	// Subscribing with topic, and with subtopic, so that we can publish messages
+	client, err := newClient(address, "clientID1", 30*time.Second)
+	assert.Nil(t, err, fmt.Sprintf("got unexpected error: %s", err))
+
+	token := client.Subscribe(topic, qos, mqttHandler(handler{false, "clientID1", msgChan}))
+	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
```
dborovcanin, line 43: Introduce some wait to give token some time to process subscription:
```go
token := client.Subscribe(topic, qos, mqttHandler(handler{false, "clientID1", msgChan}))
if ok := token.WaitTimeout(100 * time.Millisecond); !ok {
	assert.Fail(t, fmt.Sprintf("failed to subscribe to topic %s", topic))
}
assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))

token = client.Subscribe(fmt.Sprintf("%s.%s", topic, subtopic), qos, mqttHandler(handler{false, "clientID1", msgChan}))
if ok := token.WaitTimeout(100 * time.Millisecond); !ok {
	assert.Fail(t, fmt.Sprintf("failed to subscribe to topic %s", fmt.Sprintf("%s.%s", topic, subtopic)))
}
assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))

```

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: Should I add comment after: client.Disconnect(100) // 100 milliseconds

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,437 @@
...
+	if ok := token.WaitTimeout(100 * time.Millisecond); !ok {
+		assert.Fail(t, fmt.Sprintf("failed to subscribe to topic %s", topic))
+	}
+	assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+
+	token = client.Subscribe(fmt.Sprintf("%s.%s", topic, subtopic), qos, func(c mqtt.Client, m mqtt.Message) {
+		msgChan <- m.Payload()
+	})
+	if ok := token.WaitTimeout(100 * time.Millisecond); !ok {
+		assert.Fail(t, fmt.Sprintf("failed to subscribe to topic %s", topic))
```
dborovcanin: We use topic with subtopic (`fmt.Sprintf("%s.%s", topic, subtopic)`), not topic.

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,437 @@
...
+			Publisher: "clientID11",
+			Channel:   tc.channel,
+			Subtopic:  tc.subtopic,
+			Payload:   tc.payload,
+		}
+		err := pubsub.Publish(topic, expectedMsg)
+		assert.Nil(t, err, fmt.Sprintf("%s: got unexpected error: %s\n", tc.desc, err))
+
+		data, err := proto.Marshal(&expectedMsg)
+		assert.Nil(t, err, fmt.Sprintf("%s: got unexpected error: %s\n", tc.desc, err))
```
dborovcanin: Use `failed to serialize protobuf error:` instead of generic `got unexpected error:`

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,437 @@
...
+		if tc.err == nil {
+			// use pubsub to subscribe to a topic, and then publish messages to that topic.
+			expectedMsg := messaging.Message{
+				Publisher: "clientID",
+				Channel:   channel,
+				Subtopic:  subtopic,
+				Payload:   data,
+			}
+
+			// publish message, and then receive it on message Channel
```
dborovcanin: Use lovercase `c` in _message channel_ (it's not Mainflux Channel).

dborovcanin (changes requested):

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,437 @@
...
+
+	t.Cleanup(func() {
+		token := client.Unsubscribe(topic, fmt.Sprintf("%s.%s", topic, subtopic))
+		token.WaitTimeout(100 * time.Millisecond)
+		assert.Nil(t, token.Error(), fmt.Sprintf("got unexpected error: %s", token.Error()))
+
+		client.Disconnect(100)
+	})
+
+	// publish with empty topic
```
dborovcanin: // Test publish with an empty topic.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -207,7 +209,7 @@ func (sub subscription) indexOf(element string) int {
 }
 
 // Deletes a topic from the slice
-func (sub subscription) delete(topic string) bool {
+func (sub *subscription) delete(topic string) bool {
```
dborovcanin: Use `s` instead of sub. Do the same for `indexOf` and `contains` methods.

dborovcanin: Also, make comments consistent - make them a full sentence starting with a capital letter and ending with a period.

pkg/messaging/mqtt/pubsub.go
```diff
@@ -21,26 +21,28 @@ const (
...
-	errEmptyTopic             = errors.New("empty topic")
-	errEmptyID                = errors.New("empty ID")
+	ErrConnect                = errors.New("failed to connect to MQTT broker")
+	ErrSubscribeTimeout       = errors.New("failed to subscribe due to timeout reached")
+	ErrUnsubscribeTimeout     = errors.New("failed to unsubscribe due to timeout reached")
+	ErrUnsubscribeDeleteTopic = errors.New("failed to unsubscribe due to deletion of topic")
+	ErrNotSubscribed          = errors.New("not subscribed")
+	ErrEmptyTopic             = errors.New("empty topic")
+	ErrEmptyID                = errors.New("empty ID")
+	ErrFailedHandleMessage    = errors.New("failed to handle mainflux message")
```
dborovcanin: Exported fields need comments for godoc.

dborovcanin (changes requested):
We're very close. A couple of minor remarks and we're ready.

dborovcanin (dismissed):

pkg/messaging/mqtt/setup_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	container, err := pool.Run("eclipse-mosquitto", "1.6.13", nil)
+	if err != nil {
+		log.Fatalf("Could not start container: %s", err)
+	}
+
+	handleInterrupt(m, pool, container)
+
+	address = fmt.Sprintf("%s:%s", "localhost", container.GetPort("1883/tcp"))
+	pool.MaxWait = 120 * time.Second
```
drasko: This is not the big remark because this is a test, but I would consider putting these magic numbers like `120` or `1883` port number in constants somewhere. Same with the Mosquitto version.

If we ever have a need to change these (and I guess this might arrive for Mosquitto version), I would prefer that these are visibly declared at the top of the file.

drasko (dismissed):
I left a remark - I will approve, but please consider correcting a remark as well.
@dborovcanin I let you choose should we insist on this remark.

pkg/messaging/mqtt/setup_test.go
AryanGodara: Pushed these changes @drasko and @dborovcanin.
Is this format correct?

dborovcanin (dismissed):

pkg/messaging/mqtt/pubsub_test.go
```diff
@@ -0,0 +1,437 @@
...
+
+var (
+	data = []byte("payload")
+)
+
+func TestPublisher(t *testing.T) {
+	msgChan := make(chan []byte)
+
+	// Subscribing with topic, and with subtopic, so that we can publish messages.
+	client, err := newClient(address, "clientID1", 30*time.Second)
```
drasko: While we are at it, maybe move this timeout to constant as well, for better visibility/changability.

drasko (changes requested):

pkg/messaging/mqtt/pubsub_test.go
AryanGodara: Pushed, and added one for token timeout also

drasko (approved):
LGTM

drasko:
@0x6f736f646f if this is OK now, please approve

rodneyosodo (approved):
LGTM

dborovcanin (approved):

PR #1645, opened 2022-08-16
MF-1644 - Create Postman collections for API documentation

### LINKED ISSUE #1644: Create Postman collections for API documentation
**FEATURE REQUEST**

Create Postman collections for API documentation.

4. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).
Nice-to-have

EddahMurrey:
Signed-off-by: atrox <edamurrey@outlook.com>

### What does this do?
Adds mainflux postman collection
Resolves #1644 
None
### Notes
N/A


drasko (changes requested):
@EddahMurrey please rename directory to `postman`, it is simpler (no need for `collection`).

Additionally - this looks good, but it also looks huge. Can we break this into different JSON files, one per microservice?

Also, if we break it like this - will we be capable to load multiple of those at the same time in Postman (not have to load one by one, where pone would remove previous)?

@dborovcanin looks like Postman can easily import multiple Swagger files and create collection: https://learning.postman.com/docs/getting-started/importing-and-exporting-data/#importing-api-specifications. If this is the case, do we even need Postman collections in the repo (since it can be easily generated from Swagger files)?

EddahMurrey:
@drasko I think we can be able to load multiple files at the same time. Did you mean like the below picture?
Screenshot from 2022-08-17 12-34-43

drasko:
> @drasko I think we can be able to load multiple files at the same time. Did you mean like the below picture?

Yes - exactly like this.

Then, I would definitely propose that we break this big collection into smaller ones - one per microservice used in Mainflux.

drasko (changes requested):
This is good, just a few corrections:
1. We use the plural in the service name `things`, `users`, etc...
2. Some of the API functionalities are part of another microservice and should be in the same collection - for example `channels` are part of `things` microservice

api/postman/groups.json
```diff
@@ -1,7 +1,7 @@
 {
```
drasko: I think Groups and API Key are currently part of Auth microservice

drasko (changes requested):

dborovcanin:
> @dborovcanin looks like Postman can easily import multiple Swagger files and create collection: https://learning.postman.com/docs/getting-started/importing-and-exporting-data/#importing-api-specifications. If this is the case, do we even need Postman collections in the repo (since it can be easily generated from Swagger files)?

@drasko @EddahMurrey Since that's the case, we probably do not need a separate collection(s). It will only make it more complex to maintain and keep them in sync. However, in that case, we should revisit and update OpenAPI docs and add examples to them - so Postman can generate nice and usable out-of-box collections.

mteodor:
will this be used for testing?

dborovcanin:
> will this be used for testing?

Since we already have Openapi files, and they can be imported into Postman, we decided to improve the examples there and drop this pull request.

PR #1650, opened 2022-08-23
MF-1648 - Add Default MF_BROKER_URL in env file

### LINKED ISSUE #1648: Add Default MF_BROKER_URL
**BUG REPORT**

When starting addon service for example `docker/addons/influxdb-writer` it requires `MF_BROKER_URL` to have been declared as an environment variable. This is due to the fact that we don't have a default `MF_BROKER_URL` in the environment file since we were changing the `MF_BROKER_URL` for the main docker-compose file using make command on startup i.e when running `make run`

The expected results it for the docker container to startup having the `MF_BROKER_URL` variable.

image
image from [@AkshayPilane](https://github.com/AkshayPilane)

```
make run
docker-compose -f addons/influxdb-writer/docker-compose.yml up
docker-compose -f addons/influxdb-reader/docker-compose.yml up
```

Ubuntu 20.04
Mainflux version 0.13.0

A possible work around it to set the default `MF_BROKER_URL` in the `.env` file and on make run change it to either nats url or rabbitmq url

rodneyosodo:

### What does this do?
Add default `MF_BROKER_URL` in the docker env file

Resolves #1648

I have added the default `MF_BROKER_URL` to the docker env file. So with make run we aren't changing the placeholder for `MF_BROKER_URL` in the main docker-compose file but rather in the env file depending on the broker type

No

No

### Notes
N/A


docker/.env
```diff
@@ -8,6 +8,7 @@ MF_NGINX_MQTTS_PORT=8883
 
 # Message Broker
 MF_BROKER_TYPE=nats
+MF_BROKER_URL=nats://broker:${MF_NATS_PORT}
```
manuio: Should we keep `MF_NATS_URL`?  Is this replaced also for addons?

manuio (changes requested):

drasko:
@dborovcanin looks like this one is making problems for deployment on the latest images, so I think this should be taken as a priority.

docker/.env
dborovcanin: We should probably remove `MF_NATS_URL` as @manuio mentioned. Also, we should set `MF_NATS_PORT=4222` **before** `MF_BROKER_URL=nats://broker:${MF_NATS_PORT}`.

dborovcanin (changes requested):

docker/.env
rodneyosodo: No we shouldn't keep `MF_NATS_URL` as it was changed to `MF_BROKER_URL`
The problem was that we were setting it on only the main docker-compose file using make run so it wasn't propagated to addons

Makefile
```diff
@@ -119,10 +119,10 @@ rundev:
 run:
 ifeq ("$(MF_BROKER_TYPE)", "rabbitmq")
 	sed -i "s,file: brokers/.*.yml,file: brokers/rabbitmq.yml," docker/docker-compose.yml
-	sed -i "s,MF_BROKER_URL: .*,MF_BROKER_URL: $$\{MF_RABBITMQ_URL\}," docker/docker-compose.yml
```
drasko, line 122: I would probably consider just having `MF_BROKER_URL` and then let user configure it in different ways, depending on a broker type.

I.e. I am not sure if these fine-grained `MF_RABBITMQ_USER` and similar ENV vars are necessary for Mainflux, they look like overkill to me.

drasko (changes requested):

dborovcanin (approved):

Makefile
dborovcanin: I agree, but that's actually what we have - we only have one `MF_BROKER_URL`.
These `MF_RABBITMQ_...` are needed for the [broker deployment](https://github.com/mainflux/mainflux/blob/master/docker/brokers/rabbitmq.yml) anyways. This obscure way of constructing URL only prevents us from having one setup in `.env` and in RabbitMQ compose, and other values in `MF_BROKER_URL`.

dborovcanin (changes requested):
Maybe a little cleaner approach (at least from a Makefile perspective) would be to keep `MF_NATS_URL` and `MF_RABIITMQ_URL` and simply replace `MF_BROKER_URL` with one of them. Like:
```
## Nats
MF_NATS_PORT=4222
MF_NATS_URL=nats://broker:${MF_NATS_PORT}


## RabbitMQ
MF_RABBITMQ_PORT=5672
MF_RABBITMQ_HTTP_PORT=15672
MF_RABBITMQ_USER=mainflux
MF_RABBITMQ_PASS=mainflux
MF_RABBITMQ_COOKIE=mainflux
MF_RABBITMQ_VHOST=/
MF_RABBITMQ_URL=amqp://${MF_RABBITMQ_USER}:${MF_RABBITMQ_PASS}@broker:${MF_RABBITMQ_PORT}${MF_RABBITMQ_VHOST}


# Message Broker
MF_BROKER_TYPE=nats
MF_BROKER_URL=${MF_NATS_URL}
```

And in Makefile we can have:
```make
run:
ifeq ("$(MF_BROKER_TYPE)", "rabbitmq")
	sed -i "s,file: brokers/.*.yml,file: brokers/rabbitmq.yml," docker/docker-compose.yml
	sed -i "s,MF_BROKER_URL=.*,MF_BROKER_URL=$$\{MF_RABBITMQ_URL\}", docker/.env
else ifeq ("$(MF_BROKER_TYPE)", "nats")
	sed -i "s,file: brokers/.*.yml,file: brokers/nats.yml," docker/docker-compose.yml
	sed -i "s,MF_BROKER_URL=.*,MF_BROKER_URL=$$\{MF_NATS_URL\}", docker/.env 
else
	echo "Invalid broker type"; exit 1
endif
	docker-compose -f docker/docker-compose.yml up
```

drasko:
This:

```makefile
run:
ifeq ("$(MF_BROKER_TYPE)", "rabbitmq")
	sed -i "s,file: brokers/.*.yml,file: brokers/rabbitmq.yml," docker/docker-compose.yml
	sed -i "s,MF_BROKER_URL=.*,MF_BROKER_URL=$$\{MF_RABBITMQ_URL\}", docker/.env
else ifeq ("$(MF_BROKER_TYPE)", "nats")
	sed -i "s,file: brokers/.*.yml,file: brokers/nats.yml," docker/docker-compose.yml
	sed -i "s,MF_BROKER_URL=.*,MF_BROKER_URL=$$\{MF_NATS_URL\}", docker/.env
```

can probably be further contracted as `sed -i "s,file: brokers/.*.yml,file: brokers/rabbitmq.yml," docker/docker-compose.yml` can use `$(MF_BROKER_TYPE)` (which is `rabbitmq` string for example).

rodneyosodo:
> This:
> 
> ```makefile
> run:
> ifeq ("$(MF_BROKER_TYPE)", "rabbitmq")
> 	sed -i "s,file: brokers/.*.yml,file: brokers/rabbitmq.yml," docker/docker-compose.yml
> 	sed -i "s,MF_BROKER_URL=.*,MF_BROKER_URL=$$\{MF_RABBITMQ_URL\}", docker/.env
> else ifeq ("$(MF_BROKER_TYPE)", "nats")
> 	sed -i "s,file: brokers/.*.yml,file: brokers/nats.yml," docker/docker-compose.yml
> 	sed -i "s,MF_BROKER_URL=.*,MF_BROKER_URL=$$\{MF_NATS_URL\}", docker/.env
> ```
> 
> can probably be further contracted as `sed -i "s,file: brokers/.*.yml,file: brokers/rabbitmq.yml," docker/docker-compose.yml` can use `$(MF_BROKER_TYPE)` (which is `rabbitmq` string for example).

@drasko I don't get you on this one. Can you clarify again?

dborovcanin (dismissed):

manuio (dismissed):
LGTM

Makefile
drasko: @0x6f736f646f would it be possible here to use `MF_BROKER_TYPE` to construct `MF_RABBITMQ_URL` - something like:
```bash
MF_${MF_BROKER_TYPE^^}_URL
```
i.e. simply by translating `$MF_BROKER_TYPE` to uppercase and concatenating?

If yes, maybe we can avoid/simplify  if-else.

This is not necessary, I am just curious if it can be more elegant (but maybe it will be less readable).

drasko (changes requested):

rodneyosodo:
@drasko https://github.com/mainflux/mainflux/pull/1650#discussion_r974270630 This is a great suggestion but we still need to change the broker yaml file. From `NATS` to `RABBITMQ`

dborovcanin (approved):

drasko (approved):
LGTM

PR #1655, opened 2022-09-16
MF-1654 - Fix Topic Parsing in RabbitMQ Broker

### LINKED ISSUE #1654: Topic Parsing in RabbitMQ Broker
**BUG REPORT**

To send messages between different adapters for instance sending using mqtt and receiving using coap while using RabbitMQ as the internal message broker

- when you publish with mqtt adapter, mqtt adapter should receive message
- when you publish with mqtt adapter, coap adapter should receive message
- when you publish with coap adapter, mqtt adapter should receive message
- when you publish with coap adapter, coap adapter should receive message
- when you publish with http adapter, mqtt adapter should receive message
- when you publish with http adapter, coap adapter should receive message

- when you publish with mqtt adapter, mqtt adapter and coap adapter receives message only

- Setup Mainflux using RabbitMQ message broker

```bash
 MF_BROKER_TYPE=rabbitmq make mqtt coap http 
 MF_BROKER_TYPE=rabbitmq make docker_dev_mqtt docker_dev_http docker_dev_coap -j 3
 MF_BROKER_TYPE=rabbitmq make run
```
- Generate User auth token using this [Ref](https://docs.mainflux.io/api/#create-token)
- Create a test thing [Ref](https://docs.mainflux.io/api/#create-thing)
```bash
mainflux-cli things create '{"name": "testthingrabbit"}' <user_auth_token>
``` 
- Create a test channel [Ref](https://docs.mainflux.io/api/#create-channel)
```bash
mainflux-cli channels create '{"name": "testchannelrabbit"}' <user_auth_token>
``` 
- Connect thing to channel [Ref](https://docs.mainflux.io/api/#connect)
```bash
mainflux-cli things connect <testthingrabbit_id> <testchannelrabbit_id> <user_auth_token>

``` 
- Subscribe messages using MQTT adapter [Ref](https://docs.mainflux.io/messaging/#mqtt)
```bash
mosquitto_sub -u <testthingrabbit_id> -P <testthingrabbit_key> -t channels/<testchannelrabbit_id>/messages -h localhost
``` 
- Subscribe messages using COAP adapter  [Ref](https://docs.mainflux.io/messaging/#coap)
```bash
coap-cli get channels/<testchannelrabbit_id>/messages -auth <testthingrabbit_key> -o

``` 
- Publish messages using MQTT adapter [Ref](https://docs.mainflux.io/messaging/#mqtt)
```bash
mosquitto_pub -u <testthingrabbit_id> -P <testthingrabbit_key> -t channels/<testchannelrabbit_id>/messages -h localhost -m '[{"bn":"some-base-name:","bt":1.276020076001e+09, "bu":"A","bver":5, "n":"voltage","u":"V","v":120.1}, {"n":"current","t":-5,"v":1.2}, {"n":"current","t":-4,"v":1.3}]'

``` 
- Publish messages using COAP adapter [Ref](https://docs.mainflux.io/messaging/#coap)
```bash
coap-cli post channels/<testchannelrabbit_id>/messages -auth <testthingrabbit_key> -d "hello world"

``` 
- Publish messages using HTTP adapter [Ref](https://docs.mainflux.io/api/#send-messages)
```bash
curl -s -S -i -X POST -H "Authorization: Thing <testthingrabbit_key>" http://localhost/http/channels/<testchannelrabbit_id>/messages -d '[{"bn":"some-base-name:","bt":1.276020076001e+09, "bu":"A","bver":5, "n":"voltage","u":"V","v":120.1}, {"n":"current","t":-5,"v":1.2}, {"n":"current","t":-4,"v":1.3}]'

``` 
- Ubuntu 20.04
- Docker version 20.10.17, build 100c701
- docker-compose version 1.29.2, build 5becea4c

- Possible issue is topic structure difference between NATS and RabbitMQ. The system was initially built with NATS topic configuration. RabbitMQ accepts `#` as wildcard to match more words (e.g., a routing pattern of "channels.<channel_id>.#" matches any routing keys beginning with "channels.<channel_id>.messages" and "channels.<channel_id>.<any_other_subtopic>").
- With this problem I would suggest to change to Topic exchange rather than Direct exchange as currently used. This is because with topic exchange we can even poll messages from all subtopics.
- This now brings a challenge when using different brokers based on the wild cards of each broker. This is because nats and rabbitmq use different wildcard to match any topic for instance in nats it it `>` while in rabbitmq it is `#`. I suggest we have one main message format used by mainflux then the different brokers reformat the topic when publishing and subscribing or unsubscribing. Another way would be to have every broker have a way to format its message so that other services can call it. For instance during mqtt forwarding  and also when parsing URL to topic and subtopic.

rodneyosodo:
### What does this do?
Fixes topic parsed when using rabbitmq message broker

Resolves #1654 

- Instead of using direct exchange in rabbitmq this change implements topic based routing of messages
- Parses NATS topic to rabbitmq understandable topic

Yes

No

### Notes
Possible issue was topic structure difference between NATS and RabbitMQ. The system was initially built with NATS topic configuration. RabbitMQ accepts # as wildcard to match more words (e.g., a routing pattern of "channels.<channel_id>.#" matches any routing keys beginning with "channels.<channel_id>.messages" and "channels.<channel_id>.<any_other_subtopic>").
With this problem I would suggest to change to Topic exchange rather than Direct exchange as currently used. This is because with topic exchange we can even poll messages from all subtopics.
This now brings a challenge when using different brokers based on the wild cards of each broker. This is because nats and rabbitmq use different wildcard to match any topic for instance in nats it it > while in rabbitmq it is #. I suggest we have one main message format used by mainflux then the different brokers reformat the topic when publishing and subscribing or unsubscribing. Another way would be to have every broker have a way to format its message so that other services can call it. For instance during mqtt forwarding and also when parsing URL to topic and subtopic.


pkg/messaging/rabbitmq/pubsub.go
```diff
@@ -95,7 +97,7 @@ func (ps *pubsub) Subscribe(id, topic string, handler messaging.MessageHandler)
 		ps.subscriptions[topic] = s
 	}
 
-	_, err := ps.ch.QueueDeclare(topic, true, true, true, false, nil)
+	_, err := ps.ch.QueueDeclare(topic, true, false, false, false, nil)
```
dborovcanin: Inline: `if _, err := ....`

pkg/messaging/rabbitmq/publisher.go
```diff
@@ -51,16 +52,20 @@ func (pub *publisher) Publish(topic string, msg messaging.Message) error {
...
 		false,
 		amqp.Publishing{
-			Headers:     amqp.Table{},
-			ContentType: "application/octet-stream",
-			AppId:       "mainflux-publisher",
-			Body:        data,
+			Headers:      amqp.Table{},
+			ContentType:  "application/octet-stream",
+			AppId:        "mainflux-publisher",
+			DeliveryMode: 2,
```
drasko: What are those magic numbers? These need to be numerical constants, defined at the top of the file with comments explaining their meaning.

drasko (changes requested):

dborovcanin (approved):

drasko (approved):
LGTM

PR #1669, opened 2022-11-18
MF-1676 - Make Bootstrap fetch Certs data on demand

### LINKED ISSUE #1668: Auto renewal of expired certificates from the PKI issuer
**FEATURE REQUEST**

   **The certs service will renew the list of expired certificates from PKI issuer**     
   
    **nice-to-have**

### LINKED ISSUE #1677: Enable multiple certificates for a single Thing
In CERTS service
With present implementation we can't have multiple certificate for one thing 

The present CERTS Repo structure 
```sql 
CREATE TABLE IF NOT EXISTS certs (
    thing_id     TEXT NOT NULL,
    owner_id     TEXT NOT NULL,
    expire       TIMESTAMPTZ NOT NULL,
    serial       TEXT NOT NULL,
    PRIMARY KEY  (thing_id, owner_id, serial)
);
```
The primary key consist of thing_id, owner_id and serial ,
Here the serial mentions the certificate serial number , The certificate have expire .  
On renewal of certificate for a THING, then new certificate is issued and serial number of certificate will be differ from present 
Unlike thing_id and owner_id , The serial is not constant, 

So to have multiple certificate for one THING, we should have addition column something like UUID or Unique Name for each certificate. 
and the propose table CERTS 
```sql 
CREATE TABLE IF NOT EXISTS certs (
    id_or_name TEXT NOT NULL, 
    thing_id     TEXT NOT NULL,
    owner_id     TEXT NOT NULL,
    expire       TIMESTAMPTZ NOT NULL,
    serial       TEXT NOT NULL,
    PRIMARY KEY  (thing_id, owner_id, id_or_name )
);
```

arvindh123:
### What does this do?
Renew the expired cert, If the cert is not available in bootstrap and thing service, then it will remove all the certs. 

Resolves absmach/supermq#1677  and partially absmach/supermq#1678.


arvindh123:
@mainflux/maintainers 
Hello, 
At present, in certs table, one thing can have Mutiple certificate, in cert table one thing will Mutiple expired certificate, 
So if this PR tries to renew all the expired certificate of one thing, then there will be lot of certificates for a thing, 
So to solve this problem, this PR is approach in different way, One thing can have one certificate.
The cert service will renew expired thing certificate and update the serial number in cert table.

cmd/certs/main.go
```diff
@@ -40,31 +44,46 @@ import (
...
+	defMFUser       = "test@email.com"
+	defMFPass       = "123"
+
+	defThingsESURL  = "localhost:6379"
+	defThingsESPass = ""
+	defThingsESDB   = "0"
+
+	defESURL          = "localhost:6379"
+	defESPass         = ""
+	defESDB           = "0"
```
dborovcanin: Looks like these are not used.

certs/bootstrap/certs.go
```diff
@@ -0,0 +1,154 @@
...
+	token, err := c.sdk.CreateToken(user)
+	if err != nil {
+		return "", err
+	}
+	c.token = token
+
+	return c.token, nil
+}
+
+func (c *bootstrapClient) UpdateCerts(ctx context.Context, thingID, clientCert, clientKey, caCert string) error {
```
dborovcanin: Why don't we use `c.sdk.UpdateBootstrapCerts` here?

certs/certs.go
```diff
@@ -29,4 +35,13 @@ type Repository interface {
 
 	// RetrieveBySerial retrieves a certificate for a given serial ID
 	RetrieveBySerial(ctx context.Context, ownerID, serialID string) (Cert, error)
+
+	//List the certifcate need to be renew before the expiry
+	ListExpiredCerts(ctx context.Context, timeBefore time.Duration, limit, offset uint64) (Page, error)
+
+	// AutoRetrieveByThingID  retrieves list of certificate for a given thing ID without owner ID , used  for AutoRenew process
+	AutoRetrieveByThingID(ctx context.Context, thingID string) (Cert, error)
```
dborovcanin: No need for the `Auto` prefix, it sounds misleading (I know we don't take owner ID into account, but that's fine because we do not have another repo method that does).

certs/postgres/certs.go
```diff
@@ -102,6 +102,31 @@ func (cr certsRepository) Save(ctx context.Context, cert certs.Cert) (string, er
 	return cert.Serial, nil
 }
 
+func (cr certsRepository) Update(ctx context.Context, cert certs.Cert) error {
+	q := `UPDATE certs SET serial = :serial, expire = :expire	WHERE thing_id = :thing_id AND owner_id = :owner_id`
```
dborovcanin: Fix formatting.

certs/postgres/certs.go
```diff
@@ -170,6 +195,77 @@ func (cr certsRepository) RetrieveBySerial(ctx context.Context, ownerID, serialI
 	return c, nil
 }
 
+func (cr certsRepository) ListExpiredCerts(ctx context.Context, timeBefore time.Duration, limit, offset uint64) (certs.Page, error) {
+	whereClause := fmt.Sprintf("WHERE expire <= now()  - interval '%s'", timeBefore)
```
dborovcanin: Are you sure using Database server time is better than using service time? It's possible that those two are not the same and this can be a source of the sneaky errors.

certs/service.go
```diff
@@ -44,43 +55,56 @@ type Service interface {
 
 	// RevokeCert revokes a certificate for a given serial ID
 	RevokeCert(ctx context.Context, token, serialID string) (Revoke, error)
+
+	// Renew the exipred certificate from certs repo
+	RenewCerts(ctx context.Context, bsUpdateRenewCert bool) error
+
+	// Automattically trigger RenewCert function for given renew interval time
```
dborovcanin: Fix typo `Automatically` and add a comment for the next method.

dborovcanin (changes requested):

cmd/certs/main.go
arvindh123: Still Work in Progress for this variables,  ES is used to get notification of deleted things

certs/bootstrap/certs.go
arvindh123: Okay I will try to use c.sdk.UpdateBootstrapCerts

certs/certs.go
arvindh123: The repo method ```RetrieveByThing``` do the same but with owner ID 
```go
RetrieveByThing(ctx context.Context, ownerID, thingID string, offset, limit uint64) (Page, error)
```
So to identify the function with name I kept like that way. I can't able to think for another name at that time , Could you suggest a better name or prefix ?

certs/postgres/certs.go
arvindh123: Okay I agree with your point, I will change the approch

certs/bootstrap/certs.go
arvindh123: @dborovcanin, we can't use c.sdk.UpdateBootstrapCerts , since SDK simply reply with status , and we need more detailed about the response body 
https://github.com/mainflux/mainflux/blob/d73a5d53fec0313a51ca933321ea75de7ad0c90e/pkg/sdk/go/bootstrap.go#L180

certs/certs.go
dborovcanin: You can make it a single method (`RetrieveByThingID`) and make it use userID _only if present_ (i.e. ignore if empty, the same way search and filtering by some criteria usually work). On the service level, you decide how to handle authorization.

certs/bootstrap/certs.go
dborovcanin: > @dborovcanin, we can't use c.sdk.UpdateBootstrapCerts , since SDK simply reply with status , and we need more detailed about the response body
> 
> https://github.com/mainflux/mainflux/blob/d73a5d53fec0313a51ca933321ea75de7ad0c90e/pkg/sdk/go/bootstrap.go#L180

Is this only because SDK does not return return error based on the reponse, but on status code? In that case, we should update SDK to return more specifc error.

certs/api/logging.go
```diff
@@ -90,3 +90,42 @@ func (lm *loggingMiddleware) RevokeCert(ctx context.Context, token, thingID stri
 
 	return lm.svc.RevokeCert(ctx, token, thingID)
 }
+
+func (lm *loggingMiddleware) RenewCerts(ctx context.Context, bsUpdateRenewCert bool) (err error) {
```
dborovcanin: Please align the log message with the rest of the service: `Method <method name> <other useful info> took <duration> to complete <errors if any>`.

certs/bootstrap/certs.go
```diff
@@ -0,0 +1,192 @@
...
+
+var (
+	// ErrUnauthorizedAccess indicates missing or invalid credentials provided
+	// when accessing a protected resource.
+	ErrUnauthorizedAccess = errors.New("unauthorized access to Bootstrap service")
+
+	// ErrUnexpectedBSResponse indicates unexpected response from Bootstrap service.
+	ErrUnexpectedBSResponse = errors.New("unexpected Bootstrap service response")
+
+	ErrUnableToAccess = errors.New("unable to access bootstrap service")
```
dborovcanin: Add comments for exported errors.

certs/bootstrap/certs.go
dborovcanin: What about this?

certs/bootstrap/certs.go
```diff
@@ -0,0 +1,192 @@
...
+	err := json.Unmarshal(b, &content)
+	if err != nil {
+		return errors.ErrNotFound
+	}
+	if msg, ok := content["error"]; ok {
+		return errors.New(msg)
+	}
+	return errors.New(string(b))
+}
+func request(ctx context.Context, method, url string, data []byte, header map[string]string) (*http.Response, error) {
```
dborovcanin: Add an empty line between methods.

certs/certs.go
```diff
@@ -29,4 +35,13 @@ type Repository interface {
...
 	RetrieveBySerial(ctx context.Context, ownerID, serialID string) (Cert, error)
+
+	//List the certifcate need to be renew before the expiry
+	ListExpiredCerts(ctx context.Context, timeBefore time.Duration, limit, offset uint64) (Page, error)
+
+	// 	// AutoRetrieveByThingID  retrieves list of certificate for a given thing ID without owner ID , used  for AutoRenew process
+	// 	AutoRetrieveByThingID(ctx context.Context, thingID string) (Cert, error)
+
+	// // AutoRemoveByThingID removes certificate from DB for a given thing ID without owner ID , used  for AutoRenew process
+	// AutoRemoveByThingID(ctx context.Context, thingID string) error
```
dborovcanin: Remove commented code.

certs/postgres/certs.go
dborovcanin: This one doesn't seem addressed.

certs/postgres/certs.go
```diff
@@ -170,6 +224,77 @@ func (cr certsRepository) RetrieveBySerial(ctx context.Context, ownerID, serialI
...
+// 	q := `DELETE FROM certs WHERE thing_id = :thingID`
+// 	var c certs.Cert
+// 	c.ThingID = thingID
+// 	dbcrt := toDBCert(c)
+// 	if _, err := cr.db.NamedExecContext(ctx, q, dbcrt); err != nil {
+// 		return errors.Wrap(errors.ErrRemoveEntity, err)
+// 	}
+// 	return nil
+// }
+
```
dborovcanin: Remove commented code.

certs/redis/consumer/event.go
```diff
@@ -0,0 +1,14 @@
...
+
+type removeEvent struct {
+	id string
+}
+
+type updateChannelEvent struct {
+	id       string
+	name     string
+	metadata map[string]interface{}
+}
```
dborovcanin: Unused?

certs/service.go
```diff
@@ -20,7 +22,13 @@ var (
 	ErrFailedCertCreation = errors.New("failed to create client certificate")
 
 	// ErrFailedCertRevocation failed to revoke certificate
-	ErrFailedCertRevocation = errors.New("failed to revoke certificate")
+	ErrFailedCertRevocation = errors.New("failed to revoke certificate in PKI")
+
+	errThingNotInBS = errors.New("Thing not found in bootstrap service")
```
dborovcanin: Comment exported errors.

dborovcanin (changes requested):

certs/certs.go
```diff
@@ -3,7 +3,9 @@
 
 package certs
 
-import "context"
+import (
+	"context"
+)
```
dborovcanin, line 8:
```go
import context
```

certs/bootstrap/certs.go
dborovcanin: Please don't forget to address this one.

pkg/errors/types.go
```diff
@@ -34,6 +34,12 @@ var (
 	// ErrViewEntity indicates error in viewing entity or entities
 	ErrViewEntity = New("view entity failed")
 
+	// ErrScanEntity indicates error in viewing entity or entities
+	ErrScanEntity = New("failed to scan entity ")
+
+	// ErrScanCountEntity indicates error in viewing entity or entities
+	ErrScanCountEntity = New("failed to count entity")
+
```
dborovcanin: These errors are not used.

pkg/sdk/go/errors/errors.go
```diff
@@ -0,0 +1,182 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package errors
```
dborovcanin: Since this is practically a rewriting of the MF errors package (because the implementation details from our errors are not exported), we should probably move this implementation to `pkg/errors/sdk_errors.go` and do something like:

```go
package errors

import (
	"encoding/json"
	"fmt"
	"io"
	"net/http"
)

var (
	errFailedToReadBody = New("failed to read http response body ")
	errRespBodyNotJSON  = New("response body is not vaild JSON ")
	errJSONKeyNotFound  = New("response body expected error message json key not found ")
)

// SDKError type for SDK
type SDKError struct {
	*customError
	statusCode int
	status     string
}

func (e *SDKError) StatusCode() int {
	return e.statusCode
}

func (e *SDKError) Status() string {
	return e.status
}

// NewSDKErr returns the SDK error that formats as the given text.
func NewSDKErr(text string) SDKError {
	return SDKError{
		customError: &customError{
			msg: text,
			err: nil,
		},
	}
}

// NewWithStatus returns an Error that formats as the given text.
func NewWithStatus(statusCode int, status, text string) SDKError {
	return SDKError{
		statusCode: statusCode,
		status:     status,
		customError: &customError{
			msg: text,
			err: nil,
		},
	}
}

func EncodeError(resp *http.Response, errorMsgJsonKey string, expectedStatusCodes ...int) error {
	for expectedStatusCode := range expectedStatusCodes {
		if resp.StatusCode == expectedStatusCode {
			return nil
		}
	}

	b, bErr := io.ReadAll(resp.Body)
	if bErr != nil {
		e := Wrap(errFailedToReadBody, bErr)
		return Wrap(NewWithStatus(resp.StatusCode, resp.Status, ""), e)
	}

	if errorMsgJsonKey == "" {
		return NewWithStatus(resp.StatusCode, resp.Status, string(b))
	}

	err := json.Unmarshal(b, &content)
	if err != nil {
		e := Wrap(errRespBodyNotJSON, New(string(b)))

		return Wrap(NewWithStatus(resp.StatusCode, resp.Status, ""), e)
	}

	if msg, ok := content[errorMsgJsonKey]; ok {
		return NewWithStatus(resp.StatusCode, resp.Status, fmt.Sprintf("%v", msg))
	}
	e := Wrap(errJSONKeyNotFound, New(string(b)))
	return Wrap(NewWithStatus(resp.StatusCode, resp.Status, ""), e)
}
```
It's OK to have this in `pkg/errors` because that package does not have any other complex imports to pollute our SDK (and we already have it as a dependency).

pkg/sdk/go/errors/errors.go
```diff
@@ -0,0 +1,182 @@
...
+	// Status returns error Status
+	Status() string
+}
+
+var _ Error = (*customError)(nil)
+
+// customError represents a Mainflux  SDK error
+type customError struct {
+	statusCode int
+	status     string
```
dborovcanin: Do we need both code and status (since it's fixed relation between status and code)?

pkg/sdk/go/errors/errors.go
```diff
@@ -0,0 +1,182 @@
...
+func NewWithStatus(statusCode int, status, text string) Error {
+	return &customError{
+		statusCode: statusCode,
+		status:     status,
+		msg:        text,
+		err:        nil,
+	}
+}
+
+func EncodeError(resp *http.Response, errorMsgJsonKey string, expectedStatusCodes ...int) error {
```
dborovcanin: Since currently, we have only one `errorMsgJsonKey`, which is `error`, do we need this param? Also, multiple `expectedStatusCodes` seems a lot like upsert only case.

dborovcanin (changes requested):

pkg/errors/types.go
arvindh123: removed

certs/api/logging.go
arvindh123: The method returns only error 
The error contains all the details,

certs/certs.go
arvindh123: The functions no more needed removed

certs/postgres/certs.go
arvindh123: removed

certs/redis/consumer/event.go
arvindh123: removed

pkg/sdk/go/errors/errors.go
arvindh123: moved to pkg/errors

pkg/sdk/go/errors/errors.go
arvindh123: No need , removed status

pkg/sdk/go/errors/errors.go
arvindh123: removed errorMsgJsonKey

certs/certs.go
arvindh123: the  ```AutoRetrieveByThingID``` functions are no more need , the function are removed
the ```AutoRemoveByThingID``` function name changed to ```ThingCertsRevokeHandler```

certs/postgres/certs.go
arvindh123: Done, expire calculate from service time

certs/postgres/certs.go
arvindh123: I hope, I fixed the formatting

certs/service.go
arvindh123: No more this function is needed in code. It was removed

cmd/certs/main.go
arvindh123: removed unsed

certs/bootstrap/certs.go
arvindh123: done

certs/bootstrap/certs.go
arvindh123: done

certs/bootstrap/certs.go
arvindh123: yes it use sdk now

certs/certs.go
arvindh123: done

certs/service.go
arvindh123: done

arvindh123:
This PR depends on the following issues 
https://github.com/mainflux/mainflux/issues/1677
https://github.com/mainflux/mainflux/issues/1676

dborovcanin:
The Bootstrap service has been moved to the Magistrala repository.  @arvindh123  Please send this PR to Magistrala.

PR #1671, opened 2022-11-22
MF-1670 - Improve error handling in SDK

### LINKED ISSUE #1670: Improve error handling in SDK
**FEATURE REQUEST**

There is no such issue.

Currently, our SDK returns a generic error and status code of the response that caused it, not taking into account **what's the actual error return by the underlying service**. For example, when updating Thing or Channel, we return something like:
`failed to update: status 400`. This tells the end-user - exactly nothing. First, if `sdk.Update` returns an error - it's obviously _failed to update_, so that returned error that says it's failed update tells nothing. Second, status code 400 can be caused by so many things on the backend since one status code can be bound with an enormous amount of errors.
To fix this, we should update SDK so it uses the response body to read the error message service returned and remove useless generic errors such as _failed to update_.

This is a must-have because proper SDK error handling is extremely important in some cases, such as here #1669.

dborovcanin:

### What does this do?
This pull request fixes error handling on the Maifnlux SDK.

This pull request resolves #1670.

Errors returned by SDK are changed to include server errors for the respective request. 

Yes.

N/A

### Notes
N/A

### DIFF
--- a/pkg/sdk/go/sdk.go
+++ b/pkg/sdk/go/sdk.go
@@ -6,7 +6,6 @@ package sdk
 import (
 	"crypto/tls"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -15,6 +14,7 @@ import (
 
 	"github.com/mainflux/mainflux"
 	"github.com/mainflux/mainflux/internal/apiutil"
+	"github.com/mainflux/mainflux/pkg/errors"
 )
 
 const (
@@ -27,6 +27,8 @@ const (
 	// CTBinary represents binary content type.
 	CTBinary ContentType = "application/octet-stream"
 )
+
+var errEncodeError = errors.New("failed to encode response error")
 
 var (
 	// ErrFailedCreation indicates that entity creation failed.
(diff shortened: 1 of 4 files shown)


arvindh123:
Please add status code to errors.
To find the error exact, we need combination of status code and response body, because of error encoding ```(encodeError)``` in ```MakeHandler```  ,

dborovcanin:
> Please add status code to errors. To find the error exact, we need a combination of status code and response body, because of error encoding `(encodeError)` in `MakeHandler` ,

Makes sense. Since SDK is made for developers, including the status code to an error makes sense so the user can precisely determine what exactly happened. We could put a status code to the error message - but that would make things only more complex and this kind of transparency in SDK makes it harder to use. We can create separate errors (that implement our Error interface, ofc) for SDK and include the status code there. Also, potentially in the future, an error response will include more than `{"error": "message"}` JSON, so we'll need even more complex handling on the SDK which would also justify adding a new, SDK error. @drasko What do you think?

drasko:
I think this makes sense completely. I like the idea of evolved `error` structure.

dborovcanin:
Closing as a duplicate of #1674.

PR #1674, opened 2022-11-27
MF-1670 - Improve error handling in SDK

### LINKED ISSUE #1670: Improve error handling in SDK

AryanGodara:
Signed-off-by: aryan <aryangodara03@gmail.com>

This resolves #1670 

Yes

NA
### Notes
NA


pkg/sdk/go/channels_test.go
```diff
@@ -562,5 +564,10 @@ func TestDeleteChannel(t *testing.T) {
 	for _, tc := range cases {
 		err := mainfluxSDK.DeleteChannel(tc.chanID, tc.token)
 		assert.Equal(t, tc.err, err, fmt.Sprintf("%s: expected error %s, got %s", tc.desc, tc.err, err))
+		// fmt.Println()
```
drasko: Dead code

drasko (changes requested):

things/api/auth/http/endpoint.go
```diff
@@ -7,14 +7,15 @@ import (
...
 	"github.com/go-kit/kit/endpoint"
+	"github.com/mainflux/mainflux/pkg/errors"
 	"github.com/mainflux/mainflux/things"
 )
 
 func identifyEndpoint(svc things.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(identifyReq)
 		if err := req.validate(); err != nil {
-			return nil, err
```
dborovcanin: Why did you change this?

things/api/auth/http/endpoint.go
```diff
@@ -34,7 +35,7 @@ func canAccessByKeyEndpoint(svc things.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(canAccessByKeyReq)
 		if err := req.validate(); err != nil {
-			return nil, err
```
dborovcanin: And this.

pkg/sdk/go/sdk.go
```diff
@@ -408,3 +410,20 @@ func (pm PageMetadata) query() (string, error) {
 	}
 	return q.Encode(), nil
 }
+
+func encodeError(body []byte, status ...int) error {
```
dborovcanin: Extract errors to a separate file.

dborovcanin: Why does it accept multiple statuses?

pkg/sdk/go/sdk.go
```diff
@@ -408,3 +410,20 @@ func (pm PageMetadata) query() (string, error) {
...
+	}{}
+
+	if err := json.Unmarshal(body, &e); err != nil {
+		return errors.Wrap(errEncodeError, err)
+	}
+
+	if len(status) > 0 {
+		httpStatus := fmt.Sprintf("%d %s", status[0], http.StatusText(status[0]))
+		return errors.Wrap(errors.New(e.Err), errors.New(httpStatus))
+	}
```
dborovcanin: This looks obscure (which goes along with the multiple status codes remark).

dborovcanin (changes requested):

things/api/auth/http/endpoint.go
AryanGodara: The default error being returned from validate function -> which is ErrBearerToken, causes the http response to return with 500 internal server error.
But, it should ideally return Statusunauthorized. That's why I changed this to return ErrAuthentication.

things/api/auth/http/endpoint.go
AryanGodara: Same reason as above.

pkg/sdk/go/sdk.go
AryanGodara: I initially created errors with only the response body, and not the status code. So this is just so we can encode error with or without using http status. I'll change it back to using just one status.

things/api/auth/http/endpoint.go
dborovcanin: In that case, it should be changed [here](https://github.com/mainflux/mainflux/blob/933b5dedce123063c6e2d17d0c6e842d2a203de9/things/api/auth/http/transport.go#L118).

dborovcanin (changes requested):
We need a special error for SDK for the reasons mentioned here: https://github.com/mainflux/mainflux/pull/1671#issuecomment-1323651122

Something similar to this: https://github.com/mainflux/mainflux/pull/1669/files#diff-0badf386bee49d4a18b00e4b2f75a143d2d00ae36c5cd7226707dee4fa059836

things/api/auth/http/endpoint.go
AryanGodara: I made changes to transport.go, and reverted back to the original code in endpoints.go

pkg/errors/sdk_errors.go
```diff
@@ -0,0 +1,42 @@
...
+// NewSDKErrorWithStatus returns an SDK Error setting the status code.
+func NewSDKErrorWithStatus(msg string, statusCode int) SDKError {
+	return &sdkError{
+		statusCode: statusCode,
+		customError: &customError{
+			msg: msg,
+			err: nil,
+		},
+	}
+}
```
arvindh123: Hello @AryanGodara , @dborovcanin 

@dborovcanin as you suggested this approach in CERTS Auto renewal PR, 
 Here by I'm proposing the same approach for all the SDK 

Code proposal for ```pkg/errors/sdk_error.go```

```go 

// Copyright (c) Mainflux

package errors

import (
	"encoding/json"
	"io"
	"net/http"
)

var (
	errFailedToReadBody = New("failed to read http response body")
	errRespBodyNotJSON  = New("response body is not a valid JSON")
	errJSONKeyNotFound  = New("response body expected error message json key not found")
)

// SDKError is an error type for Mainflux SDK.
type SDKError interface {
	Error
	StatusCode() int
}

var _ SDKError = (*sdkError)(nil)

type sdkError struct {
	*customError
	statusCode int
}

func (ce *sdkError) StatusCode() int {
	return ce.statusCode
}

// NewSDKError returns an SDK Error that formats as the given text.
func NewSDKError(text string) SDKError {
	return &sdkError{
		customError: &customError{
			msg: text,
			err: nil,
		},
	}
}

// NewSDKErrorWithStatus returns an SDK Error setting the status code.
func NewSDKErrorWithStatus(msg string, statusCode int) SDKError {
	return &sdkError{
		statusCode: statusCode,
		customError: &customError{
			msg: msg,
			err: nil,
		},
	}
}

// CheckError will check for error in http response.
func CheckError(resp *http.Response, expectedStatusCodes ...int) error {
	for _, expectedStatusCode := range expectedStatusCodes {
		if resp.StatusCode == expectedStatusCode {
			return nil
		}
	}

	b, bErr := io.ReadAll(resp.Body)
	if bErr != nil {
		e := Wrap(errFailedToReadBody, bErr)
		return Wrap(NewSDKErrorWithStatus("", resp.StatusCode), e)
	}

	err := json.Unmarshal(b, &content)
	if err != nil {
		e := Wrap(errRespBodyNotJSON, New(string(b)))
		return Wrap(NewSDKErrorWithStatus("", resp.StatusCode), e)
	}

	if msg, ok := content["error"]; ok {
		if v, ok := msg.(string); ok {
			return NewSDKErrorWithStatus(v, resp.StatusCode)
		}
		return NewSDKErrorWithStatus("unknown error", resp.StatusCode)
	}
	e := Wrap(errJSONKeyNotFound, New(string(b)))
	return Wrap(NewSDKErrorWithStatus("", resp.StatusCode), e)
}
```

In the above code ```CheckError```  can be use in response of SDK, 


Example for at  ```pkg/sdk/go/bootstrap.go``` in  function ```AddBootstrap```

```go
func (sdk mfSDK) AddBootstrap(token string, cfg BootstrapConfig) (string, error) {
	data, err := json.Marshal(cfg)
	if err != nil {
		return "", err
	}

	url := fmt.Sprintf("%s/%s", sdk.bootstrapURL, configsEndpoint)
	req, err := http.NewRequest(http.MethodPost, url, bytes.NewReader(data))
	if err != nil {
		return "", err
	}
	
	resp, err := sdk.sendRequest(req, token, string(CTJSON))
	if err != nil {
		return "", err
	}
	defer resp.Body.Close()
        err = errors.CheckError(resp, http.StatusCreated ,   http.StatusOK)

	if err != nil {
		return "", err
	}
	body, err := ioutil.ReadAll(resp.Body)
	if err != nil {
		return "", err
	}
	id := strings.TrimPrefix(resp.Header.Get("Location"), "/things/configs/")
	return id, nil
}

```


Example for  at ```pkg/sdk/go/things.go``` in  function ```UpdateThing```

```go
func (sdk mfSDK) UpdateThing(t Thing, token string) error {
	data, err := json.Marshal(t)
	if err != nil {
		return err
	}

	url := fmt.Sprintf("%s/%s/%s", sdk.thingsURL, thingsEndpoint, t.ID)

	req, err := http.NewRequest(http.MethodPut, url, bytes.NewReader(data))
	if err != nil {
		return err
	}

	resp, err := sdk.sendRequest(req, token, string(CTJSON))
	if err != nil {
		return err
	}
	defer resp.Body.Close()
	
	return  errors.CheckError(resp,  http.StatusOK)
}

```

AryanGodara: Thanks for this @arvindh123 , I'll implement this in sdk/go and push the changes asap!

pkg/sdk/go/errors.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/mainflux/mainflux/pkg/errors"
+)
+
+var errEncodeError = errors.NewSDKError("failed to encode response error")
+
+var (
```
dborovcanin: Most of these errors are not used.

dborovcanin (changes requested):
Let's make all the SDK methods (from `sdk.go` file) return SDKError, rather than `error`. When using SDK, we want to enable quick and simple access to the response details.

pkg/sdk/go/errors.go
AryanGodara: I'll check and remove the unused ones.

pkg/errors/sdk_errors.go
AryanGodara: > Let's make all the SDK methods (from `sdk.go` file) return SDKError, rather than `error`. When using SDK, we want to enable quick and simple access to the response details.

Done, changed from error interface to SDKError interface, in the sdk/go directory.

pkg/sdk/go/bootstrap.go
```diff
@@ -46,36 +46,41 @@ type ConfigUpdateCertReq struct {
...
 	id := strings.TrimPrefix(resp.Header.Get("Location"), "/things/configs/")
 	return id, nil
 }
 
-func (sdk mfSDK) Whitelist(token string, cfg BootstrapConfig) error {
+func (sdk mfSDK) Whitelist(token string, cfg BootstrapConfig) errors.SDKError {
 	data, err := json.Marshal(BootstrapConfig{State: cfg.State})
 	if err != nil {
-		return errors.Wrap(ErrFailedWhitelist, err)
+		return errors.NewSDKError(errors.Wrap(ErrFailedWhitelist, err).Error())
```
dborovcanin: In other SDK methods, we usually have a pattern like this:
```go
data, err := json.Marshal(connIDs)
if err != nil {
    return errors.NewSDKError(err.Error())
}
```
So, without wrapping. Why is this method specific?

pkg/sdk/go/sdk.go
```diff
@@ -347,7 +299,7 @@ func NewSDK(conf Config) SDK {
 	}
 }
 
-func (sdk mfSDK) sendRequest(req *http.Request, token, contentType string) (*http.Response, error) {
+func (sdk mfSDK) sendRequest(req *http.Request, token, contentType string) (*http.Response, errors.SDKError) {
```
dborovcanin: Internal methods should probably not return SDK errors. The point of SDK error is to encapsulate service HTTP response and extract error from it if there is any.

pkg/sdk/go/errors.go
```diff
@@ -0,0 +1,57 @@
...
+	ErrCertsRemove = errors.NewSDKError("failed to remove certificate")
+
+	// ErrFailedCertUpdate failed to update certs in bootstrap config
+	ErrFailedCertUpdate = errors.NewSDKError("failed to update certs in bootstrap config")
+
+	// ErrMemberAdd failed to add member to a group.
+	ErrMemberAdd = errors.NewSDKError("failed to add member to group")
+)
+
+func encodeError(body []byte, status int) errors.SDKError {
```
dborovcanin: This is similar to `pkg/errors/sdk_errors.go` method `CheckErr`. Do we need both?

dborovcanin (changes requested):
Please work on consistency when it comes to error handling. For example, JSON marshalling error is wrapped in some cases, and not in others.

pkg/sdk/go/errors.go
AryanGodara: encodeError() simply joins the http statuscode to the error, and returns a new composite SDKError.
In Checkerror(), it checks for a range of expected statuscodes, if any one of matches with the resp.statuscode, the function doesn't return any error, else, it checks further, and returns a composite error.

So, both are used for different reasons, and hence why I had to keep both.
(The tests and other functions get very complicated and difficult-to-read, when I use just one function, for both these usecases)

pkg/sdk/go/bootstrap.go
AryanGodara: The function originally returned the wrapped error. So, the outer package of NewSDKError(...) is just to convert from `error` to `SKDError` interface.
Since both ErrFailedWhitelist, and the specific marshalling error, both are required for proper error handling, I have kept both of them. And, ErrWhitelist is used in multiple places, so I kept it.

AryanGodara: Should I remove ErrWhiteList, and remove this wrapping? Since, `Whitelist` in bootstrap, is the only function, that has its own error now

pkg/sdk/go/channels_test.go
```diff
@@ -131,26 +133,30 @@ func TestCreateChannels(t *testing.T) {
...
+			err:      errors.NewSDKErrorWithStatus(errors.ErrAuthentication.Error(), http.StatusUnauthorized),
 			res:      []sdk.Channel{},
 		},
 	}
 	for _, tc := range cases {
 		res, err := mainfluxSDK.CreateChannels(tc.channels, tc.token)
+		fmt.Println()
+		fmt.Println("expected: ", tc.err)
+		fmt.Println("received: ", err)
+		fmt.Println()
```
dborovcanin: Remove.

pkg/errors/sdk_errors.go
```diff
@@ -0,0 +1,103 @@
...
+
+	if msg, ok := content["error"]; ok {
+		if v, ok := msg.(string); ok {
+			return NewSDKErrorWithStatus(v, resp.StatusCode)
+		}
+		return NewSDKErrorWithStatus("unknown error", resp.StatusCode)
+	}
+	e := Wrap(errJSONKeyNotFound, New(string(b)))
+	return NewSDKError(Wrap(NewSDKErrorWithStatus("", resp.StatusCode), e).Error())
+}
```
dborovcanin: This looks way too complex. Something like this will do pretty much the same job:
```go
// CheckError will check for error in http response.
func CheckError(resp *http.Response, expectedStatusCodes ...int) SDKError {
	for _, expectedStatusCode := range expectedStatusCodes {
		if resp.StatusCode == expectedStatusCode {
			return nil
		}
	}

	

	if err := json.NewDecoder(resp.Body).Decode(&content); err != nil {
		return NewSDKErrorWithStatus(err.Error(), resp.StatusCode)
	}

	if msg, ok := content["error"]; ok {
		if v, ok := msg.(string); ok {
			return NewSDKErrorWithStatus(v, resp.StatusCode)
		}
		return NewSDKErrorWithStatus("unknown error", resp.StatusCode)
	}
	return NewSDKErrorWithStatus(errJSONKeyNotFound.Error(), resp.StatusCode)
}
```
We should not wrap SDK errors, SDK errors are the wrapper layer we use to wrap other errors.

pkg/sdk/go/bootstrap.go
```diff
@@ -46,36 +46,41 @@ type ConfigUpdateCertReq struct {
 	CACert     string `json:"ca_cert"`
 }
 
-func (sdk mfSDK) AddBootstrap(token string, cfg BootstrapConfig) (string, error) {
+func (sdk mfSDK) AddBootstrap(token string, cfg BootstrapConfig) (string, errors.SDKError) {
```
dborovcanin, line 47: This is an example of how we can use `errors.CheckError` insetad of `encodeError`:
```suggestion
func (sdk mfSDK) ViewBootstrap(token, id string) (BootstrapConfig, errors.SDKError) {
	url := fmt.Sprintf("%s/%s/%s", sdk.bootstrapURL, configsEndpoint, id)
	req, err := http.NewRequest(http.MethodGet, url, nil)
	if err != nil {
		return BootstrapConfig{}, errors.NewSDKError(err.Error())
	}

	resp, err := sdk.sendRequest(req, token, string(CTJSON))
	if err != nil {
		return BootstrapConfig{}, errors.NewSDKError(err.Error())
	}
	defer resp.Body.Close()

	if err := errors.CheckError(resp, http.StatusOK); err != nil {
		return BootstrapConfig{}, err
	}

	var bc BootstrapConfig
	if err := json.NewDecoder(resp.Body).Decode(&bc); err != nil {
		return BootstrapConfig{}, errors.NewSDKError(err.Error())
	}

	return bc, nil
}
```

pkg/sdk/go/errors.go
dborovcanin: I am not sure this is the case. Please check the comment I left on `ViewBootstrap` for reference.

dborovcanin (changes requested):
There is no need for unexported methods to return SDK errors.

pkg/errors/sdk_errors.go
```diff
@@ -0,0 +1,92 @@
...
+package errors
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+var (
+	errJSONKeyNotFound = New("response body expected error message json key not found")
+)
```
dborovcanin: Inline:
```go
var errJSONKeyNotFound = New("response body expected error message not found")
```
Also, no need for "json key" construct in the message.

pkg/errors/sdk_errors.go
```diff
@@ -0,0 +1,92 @@
...
+			return nil
+		}
+	}
+
+	var content map[string]interface{}
+	if err := json.NewDecoder(resp.Body).Decode(&content); err != nil {
+		return NewSDKErrorWithStatus(err.Error(), resp.StatusCode)
+	}
+
+	if msg, ok := content["error"]; ok {
```
dborovcanin: Extract "error" to the constant.

pkg/errors/sdk_errors.go
```diff
@@ -0,0 +1,92 @@
...
+		if v, ok := msg.(string); ok {
+			return NewSDKErrorWithStatus(v, resp.StatusCode)
+		}
+		return NewSDKErrorWithStatus("unknown error", resp.StatusCode)
+	}
+	return NewSDKErrorWithStatus(errJSONKeyNotFound.Error(), resp.StatusCode)
+
```
dborovcanin, line 93: Remove the empty line.

pkg/errors/sdk_errors.go
```diff
@@ -0,0 +1,92 @@
...
+		return NewSDKErrorWithStatus(err.Error(), resp.StatusCode)
+	}
+
+	if msg, ok := content["error"]; ok {
+		if v, ok := msg.(string); ok {
+			return NewSDKErrorWithStatus(v, resp.StatusCode)
+		}
+		return NewSDKErrorWithStatus("unknown error", resp.StatusCode)
+	}
+	return NewSDKErrorWithStatus(errJSONKeyNotFound.Error(), resp.StatusCode)
```
dborovcanin: Since `errJSONKeyNotFound` is only used here, you don't need to make it an error, but rather a string constant. The same goes for `"unknown error"`.

pkg/errors/sdk_errors.go
dborovcanin: This should be a string, see comments below.

pkg/sdk/go/bootstrap.go
dborovcanin: The thing is that errors such as `ErrFailedWhitelist` or `ErrFailed<MethodName>` make no sense here. If the method is `Whitelist`, and it returns an error -  there is not much point in having it wrapped t something as trivial as "failed to whitelist". That's self-evident. The caller, however, can do a wrap to indicate the actual point of failure, but that's not the case here.

pkg/sdk/go/certs.go
```diff
@@ -44,47 +47,47 @@ func (sdk mfSDK) IssueCert(thingID string, keyBits int, keyType, valid, token st
...
 	case http.StatusForbidden:
-		return errors.ErrAuthorization
+		return errors.NewSDKError(errors.ErrAuthorization.Error())
 	default:
 		return ErrCertsRemove
 	}
 }
 
-func (sdk mfSDK) RevokeCert(thingID, certID string, token string) error {
+func (sdk mfSDK) RevokeCert(thingID, certID string, token string) errors.SDKError {
```
dborovcanin, line 70: We need to implement this one, but that's for a separate PR.

provision/service.go
```diff
@@ -311,9 +311,9 @@ func (ps *provisionService) updateGateway(token string, bs SDK.BootstrapConfig,
 	gw.CfgID = bs.MFThing
 	gw.Type = gateway
 
-	th, err := ps.sdk.Thing(bs.MFThing, token)
-	if err != nil {
-		return errors.Wrap(ErrGatewayUpdate, err)
+	th, sdkerr := ps.sdk.Thing(bs.MFThing, token)
+	if sdkerr != nil {
+		return errors.Wrap(ErrGatewayUpdate, sdkerr)
```
dborovcanin, line 317: No need to rename.

dborovcanin (changes requested):

pkg/sdk/go/bootstrap.go
arvindh123: > The thing is that errors such as `ErrFailedWhitelist` or `ErrFailed<MethodName>` make no sense here. If the method is `Whitelist`, and it returns an error - there is not much point in having it wrapped t something as trivial as "failed to whitelist". That's self-evident. The caller, however, can do a wrap to indicate the actual point of failure, but that's not the case here.

By indicating the actual point of failure like error `ErrFailedWhitelist` , will helps to make decision in automations,  by checking the error contains `ErrFailedWhitelist` automation program can take decisions.

AryanGodara: Yes, like @arvindh123 mentioned. This specific error `errwhitelist` is used in provision/service also. That's why I didn't remove it.

provision/service.go
AryanGodara: In the next line, `err` variable-name is used again for `json.Marshal`
But since err is previously of `SDKError interface`, it can't be used to store marshall error of `error interface`.
That's why I had to rename.
But I'll remove the `sdkerr` var from other places where it's not required anymore

pkg/errors/sdk_errors.go
```diff
@@ -80,15 +81,15 @@ func CheckError(resp *http.Response, expectedStatusCodes ...int) SDKError {
...
+		return NewSDKErrorWithStatus(err, resp.StatusCode)
 	}
 
 	if msg, ok := content[err]; ok {
 		if v, ok := msg.(string); ok {
-			return NewSDKErrorWithStatus(v, resp.StatusCode)
+			return NewSDKErrorWithStatus(errors.New(v), resp.StatusCode)
 		}
-		return NewSDKErrorWithStatus(unknownErr, resp.StatusCode)
+		return NewSDKErrorWithStatus(errors.New(unknownErr), resp.StatusCode)
```
dborovcanin: Since `NewSDKError` and `NewSDKErrorWithStatus` now accept error - `unknownErr` and `jsonKeyNotFound` should be errors again.

pkg/sdk/go/certs.go
```diff
@@ -61,13 +61,13 @@ func (sdk mfSDK) RemoveCert(id, token string) errors.SDKError {
...
 	}
 	switch res.StatusCode {
 	case http.StatusNoContent:
 		return nil
 	case http.StatusForbidden:
-		return errors.NewSDKError(errors.ErrAuthorization.Error())
+		return errors.NewSDKError(errors.ErrAuthorization)
 	default:
 		return ErrCertsRemove
 	}
```
dborovcanin: You can replace with:
```go
return errors.CheckError(res, http.StatusNoContent)
```

pkg/sdk/go/errors.go
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package sdk
+
+import (
+	"github.com/mainflux/mainflux/pkg/errors"
+)
+
+var (
```
dborovcanin: Some of these errors are not used.

pkg/sdk/go/errors.go
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package sdk
+
+import (
+	"github.com/mainflux/mainflux/pkg/errors"
+)
```
dborovcanin: Inline.

dborovcanin (changes requested):

pkg/sdk/go/errors.go
AryanGodara: I have left just 4, which are being used. I can remove `ErrCerts`.
And for the others, I can move them to errors/types.go , in the errors package.

provision/service.go
```diff
@@ -382,7 +384,7 @@ func (ps *provisionService) recover(e *error, ths *[]SDK.Thing, chs *[]SDK.Chann
 		}
 	}
 
-	if errors.Contains(err, SDK.ErrFailedWhitelist) || errors.Contains(err, ErrGatewayUpdate) {
+	if errors.Contains(err, ErrFailedWhitelist) || errors.Contains(err, ErrGatewayUpdate) {
```
dborovcanin: Let's stick to `Things` terminology: `ErrThingUpdate`.

pkg/errors/types.go
```diff
@@ -42,4 +42,10 @@ var (
 
 	// ErrScanMetadata indicates problem with metadata in db
 	ErrScanMetadata = New("failed to scan metadata in db")
+
+	// ErrFailedFetch indicates that fetching of entity data failed.
+	ErrFailedFetch = New("failed to fetch entity")
+
+	// ErrFailedConnect indicates that connecting thing to channel failed.
+	ErrFailedConnect = New("failed to connect thing to channel")
```
dborovcanin: Are these errors used? I see them used in the context of `errors.Contains` but they are never returned (directly or as a wrap), so those checks make no sense. Please double-check.

dborovcanin (changes requested):

pkg/errors/types.go
AryanGodara: Yes, it actually was useless to keep these errors. I've removed these two.

provision/service.go
AryanGodara: I updated the error

tools/provision/provision.go
```diff
@@ -159,7 +160,7 @@ func Provision(conf Config) {
 
 		if conf.SSL {
 			var priv interface{}
-			priv, err = rsa.GenerateKey(rand.Reader, rsaBits)
+			priv, _ = rsa.GenerateKey(rand.Reader, rsaBits)
```
dborovcanin, line 163: Why is this changed, and why is the error extracted to the var at the top?

pkg/errors/sdk_errors.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (ce *sdkError) Msg() string {
+	return ce.msg
+}
+
+func (ce *sdkError) Err() Error {
+	return ce.err
+}
+
```
dborovcanin: No need for these. Eventually, you can change `Error` to something like:
```go
func (ce *sdkError) Error() string {
	if ce == nil {
		return ""
	}
	if ce.customError == nil {
		return http.StatusText(ce.statusCode)
	}
	return fmt.Sprintf("Status: %s: %s", http.StatusText(ce.statusCode), ce.customError.Error())
}
```

dborovcanin (changes requested):

tools/provision/provision.go
AryanGodara: It was a problem with interfaces. `error` can be converted to `sdkerror` but not the reverse.
So, if I had not defined `var err error`, then 
```
token, err := s.CreateToken(user)
```
would make err of `sdkerror`. But later, we have to use err in sending http requests, etc.
That's why to avoid creating new variable of `sdkerr`, I defined err var at the top.

As for removing the `err` to `_`

```
var priv interface{}
			priv, err = rsa.GenerateKey(rand.Reader, rsaBits)
			priv, _ = rsa.GenerateKey(rand.Reader, rsaBits)

			notBefore := time.Now()
			validFor, err := time.ParseDuration(ttl)
```
We weren't actually using that value of err, so it was showing a 'problem' as that value of err was never being used. So I removed it for the time being.

pkg/errors/sdk_errors.go
AryanGodara: But in this, shouldn't we also return the `ce.msg`?
Does this work?
```
func (ce *sdkError) Error() string {
	if ce == nil {
		return ""
	}
	if ce.customError == nil {
		return http.StatusText(ce.statusCode)
	}
	if ce.err == nil {
		return fmt.Sprintf("Status: %s: %s", http.StatusText(ce.statusCode), ce.msg)
	}
	if ce.msg == "" {
		return fmt.Sprintf("Status: %s: %s", http.StatusText(ce.statusCode), ce.customError.Error())
	}

	return fmt.Sprintf("Status: %s: %s: %s", http.StatusText(ce.statusCode), ce.customError.Error(), ce.msg)
}
```

dborovcanin: `ce.msg` is returned anyways. Please take a look at the implementation of `customError.Error()`.

arvindh123:
Hello @AryanGodara @dborovcanin 

I feel lot of same code is repeated in SDK , So I propose the function of similar kind below , 
I hope similar kind of function reduce reptation code in SDK.

```go
func (sdk mfSDK) sendRequestAndGetBodyOrError(method, url string, data []byte, key, contentType string, expectedResponseCode ...http.StatusCode) ([]byte, sdkError) {
	req, err := http.NewRequest(method, url, bytes.NewReader(data))
	if err != nil {
		return []byte{}, errorWrap("failed to create new HTTP request", err)
	}
	if key != "" {
		req.Header.Set("Authorization", apiutil.ThingPrefix+key)
	}
	if contentType != "" {
		req.Header.Add("Content-Type", contentType)
	}
	resp, err := sdk.client.Do(req)
	if err != nil {
		return []byte{}, errorWrap("failed to send HTTP request", err)
	}
	defer resp.Body.Close()
	err = checkSDKError(resp,expectedResponseCode)
	if err != nil {
		return []byte{},  errorWrap("unexpected status", err)  
	}
	body, err := ioutil.ReadAll(resp.Body)
	if err != nil {
		return []byte{}, errorWrap("failed to parse body", err) 
	}
	return  body , nil 
}
```
Above functions Function inputs:  method, url, data bytes, key and content type,  
Function output: body as bytes and error 

The above proposed kind of function could be used to replace   `func (sdk mfSDK) sendRequest(req *http.Request, token, contentType string) (*http.Response, error) `
https://github.com/mainflux/mainflux/blob/ff4df7a73145679c853479c812c7c7f1a75a172f/pkg/sdk/go/sdk.go#L302-L317

dborovcanin:
> ```go
> 	req, err := http.NewRequest(method, url, bytes.NewReader(data))
> 	if err != nil {
> 		return []byte{}, errorWrap("failed to create new HTTP request", err)
> 	}
> 	if key != "" {
> 		req.Header.Set("Authorization", apiutil.ThingPrefix+key)
> 	}
> 	if contentType != "" {
> 		req.Header.Add("Content-Type", contentType)
> 	}
> 	resp, err := sdk.client.Do(req)
> 	if err != nil {
> 		return []byte{}, errorWrap("failed to send HTTP request", err)
> 	}
> 	defer resp.Body.Close()
> 	err = checkSDKError(resp,expectedResponseCode)
> 	if err != nil {
> 		return []byte{},  errorWrap("unexpected status", err)  
> 	}
> 	body, err := ioutil.ReadAll(resp.Body)
> 	if err != nil {
> 		return []byte{}, errorWrap("failed to parse body", err) 
> 	}
> 	return  body , nil 
> ```

Makes a lot of sense. We should replace `sendRequest` with the implementation @arvindh123 proposed.

pkg/errors/sdk_errors.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+)
+
+var (
+	ErrJSONKeyNotFound = errors.New("response body expected error message json key not found")
```
dborovcanin: Add comments to exported errors.

pkg/errors/sdk_errors.go
```diff
@@ -0,0 +1,90 @@
...
+	ErrUnknown         = errors.New("unknown error")
+)
+
+const err = "error"
```
dborovcanin, line 13: Improve the name, something like `jsonErrKey` would be better.

pkg/errors/sdk_errors.go
```diff
@@ -0,0 +1,90 @@
...
+	return &sdkError{
+		statusCode: statusCode,
+		customError: &customError{
+			msg: err.Error(),
+			err: nil,
+		},
+	}
+}
+
+// CheckError will check for error in http response.
```
dborovcanin: Improve comment. Explain that in some cases (upsert), multiple status codes are valid.

pkg/sdk/go/sdk.go
```diff
@@ -347,6 +301,38 @@ func NewSDK(conf Config) SDK {
 	}
 }
 
+func (sdk mfSDK) sendRequestAndGetBodyOrError(method, url string, data []byte, token, contentType string, expectedResponseCode ...int) ([]byte, errors.SDKError) {
```
dborovcanin: Why doesn't this replace `sendRequest`?

dborovcanin (changes requested):

tools/provision/provision.go
dborovcanin: Since `sdkerror` implements  `error`, so it's compatible, but since it's not handled anyway, it's a separate issue in itself and should not be a part of this PR. I'll open a new issue and resolve the comment.

pkg/sdk/go/sdk.go
AryanGodara: There are 6 `create` functions which use response header "Location" to create ID, instead of using response body.
We need to use another function for this, something like sendRequestAndGetHeadersorError.

dborovcanin: Please just rename, maybe something like `processRequest`.

dborovcanin (changes requested):

pkg/errors/sdk_errors.go
```diff
@@ -0,0 +1,87 @@
...
+		customError: &customError{
+			msg: err.Error(),
+			err: nil,
+		},
+	}
+}
+
+// CheckError will check the http response status code and matches it with the given status codes.
+// Since multiple status codes can be valid, we can pass multiple status codes to the function.
+// The function then checks for errors in the http response.
```
dborovcanin: ...in HTTP response...

pkg/errors/types.go
```diff
@@ -28,18 +28,24 @@ var (
...
 
-	// ErrScanMetadata indicates problem with metadata in db
+	// ErrScanMetadata indicates problem with metadata in db.
 	ErrScanMetadata = New("failed to scan metadata in db")
+
+	// ErrJSONErrKey indicates response body did not contain erorr message.
+	ErrJSONErrKey = New("response body expected error message json key not found")
+
+	// ErrUnknown indicates that an unknown error was found in the response body.
+	ErrUnknown = New("unknown error")
```
dborovcanin: These are internal errors and should not be exported. You can use separate `var` block for them.

dborovcanin (changes requested):

pkg/sdk/go/sdk.go
```diff
@@ -347,28 +302,40 @@ func NewSDK(conf Config) SDK {
 	}
 }
 
-func (sdk mfSDK) sendRequest(req *http.Request, token, contentType string) (*http.Response, error) {
-	if token != "" {
-		req.Header.Set("Authorization", apiutil.BearerPrefix+token)
+func (sdk mfSDK) processRequest(method, url string, data []byte, token, contentType string, expectedResponseCode ...int) ([]byte, map[string][]string, errors.SDKError) {
```
dborovcanin: Change signature to:
```go
func (sdk mfSDK) processRequest(method, url, token, contentType string, data []byte, expectedRespCodes ...int) (http.Header, []byte, errors.SDKError) {...}
```
While it would be good not to expose any `HTTP` details, it's cleaner to return headers directly. Also, a method that accepts 6 parameters and returns 3 results doesn't sound like a good design, but, since it's a pretty simple implementation, we can accept it.

pkg/sdk/go/sdk.go
```diff
@@ -347,28 +302,40 @@ func NewSDK(conf Config) SDK {
...
+		return []byte{}, make(map[string][]string), errors.NewSDKError(err)
 	}
 
+	if token != "" {
+		if strings.Contains(token, apiutil.ThingPrefix) {
+			req.Header.Set("Authorization", token)
+		} else {
+			req.Header.Set("Authorization", apiutil.BearerPrefix+token)
+		}
+	}
```
dborovcanin: Avoid using else:
```go
	if token != "" {
		if !strings.Contains(token, apiutil.ThingPrefix) {
			token = apiutil.BearerPrefix + token
		}
		req.Header.Set("Authorization", token)
	}
```

pkg/sdk/go/sdk.go
dborovcanin: Add a short comment explaining what this method does since its signature might look a little intimidating.

dborovcanin (changes requested):

pkg/errors/sdk_errors.go
AryanGodara: I changed this to errJSONErrKey (added err prefix, since we're using it as an error now)
I'll change it to `errJSONKey`

dborovcanin (approved):

drasko (approved):
LGTM

PR #1694, opened 2023-01-09
NOISSUE - Update SDK and Things Service

rodneyosodo:
### What does this do?
1. Makes update thing to update either name or metadata and not both
2. Corrects pagemetadata for things and channels to include disconnected and connected
3. Adds reader URL for reading messages

N/A

Listed

Yes

No

### Notes
N/A

### DIFF
--- a/cli/channels.go
+++ b/cli/channels.go
@@ -156,7 +156,7 @@ var cmdChannels = []cobra.Command{
 			pm := mfxsdk.PageMetadata{
 				Offset:       uint64(Offset),
 				Limit:        uint64(Limit),
-				Disconnected: false,
+				Disconnected: true,
 			}
 			cl, err := sdk.ThingsByChannel(args[0], pm, args[1])
 			if err != nil {
--- a/cli/things.go
+++ b/cli/things.go
@@ -189,7 +189,7 @@ var cmdThings = []cobra.Command{
 			pm := mfxsdk.PageMetadata{
 				Offset:       uint64(Offset),
 				Limit:        uint64(Limit),
-				Disconnected: true,
+				Disconnected: false,
 			}
 			cl, err := sdk.ChannelsByThing(args[0], pm, args[1])
 			if err != nil {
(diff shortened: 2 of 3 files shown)


things/postgres/things.go
```diff
@@ -73,7 +73,21 @@ func (tr thingRepository) Save(ctx context.Context, ths ...things.Thing) ([]thin
 }
 
 func (tr thingRepository) Update(ctx context.Context, t things.Thing) error {
-	q := `UPDATE things SET name = :name, metadata = :metadata WHERE id = :id;`
+	var query []string
+	var upq string
+	if t.Name != "" {
```
manuio: Can we update with empty name or empty metadata anymore?

rodneyosodo: No

manuio: This means that if the name or the metadata has been edited there is no way to revert it to an empty value. Is this what we want?

rodneyosodo: Let me revert

cli/channels.go
```diff
@@ -156,7 +156,7 @@ var cmdChannels = []cobra.Command{
 			pm := mfxsdk.PageMetadata{
 				Offset:       uint64(Offset),
 				Limit:        uint64(Limit),
-				Disconnected: false,
+				Disconnected: true,
```
manuio, line 159: Isn't `List of Channels connected to Thing` what we are using here? Why are you inverting this value?

cli/things.go
```diff
@@ -189,7 +189,7 @@ var cmdThings = []cobra.Command{
 			pm := mfxsdk.PageMetadata{
 				Offset:       uint64(Offset),
 				Limit:        uint64(Limit),
-				Disconnected: true,
+				Disconnected: false,
```
manuio, line 192: Same here, why are you inverting the value if it's to list `non-connected` ?

manuio (changes requested):

cli/channels.go
rodneyosodo: Here in `List of Things connected to a Channel`
when disconnected is set to true it returns a list of things not connected to that channel and when disconnected is set to false it returns a list of things connected to the channel

cli/things.go
rodneyosodo: Here `disconnected` should be `false` meaning `connected` it `true`

cli/channels.go
manuio: Right, my bad. I got confused :)

manuio (approved):
LGTM

drasko (approved):
LGTM

PR #1706, opened 2023-01-18
NOISSUE - Update Nginx Docker Image

rodneyosodo:
Signed-off-by: rodneyosodo <socials@rodneyosodo.com>

### What does this do?
Updates nginx docker image from `nginx:1.20.0-alpine` to `nginx:1.23.3-alpine`.  In doing so, it fixes typo is `docker/ssl/Makefile` and also return error of type `ErrBearerKey` on publish request instead of `ErrBearerToken`

N/A

No

No

No

### Notes
N/A


docker/ssl/authorization.js
```diff
@@ -167,10 +167,12 @@ function parseCert(cert, key) {
...
-                return pair[1];
+                return "Thing " + pair[1].replace("\\", "");
             }
         }
     }
 
     return '';
 }
+
+export default {setKey,authenticate};
```
dborovcanin: Add an empty line to the end of the file.

drasko (dismissed):
LGTM

dborovcanin (changes requested):
Can you also generate CA and server certs since the old ones are expired?
(You can use Makefile from `docker/ssl` with commands `make ca` and `make server_cert`).

docker/ssl/authorization.js
```diff
@@ -167,10 +167,12 @@ function parseCert(cert, key) {
         for (var i = 0; i < pairs.length; i++) {
             var pair = pairs[i].split('=');
             if (pair[0].toUpperCase() == key) {
-                return pair[1];
+                return "Thing " + pair[1].replace("\\", "");
```
dborovcanin: Please check if the MQTT adapter handles it correctly in both cases when a prefix is sent in the password and when it's not.

dborovcanin (changes requested):

docker/ssl/authorization.js
rodneyosodo: I have updated

docker/ssl/authorization.js
```diff
@@ -29,8 +29,8 @@ function authenticate(s) {
 
         var pass = parsePackage(s, data);
 
-        if (!clientKey.length || pass !== clientKey) {
-            s.error('Cert CN (' + clientKey + ') does not match client password');
+        if (!clientKey.length || !clientKey.includes(pass) ) {
```
dborovcanin: Is there an option to use it like `endsWith` which is more precise for ignoring prefixes?

dborovcanin (changes requested):

dborovcanin (approved):

drasko (approved):
LGTM

PR #1586, opened 2022-03-30
MF-1584 Upgrade InfluxDB from 1.x to 2.x

fthdrmzzz:
Signed-off-by: fatih <fatihdurmaz@sabanciuniv.edu>

### What does this do?
This upgrades the InfluxDB from Version 1.x to Version 2.x. This also updates writers and readers for InfluxDB upgrade.
### Which issue(s) does this PR fix/relate to?
This resolves #1584

PR #1601, opened 2022-05-03
MF-1556  - Move the most used functions in main.go to internal package

arvindh123:
### What does this do?
   This pull request for moving the HTTP Server and GRPC server from cmd/service/main.go to internal/mfserver  package 

### Which issue(s) does this PR fix/relate to?
Resolve partially  #1556  because need to check other functions feasibility  to move to internal package 

### List any changes that modify/break current functionality
N/A  

### Have you included tests for your changes?
N/A  

### Did you document any new/modified functionality?
N/A  

### Notes
N/A

PR #1709, opened 2023-01-24
MF-1584 - Upgrade InfluxDB from 1.x to 2.x

### LINKED ISSUE #1584: Upgrade InfluxDB driver from InfluxQL to Flux
**ENHANCEMENT**

I tried to implement string comparators for influxDB readers in #1573 . We should switch from influxQL to flux in order to implement comparators. Because In InfluxQL it is not supported. 

3. Indicate the importance of this enhancement to you (must-have, should-have, nice-to-have).
This is a must-have for #1573

AryanGodara:
### What does this do?
This upgrades the InfluxDB from Version 1.x to Version 2.x. This also updates writers and readers for the InfluxDB upgrade.

This PR is a continuation of the PR (https://github.com/mainflux/mainflux/pull/1586) by @fthdrmzzz.
This PR resolves #1584.

Support for InfluxDB 1.x is dropped in favor of 2.x.

Yes.

Yes.

### Notes
N/A


drasko (dismissed):
LGTM

consumers/writers/influxdb/consumer.go
```diff
@@ -21,65 +23,61 @@ var errSaveMessage = errors.New("failed to save message to influxdb database")
 
 var _ consumers.Consumer = (*influxRepo)(nil)
 
+type RepoConfig struct {
+	Bucket string
+	Org    string
+}
```
dborovcanin, line 30: Add an empty line.

consumers/writers/influxdb/consumer.go
```diff
@@ -12,7 +13,8 @@ import (
 	"github.com/mainflux/mainflux/pkg/transformers/json"
 	"github.com/mainflux/mainflux/pkg/transformers/senml"
 
-	influxdata "github.com/influxdata/influxdb/client/v2"
+	influxdata "github.com/influxdata/influxdb-client-go/v2"
+	influxdatawrite "github.com/influxdata/influxdb-client-go/v2/api/write"
```
dborovcanin: No need for package alias here.

readers/influxdb/messages_test.go
```diff
@@ -72,14 +74,13 @@ func TestReadAll(t *testing.T) {
...
-	rand.Seed(time.Now().UnixNano())
-	to := msgsNum
-	now := float64(rand.Intn(to) + offset)
+	now := time.Now().Unix() * 1e9
 
 	for i := 0; i < msgsNum; i++ {
 		// Mix possible values as well as value sum.
 		msg := m
-		msg.Time = now - float64(i)
+		// now is unixnano, now/1e9 is seconds
```
dborovcanin: Why don't you set `now` immediately rather than multiplying and then dividing? Like:
```go
now := float64(time.Now().Unix())
for i := 0; i < msgsNum; i++ {
// Mix possible values as well as value sum.
msg := m
msg.Time = now - messageDelay*float64(i)
```

dborovcanin (changes requested):

dborovcanin (changes requested):
Please align with the changes on `master` introduced in #1601.

rodneyosodo (changes requested):
Fix the `total` number of messages
I published 10 messages and I get
```sh
HTTP/1.1 200 OK
Date: Fri, 17 Feb 2023 12:37:36 GMT
Content-Length: 629

{"offset":0,"limit":10,"format":"messages","total":3,"messages":
```

rodneyosodo:
@AryanGodara check if this is true -> influxdb-reader will returns 3 messages even if you published 5 messages

http/adapter.go
```diff
@@ -43,6 +47,9 @@ func (as *adapterService) Publish(ctx context.Context, token string, msg *messag
 		return err
 	}
 	msg.Publisher = thid.GetValue()
+	fmt.Println()
+	fmt.Println("Called as.pubslisher.publish")
+	fmt.Println()
```
arvindh123: Do we need these print line?

http/adapter.go
```diff
@@ -34,6 +35,9 @@ func New(publisher messaging.Publisher, things mainflux.ThingsServiceClient) Ser
 }
 
 func (as *adapterService) Publish(ctx context.Context, token string, msg *messaging.Message) error {
+	fmt.Println()
+	fmt.Println("Inside svc.Publish")
+	fmt.Println()
```
arvindh123: Do we need these print line?

http/api/endpoint.go
```diff
@@ -5,20 +5,29 @@ package api
...
+	"fmt"
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/mainflux/mainflux/http"
 )
 
 func sendMessageEndpoint(svc http.Service) endpoint.Endpoint {
+	fmt.Println()
+	fmt.Println("Got into http, sending message, sendMessageEndpoint")
+	fmt.Println()
```
arvindh123: Do we need these print line?

http/api/endpoint.go
```diff
@@ -5,20 +5,29 @@ package api
...
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(publishReq)
+		fmt.Println()
+		fmt.Println("Inside func returned by sendMessageEndpoint: req = \n", req)
+		fmt.Println()
```
arvindh123: Do we need these print line?

http/api/endpoint.go
```diff
@@ -5,20 +5,29 @@ package api
...
 
 		if err := req.validate(); err != nil {
 			return nil, err
 		}
 
-		err := svc.Publish(ctx, req.token, req.msg)
-		return nil, err
+		fmt.Println()
+		fmt.Println("validate func returned success, called svc.Publish")
+		fmt.Println()
```
arvindh123: Do we need these print line?

http/adapter.go
AryanGodara: This must be while merging, I'll remove them asap!

arvindh123 (commented):
The Influx logic looks good for me and left some comments and code change request

arvindh123 (dismissed):
LGTM

readers/influxdb/messages.go
```diff
@@ -89,62 +100,80 @@ func (repo *influxRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
...
 }
 
-func (repo *influxRepository) count(measurement, condition string) (uint64, error) {
-	cmd := fmt.Sprintf(`SELECT COUNT(*) FROM %s WHERE %s`, measurement, condition)
-	q := influxdata.Query{
-		Command:  cmd,
-		Database: repo.database,
-	}
+func (repo *influxRepository) count(measurement, condition string, timeRange string) (uint64, error) {
+
```
dborovcanin: Remove the empty line.

consumers/writers/influxdb/consumer.go
```diff
@@ -21,65 +23,62 @@ var errSaveMessage = errors.New("failed to save message to influxdb database")
...
 	}
 	if err != nil {
 		return err
 	}
-
-	if err := repo.client.Write(pts); err != nil {
-		return errors.Wrap(errSaveMessage, err)
-	}
-	return nil
+	writeAPI := repo.client.WriteAPIBlocking(repo.cfg.Org, repo.cfg.Bucket)
```
dborovcanin: Let's switch to async API for writing.

dborovcanin (changes requested):

internal/clients/influxdb/influxdb.go
```diff
@@ -36,16 +40,9 @@ func Setup(envPrefix string) (client.HTTPClient, error) {
...
-		Addr:      address,
-		Username:  config.Username,
-		Password:  config.Password,
-		UserAgent: config.UserAgent,
-		Timeout:   config.Timeout,
-	}
-	client, err := client.NewHTTPClient(clientConfig)
+func Connect(config Config) (influxdb2.Client, error) {
+	client := influxdb2.NewClient(config.DBUrl, config.Token)
+	_, err := client.Ready(context.Background())
```
dborovcanin: Inline:
```go
if _, err := client.Ready(context.Background()); err != nil {
....
}
```

dborovcanin: Also, are you sure that using `context.Background()` is the correct way? Should we put some deadline (maybe pass context from main, and use it to spawn `context.WithDeadline(ctx, deadline)`?

consumers/writers/influxdb/consumer.go
```diff
@@ -21,65 +23,85 @@ var errSaveMessage = errors.New("failed to save message to influxdb database")
...
+				err = e
+			}
+		}()
+
+		for _, pt := range pts {
+			writeAPI.WritePoint(pt)
+		}
+
+		writeAPI.Flush()
+		repo.client.Close()
```
dborovcanin: Flushing after each write call does not make much sense. Especially since flushing is usually done automatically after a flush interval (default is 1000ms). Also closing the client after writing, and setting the error this way makes no sense.

consumers/writers/influxdb/consumer.go
```diff
@@ -21,65 +23,85 @@ var errSaveMessage = errors.New("failed to save message to influxdb database")
...
 	}
 	if err != nil {
 		return err
 	}
 
-	if err := repo.client.Write(pts); err != nil {
-		return errors.Wrap(errSaveMessage, err)
+	switch repo.async {
+	case true:
+		writeAPI := repo.client.WriteAPI(repo.cfg.Org, repo.cfg.Bucket)
```
dborovcanin: Since we always use the same org and bucket (see the implementation of `WriteAPI` method), maybe this is not needed and we can extract API as a field of `influxRepo`.

consumers/writers/influxdb/consumer.go
```diff
@@ -21,65 +23,85 @@ var errSaveMessage = errors.New("failed to save message to influxdb database")
...
+		}()
+
+		for _, pt := range pts {
+			writeAPI.WritePoint(pt)
+		}
+
+		writeAPI.Flush()
+		repo.client.Close()
+	default:
+		writeAPI := repo.client.WriteAPIBlocking(repo.cfg.Org, repo.cfg.Bucket)
```
dborovcanin: Since we always use the same org and bucket (see the implementation of the `WriteAPIBlocking` method), maybe this is not needed and we can extract API as a field of influxRepo.

dborovcanin (changes requested):
Our `Consumer` interface is unsuitable for async operations, and we should probably modify it. There is no reason to assume messages will be consumed synchronously, and we should support async design, while also enabling a sync approach so the one who implements it has a choice. Consider changing it to something like:
```go
type Consumer interface {
	// Consume method is used to consume received messages.
	// An error channel is used to handle errors, so it supports the async approach.
	Consume(messages interface{}, errs chan<- (error))
}
```

internal/clients/influxdb/influxdb.go
AryanGodara: Should this deadline be the same as the `Timeout` inside Config struct?

consumers/writers/influxdb/consumer.go
```diff
@@ -29,17 +31,21 @@ type RepoConfig struct {
...
 // New returns new InfluxDB writer.
 func New(client influxdb2.Client, config RepoConfig, async bool) consumers.Consumer {
 	return &influxRepo{
-		client: client,
-		cfg:    config,
-		async:  async,
+		client:           client,
+		cfg:              config,
+		async:            async,
+		writeAPI:         client.WriteAPI(config.Org, config.Bucket),
```
dborovcanin: Let's support only blocking API until #1742 is merged and we'll add async support later.

consumers/writers/influxdb/consumer.go
```diff
@@ -58,24 +64,28 @@ func (repo *influxRepo) Consume(message interface{}) error {
...
+				select {
+				case err := <-errCh:
+					fmt.Println(err) // Send this error to channel in Consume parameter (not yet implemented)
+				case <-ch:
+					return
+				}
 			}
-		}()
+
+		}(ch)
```
dborovcanin: Avoid, this is not a way to handle async anyways. We'll fix it after #1742.

cmd/influxdb-reader/main.go
```diff
@@ -73,7 +73,7 @@ func main() {
 		Org:    influxDBConfig.Org,
 	}
 
-	client, err := influxDBClient.Connect(influxDBConfig)
+	client, err := influxDBClient.Connect(influxDBConfig, context.Background())
```
dborovcanin: Use `ctx` from L38.

cmd/influxdb-writer/main.go
```diff
@@ -68,7 +68,7 @@ func main() {
 		Org:    influxDBConfig.Org,
 	}
 
-	client, err := influxDBClient.Connect(influxDBConfig)
+	client, err := influxDBClient.Connect(influxDBConfig, context.Background())
```
dborovcanin: Same here, us `ctx` from before rather than spawning a new background context.

dborovcanin (changes requested):

internal/clients/influxdb/influxdb.go
```diff
@@ -21,32 +21,30 @@ type Config struct {
...
-		Username:  config.Username,
-		Password:  config.Password,
-		UserAgent: config.UserAgent,
-		Timeout:   config.Timeout,
-	}
-	client, err := client.NewHTTPClient(clientConfig)
-	if err != nil {
+func Connect(config Config, ctx context.Context) (influxdb2.Client, error) {
+	client := influxdb2.NewClient(config.DBUrl, config.Token)
+	ctx, cancel := context.WithTimeout(ctx, config.Timeout)
```
dborovcanin, line 40: Looks like this timeout is not configured correctly (or does not have the correct default value), so `Ready` always fails due to expired context.

dborovcanin (changes requested):

cmd/influxdb-reader/main.go
```diff
@@ -66,13 +67,22 @@ func main() {
 	if err := env.Parse(&influxDBConfig, env.Options{Prefix: envPrefixInfluxdb}); err != nil {
 		logger.Fatal(fmt.Sprintf("failed to load InfluxDB client configuration from environment variable : %s", err))
 	}
-	client, err := influxDBClient.Connect(influxDBConfig)
+	influxDBConfig.DBUrl = fmt.Sprintf("%s://%s:%s", influxDBConfig.Protocol, influxDBConfig.Host, influxDBConfig.Port)
+	if influxDBConfig.Timeout == 0 {
+		influxDBConfig.Timeout = 1000 * time.Millisecond
```
rodneyosodo: `1000 * time.Millisecond` should be exported as a constant

rodneyosodo (changes requested):

cmd/influxdb-writer/main.go
```diff
@@ -61,13 +62,22 @@ func main() {
 	if err := env.Parse(&influxDBConfig, env.Options{Prefix: envPrefixInfluxdb}); err != nil {
 		logger.Fatal(fmt.Sprintf("failed to load InfluxDB client configuration from environment variable : %s", err))
 	}
-	client, err := influxDBClient.Connect(influxDBConfig)
+	influxDBConfig.DBUrl = fmt.Sprintf("%s://%s:%s", influxDBConfig.Protocol, influxDBConfig.Host, influxDBConfig.Port)
+	if influxDBConfig.Timeout == 0 {
+		influxDBConfig.Timeout = 1000 * time.Millisecond
```
dborovcanin: You can set the default value in `/internal/clients/influxdb/influxdb.go`like:
```go
type Config struct {
	Protocol           string        `env:"PROTOCOL"              envDefault:"http"`
	Host               string        `env:"HOST"                  envDefault:"localhost"`
	Port               string        `env:"PORT"                  envDefault:"8086"`
	Username           string        `env:"ADMIN_USER"            envDefault:"mainflux"`
	Password           string        `env:"ADMIN_PASSWORD"        envDefault:"mainflux"`
	DbName             string        `env:"DB"                    envDefault:"mainflux"`
	Bucket             string        `env:"BUCKET"                envDefault:"mainflux-bucket"`
	Org                string        `env:"ORG"                   envDefault:"mainflux"`
	Token              string        `env:"TOKEN"                 envDefault:"mainflux-token"`
	DBUrl              string        `env:"DBURL"                   envDefault:""`
	UserAgent          string        `env:"USER_AGENT"            envDefault:"InfluxDBClient"`
	Timeout            time.Duration `env:"TIMEOUT"  envDefault:"1s"` // Influxdb client configuration by default there is no timeout duration , this field will not have fallback default timeout duration Reference: https://pkg.go.dev/github.com/influxdata/influxdb@v1.10.0/client/v2#HTTPConfig
	InsecureSkipVerify bool          `env:"INSECURE_SKIP_VERIFY"  envDefault:"false"`
}
```

dborovcanin: Also, please align these tags. :)

dborovcanin (changes requested):

cmd/influxdb-writer/main.go
AryanGodara: Thanks @dborovcanin , I wasn't able to find this config struct for some reason, hence the use of extra line.

dborovcanin (dismissed):

docker/.env
```diff
@@ -240,10 +240,15 @@ MF_CASSANDRA_READER_SERVER_KEY=
 
 ### InfluxDB
```
rodneyosodo, line 238: - add `MF_INFLUXDB_PROTOCOL` and propagate it to docker-compose file
- add `MF_INFLUXDB_TIMEOUT` and propagate it to docker-compose file

docker/.env
```diff
@@ -240,10 +240,15 @@ MF_CASSANDRA_READER_SERVER_KEY=
...
 MF_INFLUXDB_PORT=8086
-MF_INFLUXDB_DB=mainflux
+MF_INFLUXDB_HOST=mainflux-influxdb
 MF_INFLUXDB_ADMIN_USER=mainflux
 MF_INFLUXDB_ADMIN_PASSWORD=mainflux
-MF_INFLUXDB_HTTP_AUTH_ENABLED=true
+MF_INFLUXDB_ADMIN_URL="http://localhost:8086"
```
rodneyosodo: is `MF_INFLUXDB_ADMIN_URL` used anywhere? If not remove it.

internal/clients/influxdb/influxdb.go
```diff
@@ -21,32 +21,30 @@ type Config struct {
 	Username           string        `env:"ADMIN_USER"            envDefault:"mainflux"`
 	Password           string        `env:"ADMIN_PASSWORD"        envDefault:"mainflux"`
 	DbName             string        `env:"DB"                    envDefault:"mainflux"`
+	Bucket             string        `env:"BUCKET"                envDefault:"mainflux-bucket"`
+	Org                string        `env:"ORG"                   envDefault:"mainflux"`
+	Token              string        `env:"TOKEN"                 envDefault:"mainflux-token"`
+	DBUrl              string        `env:"DBURL"                 envDefault:""`
 	UserAgent          string        `env:"USER_AGENT"            envDefault:"InfluxDBClient"`
```
rodneyosodo: is `UserAgent` and `InsecureSkipVerify` used anywhere? If not remove them.

rodneyosodo (changes requested):
Some minor changes on env variables

docker/.env
AryanGodara: No, we define the port 8086 separately, so I will remote this

rodneyosodo (changes requested):
Don't forget to update `consumers/writers/influxdb/README.md`

consumers/writers/influxdb/README.md
```diff
@@ -12,7 +12,7 @@ default values.
 | ----------------------------- | --------------------------------------------------------------------------------- | ---------------------- |
```
rodneyosodo, line 12: Add docs for 

- MF_INFLUXDB_HOST
- MF_INFLUXDB_PROTOCOL
- MF_INFLUXDB_TIMEOUT
- MF_INFLUXDB_ORG
- MF_INFLUXDB_BUCKET
- MF_INFLUXDB_TOKEN
- MF_INFLUXDB_HTTP_ENABLED
- MF_INFLUXDB_INIT_MODE

rodneyosodo: Also here https://github.com/AryanGodara/mainflux/blob/influx-db-update/consumers/writers/influxdb/README.md?plain=1#L41

rodneyosodo (changes requested):

docker/.env
```diff
@@ -240,23 +237,29 @@ MF_CASSANDRA_READER_SERVER_KEY=
 
 ### InfluxDB
 MF_INFLUXDB_PORT=8086
-MF_INFLUXDB_DB=mainflux
+MF_INFLUXDB_HOST=mainflux-influxdb
 MF_INFLUXDB_ADMIN_USER=mainflux
 MF_INFLUXDB_ADMIN_PASSWORD=mainflux
-MF_INFLUXDB_HTTP_AUTH_ENABLED=true
+MF_INFLUXDB_PROTOCOL=http
+MF_INFLUXDB_TIMEOUT==1s
```
rodneyosodo: Fix this

AryanGodara: added auth_enabled back

consumers/writers/influxdb/README.md
```diff
@@ -12,12 +12,20 @@ default values.
...
 | MF_INFLUX_WRITER_DB_HOST      | InfluxDB host                                                                     | localhost              |
 | MF_INFLUXDB_PORT              | Default port of InfluxDB database                                                 | 8086                   |
 | MF_INFLUXDB_ADMIN_USER        | Default user of InfluxDB database                                                 | mainflux               |
 | MF_INFLUXDB_ADMIN_PASSWORD    | Default password of InfluxDB user                                                 | mainflux               |
 | MF_INFLUXDB_DB                | InfluxDB database name                                                            | mainflux               |
+| MF_INFLUXDB_HOST              | InfluxDB host name                                                                | mainflux-influxdb      |
+| MF_INFLUXDB_PROTOCOL          | InfluxDB protocol                                                                 | http                   |
+| MF_INFLUXDB_TIMEOUT           | InfluxDB timeout value                                                            | 1s                     |
```
dborovcanin: Improve descriptions - this is client connection readiness timeout. Also, since Reader is using the same internal client, Reader README probably needs an update, too.

dborovcanin (changes requested):
Almost there. I'll approve as soon as the remark is resolved.

consumers/writers/influxdb/README.md
```diff
@@ -20,12 +20,12 @@ default values.
 | MF_INFLUXDB_DB                | InfluxDB database name                                                            | mainflux               |
 | MF_INFLUXDB_HOST              | InfluxDB host name                                                                | mainflux-influxdb      |
 | MF_INFLUXDB_PROTOCOL          | InfluxDB protocol                                                                 | http                   |
-| MF_INFLUXDB_TIMEOUT           | InfluxDB timeout value                                                            | 1s                     |
-| MF_INFLUXDB_ORG               | InfluxDB org name                                                                 | mainflux               |
+| MF_INFLUXDB_TIMEOUT           | InfluxDB client connection readiness timeout                                      | 1s                     |
+| MF_INFLUXDB_ORG               | InfluxDB organisation name                                                        | mainflux               |
```
dborovcanin: Fix typo.

readers/influxdb/README.md
```diff
@@ -8,24 +8,32 @@ The service is configured using the environment variables presented in the
...
+| MF_INFLUX_READER_PORT        | Service HTTP port                                   | 8905              |
+| MF_INFLUXDB_HOST             | InfluxDB host                                       | localhost         |
+| MF_INFLUXDB_PORT             | Default port of InfluxDB database                   | 8086              |
+| MF_INFLUXDB_ADMIN_USER       | Default user of InfluxDB database                   | mainflux          |
+| MF_INFLUXDB_ADMIN_PASSWORD   | Default password of InfluxDB user                   | mainflux          |
+| MF_INFLUXDB_DB               | InfluxDB database name                              | mainflux          |
+| MF_INFLUXDB_HOST             | InfluxDB host name                                  | mainflux-influxdb |
+| MF_INFLUXDB_PROTOCOL         | InfluxDB protocol                                   | http              |
+| MF_INFLUXDB_TIMEOUT          | InfluxDB client connection readiness timeout        | 1s                |
+| MF_INFLUXDB_ORG              | InfluxDB organisation name                          | mainflux          |
```
dborovcanin: Same here.

readers/influxdb/README.md
```diff
@@ -76,6 +91,11 @@ docker-compose -f docker/docker-compose.yml up -d
 docker-compose -f docker/addons/influxdb-reader/docker-compose.yml up -d
 ```
 
+And, to use the default .env file, execute the following command:
```
dborovcanin, line 94: Add an empty line between the text and code block.

readers/influxdb/README.md
```diff
@@ -76,6 +91,11 @@ docker-compose -f docker/docker-compose.yml up -d
...
+```bash
+docker-compose -f docker/addons/influxdb-reader/docker-compose.yml up --env-file docker/.env -d
+```
+
```
dborovcanin, line 99: Is this `[doc]` link at the end needed, looks like we don't use it?

dborovcanin (changes requested):

readers/influxdb/README.md
AryanGodara: https://docs.mainflux.io/api/
Should we use this link?
It points directly to API reference
Or do I remove the link altogether?

dborovcanin: This is just a link, it's not used anywhere in the file (there is no `[text](doc)` pattern). It does make sense to keep it, but we need to add a ref like this:
```
Official docs can be found [here](doc).
```

rodneyosodo (approved):

dborovcanin (approved):

SammyOina (approved):

drasko (approved):
LGTM

PR #1713, opened 2023-01-31
NOISSUE - Change Things to Client Entity

rodneyosodo:
Signed-off-by: rodneyosodo <socials@rodneyosodo.com>

### What does this do?
Change things service to clients service.

N/A

Changes things to clients

Yes

No

### Notes
N/A


rodneyosodo:
What currently works
- Bulk provision of things, channels
- Create a single things and channel
- Retrieve thing/channel info
- Update thing/channel info
- Change thing/channel status

cmd/postgres-reader/main.go
```diff
@@ -81,7 +81,7 @@ func main() {
 	if err := env.Parse(&httpServerConfig, env.Options{Prefix: envPrefixHttp, AltPrefix: envPrefix}); err != nil {
 		log.Fatalf("failed to load %s HTTP server configuration : %s", svcName, err.Error())
 	}
-	hs := httpserver.New(ctx, cancel, svcName, httpServerConfig, api.MakeHandler(repo, tc, auth, svcName), logger)
+	hs := httpserver.New(ctx, cancel, svcName, httpServerConfig, api.MakeHandler(repo, tc, auth, svcName, ), logger)
```
dborovcanin, line 84: Remove this change.

certs/service_test.go
```diff
@@ -1,414 +0,0 @@
-// Copyright (c) Mainflux
```
dborovcanin: Do not remove tests. It's OK if they fail for the time being.

readers/api/endpoint_test.go
```diff
@@ -1,758 +0,0 @@
-// Copyright (c) Mainflux
```
dborovcanin: Same.

auth.proto
```diff
@@ -9,13 +9,6 @@ import "google/protobuf/empty.proto";
 
 option go_package = "./mainflux";
 
-service ThingsService {
```
dborovcanin, line 12: All proto and generated files should be removed from the root at the moment.

dborovcanin (changes requested):

certs/service_test.go
rodneyosodo: It isn't about failing or passing. We have different testing methods. For clients, we don't use maps

things/policies/api/http/transport.go
```diff
@@ -0,0 +1,207 @@
...
+
+func decodeConnectThing(_ context.Context, r *http.Request) (interface{}, error) {
+	req := createPolicyReq{
+		token:    apiutil.ExtractBearerToken(r),
+		GroupID:  bone.GetValue(r, "chanId"),
+		ClientID: bone.GetValue(r, "thingId"),
+	}
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return nil, errors.Wrap(errors.ErrMalformedEntity, err)
+	}
```
dborovcanin, line 99: No need for this since we already extracted all the required data from path params.

cmd/things/main.go
```diff
@@ -103,48 +115,30 @@ func main() {
...
 	}
-	hs1 := httpserver.New(ctx, cancel, "thing-http", httpServerConfig, thhttpapi.MakeHandler(thingsTracer, svc, logger), logger)
+	hsc := httpserver.New(ctx, cancel, "things-clients", httpServerConfig, capi.MakeHandler(csvc, mux, logger), logger)
 
-	// Create new things auth http server
-	authHttpServerConfig := server.Config{Port: defSvcAuthHttpPort}
-	if err := env.Parse(&authHttpServerConfig, env.Options{Prefix: envPrefixAuthHttp, AltPrefix: envPrefix}); err != nil {
-		logger.Fatal(fmt.Sprintf("failed to load %s gRPC server configuration : %s", svcName, err))
+	httpServerConfig = server.Config{Port: defSvcHttpPort}
+	if err := env.Parse(&httpServerConfig, env.Options{Prefix: envPrefixHttp, AltPrefix: envPrefix}); err != nil {
```
dborovcanin, line 128: Do we need 3 calls of 
```go
if err := env.Parse(&httpServerConfig, env.Options{Prefix: envPrefixHttp, AltPrefix: envPrefix}); err != nil {
...
}
```
snippet?

dborovcanin (changes requested):
Sending message with HTTP adapter does not work:
```
{"level":"warn","message":"Method publish to channel 6f4b4ef8-5f86-4f08-b054-e94eeb18cde2.ok.topic1 took 723.933µs to complete with error: rpc error: code = Unimplemented desc = method AuthorizeByKey not implemented.","ts":"2023-03-01T13:25:57.801247521Z"}
```
Check Things gPRC client config, looks like something is wrong with it.
MQTT publishing works.

cmd/things/main.go
rodneyosodo: No. This is solved #1733

things/policies/api/http/transport.go
rodneyosodo: We need it. Since when we connect we can specify the action or not. If we passing in the action i.e `m_read` the clientID will have only that action. If we don't we set `m_read` and `m_write` explicity as demonstrated here https://github.com/rodneyosodo/mainflux/blob/83ab79fe871424cc301f108a376954d56fa564e8/things/policies/api/http/endpoints.go#L73

rodneyosodo:
Regarding this https://github.com/mainflux/mainflux/pull/1713#pullrequestreview-1319820515 it is fixed by https://github.com/mainflux/mainflux/pull/1733/commits/08e5c8da2e0e1fa72927ee813da9c7ee26a00405

things/policies/api/http/transport.go
dborovcanin: Got it. Please take care of the empty body here, since decode will fail if you send no body, and it is a valid request (since we have a fallback).

rodneyosodo: This is fixed by [`45fdec5` (#1735)](https://github.com/mainflux/mainflux/pull/1735/commits/45fdec5a69f8498a824e476aa031dd91985e06bc) - Reference to code is https://github.com/rodneyosodo/mainflux/blob/45fdec5a69f8498a824e476aa031dd91985e06bc/things/policies/api/http/transport.go#L104


I encountered the problem while fixing tests

PR #1716, opened 2023-02-03
MF-1506 - Group-based Access Control

### LINKED ISSUE #1506: Group-based Access Control

### LINKED ISSUE #1813: Migrate from OpenTracing to OpenTelemetry
I have a question: currently, lib jaeger-client-go is DEPRECATED! and move to OpenTelemetry SDK. Does the team have a plan to migrate jaeger-client-go to OpenTelemetry?
https://github.com/jaegertracing/jaeger-client-go

dborovcanin:
### What does this do?
This pull request moves Groups to domain services and introduces group-based policies for access control.

This pull request resolves #1506.
Also resolves #1813.

### List any changes that modify/break current functionality.
There are both changes in the architecture and on the code level. 
Things and Users, while being different services, are somewhat unified to the `Clients` codebase.
For more info, see [the linked issue](https://github.com/mainflux/mainflux/issues/1506).

Yes.

Yes.

### Notes
The goal is that this pull request is merged for release 0.13.0, so we can provide potential fixes and updates and test it in real-world cases before Mainflux 1.0 release. Since this one introduces architectural changes, we're planning to add migration scripts.


cmd/bootstrap/main.go
```diff
@@ -25,7 +20,13 @@ import (
 	httpserver "github.com/mainflux/mainflux/internal/server/http"
 	mflog "github.com/mainflux/mainflux/logger"
 	mfsdk "github.com/mainflux/mainflux/pkg/sdk/go"
+	"github.com/mainflux/mainflux/users/policies"
 	"golang.org/x/sync/errgroup"
+
+	r "github.com/go-redis/redis/v8"
```
dborovcanin: Use `redis`, it's short enough.

cmd/bootstrap/main.go
```diff
@@ -25,7 +20,13 @@ import (
...
+	"github.com/jmoiron/sqlx"
+	"github.com/mainflux/mainflux/bootstrap"
+	api "github.com/mainflux/mainflux/bootstrap/api"
```
dborovcanin: No need for import alias since package name is already `api`.

things/policies/thingsauth.proto
```diff
@@ -0,0 +1,33 @@
...
+
+option go_package = "./policies";
+
+service ThingsService {
+    rpc Authorize(TAuthorizeReq) returns (TAuthorizeRes) {}
+    rpc AuthorizeByKey(TAuthorizeReq) returns (ClientID) {}
+    rpc Identify(Key) returns (ClientID) {}
+}
+
+message TAuthorizeReq {
```
dborovcanin: No need for `T` prefix.

internal/clients/grpc/auth/client.go
```diff
@@ -1,28 +1,28 @@
...
-	"github.com/mainflux/mainflux"
-	authapi "github.com/mainflux/mainflux/auth/api/grpc"
 	grpcClient "github.com/mainflux/mainflux/internal/clients/grpc"
 	"github.com/mainflux/mainflux/internal/env"
 	"github.com/mainflux/mainflux/pkg/errors"
+	"github.com/mainflux/mainflux/users/policies"
+	authapi "github.com/mainflux/mainflux/users/policies/api/grpc"
 )
 
 const envAuthGrpcPrefix = "MF_AUTH_GRPC_"
```
arvindh123, line 15: This can be renamed to `MF_USERS_GRPC_` , Since `MF_USERS_GRPC_PORT` is used in `users` service for gRPC port 

The package can be renamed to `users`

internal/clients/grpc/auth/client.go
```diff
@@ -1,28 +1,28 @@
 package auth
```
arvindh123, line 5: The package can be renamed to `users`

cmd/bootstrap/main.go
```diff
@@ -115,7 +116,7 @@ func main() {
 	}
```
arvindh123, line 127: At present we have auth client 
```go
	auth, authHandler, err := authClient.Setup(envPrefix, cfg.JaegerURL)
	if err != nil {
		logger.Fatal(err.Error())
	}
	defer authHandler.Close()
```
Instead of above code we can have below code , 

```go
	users, usersHandler, err := usersClient.Setup(envPrefix, cfg.JaegerURL)
	if err != nil {
		logger.Fatal(err.Error())
	}
	defer usersHandler.Close()
```
`internal/clients/auth` have default env var `MF_AUTH_GRPC_` 
It should we changed to `MF_USERS_GRPC_` , Since `MF_USERS_GRPC_PORT` is used in users service for gRPC port 
and 
also we can rename the package from `internal/clients/auth` to `internal/clients/users`.

drasko (dismissed):
LGTM

drasko (dismissed):
LGTM

drasko (approved):
LGTM

PR #1728, opened 2023-02-21
NOISSUE - Logger Fatal method returns no value

dborovcanin:
### What does this do?
This pull request makes Mainflux Logger Fatal method return no value.

There is no such issue.

There are no changes in current functionalities.

Yes, tests for Fatal are added and other tests are refactored.

N/A

### Notes
N/A


cmd/auth/main.go
```diff
@@ -61,54 +61,54 @@ func main() {
...
 	cfg := config{}
 	if err := env.Parse(&cfg); err != nil {
-		log.Fatalf("failed to load %s configuration : %s", svcName, err.Error())
+		log.Fatalf("failed to load %s configuration : %s", svcName, err)
 	}
 
 	logger, err := logger.New(os.Stdout, cfg.LogLevel)
 	if err != nil {
-		log.Fatalf(err.Error())
+		log.Fatal(fmt.Sprintf("failed to init logger: %s", err))
```
drasko: Is this `logger` or `log`?

drasko (changes requested):

cmd/auth/main.go
dborovcanin: This is a native Go `log` because `logger` was not initialized properly. But let me also rename it because we have:
```go
logger, err := logger.New(os.Stdout, cfg.LogLevel)
```

logger/logger_test.go
```diff
@@ -37,90 +46,212 @@ func (writer *mockWriter) Read() (logMsg, error) {
...
+		// actually running the necessary Fatal call.
+		cmd.Env = append(os.Environ(), fmt.Sprintf("%s=%s", testFlag, testFlagVal))
+		cmd.Stderr = &writer
+		cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", testMsg, tc.input))
+		err := cmd.Run()
+		if e, ok := err.(*exec.ExitError); ok && !e.Success() {
+			res, err := writer.Read()
+			require.Nil(t, err, "required successful buffer read")
+			assert.Equal(t, e.ExitCode(), 1, fmt.Sprintf("%s: expected exit code %d, got %d", tc.desc, 1, e.ExitCode()))
+			assert.Equal(t, res, tc.output, fmt.Sprintf("%s: expected output %s got %s", tc.desc, tc.output, res))
```
drasko: When you assert, do you put first what you expect and then what you got as the params? As in these two lines this seems to be inverse and is a little bit confusing - because it is not consistent.

I will approve, but please change if you see fit.

drasko (dismissed):
LGTM

drasko (dismissed):
LGTM

drasko (approved):
LGTM

PR #1734, opened 2023-03-01
NOISSUE - Add Tests To Users Service

rodneyosodo:
### What does this do?
Fixes failing tests on users service
Adds pkg tests for users service

NOISSUE

No

Yes

No

### Notes
N/A


pkg/sdk/go/users_test.go
```diff
@@ -771,17 +776,19 @@ func TestUpdateClientIdentity(t *testing.T) {
 	}
 
 	for _, tc := range cases {
-		repoCall := pRepo.On("Evaluate", mock.Anything, mock.Anything, mock.Anything).Return(nil)
-		repoCall1 := cRepo.On("RetrieveByID", mock.Anything, mock.Anything).Return(convertClient(tc.client), tc.err)
-		repoCall2 := cRepo.On("UpdateIdentity", mock.Anything, mock.Anything).Return(convertClient(tc.response), tc.err)
+		repoCall := pRepo.On("CheckAdmin", mock.Anything, mock.Anything).Return(nil)
+		repoCall1 := pRepo.On("Evaluate", mock.Anything, mock.Anything, mock.Anything).Return(nil)
+		repoCall2 := cRepo.On("RetrieveByID", mock.Anything, mock.Anything).Return(convertClient(tc.client), tc.err)
+		repoCall3 := cRepo.On("UpdateIdentity", mock.Anything, mock.Anything).Return(convertClient(tc.response), tc.err)
```
dborovcanin: Let's avoid using `mock.Anything` since that does not assume anything about the actual parameters, while in real scenarios, we usually know those values. Also, let's test whether these calls are actually executed and in which order (`After`, `Before`, `Times`, `parent.AssertCalled`)...

users/clients/mocks/authn.go
```diff
@@ -74,6 +74,6 @@ func (svc authServiceMock) AddPolicy(ctx context.Context, in *policies.AddPolicy
 }
 
 func (svc authServiceMock) DeletePolicy(ctx context.Context, in *policies.DeletePolicyReq, opts ...grpc.CallOption) (*policies.DeletePolicyRes, error) {
-	// Not implemented
+	// Not implemented correctly
```
dborovcanin: Implement it or we risk forgetting it.

dborovcanin (changes requested):

pkg/sdk/go/users_test.go
rodneyosodo: From the transport layer is is difficult to check the actual things pass this is same as for `parent.AssertCalled`. Would be checking the number of times it was called okay `Parent.AssertNumberOfCalls`?

users/clients/mocks/authn.go
rodneyosodo: Fixing it

pkg/sdk/go/users_test.go
dborovcanin: Yes, that should be sufficient.

internal/api/common.go
```diff
@@ -24,6 +24,7 @@ const (
 	ClientKey        = "client"
 	GroupKey         = "group"
 	ActionKey        = "action"
+	IdentityKey      = "identifier"
```
dborovcanin: Remove `IdentifierKey` if not used.

pkg/sdk/go/groups_test.go
```diff
@@ -0,0 +1,766 @@
...
+	}
+	groupSDK := sdk.NewSDK(conf)
+	cases := []struct {
+		desc  string
+		group sdk.Group
+		token string
+		err   errors.SDKError
+	}{
+		{
+			desc:  "register new group",
```
dborovcanin: `create group successfully`

users/clients/api/logging.go
```diff
@@ -95,7 +95,7 @@ func (lm *loggingMiddleware) ListClients(ctx context.Context, token string, pm c
 
 func (lm *loggingMiddleware) UpdateClient(ctx context.Context, token string, client clients.Client) (c clients.Client, err error) {
 	defer func(begin time.Time) {
-		message := fmt.Sprintf("Method update_client_name_and_metadata for token %s took %s to complete", token, time.Since(begin))
+		message := fmt.Sprintf("Method update_client_name_and_metadata for client of id %s using token %s took %s to complete", c.ID, token, time.Since(begin))
```
dborovcanin: `...with id %s using token...`

dborovcanin: Check for the same patter in other logs.

users/policies/api/grpc/client.go
```diff
@@ -113,7 +113,7 @@ func (client grpcClient) Issue(ctx context.Context, req *policies.IssueReq, _ ..
 	ctx, close := context.WithTimeout(ctx, client.timeout)
 	defer close()
 
-	res, err := client.issue(ctx, issueReq{id: req.GetId(), email: req.GetEmail()})
+	res, err := client.issue(ctx, issueReq{email: req.GetEmail(), password: req.GetPassword()})
```
dborovcanin: Extract `issueRequest{}` to the variable.

dborovcanin (changes requested):

users/policies/service_test.go
```diff
@@ -154,23 +158,27 @@ func TestAddPolicy(t *testing.T) {
 
 	for _, tc := range cases {
 		repoCall := pRepo.On("CheckAdmin", context.Background(), mock.Anything).Return(nil)
-		repoCall1 := pRepo.On("Evaluate", context.Background(), "client", mock.Anything).Return(nil)
-		repoCall2 := pRepo.On("Update", context.Background(), tc.policy).Return(tc.err)
+		repoCall1 := pRepo.On("Retrieve", context.Background(), mock.Anything).Return(tc.page, nil)
+		repoCall2 := pRepo.On("Update", context.Background(), mock.Anything).Return(tc.err)
 		repoCall3 := pRepo.On("Save", context.Background(), mock.Anything).Return(tc.err)
-		repoCall4 := pRepo.On("Retrieve", context.Background(), mock.Anything).Return(tc.page, nil)
+		fmt.Println(tc.policy)
```
dborovcanin: Remove log.

users/policies/service_test.go
```diff
@@ -154,23 +158,27 @@ func TestAddPolicy(t *testing.T) {
 
 	for _, tc := range cases {
 		repoCall := pRepo.On("CheckAdmin", context.Background(), mock.Anything).Return(nil)
-		repoCall1 := pRepo.On("Evaluate", context.Background(), "client", mock.Anything).Return(nil)
-		repoCall2 := pRepo.On("Update", context.Background(), tc.policy).Return(tc.err)
+		repoCall1 := pRepo.On("Retrieve", context.Background(), mock.Anything).Return(tc.page, nil)
```
dborovcanin, line 141: Retrieve, Save, and Update make sense, and so does `CheckAdmin`. However, `Authorize is not called here. Can you point out where the `CheckAdmin` call comes from? Also, can that one be embedded in Evaluate policy call?

users/groups/service_test.go
```diff
@@ -267,11 +265,15 @@ func TestUpdateGroup(t *testing.T) {
...
 	for _, tc := range cases {
-		repoCall := pRepo.On("Evaluate", context.Background(), "group", mock.Anything).Return(nil)
+		repoCall := pRepo.On("CheckAdmin", context.Background(), mock.Anything).Return(nil)
 		repoCall1 := gRepo.On("Update", context.Background(), mock.Anything).Return(tc.response, tc.err)
 		expectedGroup, err := svc.UpdateGroup(context.Background(), tc.token, tc.group)
 		assert.True(t, errors.Contains(err, tc.err), fmt.Sprintf("%s: expected %s got %s\n", tc.desc, tc.err, err))
 		assert.Equal(t, tc.response, expectedGroup, fmt.Sprintf("%s: expected %v got %v\n", tc.desc, tc.response, expectedGroup))
+		if tc.err == nil {
+			repoCall.Parent.AssertCalled(t, "CheckAdmin", context.Background(), mock.Anything)
+			repoCall1.Parent.AssertCalled(t, "Update", context.Background(), mock.Anything)
```
dborovcanin: `AssertCalled` does not do assertion itself, but **returns a bool value** that needs to be checked. Either way, it looks like this `if err == nil` is irrelevant, since these two methods are always called.

dborovcanin (changes requested):

users/policies/service_test.go
rodneyosodo: `Authorize` on service layer contains both `CheckAdmin` which comes before `Evaluate`. We are using a token from an admin user. If it would have been a normal user we would have both `CheckAdmin` and `Evaluate`

users/groups/service_test.go
rodneyosodo: Actually no. Since we have an `"update group name with invalid token"` it would fail on identify

rodneyosodo: Chaning it to something like this

```go
if tc.err == nil {
	if !repoCall.Parent.AssertCalled(t, "CheckAdmin", context.Background(), mock.Anything) {
		assert.Fail(t, fmt.Sprintf("CheckAdmin was not called on %s", tc.desc))
	}
	if !repoCall1.Parent.AssertCalled(t, "Update", context.Background(), mock.Anything) {
		assert.Fail(t, fmt.Sprintf("Update was not called on %s", tc.desc))
	}
}
```

dborovcanin: You can use constructs:
```go
c := repoCall.Parent.AssertCalled(t, "Save", context.Background(), mock.Anything)
assert.True(t, c, fmt.Sprintf("Save was not called on %s", tc.desc))
```
instead of `if`.

dborovcanin (approved):

PR #1740, opened 2023-03-07
NOISSUE - Fix Send Response for Publishing Message

rodneyosodo:
### What does this do?
Adds functionality to send response for successfully publishing messages.

NOISSUE

Previously, we only sent response when we encountered an error. This meant when you try ab publish message using coap-cli it would not automatically close after successfully publishing the message. It would close after the timeout has reached. This was because we were not returning a response upon successful publishing of message

No

No

### Notes
How to reproduce the problem would be to try and publish message using coap-cli

```bash
coap-cli post channels/<channel_id>/messages -auth <thing_key> -d 'message'
```

### DIFF
--- a/coap/api/transport.go
+++ b/coap/api/transport.go
@@ -79,19 +79,17 @@ func handler(w mux.ResponseWriter, m *mux.Message) {
 		Context: m.Context,
 		Options: make(message.Options, 0, 16),
 	}
-
+	defer sendResp(w, &resp)
 	msg, err := decodeMessage(m)
 	if err != nil {
 		logger.Warn(fmt.Sprintf("Error decoding message: %s", err))
 		resp.Code = codes.BadRequest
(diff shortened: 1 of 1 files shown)


SammyOina (dismissed):

dborovcanin (dismissed):

coap/api/transport.go
```diff
@@ -114,8 +114,8 @@ func handler(w mux.ResponseWriter, m *mux.Message) {
 		default:
 			resp.Code = codes.InternalServerError
 		}
-		sendResp(w, &resp)
 	}
+	sendResp(w, &resp)
```
dborovcanin: Consider using `defer sendResp(w, &resp)` as a little more elegant alternative.

drasko (approved):
LGTM

dborovcanin (approved):

PR #1742, opened 2023-03-09
NOISSUE -  Blocking and Async Consumer Interface

AryanGodara:
Add the support for blocking (synchronous) and non-blocking(asynchronous) messaging to the Consumer interface

There is no such issue.

There are no such changes. There is an async API added to support async consumers. It is very unlikely that one consumer will support both sync and async mode in the same API. We did not want to sacrifice the consumer interface by artificially adding async support, so we created a separate, async interface.

No.

No.

### Notes
N/A


consumers/notifiers/service.go
```diff
@@ -127,3 +127,36 @@ func (ns *notifierService) Consume(message interface{}) error {
...
+
+	var to []string
+	for _, sub := range page.Subscriptions {
+		to = append(to, sub.Contact)
+	}
+	if len(to) > 0 {
+		err := ns.notifier.Notify(ns.from, to, msg)
+		if err != nil {
+			errs <- errors.Wrap(ErrNotify, err)
+		}
```
SammyOina:
```suggestion
		if err := ns.notifier.Notify(ns.from, to, msg);err != nil {
			errs <- errors.Wrap(ErrNotify, err)
		}
```

SammyOina (changes requested):

consumers/writers/api/logging.go
```diff
@@ -38,5 +38,21 @@ func (lm *loggingMiddleware) Consume(msgs interface{}) (err error) {
...
+	ch := make(chan error)
+	var err error
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method consume took %s to complete", time.Since(begin))
+		if err != nil {
+			lm.logger.Warn(fmt.Sprintf("%s with error: %s.", message, err))
+			return
+		}
+		lm.logger.Info(fmt.Sprintf("%s without errors.", message))
+	}(time.Now())
```
dborovcanin: This method **is blocking** and `consumer.Consume` will practically never (successfully) end, so logging this way does not make sense.

consumers/writers/api/metrics.go
```diff
@@ -30,10 +30,18 @@ func MetricsMiddleware(consumer consumers.Consumer, counter metrics.Counter, lat
...
+func (mm *metricsMiddleware) ConsumeBlocking(msgs interface{}) error {
 	defer func(begin time.Time) {
 		mm.counter.With("method", "consume").Add(1)
 		mm.latency.With("method", "consume").Observe(time.Since(begin).Seconds())
 	}(time.Now())
-	return mm.consumer.Consume(msgs)
+	return mm.consumer.ConsumeBlocking(msgs)
+}
+
+func (mm *metricsMiddleware) Consume(msgs interface{}, errs chan<- error) {
```
dborovcanin: Same here.

consumers/consumer.go
```diff
@@ -5,7 +5,11 @@ package consumers
 
 // Consumer specifies message consuming API.
 type Consumer interface {
-	// Consume method is used to consumed received messages.
-	// A non-nil error is returned to indicate operation failure.
-	Consume(messages interface{}) error
+	// ConsumeBlocking method is used to consume received messages synchronously
```
dborovcanin: Consider using `ConsumerBlocking` and `ConsumeAsync`.

consumers/messages.go
```diff
@@ -59,7 +59,10 @@ func handle(t transformers.Transformer, c Consumer) handleFunc {
 				return err
 			}
 		}
-		return c.Consume(m)
+		errs := make(chan error)
+		c.Consume(m, errs)
```
dborovcanin: This will not work. You need to call the correct method - i.e. you need an indicator of which mode you want to use.

dborovcanin (changes requested):

consumers/messages.go
AryanGodara: Can I add a `async bool` variable to the consumer interface, which returns which mode is being used? @dborovcanin

consumers/writers/api/logging.go
AryanGodara: I have updated these in previous commit

consumers/messages.go
dborovcanin: No need to change the interface, but you can change `Start` method of `conusmers/messages.go` to accept this bool value.

consumers/writers/api/logging.go
```diff
@@ -13,30 +13,61 @@ import (
...
+		message := fmt.Sprintf("Method consume took %s to complete", time.Since(begin))
+		if err != nil {
+			alm.logger.Warn(fmt.Sprintf("%s with error: %s.", message, err))
+			return
+		}
+		alm.logger.Info(fmt.Sprintf("%s without errors.", message))
+	}(time.Now())
+
+	go alm.asyncConsumer.ConsumeAsync(msgs, ch)
+	err = <-ch
```
dborovcanin: This will log only a single error. We need an infinite loop in a separate routine.

consumers/consumer.go
```diff
@@ -4,8 +4,14 @@
...
 
 // Consumer specifies message consuming API.
-type Consumer interface {
-	// Consume method is used to consumed received messages.
-	// A non-nil error is returned to indicate operation failure.
-	Consume(messages interface{}) error
+type AsyncConsumer interface {
+	// Consume method is used to consume received messages.
+	// An error channel is used to handle errors, so it supports the async approach.
+	ConsumeAsync(messages interface{}, errs chan<- error)
```
dborovcanin: Consider adding the `Errors() <- chan error` method instead of receiving `errs` chan.

dborovcanin (changes requested):

consumers/consumer.go
```diff
@@ -4,8 +4,15 @@
 package consumers
 
 // Consumer specifies message consuming API.
-type Consumer interface {
-	// Consume method is used to consumed received messages.
-	// A non-nil error is returned to indicate operation failure.
-	Consume(messages interface{}) error
+type AsyncConsumer interface {
+	// Consume method is used to consume received messages.
+	// An error channel is used to handle errors, so it supports the async approach.
```
dborovcanin: Move the error doc comment above `Errors()`. These comments are not here only for Godoc, but also to provide useful information: is the operation supposed to be safe for async use, how it is supposed to be used, what are the design limitations and potential errors... Something like the InfluxDB:
> 	// Errors returns a channel for reading errors that occur during async writes.
	// Must be called before performing any writes for errors to be collected.
	// The chan is unbuffered and must be drained or the consumer will block.

consumers/consumer.go
```diff
@@ -4,8 +4,15 @@
...
+	ConsumeAsync(messages interface{})
+	Errors() <-chan error
+}
+
+type SyncConsumer interface {
+	// ConsumeBlocking method is used to consume received messages synchronously
```
dborovcanin: Add full stops to be consistent whit other interface definitions.

consumers/writers/api/logging.go
```diff
@@ -13,30 +13,65 @@ import (
...
+		message := fmt.Sprintf("Method ConsumeAsync took %s to complete", time.Since(begin))
+		if err != nil {
+			alm.logger.Warn(fmt.Sprintf("%s with error: %s.", message, err))
+			return
+		}
+		alm.logger.Info(fmt.Sprintf("%s without errors.", message))
+	}(time.Now())
+
+	go alm.asyncConsumer.ConsumeAsync(msgs)
+	err = <-ch
```
dborovcanin: Not only this is incorrect because errors need to be processed **in an infinite loop**, but this will also cause **consumer to block if the errors channel is unbuffered**. I'd suggest that you remove AsyncLogger (and handle logging in the `main.go`) because, unlike for sync operations, the error consumer is an integral part of the async consumers and is not only the logger but also the error handler.

consumers/writers/api/metrics.go
```diff
@@ -12,28 +12,61 @@ import (
...
 
-// MetricsMiddleware returns new message repository
+// AsyncMetricsMiddleware returns new message repository
 // with Save method wrapped to expose metrics.
-func MetricsMiddleware(consumer consumers.Consumer, counter metrics.Counter, latency metrics.Histogram) consumers.Consumer {
-	return &metricsMiddleware{
-		counter:  counter,
-		latency:  latency,
-		consumer: consumer,
+func AsyncMetricsMiddleware(asyncConsumer consumers.AsyncConsumer, counter metrics.Counter, latency metrics.Histogram) consumers.AsyncConsumer {
```
dborovcanin: These async metrics make no sense since it's only one call and latency is very difficult to measure. Remove it.

consumers/writers/influxdb/consumer.go
```diff
@@ -32,19 +33,76 @@ type RepoConfig struct {
...
+				return
+			}
+		}
+	}(done)
+
+	for _, pt := range pts {
+		repo.writeAPI.WritePoint(pt)
+	}
+
+	repo.writeAPI.Flush()
```
dborovcanin, line 94: Instead of manually flushing, set a flush interval and let the client take care of it. Looks like defaults are pretty good for us, so you can just omit this flush.
(Defaults:
```go
// DefaultOptions returns Options object with default values
func DefaultOptions() *Options {
	return &Options{batchSize: 5_000, flushInterval: 1_000, precision: time.Nanosecond, useGZip: false, retryBufferLimit: 50_000, defaultTags: make(map[string]string),
		maxRetries: 5, retryInterval: 5_000, maxRetryInterval: 125_000, maxRetryTime: 180_000, exponentialBase: 2}
}
```
)

consumers/writers/influxdb/consumer.go
```diff
@@ -32,19 +33,76 @@ type RepoConfig struct {
...
+}
+
+func (repo *influxRepo) Errors() <-chan error {
+	if repo.errCh != nil {
+		return repo.errCh
+	}
+
+	return make(<-chan error, 1)
```
dborovcanin: This is wrong and extremely dangerous. I let you explain why. :)

consumers/writers/influxdb/consumer_test.go
```diff
@@ -118,107 +118,103 @@ func queryDB(fluxQuery string) (int, error) {
...
-				msgsNum:      1,
-				expectedSize: 1,
-			},
-			{
-				desc:         "save a batch of messages",
-				msgsNum:      streamsSize,
-				expectedSize: streamsSize,
-			},
-		}
+	// Testing both async and sync
```
dborovcanin: You can separate Async and Sync cases, it's a little cleaner solution. The same goes for JSON and Readers tests.

dborovcanin (changes requested):

consumers/writers/influxdb/consumer.go
AryanGodara: I was worried about this one 😅 . I wanted to set `panic("unimplemented")` if repo.errCh == nil (ie, for sync consumer) or just have two separate repo structs (one for sync and one for async).

consumers/writers/influxdb/consumer.go
AryanGodara: Tests fail if I don't flush at the end. The messages don't get saved.
I tried default options with `SetFlushInterval(time.Millisecond)` to 1000, 100, 10, and 1. And all of these still cause the tests to fail.

cmd/influxdb-writer/main.go
```diff
@@ -73,9 +72,16 @@ func main() {
 	}
 	defer client.Close()
 
-	repo := newService(client, repocfg, logger)
+	repo := influxdb.NewAsync(client, repocfg)
+
+	go func(log mflog.Logger) {
+		errCh := repo.Errors()
+		for err := range errCh {
+			log.Error(err.Error())
```
dborovcanin: Use `log.Error(err)` in case the writer publishes `nil` error to, say, indicate it is closed.

consumers/writers/api/logging.go
```diff
@@ -13,30 +13,65 @@ import (
...
 )
 
-var _ consumers.Consumer = (*loggingMiddleware)(nil)
+// var _ consumers.AsyncConsumer = (*asyncLoggingMiddleware)(nil)
+var _ consumers.SyncConsumer = (*syncLoggingMiddleware)(nil)
 
-type loggingMiddleware struct {
-	logger   log.Logger
-	consumer consumers.Consumer
+// type asyncLoggingMiddleware struct {
```
dborovcanin: Remove commented code.

consumers/writers/api/logging.go
```diff
@@ -13,30 +13,65 @@ import (
...
+	syncConsumer consumers.SyncConsumer
 }
 
-// LoggingMiddleware adds logging facilities to the adapter.
-func LoggingMiddleware(consumer consumers.Consumer, logger log.Logger) consumers.Consumer {
-	return &loggingMiddleware{
-		logger:   logger,
-		consumer: consumer,
+// SyncLoggingMiddleware adds logging facilities to the adapter.
+func SyncLoggingMiddleware(syncConsumer consumers.SyncConsumer, logger log.Logger) consumers.SyncConsumer {
```
dborovcanin: No need for `Sync` prefix - we only have these for `Sync` consumers.  Mention that in, say, `doc.go` file comment that you add to this package.

consumers/writers/api/metrics.go
```diff
@@ -12,28 +12,61 @@ import (
...
 
-var _ consumers.Consumer = (*metricsMiddleware)(nil)
+// var _ consumers.AsyncConsumer = (*asyncMetricsMiddleware)(nil)
+var _ consumers.SyncConsumer = (*syncMetricsMiddleware)(nil)
 
-type metricsMiddleware struct {
-	counter  metrics.Counter
-	latency  metrics.Histogram
-	consumer consumers.Consumer
+// type asyncMetricsMiddleware struct {
```
dborovcanin: Same, remove unused code, and no need for `Sync` prefix.

consumers/writers/influxdb/consumer.go
dborovcanin: Tested, looks like a bug in InfluxDB client.

consumers/writers/influxdb/consumer_test.go
```diff
@@ -118,119 +118,141 @@ func queryDB(fluxQuery string) (int, error) {
...
-				case 3:
-					msg.DataValue = &dataV
-				case 4:
-					msg.Sum = &sum
-				}
-
-				msg.Time = float64(now)/float64(1e9) - float64(i)
-				msgs = append(msgs, msg)
-			}
+	// Test async
```
dborovcanin: Separate Sync and async tests, since it's also separated on the interface level (`TestSyncConsumer` and `TestAsyncConsumer`).

go.mod
```diff
@@ -99,6 +99,7 @@ require (
 	github.com/hashicorp/yamux v0.1.1 // indirect
 	github.com/imdario/mergo v0.3.13 // indirect
 	github.com/inconshreveable/mousetrap v1.0.1 // indirect
+	github.com/influxdata/influxdb-client-go v1.4.0 // indirect
```
dborovcanin: You can also update `github.com/influxdata/influxdb-client-go/v2` to ` v2.12.3`.

dborovcanin (changes requested):

cmd/influxdb-writer/main.go
AryanGodara: log.Error() expects a string, so I put a err !=nil check

consumers/messages.go
```diff
@@ -33,7 +33,7 @@ var (
 // Start method starts consuming messages received from Message broker.
 // This method transforms messages to SenML format before
 // using MessageRepository to store them.
-func Start(id string, sub messaging.Subscriber, consumer Consumer, configPath string, logger logger.Logger) error {
+func Start(id string, sub messaging.Subscriber, consumer interface{}, configPath string, logger logger.Logger, async bool) error {
```
SammyOina:
```suggestion
func Start(id string, sub messaging.Subscriber, consumer interface{}, configPath string, logger logger.Logger) error {
```
async is unused

consumers/writers/api/logging.go
```diff
@@ -13,30 +13,30 @@ import (
...
 type loggingMiddleware struct {
-	logger   log.Logger
-	consumer consumers.Consumer
+	logger       log.Logger
+	syncConsumer consumers.SyncConsumer
 }
 
-// LoggingMiddleware adds logging facilities to the adapter.
-func LoggingMiddleware(consumer consumers.Consumer, logger log.Logger) consumers.Consumer {
+// SyncLoggingMiddleware adds logging facilities to the adapter.
```
SammyOina:
```suggestion
// LoggingMiddleware adds logging facilities to the adapter.
```

consumers/writers/influxdb/consumer.go
```diff
@@ -32,19 +33,75 @@ type RepoConfig struct {
...
+	}
+
+	repo.writeAPI.Flush()
+}
+
+func (repo *influxRepo) Errors() <-chan error {
+	if repo.errCh != nil {
+		return repo.errCh
+	}
+	panic("unimplemented")
```
SammyOina:
```suggestion
	
```

consumers/writers/mongodb/consumer.go
```diff
@@ -18,25 +18,28 @@ const senmlCollection string = "messages"
...
 	switch m := message.(type) {
 	case json.Messages:
 		return repo.saveJSON(m)
 	default:
 		return repo.saveSenml(m)
 	}
+}
 
+func (repo *mongoRepo) ConsumeAsync(message interface{}, errs chan<- error) {
+	panic("unimplemented")
```
SammyOina:
```suggestion
	
```

internal/clients/influxdb/influxdb.go
```diff
@@ -15,14 +15,19 @@ var (
...
+	Port               string        `env:"PORT"                  envDefault:"8086"`
+	Username           string        `env:"ADMIN_USER"            envDefault:"mainflux"`
+	Password           string        `env:"ADMIN_PASSWORD"        envDefault:"mainflux"`
+	DbName             string        `env:"DB"                    envDefault:"mainflux"`
+	Bucket             string        `env:"BUCKET"                envDefault:"mainflux-bucket"`
+	Org                string        `env:"ORG"                   envDefault:"mainflux"`
+	Token              string        `env:"TOKEN"                 envDefault:"mainflux-token"`
+	DBUrl              string        `env:"DBURL"                 envDefault:""`
+	UserAgent          string        `env:"USER_AGENT"            envDefault:"InfluxDBClient"`
+	Timeout            time.Duration `env:"TIMEOUT"` // Influxdb client configuration by default there is no timeout duration , this field will not have fallback default timeout duration Reference: https://pkg.go.dev/github.com/influxdata/influxdb@v1.10.0/client/v2#HTTPConfig
```
SammyOina:
```suggestion
	Timeout            time.Duration `env:"TIMEOUT"` // Influxdb client configuration by default has no timeout duration, this field will not have a fallback default timeout duration. Reference: https://pkg.go.dev/github.com/influxdata/influxdb@v1.10.0/client/v2#HTTPConfig
```

SammyOina (changes requested):

consumers/consumer.go
```diff
@@ -4,8 +4,16 @@
...
-	// Consume method is used to consumed received messages.
+type AsyncConsumer interface {
+	// Consume method is used to consume received messages.
+	ConsumeAsync(messages interface{})
+
+	// An error channel is used to handle errors, so it supports the async approach.
+	Errors() <-chan error
+}
+
+type SyncConsumer interface {
```
dborovcanin: While I like the `SyncConsumer` naming as opposed to `AsyncConsumer`, let's rename it to `BlockingConsumer`, because `Sync` terminology is not correct - this is not synced with anything, it is just blocking (and the method name is `ConsumeBlocking` which is correct).

consumers/messages.go
```diff
@@ -59,7 +82,10 @@ func handle(t transformers.Transformer, c Consumer) handleFunc {
 				return err
 			}
 		}
-		return c.Consume(m)
+
+		errs := ac.Errors()
+		ac.ConsumeAsync(m)
+		return <-errs
```
dborovcanin: There is no point in waiting for errors - this results in a blocking operation, which is counter-productive. It also makes the writer required to send `nil` errors in case of a successful operation. We handle errors elsewhere, so simply return nil here. Subscriber Handler is successful, and Consumer might fail, but those errors are handled in a separate routine.

consumers/notifiers/service.go
```diff
@@ -127,3 +127,35 @@ func (ns *notifierService) Consume(message interface{}) error {
 
 	return nil
 }
+
+func (ns *notifierService) ConsumeAsync(message interface{}, errs chan<- error) {
```
dborovcanin: `ConsumerAsync` makes a lot of sense in notifiers, but this does not implement the correct interface. :)
Please add checks `var _ consumers.AsyncConsumer = (*notifierService)(nil)`.

internal/clients/influxdb/influxdb.go
```diff
@@ -36,7 +41,11 @@ func Setup(envPrefix string, ctx context.Context) (influxdb2.Client, error) {
 
 // Connect create InfluxDB client and connect to InfluxDB server
 func Connect(config Config, ctx context.Context) (influxdb2.Client, error) {
-	client := influxdb2.NewClient(config.DBUrl, config.Token)
+	// client := influxdb2.NewClient(config.DBUrl, config.Token)
```
dborovcanin: Remove commented code.

readers/influxdb/messages_test.go
```diff
@@ -50,7 +50,9 @@ var (
 )
 
 func TestReadSenml(t *testing.T) {
-	writer := iwriter.New(client, repoCfg, true)
+	// Testing both async and sync
```
dborovcanin: This is a Reader test, no need to use both types of Consumers.

dborovcanin (changes requested):

SammyOina (dismissed):

readers/influxdb/messages_test.go
AryanGodara: Should I use sync or async consumer for this test?

AryanGodara: I used async

consumers/writers/api/logging.go
```diff
@@ -13,30 +13,30 @@ import (
...
 )
 
-var _ consumers.Consumer = (*loggingMiddleware)(nil)
+var _ consumers.BlockingConsumer = (*loggingMiddleware)(nil)
 
 type loggingMiddleware struct {
-	logger   log.Logger
-	consumer consumers.Consumer
+	logger       log.Logger
+	syncConsumer consumers.BlockingConsumer
```
dborovcanin: Use simply `consumer` instead of `syncConsumer`. Also `func(lm *loggingMiddleware)` instead of `func(slm *loggingMiddleware)` as a receiver.

consumers/writers/api/metrics.go
```diff
@@ -12,28 +12,28 @@ import (
...
-var _ consumers.Consumer = (*metricsMiddleware)(nil)
+var _ consumers.BlockingConsumer = (*metricsMiddleware)(nil)
 
 type metricsMiddleware struct {
-	counter  metrics.Counter
-	latency  metrics.Histogram
-	consumer consumers.Consumer
+	counter      metrics.Counter
+	latency      metrics.Histogram
+	syncConsumer consumers.BlockingConsumer
```
dborovcanin: Same here: `consumer` and `mm` instead of `smm`.

consumers/consumer.go
```diff
@@ -3,9 +3,18 @@
 
 package consumers
 
-// Consumer specifies message consuming API.
-type Consumer interface {
-	// Consume method is used to consumed received messages.
+// AsyncConsumer specifies asynchronous message consuming API.
```
dborovcanin: Improve comments, I already shared an example from InfluxDB code.

consumers/consumer.go
```diff
@@ -3,9 +3,18 @@
...
+type AsyncConsumer interface {
+	// Consume method is used to consume received messages.
+	ConsumeAsync(messages interface{})
+
+	// An error channel is used to handle errors, so it supports the async approach.
```
dborovcanin: It's worth mentioning that the Error channel may receive `nil` error to indicate the success of async execution.

consumers/writers/influxdb/consumer_test.go
```diff
@@ -227,10 +278,107 @@ func TestSaveJSON(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range cases {
+		err := resetBucket()
+		assert.Nil(t, err, fmt.Sprintf("Cleaning data from InfluxDB expected to succeed: %s.\n", err))
+
+		errs := asyncRepo.Errors()
+		asyncRepo.ConsumeAsync(msgs)
+		err = <-errs
```
dborovcanin:
```suggestion
		asyncRepo.ConsumeAsync(msgs)
		err = <- asyncRepo.Errors()
```

dborovcanin: If something is wrong, `err = <- asyncRepo.Errors()` might get stuck, so probably something like:
```go
tmr := time.NewTimer(1 * time.Millisecond)
		select {
		case err = <-asyncRepo.Errors():
			break
		case <-tmr.C:
			// log error, fail, but continue
		}
``` 
is needed.

dborovcanin (changes requested):
Almost there.

dborovcanin (dismissed):

SammyOina (approved):

dborovcanin (approved):

drasko (approved):
LGTM

PR #1747, opened 2023-03-15
Github Actions PR

dborovcanin:


dborovcanin:
Accidentally sent it to `mainflux` upstream instead of fork.

PR #1766, opened 2023-04-12
NOISSUE - Set To MF_USERS_GRPC_ Var to Docker-Compose

rodneyosodo:
### What does this do?
Sets `MF_USERS_GRPC_` to the `MF_AUTH_GRPC_` variable in docker-compose file and README docs

Noissue

Env vars

No

Yes

### Notes
N/A


dborovcanin (commented):
Let's think about this one because Users act like Auth providers so this change might not be necessary.

SammyOina:
exclude merge commits

rodneyosodo:
> Let's think about this one because Users act like Auth providers so this change might not be necessary.

Repurposing this so that we have `MF_AUTH_` set as an environment variable

rodneyosodo:
CI is fixed by [`6e919ba` (#1781)](https://github.com/mainflux/mainflux/pull/1781/commits/6e919ba872434c328af544e22410687ea0a11490)

dborovcanin (approved):

PR #1781, opened 2023-04-28
NOISSUE - Seperate Identify and Authorize

rodneyosodo:
### What does this do?
Seperate Identify and Authorize

Noissue

Instead of passing in token as subject when doing authorize, we passing in clientID after doing identify

No

No

### Notes
N/A

### DIFF
--- a/pkg/sdk/go/users_test.go
+++ b/pkg/sdk/go/users_test.go
@@ -355,15 +355,17 @@ func TestListClients(t *testing.T) {
 			Tag:      tc.tag,
 		}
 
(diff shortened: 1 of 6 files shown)


drasko (commented):
Why did we separate those? I guess there are some cases where you need only to authenticate (identify), not to authorize, right?

BTW, @dborovcanin would it make sense to add one function `authX()`, or `identifyAndAuthorize()` or something like this, that would do both operations atomically, for the cases where this is needed (I see that there are code repetitions)?

SammyOina (approved):

PR #1790, opened 2023-05-11
NOISSUE - Fix Add Policy If Policy Exists

rodneyosodo:
### What does this do?
If policy exists in the database we need to update it with only the new fields.

Noissue

When policy exist we update it rather than creating another one

No

No

### Notes
Changed the primary key constain to be only on `subject` `object` rather than having it on `subject` `object` `action`

### DIFF
--- a/users/policies/service.go
+++ b/users/policies/service.go
@@ -83,20 +83,8 @@ func (svc service) AddPolicy(ctx context.Context, token string, p Policy) error 
 		return err
 	}
 
-	// If the policy already exists, update it appending the new actions
+	// If the policy already exists, replace the actions
 	if len(page.Policies) == 1 {
-		p.Actions = append(p.Actions, page.Policies[0].Actions...)
-
-		isUnique := make(map[string]bool)
-		var uniqueActions []string
-		for _, action := range p.Actions {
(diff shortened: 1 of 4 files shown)


users/policies/service.go
```diff
@@ -72,31 +73,35 @@ func (svc service) AddPolicy(ctx context.Context, token string, p Policy) error
...
 	}
 
-	page, err := svc.policies.Retrieve(ctx, Page{Subject: p.Subject, Object: p.Object})
+	pm := Page{Subject: p.Subject, Object: p.Object, Offset: 0, Limit: 1}
+	page, err := svc.policies.Retrieve(ctx, pm)
 	if err != nil {
 		return err
 	}
+
+	// If the policy already exists, update it appending the new actions
```
dborovcanin: No need to make it complex, just replace actions with new ones. If someone wants to update the policy, they will use the update endpoint.

dborovcanin (changes requested):

users/policies/service.go
rodneyosodo: Okay

dborovcanin (approved):

PR #1792, opened 2023-05-11
NOISSUE - Update Add and Delete Policies

rodneyosodo:
### What does this do?
AddPolicy adds a policy is added if:
1. The client is an admin
2. The client has `g_add` action on the object.
3. The client is the owner of the subject.

checkPolicy checks for the following:
1. Check if the client is an admin
2. Check if the client is the owner of the policy

Noissue

No

Yes

### Notes
N/A


dborovcanin (changes requested):
Add in a new policy flow goes like this:
1. Check if I'm Admin
2. Check if I'm a member of the group with right to add members (not necessarily the owner of the group)
3. Check if I'm adding a thing or a user (we'll need that info in the request, default is thing type):
    - if user, use gRPC to check if the user exits (in the future - if the user exists and is in the same scope)
    - if thing, check if the thing is owned by me (in the future, maybe some other checks such if the thing is already shared with me)

things/policies/service.go
```diff
@@ -178,3 +213,29 @@ func (svc service) hasThing(ctx context.Context, p Policy) (string, error) {
...
+		return "", errors.Wrap(errors.ErrAuthorization, err)
+	}
+	return res.GetId(), nil
+}
+
+func (svc service) checkAdmin(ctx context.Context, id string) error {
+	req := &upolicies.AuthorizeReq{
+		Sub:        id,
+		Obj:        thingsObjectKey,
+		Act:        "c_update", // TODO: remove since this is not used
```
drasko: Do we need this TODO?

things/groups/service.go
```diff
@@ -181,31 +182,37 @@ func (svc service) changeGroupStatus(ctx context.Context, token string, group gr
...
 	}
+
+	policy := tpolicies.Policy{Subject: subject, Object: object, Actions: []string{action}}
+	if err := policy.Validate(); err != nil {
+		return err
+	}
+	return svc.policies.Authorize(ctx, entityType, policy)
+}
+
+// TODO : Only accept token as parameter since object and action are irrelevant.
```
drasko: Do we need this TODO?

things/clients/service.go
```diff
@@ -250,50 +275,67 @@ func (svc service) Identify(ctx context.Context, key string) (string, error) {
...
-	req := &policies.AuthorizeReq{
+
+	policy := tpolicies.Policy{Subject: subject, Object: object, Actions: []string{action}}
+	if err := policy.Validate(); err != nil {
+		return err
+	}
+	return svc.policies.Authorize(ctx, clientEntityType, policy)
+}
+
+// TODO : Only accept token as parameter since object and action are irrelevant.
```
drasko, line 333: Do we need this TODO?

drasko (changes requested):

things/policies/service.go
rodneyosodo: Yes. @dborovcanin had suggested we move admin check to `JWT` layer so that we have role information in the `JWT` token

dborovcanin (changes requested):
@rodneyosodo Please check @drasko's comments and address (or answer) them.

dborovcanin (approved):

PR #1794, opened 2023-05-12
NOISSUE - Fix List Users When It is Admin

rodneyosodo:
### What does this do?
List all users when it is admin

Noissue

No

Yes

No

### Notes
N/A

### DIFF
--- a/users/clients/service.go
+++ b/users/clients/service.go
@@ -173,9 +173,8 @@ func (svc service) ListClients(ctx context.Context, token string, pm mfclients.P
 	pm.Action = "c_list"
 
 	// If the user is admin, fetch all things from database.
-	pol := policies.Policy{Subject: id, Object: clientsObjectKey, Actions: []string{listRelationKey}}
-	if err := svc.authorize(ctx, clientsObjectKey, pol); err == nil {
-		pm.Owner = ""
+	p := policies.Policy{Subject: id, Object: clientsObjectKey, Actions: []string{listRelationKey}}
+	if err := svc.authorize(ctx, clientsObjectKey, p); err == nil {
 		pm.SharedBy = ""
 		pm.Action = ""
 	}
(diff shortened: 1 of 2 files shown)


users/clients/service.go
```diff
@@ -163,13 +163,24 @@ func (svc service) ListClients(ctx context.Context, token string, pm mfclients.P
 	if err != nil {
 		return mfclients.ClientsPage{}, err
 	}
+
+	// If the user is admin, fetch all things from database.
+	pol := policies.Policy{Subject: id, Object: clientsObjectKey, Actions: []string{listRelationKey}}
+	if err := svc.authorize(ctx, clientsObjectKey, pol); err == nil {
+		pm.Owner = ""
+		pm.SharedBy = ""
+		goto QUERY_DB
```
dborovcanin: Avoid using labels and `goto`.

dborovcanin (changes requested):

users/clients/service.go
```diff
@@ -163,13 +163,23 @@ func (svc service) ListClients(ctx context.Context, token string, pm mfclients.P
...
 	if pm.SharedBy == MyKey {
 		pm.SharedBy = id
 	}
 	if pm.Owner == MyKey {
 		pm.Owner = id
 	}
 	pm.Action = "c_list"
+
+	// If the user is admin, fetch all things from database.
+	pol := policies.Policy{Subject: id, Object: clientsObjectKey, Actions: []string{listRelationKey}}
```
dborovcanin: Use `p` instead of `pol`. Does this mean that we disable filtering by, say, owner for admin?

dborovcanin (changes requested):

users/clients/service.go
rodneyosodo: Currently we don't filter by `owner`. This is because visibility changes the owner field

dborovcanin: We need to provide that filter in the future. Please open a ticket for that.

dborovcanin (approved):

PR #1795, opened 2023-05-15
NOISSUE - Fix Event Sourcing and Authorization Cache

rodneyosodo:

### What does this do?
1. It enables event sourcing for things service. This for all service operations in the clients, groups and policies domain

image

2. It fixes error on Authorization when publishing message on HTTP adapter endpoint

image

3. It removed timestamp on mqtt events as the timestamps are generated automatically and they can be referenced using the `id` field

image

4. Enabled event sourcing for changing things secret but we don't send the secret over the wire.

Noissue

Listed above.

Manually tested with the attached screenshots

No

### Notes
N/A

### DIFF
--- a/bootstrap/postgres/configs.go
+++ b/bootstrap/postgres/configs.go
@@ -558,7 +558,7 @@ func nullString(s string) sql.NullString {
 }
 
 func nullTime(t time.Time) sql.NullTime {
-	if t == (time.Time{}) {
+	if t.IsZero() {
 		return sql.NullTime{}
 	}
 
--- a/bootstrap/postgres/configs.go
+++ b/bootstrap/postgres/configs.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/jackc/pgerrcode"
 	"github.com/jackc/pgtype"
@@ -15,6 +16,7 @@ import (
 	"github.com/jmoiron/sqlx"
 	"github.com/mainflux/mainflux/bootstrap"
 	mflog "github.com/mainflux/mainflux/logger"
+	"github.com/mainflux/mainflux/pkg/clients"
 	"github.com/mainflux/mainflux/pkg/errors"
 )
 
@@ -391,7 +393,8 @@ func (cr configRepository) UpdateChannel(c bootstrap.Channel) error {
 		return errors.Wrap(errors.ErrUpdateEntity, err)
 	}
 
-	q := `UPDATE channels SET name = :name, metadata = :metadata WHERE mainflux_channel = :mainflux_channel`
+	q := `UPDATE channels SET name = :name, metadata = :metadata, updated_at = :updated_at, updated_by = :updated_by 
+			WHERE mainflux_channel = :mainflux_channel`
 	if _, err = cr.db.NamedExec(q, dbch); err != nil {
 		return errors.Wrap(errUpdateChannels, err)
 	}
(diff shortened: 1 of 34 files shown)


things/clients/redis/events.go
```diff
@@ -17,75 +29,289 @@ var (
...
+		"actions":   sce.actions,
+		"user_ids":  sce.userIDs,
+	}
+}
+
+type viewClientEvent struct {
+	mfclients.Client
+}
+
+func (vce viewClientEvent) Encode() map[string]interface{} {
```
dborovcanin: Would simple JSON encoding work here?

things/clients/redis/events.go
```diff
@@ -17,75 +29,289 @@ var (
...
+	}
+	if len(uce.Tags) > 0 {
+		tags := fmt.Sprintf("[%s]", strings.Join(uce.Tags, ","))
+		val["tags"] = tags
+	}
+	if uce.Credentials.Identity != "" {
+		val["identity"] = uce.Credentials.Identity
+	}
+	if uce.Metadata != nil {
+		metadata, err := json.Marshal(uce.Metadata)
```
dborovcanin, line 102: We're masking the error here.

dborovcanin (changes requested):

things/clients/redis/events.go
rodneyosodo: I think we should do it that way.

rodneyosodo: I have tried out different approached but none has yielded

```go
var val map[string]interface{}
data, err := json.Marshal(cce)
if err != nil {
	return val
}
if err := json.Unmarshal(data, &val); err != nil {
	return val
}
val["operation"] = clientCreate
return val
```

I get this error ` redis: can't marshal map[string]interface {} (implement encoding.BinaryMarshaler).`, even after adding `BinaryMarshaler` and `BinaryUnmarshaler` interfaces to clients i.e

```go
// MarshalBinary implements the encoding.BinaryMarshaler interface.
func (c *Client) MarshalBinary() ([]byte, error) {
	return json.Marshal(c)
}

// UnmarshalBinary implements the encoding.BinaryUnmarshaler interface.
func (c *Client) UnmarshalBinary(data []byte) error {
	return json.Unmarshal(data, c)
}
```

things/clients/redis/events.go
rodneyosodo: Yes

dborovcanin: Please don't mask the error. :)

rodneyosodo: Should `Encode()` return `(error, map[string]interface{})` ?

bootstrap/postgres/configs.go
```diff
@@ -555,6 +557,17 @@ func nullString(s string) sql.NullString {
 	}
 }
 
+func nullTime(t time.Time) sql.NullTime {
+	if t == (time.Time{}) {
+		return sql.NullTime{}
+	}
```
dborovcanin: Use `t.IsZero()` instead.

cmd/things/main.go
```diff
@@ -157,21 +165,25 @@ func main() {
...
+func newService(db *sqlx.DB, auth upolicies.AuthServiceClient, cacheClient *redis.Client, esClient *redis.Client, tracer trace.Tracer, logger mflog.Logger) (clients.Service, groups.Service, tpolicies.Service) {
 	database := postgres.NewDatabase(db, tracer)
 	cRepo := cpostgres.NewRepository(database)
 	gRepo := gpostgres.NewRepository(database)
 	pRepo := ppostgres.NewRepository(database)
 
 	idp := uuid.New()
 
-	policyCache := redischcache.NewCache(cacheClient)
+	policyCache := redispcache.NewCache(cacheClient)
```
dborovcanin, line 176: tingCache is removed now.

things/clients/redis/events.go
dborovcanin: Yes, just in reverse order.

dborovcanin (changes requested):

cmd/things/main.go
rodneyosodo: It is still used in `things.Idenitify` for mqtt

rodneyosodo: Do we need identify since it is only used in `mqtt` to check is the `username` is the same as the `id`?
https://github.com/mainflux/mainflux/blob/72e18016901e26245dbc4f5cbe6646a315de1e74/mqtt/handler.go#LL87C1-L95C1

dborovcanin: Let's keep it for the time being.

dborovcanin (approved):

PR #1796, opened 2023-05-16
NOISSUE - Add tests for env pkg

SammyOina:

### What does this do?
Add unit tests for env package

No issue

None

Yes

No

### DIFF
--- a/internal/env/parser_test.go
+++ b/internal/env/parser_test.go
@@ -14,11 +14,124 @@ import (
 
 var errNotDuration error = errors.New("unable to parse duration")
 
-func TestParse(t *testing.T) {
+func TestParseServerConfig(t *testing.T) {
 	tests := []struct {
-		name           string
-		config         interface{}
-		expectedConfig interface{}
+		description    string
+		config         *server.Config
+		expectedConfig *server.Config
+		options        []Options
+		err            error
+	}{
+		{
+			"Parsing with Server Config",
+			&server.Config{},
+			&server.Config{
+				Host:     "localhost",
+				Port:     "8080",
+				CertFile: "cert",
+				KeyFile:  "key",
+			},
+			[]Options{
+				{
+					Environment: map[string]string{
+						"HOST":        "localhost",
+						"PORT":        "8080",
+						"SERVER_CERT": "cert",
+						"SERVER_KEY":  "key",
+					},
+				},
+			},
+			nil,
+		},
+		{
+			"Parsing with Server Config with Prefix",
+			&server.Config{},
+			&server.Config{
+				Host:     "localhost",
+				Port:     "8080",
+				CertFile: "cert",
+				KeyFile:  "key",
+			},
+			[]Options{
+				{
+					Environment: map[string]string{
+						"MF-HOST":        "localhost",
+						"MF-PORT":        "8080",
+						"MF-SERVER_CERT": "cert",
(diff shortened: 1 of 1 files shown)


dborovcanin (changes requested):
Make it consistent with other tests in the codebase (use slices for cases and run them in the loop).

drasko (dismissed):
LGTM

internal/env/parser_test.go
```diff
@@ -0,0 +1,129 @@
...
+					},
+					Prefix:    "MF-",
+					AltPrefix: "MF-GRPC-",
+				},
+			},
+			errNotDuration,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
```
dborovcanin: No need for `t.Run` call.

internal/env/parser_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/mainflux/mainflux/pkg/errors"
+
+	"github.com/mainflux/mainflux/internal/clients/grpc"
+	"github.com/mainflux/mainflux/internal/server"
+	"github.com/stretchr/testify/assert"
+)
+
+var errNotDuration error = errors.New("unable to parse duration")
+
+func TestParse(t *testing.T) {
```
dborovcanin: Split gRPC, HTTP, and other configs to separate cases lists, or even separate tests.

dborovcanin (changes requested):

internal/env/parser_test.go
```diff
@@ -0,0 +1,311 @@
...
+			nil,
+		},
+	}
+	for _, test := range tests {
+		err := Parse(test.config, test.options...)
+		if test.err != nil {
+			assert.Error(t, err, "expected error but got nil")
+		} else {
+			assert.NoError(t, err, fmt.Sprintf("expected no error but got %v", err))
+		}
```
dborovcanin:
```suggestion
assert.Equal(t, tc.err, err, fmt.Sprintf("expected error %s, but got %s", tc.err, err))
```

internal/env/parser_test.go
```diff
@@ -0,0 +1,311 @@
...
```
dborovcanin: Same here.

internal/env/parser_test.go
```diff
@@ -0,0 +1,311 @@
...
+		},
+	}
+
+	for _, test := range tests {
+		err := Parse(test.config, test.options...)
+		if test.err != nil {
+			assert.Error(t, err, "expected error but got nil")
+		} else {
+			assert.NoError(t, err, fmt.Sprintf("expected no error but got %v", err))
+		}
```
dborovcanin: And here.

dborovcanin (changes requested):

internal/env/parser_test.go
SammyOina: The errors generated are concatenated, they are also not exported from the package so I can't use errors.Contains(). I could copy the entire  error and assert equality, but this current assertion still achieves similar results

dborovcanin: Makes sense, wrapping errors would not bring benefits here. Please use `switch` instead of `if/else`, just to be aligned with the rest of the code.

dborovcanin (approved):

drasko (approved):
LGTM

PR #1798, opened 2023-05-22
MF-1783 - Improve HTTP / gRPC codes when attempting to publish with nonexistent thing key / channel id

### LINKED ISSUE #1783: Improve HTTP / gRPC codes when attempting to publish with nonexistent thing key / channel id
**BUG REPORT**

Receive informative error messages and codes when publishing messages with a thing key which is not recorded in things-db, or to a channel that is not recorded in things-db.

When publishing via the HTTP adapter:
- publishing to a non-existent channel should receive a 404 Not Found error
- publishing with a non-existent thing key should possibly receive a 422 Unprocessable content error, but at least some 4XX error

I am less familiar with the conventions for the other available adapters, but they should also be following their appropriate conventions in these scenarios.

Additionally, under the hood, the `things` service should be returning a gRPC code of 5 (Not Found) in the first case. Again, the second case is more subjective, but perhaps a code of 9 (Failed Precondition) ?

4. What are the received results?

Both scenarios produce a 500 Internal Server Error in the HTTP adapter.

`things` also returns a gRPC code of 13 (Internal) in the case of publishing with a non-existent thing key.

This pattern does not provide useful or accurate information to the client to support them in fixing their configuration error(s)

5. What are the steps to reproduce the issue?

Attempt to publish a message via the HTTP adapter with a known-to-be-incorrect channel ID or thing key (for example, by changing the last character of an existing key/id at random)

6. In what environment did you encounter the issue?

Local docker environment built off the `latest` tag

SammyOina:
### What does this do?
Improves error reports for invalid channels and things.

Resolves #1783 

Non-existent channel id and thing key returned 500 internal error which is now switched to 404 not found.

No

No

### DIFF
--- a/http/api/transport.go
+++ b/http/api/transport.go
@@ -78,7 +78,7 @@ func parseSubtopic(subtopic string) (string, error) {
 	if err != nil {
 		return "", errMalformedSubtopic
 	}
-	subtopic = strings.Replace(subtopic, "/", ".", -1)
+	subtopic = strings.ReplaceAll(subtopic, "/", ".")
 
 	elems := strings.Split(subtopic, ".")
 	filteredElems := []string{}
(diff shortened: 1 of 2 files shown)


dborovcanin (approved):

dborovcanin (approved):

drasko (approved):
LGTM

PR #1821, opened 2023-06-16
NOISSUE - Improve graceful shutdown and code quality

SammyOina:
### What does this do?
- Improve graceful stop
- Remove duplicate cases
- Remove single case switch 

None

None

No

No

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -5,7 +5,6 @@ package cli
 
 import (
 	"fmt"
-	"log"
 	"os"
 
 	"github.com/mainflux/mainflux/pkg/errors"
@@ -33,21 +32,21 @@ func read(file string) (Config, error) {
 	return c, nil
 }
 
-func ParseConfig() {
+func ParseConfig() error {
 	if ConfigPath == "" {
 		// No config file
-		return
+		return errors.New("empty config path")
 	}
 
 	if _, err := os.Stat(ConfigPath); os.IsNotExist(err) {
 		errConfigNotFound := errors.Wrap(errors.New("config file was not found"), err)
 		logError(errConfigNotFound)
-		return
+		return errConfigNotFound
 	}
 
 	config, err := read(ConfigPath)
 	if err != nil {
-		log.Fatal(err)
+		return err
 	}
 
 	if config.Offset != 0 {
@@ -65,4 +64,5 @@ func ParseConfig() {
 	if config.RawOutput {
 		RawOutput = config.RawOutput
 	}
+	return nil
 }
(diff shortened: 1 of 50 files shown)


cmd/bootstrap/main.go
```diff
@@ -73,19 +73,27 @@ func main() {
...
+	defer func() {
+		os.Exit(exitCode)
+	}()
 	defer db.Close()
 
 	// Create new redis client for bootstrap event store
 	esClient, err := redisClient.Setup(envPrefixES)
 	if err != nil {
-		logger.Fatal(fmt.Sprintf("failed to setup %s bootstrap event store redis client : %s", svcName, err))
+		logger.Error(fmt.Sprintf("failed to setup %s bootstrap event store redis client : %s", svcName, err))
```
arvindh123, line 97: By improving the logger function can achieve the same with less line 

example:

```go
logger.Fatal(fmt.Sprintf("failed to setup %s bootstrap event store redis client : %s", svcName, err), 1)
```
in `logger/logger.go`
```go
func (l logger) Fatal(msg string, ec ...int) {
	_ = l.kitLogger.Log("fatal", msg)
	if len(ec) > 1 {
		os.Exit(ec[0])
		return
	}
	os.Exit(1)
	return
}

```

SammyOina: certainly but I was looking at leaving the logger.Fatal as analogous to the standard library log.Fatal

dborovcanin: Let's keep it like @SammyOina suggested since the whole point of the PR is to avoid Fatal calls; i.e. to have a single Exit call.

arvindh123:
@dborovcanin  @SammyOina  
Same Code is repeating in all `main.go`    
So only proposed `log.Fatalf` , Since we don't to change analogous of `log.Fatalf`  
Can we have some other name with same function like `log.ErrorExit`

arvindh123:
@SammyOina Why we are removing single switch case ?

cli/config.go
```diff
@@ -33,21 +32,21 @@ func read(file string) (Config, error) {
 	return c, nil
 }
 
-func ParseConfig() {
+func ParseConfig() error {
 	if ConfigPath == "" {
 		// No config file
-		return
+		return errors.New("empty config path")
 	}
```
arvindh123: @SammyOina, CLI is not working getting error `empty config path`

SammyOina: ok, i'll return nill instead

arvindh123: @SammyOina At Two places we need to return nil 
https://github.com/SammyOina/mainflux/blob/code-quality/cli/config.go#L44
https://github.com/SammyOina/mainflux/blob/code-quality/cli/config.go#L38

arvindh123:
LGTM

drasko (approved):
LGTM

PR #1824, opened 2023-06-19
NOISSUE - Use Insert On Conflict For Policies

rodneyosodo:
### What does this do?
Uses `INSERT ... ON CONFLICT (subject, object) DO UPDATE SET actions = :actions, ...` rather than `INSERT ...`.  This is useful when adding an already existing policy as we only update the actions rather than returning `ErrExistingConflict`

No issue

Listed above. Moves AddPolicy to one repository call, `SAVE` rather than 2 possible repository calls `RETRIEVE` and `UPDATE`.

Yes

No

### Notes
Done for both things and users policies

### DIFF
--- a/pkg/sdk/go/policies_test.go
+++ b/pkg/sdk/go/policies_test.go
@@ -28,8 +28,6 @@ import (
 	"github.com/stretchr/testify/mock"
 )
 
-const addExistingPolicyDesc = "add existing policy"
-
 var utadminPolicy = umocks.SubjectSet{Subject: "things", Relation: []string{"g_add"}}
 
 func newPolicyServer(svc upolicies.Service) *httptest.Server {
@@ -75,7 +73,7 @@ func TestCreatePolicy(t *testing.T) {
 			err:   nil,
 		},
 		{
-			desc: addExistingPolicyDesc,
+			desc: "add existing policy",
 			policy: sdk.Policy{
 				Subject: subject,
 				Object:  object,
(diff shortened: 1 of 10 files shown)


pkg/sdk/go/policies_test.go
```diff
@@ -75,7 +73,7 @@ func TestCreatePolicy(t *testing.T) {
 			err:   nil,
 		},
 		{
-			desc: addExistingPolicyDesc,
+			desc: "add existing policy",
```
drasko, line 76: Why did we decide to use hard-coded string, rather then var delclaration, as previously?

things/policies/service.go
```diff
@@ -109,26 +109,14 @@ func (svc service) AddPolicy(ctx context.Context, token string, p Policy) (Polic
...
-		return svc.policies.Update(ctx, p)
-	}
 
 	p.OwnerID = userID
 	p.CreatedAt = time.Now()
 
+	// incase the policy exists, use these for update.
+	p.UpdatedAt = time.Now()
+	p.UpdatedBy = userID
+
```
drasko, line 119: Do we need to do something with cache here?

drasko (changes requested):

pkg/sdk/go/policies_test.go
rodneyosodo: `gocost` suggest declaring a variable as const when you use it more than 5 times. Initially, we were using it more than 5 times but currently, we use it 4 times.

things/policies/service.go
rodneyosodo: I don't think so. I added it because the full query for an update operation if any is to also update the `updated_at` and `updated_by` field. It we don't have them it will be empty yet we have done an update operation.

pkg/sdk/go/policies_test.go
dborovcanin: While it might sound counter-intuitive, I prefer literal over constant for readability in this context.

things/policies/service.go
dborovcanin: In fact, @drasko is right. Since this is effectively a policy update, **we do need to invalidate the cache before to prevent incorrect actions**.

dborovcanin (changes requested):

dborovcanin (approved):

drasko (approved):
LGTM

PR #1825, opened 2023-06-19
NOISSUE - Change Share Client To Be AddPolicy

rodneyosodo:
### What does this do?
Removes the `ShareClient` service function as it is basically the `AddPolicy` function on the policies sub-service.
It is addPolicy as it only requires `channelID`, `userID` and `actions`. Things to be shared are supposed to be already added to the `channelID` or will be added later on.

No issue

Removes service layer `ShareClient` function on the things services.

Yes

No

### Notes
Should we remove also the API endpoint?

### DIFF
--- a/pkg/sdk/go/sdk.go
+++ b/pkg/sdk/go/sdk.go
@@ -374,7 +374,7 @@ type SDK interface {
 	// already created a group and added things to it. It also assumes that
 	// you have required policy to share a thing with the specified user.
 	//
-	// Under the hoold `ShareThing` calls the `Connect` method with the
+	// The `ShareThing` method calls the `Connect` method with the
 	// subject as `userID` rather than `thingID`.
 	//
 	// example:
--- a/pkg/sdk/go/sdk.go
+++ b/pkg/sdk/go/sdk.go
@@ -374,6 +374,9 @@ type SDK interface {
 	// already created a group and added things to it. It also assumes that
 	// you have required policy to share a thing with the specified user.
 	//
+	// Under the hoold `ShareThing` calls the `Connect` method with the
+	// subject as `userID` rather than `thingID`.
+	//
 	// example:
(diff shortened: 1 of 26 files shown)


things/clients/api/transport.go
```diff
@@ -18,95 +18,96 @@ import (
...
-		otelkit.EndpointMiddleware(otelkit.WithOperation("update_thing_tags"))(updateClientTagsEndpoint(svc)),
+		otelkit.EndpointMiddleware(otelkit.WithOperation("update_thing_tags"))(updateClientTagsEndpoint(csvc)),
 		decodeUpdateClientTags,
 		api.EncodeResponse,
 		opts...,
 	))
 
-	mux.Post("/things/:thingID/share", kithttp.NewServer(
-		otelkit.EndpointMiddleware(otelkit.WithOperation("share_thing"))(shareClientEndpoint(svc)),
+	mux.Post("/things/share", kithttp.NewServer(
```
dborovcanin: No need to expose this API at all; since it's already exposed at Policies. UI/CLI will take care of using the correct endpoint.

dborovcanin (changes requested):

pkg/sdk/go/sdk.go
```diff
@@ -370,12 +370,17 @@ type SDK interface {
 	//  fmt.Println(id)
 	IdentifyThing(key string) (string, errors.SDKError)
 
-	// ShareThing shares thing with other user.
+	// ShareThing shares thing with other user. It assumes that you have
+	// already created a group and added things to it. It also assumes that
+	// you have required policy to share a thing with the specified user.
+	//
+	// Under the hoold `ShareThing` calls the `Connect` method with the
```
dborovcanin: Typo `hood`.  But no need for that at all: `It calls Connect method with...` is sufficient.

dborovcanin (changes requested):

dborovcanin (approved):

drasko (approved):
LGTM

PR #1836, opened 2023-06-28
MF-703 - Reliably Publish Event Messages to Redis

### LINKED ISSUE #703: Enable reliable publishing of event messages to Redis Streams
**FEATURE REQUEST**

Currently there are mqtt adapter and bootstrap service publishing event messages to Redis Streams. Some of these events are status changes so it's very important to have them successfully published.
If the Redis Streams servers can not accept messages, the service/adapter should have a way to store those unpublished events and retry.

Since the events notifying about status changes are important for having an accurate overview of the system, I think that this feature is a must have one.

rodneyosodo:
### What does this do?
I have introduced a mechanism for storing unpublished events and implementing retries. We use memory to save unpublished events when the Redis server is unavailable. Then, we periodically attempt to publish the stored events when the Redis server becomes available again.

1. When an event needs to be published, check if the Redis server is reachable. Added a `100ms` timeout so stop retrying to connect to Redis when it is down
2. If the Redis server is reachable, publish the event.
3. If the Redis server is unreachable, store the event in memory using slice.
4. Periodically check the connection to the Redis server, after every minute, and retry publishing the stored events when the connection is re-established.
5. Once an event is successfully published, remove it from memory.

Resolves #703 

added `100ms` timeout to check if Redis is reachable

No

No

### Notes
Tested by pausing the `es-redis` container and resuming it after doing some operations using the `e2e` tool.

### DIFF
--- a/internal/clients/redis/producer.go
+++ b/internal/clients/redis/producer.go
@@ -67,7 +67,13 @@ func (es *eventStore) Publish(ctx context.Context, event Event) error {
 		es.mu.Lock()
 		defer es.mu.Unlock()
 
-		es.unpublishedEvents <- record
+		select {
+		case es.unpublishedEvents <- record:
+		default:
+			// If the channel is full (rarely happens), drop the events.
+			return nil
+		}
+
 		return nil
 	}
 
@@ -75,6 +81,8 @@ func (es *eventStore) Publish(ctx context.Context, event Event) error {
 }
 
 func (es *eventStore) StartPublishingRoutine(ctx context.Context) {
+	defer close(es.unpublishedEvents)
+
 	ticker := time.NewTicker(unpublishedEventsCheckInterval)
 	defer ticker.Stop()
 
--- a/internal/clients/redis/producer.go
+++ b/internal/clients/redis/producer.go
@@ -55,6 +55,8 @@ func (es *eventStore) Publish(ctx context.Context, event Event) error {
 	if err != nil {
 		return err
 	}
+	values["occurred_at"] = time.Now().UnixNano()
+
 	record := &redis.XAddArgs{
 		Stream:       es.streamID,
 		MaxLenApprox: es.streamLen,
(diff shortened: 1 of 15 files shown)


internal/clients/redis/producer.go
```diff
@@ -0,0 +1,98 @@
...
+	Publish(ctx context.Context, event Event) error
+
+	// StartPublishingRoutine starts routine that checks for unpublished events
+	// and publishes them to redis stream.
+	StartPublishingRoutine(ctx context.Context)
+}
+
+type eventStore struct {
+	client            *redis.Client
+	unpublishedEvents []*redis.XAddArgs
```
SammyOina: this can grow unexpectedly large sizes. I think a different approach is required for memory management. Perhaps a channel or some external queue like nats might serve better

SammyOina (changes requested):

dborovcanin (changes requested):
I'm afraid that reliable publishing needs a little bit more than what we have. The buffer mechanism is not bad in itself, but what it effectively introduce is - the reordering of events. This topic is very advanced and out of the scope of this PR. What we can do is - provide a mechanism to mark these events "dirty" or "out of order" maybe by adding a field such as `event_time` that will represent the time event occurred and will be different from the time the event is published on Redis.

internal/clients/redis/producer.go
rodneyosodo: Currently, nats is configured as fire and forget unless we move to Jetstream. I have implemented a channel with a max of 1M capacity.

rodneyosodo: Before publishing the memory footprint of the container was 17MB after publishing before resuming Redis it was 49MB

dborovcanin (changes requested):
Please fix tests and test locally.

SammyOina (approved):

drasko:
@dborovcanin please review this one

SammyOina (approved):
LGTM

drasko (approved):
LGTM

PR #1860, opened 2023-07-18
NOISSUE - Check For Policy Action Before Adding Policies

rodneyosodo:
### What does this do?
Check for policy actions before adding policy. If the current users has `g_add` and `c_list` policies on a group, they can only add members to that group with either `g_add` or `c_list` actions. If they try to add a member with `c_update` it should fail unless they are the owner of that group.

No issue

When adding members to a group, add members with equal policy like yours or subset of your policies.

Yes

No

### Notes
How to reproduce this while testing is
1. Create 3 different users `a`, `b` and `c`.
2. Create a group with user `a`
3. Add user `b` to the group with only `g_add` actions while using the token of user `a`
4. Try to add user `c` to the group with `c_update` actions while using the token of user `b`


things/policies/service.go
```diff
@@ -127,8 +127,21 @@ func (svc service) AddPolicy(ctx context.Context, token string, p Policy) (Polic
 	}
 
 	// If the client has `g_add` action on the object or is the owner of the object, add the policy
-	ar := AccessRequest{Subject: userID, Object: p.Object, Action: "g_add"}
-	if _, err := svc.policies.EvaluateGroupAccess(ctx, ar); err == nil {
+	areq := AccessRequest{Subject: userID, Object: p.Object, Action: "g_add"}
+	if _, err := svc.policies.EvaluateGroupAccess(ctx, areq); err == nil {
+		page := Page{Subject: userID, Object: p.Object, Offset: 0, Limit: 1}
+		pol, err := svc.policies.Retrieve(ctx, page)
```
dborovcanin: Is this fetch necessary and why don't we use the policy from `EvaluateGroupAccess`?

dborovcanin (changes requested):

things/policies/service.go
rodneyosodo: Because it was returning only `subject`. I have changed it to return `SOA`

things/policies/policies.go
```diff
@@ -135,3 +136,22 @@ func ValidateAction(act string) bool {
...
+func checkActions(currentActions, incomingActions []string) error {
+	for _, action := range incomingActions {
+		if !contains(currentActions, action) {
+			return errors.ErrAuthorization
+		}
+	}
+	return nil
+}
+
+func contains(actions []string, action string) bool {
```
SammyOina: you can use golang.org/x/exp/slices with the slices.Contains() method

users/policies/policies.go
```diff
@@ -208,3 +209,22 @@ func removeDuplicates(slice []string) []string {
...
+func checkActions(currentActions, incomingActions []string) error {
+	for _, action := range incomingActions {
+		if !contains(currentActions, action) {
+			return errors.ErrAuthorization
+		}
+	}
+	return nil
+}
+
+func contains(actions []string, action string) bool {
```
SammyOina: same here

SammyOina (changes requested):

SammyOina (approved):

drasko (approved):
LGTM

PR #1867, opened 2023-07-21
NOISSUE - Set subtopic for messages published outside adapters 

SammyOina:
Signed-off-by: SammyOina <sammyoina@gmail.com>
### What does this do?
Sets message subtopic to subject/routing key when not set as in the case for messages published using external clients as in heartbeat service of https://github.com/mainflux/agent/pull/56.

To test on nats:
```
git clone https://github.com/nats-io/nats.go
cd github.com/nats-io/nats.go/examples/nats-pub
go run main.go -s http://localhost:4222 heartbeat.<service-name>.<service-type> ""; 
```
Using rabbitmq:
```
package main

import (
	"context"
	"log"
	"time"

	amqp "github.com/rabbitmq/amqp091-go"
)

func failOnError(err error, msg string) {
	if err != nil {
		log.Panicf("%s: %s", msg, err)
	}
}

func main() {
	conn, err := amqp.Dial("amqp://mainflux:mainflux@localhost:5672/")
	failOnError(err, "Failed to connect to RabbitMQ")
	defer conn.Close()
	ch, err := conn.Channel()
	failOnError(err, "Failed to open a channel")
	defer ch.Close()

	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()

	body := ""
	err = ch.PublishWithContext(ctx,
		"mainflux-exchange",  // exchange
		"heartbeat.svc.name", // routing key
		false,                // mandatory
		false,                // immediate
		amqp.Publishing{
			ContentType: "text/plain",
			Body:        []byte(body),
		})
	failOnError(err, "Failed to publish a message")
	log.Printf(" [x] Sent %s\n", body)
}
```

None

No


drasko:
I think it is safe to close this one, as `Subject` can be derived from broker's metadata in the subscription handler.

pkg/messaging/nats/pubsub.go
```diff
@@ -162,6 +162,9 @@ func (ps *pubsub) natsHandler(h messaging.MessageHandler) broker.MsgHandler {
 			ps.logger.Warn(fmt.Sprintf("Failed to unmarshal received message: %s", err))
 			return
 		}
+		if msg.Subtopic == "" {
+			msg.Subtopic = m.Subject // set for messages not published using adapters.
```
drasko: Why is this needed? This is confusing.

If there is no subtopic, it should stay an empty string.

drasko (changes requested):

SammyOina:
I am going to update the example in the docs instead which should have the same result

PR #1872, opened 2023-07-28
NOISSUE-Add resetPassword to the SDK

ianmuchyri:
### What does this do?
Add `ResetPasswordRequest` and `ResetPassword` methods to SDK functions
No Issue
None
No
No


users/clients/api/transport.go
```diff
@@ -301,7 +301,6 @@ func decodePasswordResetRequest(_ context.Context, r *http.Request) (interface{}
 		return nil, errors.Wrap(errors.ErrMalformedEntity, err)
 	}
 
-	req.Host = r.Header.Get("Referer")
```
SammyOina: why is this removed

ianmuchyri: It kept on resulting in the `req.host` being blank. I pass the `host` in the body.

docker/.env
```diff
@@ -58,17 +58,17 @@ MF_USERS_REFRESH_TOKEN_DURATION=24h
 MF_USERS_ADMIN_EMAIL=admin@example.com
 MF_USERS_ADMIN_PASSWORD=12345678
 MF_USERS_RESET_PWD_TEMPLATE=users.tmpl
-MF_USERS_PASS_REGEX=^.{8,}$$
+MF_USERS_PASS_REGEX=^.{8,}$
 MF_USERS_INSTANCE_ID=
 
 ### Email utility
-MF_EMAIL_HOST=smtp.mailtrap.io
```
rodneyosodo: This can be left unchanged

pkg/sdk/go/sdk.go
```diff
@@ -200,6 +200,20 @@ type SDK interface {
 	//  fmt.Println(user)
 	UpdateUserOwner(user User, token string) (User, errors.SDKError)
 
+	//ResetPasswordRequest sends a password request email to a user.
```
rodneyosodo:
```suggestion
	// ResetPasswordRequest sends a password request email to a user.
```

pkg/sdk/go/sdk.go
```diff
@@ -200,6 +200,20 @@ type SDK interface {
...
+	//
+	// example:
+	//err:= sdk.ResetPasswordRequest("example@email.com")
+	//fmt.Println(err)
+	ResetPasswordRequest(email string) errors.SDKError
+
+	//ResetPassword changes a user's password to the one passed in the argument.
```
rodneyosodo:
```suggestion
	// ResetPassword changes a user's password to the one passed in the argument.
```

rodneyosodo (changes requested):

docker/.env
```diff
@@ -58,17 +58,17 @@ MF_USERS_REFRESH_TOKEN_DURATION=24h
 MF_USERS_ADMIN_EMAIL=admin@example.com
 MF_USERS_ADMIN_PASSWORD=12345678
 MF_USERS_RESET_PWD_TEMPLATE=users.tmpl
-MF_USERS_PASS_REGEX=^.{8,}$$
+MF_USERS_PASS_REGEX=^.{8,}$
 MF_USERS_INSTANCE_ID=
 
 ### Email utility
 MF_EMAIL_HOST=smtp.mailtrap.io
-MF_EMAIL_PORT=2525
```
rodneyosodo: same to this. change `MF_EMAIL_TEMPLATE`

users/clients/api/transport.go
rodneyosodo: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer

pkg/sdk/go/sdk.go
```diff
@@ -200,6 +200,20 @@ type SDK interface {
...
+	//
+	// example:
+	//err:= sdk.ResetPasswordRequest("example@email.com")
+	//fmt.Println(err)
+	ResetPasswordRequest(email string) errors.SDKError
+
+	// ResetPassword changes a user's password to the one passed in the argument.
+	//
+	// example:
+	//err:=sdk.ResetPassword("token","password","password")
```
rodneyosodo:
```suggestion
	//  err:=sdk.ResetPassword("token","password","password")
```
Check on the formatting of other go docs you have provided.

rodneyosodo (changes requested):

users/clients/api/transport.go
ianmuchyri: I know what the referer is. In this case I kept on getting the error `missing host` in Mainflux-Users service. That's why I removed it from here.

docker/.env
```diff
@@ -58,17 +58,17 @@ MF_USERS_REFRESH_TOKEN_DURATION=24h
...
 MF_USERS_ADMIN_PASSWORD=12345678
 MF_USERS_RESET_PWD_TEMPLATE=users.tmpl
-MF_USERS_PASS_REGEX=^.{8,}$$
+MF_USERS_PASS_REGEX=^.{8,}$
 MF_USERS_INSTANCE_ID=
 
 ### Email utility
 MF_EMAIL_HOST=smtp.mailtrap.io
 MF_EMAIL_PORT=2525
-MF_EMAIL_USERNAME=18bf7f70705139
```
rodneyosodo: same

docker/.env
```diff
@@ -58,17 +58,17 @@ MF_USERS_REFRESH_TOKEN_DURATION=24h
...
-MF_EMAIL_PASSWORD=2b0d302e775b1e
```
rodneyosodo, line 80: same

docker/.env
```diff
@@ -58,17 +58,17 @@ MF_USERS_REFRESH_TOKEN_DURATION=24h
...
-MF_EMAIL_FROM_ADDRESS=from@example.com
```
rodneyosodo, line 81: same

docker/.env
```diff
@@ -58,17 +58,17 @@ MF_USERS_REFRESH_TOKEN_DURATION=24h
...
-MF_EMAIL_FROM_NAME=Example
```
rodneyosodo, line 82: same

docker/.env
```diff
@@ -58,17 +58,17 @@ MF_USERS_REFRESH_TOKEN_DURATION=24h
...
-MF_EMAIL_TEMPLATE=smtp-notifier.tmpl
+MF_EMAIL_USERNAME=
+MF_EMAIL_PASSWORD=
+MF_EMAIL_FROM_ADDRESS=Mainflux@example.com
+MF_EMAIL_FROM_NAME=Mainflux
+MF_EMAIL_TEMPLATE=email.tmpl
 MF_EMAIL_INSTANCE_ID=
```
rodneyosodo: Also remove this

rodneyosodo (changes requested):

users/clients/api/transport.go
rodneyosodo: On the request you can pass it as a header
```go
req.Header.Add("Referer", "<referer url>")
```

ianmuchyri: This would take in the request url of the sdk `processRequest ` method, not the request of my UI, hence it would not work. I pass the host as an sdk configuration. Now not unless we add it here in the header https://github.com/mainflux/mainflux/blob/57d47fea66193b762f6ac68db757d5fc26d7bc82/pkg/sdk/go/sdk.go#L1005

rodneyosodo: You can change process Request to accept headers

ianmuchyri: This would change `processRequest` just for one function. I only need Referer for this function only. And I can pass it in the body together with the email and decode it here: https://github.com/mainflux/mainflux/blob/57d47fea66193b762f6ac68db757d5fc26d7bc82/users/clients/api/transport.go#L300, that way it will only be in that method, as I don't see the need of adding a `referer` header in the processRequest as other methods don't need it. Is there any particular reason you really need this line: https://github.com/mainflux/mainflux/blob/57d47fea66193b762f6ac68db757d5fc26d7bc82/users/clients/api/transport.go#L304

SammyOina: I think it's ok to extend processRequest even if it is only used in one method. Other methods can benefit in the future

pkg/sdk/go/sdk.go
```diff
@@ -1007,6 +1024,8 @@ func (sdk mfSDK) processRequest(method, url, token, contentType string, data []b
 		return make(http.Header), []byte{}, errors.NewSDKError(err)
 	}
 
+	req.Header.Add("Referer", sdk.HostURL)
```
drasko: Do not do it like this. Rather change `processRequest()` function in that way that we can pass any arbitrary header as a parameter.

It is not needed that we always add header, especially `Referer` header. This function should only add header when it is passed via parametr, and it should be key-value map, so that I can set any number of arbitrary headers.

drasko (changes requested):

pkg/sdk/go/bootstrap.go
```diff
@@ -97,7 +97,10 @@ func (sdk mfSDK) AddBootstrap(cfg BootstrapConfig, token string) (string, errors
 
 	url := fmt.Sprintf("%s/%s", sdk.bootstrapURL, configsEndpoint)
 
-	headers, _, sdkerr := sdk.processRequest(http.MethodPost, url, token, string(CTJSON), data, http.StatusOK, http.StatusCreated)
+	header := make(map[string]string)
+	header["Content-Type"] = string(CTJSON)
```
rodneyosodo: Since this repeats for most part you can leave it at `processRequest`

https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

rodneyosodo (changes requested):

pkg/sdk/go/bootstrap.go
drasko: Yes, good idea - we can have a default CT, which is in the `processRequest()` function, and it is set if I do not provide any CT header. However - if I provide CT header, then this one is taken into account (not the default one).

ianmuchyri: Okay, I'll add `req.Header.Add("Content-Type", string(CTJSON))` to `processRequest()` before the headers `for` loop. In case someone provides their own CT in the headers, it will automatically be appended then.

pkg/sdk/go/users.go
```diff
@@ -188,6 +189,37 @@ func (sdk mfSDK) UpdateUserIdentity(user User, token string) (User, errors.SDKEr
...
+func (sdk mfSDK) ResetPasswordRequest(email string) errors.SDKError {
+	var rpr = resetPasswordRequestreq{Email: email}
+
+	data, err := json.Marshal(rpr)
+	if err != nil {
+		return errors.NewSDKError(err)
+	}
+	url := fmt.Sprintf("%s/%s/reset-request", sdk.usersURL, PasswordResetEndpoint)
+
+	header := make(map[string]string)
```
rodneyosodo:
```suggestion
	var header = make(map[string]string)
```
It is more go like

pkg/sdk/go/sdk.go
```diff
@@ -1001,21 +1018,24 @@ func NewSDK(conf Config) SDK {
...
 // processRequest creates and send a new HTTP request, and checks for errors in the HTTP response.
 // It then returns the response headers, the response body, and the associated error(s) (if any).
-func (sdk mfSDK) processRequest(method, url, token, contentType string, data []byte, expectedRespCodes ...int) (http.Header, []byte, errors.SDKError) {
+func (sdk mfSDK) processRequest(method, url, token string, data []byte, requestHeaders map[string]string, expectedRespCodes ...int) (http.Header, []byte, errors.SDKError) {
 	req, err := http.NewRequest(method, url, bytes.NewReader(data))
 	if err != nil {
 		return make(http.Header), []byte{}, errors.NewSDKError(err)
 	}
 
+	req.Header.Add("Content-Type", string(CTJSON))
```
drasko, line 1029: Add only if `Content-Type` header is not set in the parameter. Also, add a comment that this is a default value.

drasko (changes requested):

pkg/sdk/go/sdk.go
```diff
@@ -1001,21 +1018,24 @@ func NewSDK(conf Config) SDK {
 
 // processRequest creates and send a new HTTP request, and checks for errors in the HTTP response.
 // It then returns the response headers, the response body, and the associated error(s) (if any).
-func (sdk mfSDK) processRequest(method, url, token, contentType string, data []byte, expectedRespCodes ...int) (http.Header, []byte, errors.SDKError) {
+func (sdk mfSDK) processRequest(method, url, token string, data []byte, requestHeaders map[string]string, expectedRespCodes ...int) (http.Header, []byte, errors.SDKError) {
```
rodneyosodo:
```suggestion
func (sdk mfSDK) processRequest(method, url, token, contentType string, data []byte, headers map[string]string, expectedRespCodes ...int) (http.Header, []byte, errors.SDKError) {
```
I thought this would be the new function signature

rodneyosodo (changes requested):

pkg/sdk/go/sdk.go
ianmuchyri: In this case, it is overwritten in the next lines if `Content-Type ` is set in the header parameter. This is because `request.Header.Add()` appends any existing value associated with the key.

pkg/sdk/go/sdk.go
ianmuchyri: There is no need to pass `contentType` as a separate argument since its also a header

pkg/sdk/go/sdk.go
```diff
@@ -1001,21 +1018,26 @@ func NewSDK(conf Config) SDK {
...
```
rodneyosodo:
```suggestion
func (sdk mfSDK) processRequest(method, url, token string, data []byte, headers map[string]string, expectedRespCodes ...int) (http.Header, []byte, errors.SDKError) {
```

rodneyosodo (changes requested):
Also add CLI functionality

pkg/sdk/go/users.go
```diff
@@ -188,6 +189,37 @@ func (sdk mfSDK) UpdateUserIdentity(user User, token string) (User, errors.SDKEr
 	return user, nil
 }
 
+func (sdk mfSDK) ResetPasswordRequest(email string) errors.SDKError {
```
mteodor, line 192: what is the flow to reset a password, what is the sequence of commands to reset password with cli?

docker/docker-compose.yml
```diff
@@ -184,8 +184,8 @@ services:
     networks:
       - mainflux-base-net
     volumes:
-      - ./templates/${MF_USERS_RESET_PWD_TEMPLATE}:/email.tmpl     
-  
+      - ./templates/${MF_EMAIL_TEMPLATE}:/email.tmpl
```
rodneyosodo: This should be left unchanged

rodneyosodo (changes requested):

pkg/sdk/go/users.go
ianmuchyri: set in the current commit

cli/users.go
```diff
@@ -232,6 +232,47 @@ var cmdUsers = []cobra.Command{
 			logJSON(user)
 		},
 	},
+	{
+		Use:   "resetpasswordrequest <email> <host>",
+		Short: "Reset password request",
+		Long: "Reset password request\n" +
+			"Usage:\n" +
+			"\tmainflux-cli users resetpasswordrequest email host\n",
```
rodneyosodo:
```suggestion
			"\tmainflux-cli users resetpasswordrequest example@mail.com http://localhost\n",
```

cli/users.go
```diff
@@ -232,6 +232,47 @@ var cmdUsers = []cobra.Command{
...
+
+			logOK()
+		},
+	},
+	{
+		Use:   "resetpassword <password> <confpass> <user_auth_token>",
+		Short: "Reset password",
+		Long: "Reset password\n" +
+			"Usage:\n" +
+			"\tmainflux-cli users resetpassword password confpass $USERTOKEN\n",
```
rodneyosodo:
```suggestion
			"\tmainflux-cli users resetpassword 12345678 12345678 $REQUESTTOKEN\n",
```

rodneyosodo (changes requested):

cli/users.go
```diff
@@ -232,6 +232,47 @@ var cmdUsers = []cobra.Command{
 			logJSON(user)
 		},
 	},
+	{
+		Use:   "resetpasswordrequest <email> <host>",
```
rodneyosodo: Remove host

cli/users.go
```diff
@@ -232,6 +232,47 @@ var cmdUsers = []cobra.Command{
...
+			if err != nil {
+				logError(err)
+				return
+			}
+
+			logOK()
+		},
+	},
+	{
+		Use:   "resetpassword <password> <confpass> <user_auth_token>",
```
rodneyosodo:
```suggestion
		Use:   "resetpassword <password> <confpass> <password_request_token>",
```

cmd/cli/main.go
```diff
@@ -19,6 +19,7 @@ const (
 	defThingsURL    string = defURL + ":9000"
 	defBootstrapURL string = defURL + ":9013"
 	defCertsURL     string = defURL + ":9019"
+	defHostURL      string = "http://localhost"
```
rodneyosodo: Isn't this the same as `defURL`?

pkg/sdk/go/sdk.go
```diff
@@ -1001,21 +1018,26 @@ func NewSDK(conf Config) SDK {
...
 // It then returns the response headers, the response body, and the associated error(s) (if any).
-func (sdk mfSDK) processRequest(method, url, token, contentType string, data []byte, expectedRespCodes ...int) (http.Header, []byte, errors.SDKError) {
+func (sdk mfSDK) processRequest(method, url, token string, data []byte, headers map[string]string, expectedRespCodes ...int) (http.Header, []byte, errors.SDKError) {
 	req, err := http.NewRequest(method, url, bytes.NewReader(data))
 	if err != nil {
 		return make(http.Header), []byte{}, errors.NewSDKError(err)
 	}
 
+	// Sets a default value for the Content-Type.
+	// Overridden if Content-Type is passes in the headers arguments.
```
drasko: Typo: `passed`

drasko (changes requested):

rodneyosodo (changes requested):

cli/users.go
```diff
@@ -232,6 +232,47 @@ var cmdUsers = []cobra.Command{
...
+			"Usage:\n" +
+			"\tmainflux-cli users resetpasswordrequest example@mail.com\n",
+		Run: func(cmd *cobra.Command, args []string) {
+			if len(args) != 1 {
+				logUsage(cmd.Use)
+				return
+			}
+
+			err := sdk.ResetPasswordRequest(args[0])
+			if err != nil {
```
rodneyosodo: inline

rodneyosodo (changes requested):

cli/users.go
```diff
@@ -232,6 +232,47 @@ var cmdUsers = []cobra.Command{
...
+		Short: "Reset password",
+		Long: "Reset password\n" +
+			"Usage:\n" +
+			"\tmainflux-cli users resetpassword 12345678 12345678 $REQUESTTOKEN\n",
+		Run: func(cmd *cobra.Command, args []string) {
+			if len(args) != 3 {
+				logUsage(cmd.Use)
+				return
+			}
+			err := sdk.ResetPassword(args[0], args[1], args[2])
```
rodneyosodo: inline

pkg/sdk/go/sdk.go
```diff
@@ -200,6 +200,20 @@ type SDK interface {
...
+	//
+	// example:
+	//  err:= sdk.ResetPasswordRequest("example@email.com")
+	//  fmt.Println(err)
+	ResetPasswordRequest(email string) errors.SDKError
+
+	// ResetPassword changes a user's password to the one passed in the argument.
+	//
+	// example:
+	//  err:=sdk.ResetPassword("password","password","token")
```
rodneyosodo:
```suggestion
	//  err := sdk.ResetPassword("password","password","token")
```

pkg/sdk/go/sdk.go
```diff
@@ -200,6 +200,20 @@ type SDK interface {
 	//  fmt.Println(user)
 	UpdateUserOwner(user User, token string) (User, errors.SDKError)
 
+	// ResetPasswordRequest sends a password request email to a user.
+	//
+	// example:
+	//  err:= sdk.ResetPasswordRequest("example@email.com")
```
rodneyosodo:
```suggestion
	//  err := sdk.ResetPasswordRequest("example@email.com")
```

rodneyosodo (changes requested):

pkg/sdk/go/sdk.go
```diff
@@ -200,6 +200,20 @@ type SDK interface {
...
+	//
+	// example:
+	//  err := sdk.ResetPasswordRequest("example@email.com")
+	//  fmt.Println(err)
+	ResetPasswordRequest(email string) errors.SDKError
+
+	// ResetPassword changes a user's password to the one passed in the argument.
+	//
+	// example:
+	//  err :=sdk.ResetPassword("password","password","token")
```
rodneyosodo:
```suggestion
	//  err := sdk.ResetPassword("password","password","token")
```

rodneyosodo (changes requested):

cli/users.go
```diff
@@ -232,6 +232,46 @@ var cmdUsers = []cobra.Command{
 			logJSON(user)
 		},
 	},
+	{
+		Use:   "resetpasswordrequest <email>",
+		Short: "Reset password request",
```
rodneyosodo:
```suggestion
		Short: "Send reset password request",
```
Same as Long description

rodneyosodo (changes requested):

rodneyosodo (approved):
LGTM

drasko (approved):
LGTM

Issue #1890, opened 2023-08-16
Errors.Contains() not working as expected


ianmuchyri:
**BUG REPORT**

1. What were you trying to achieve?
 Determine if a certain statement(error) is contained in another error.
2. What are the expected results?
For example: If a user tries to log in with an expired token:
```
errAuthentication : =  errors.New("failed to perform authentication over the entity")
bool := errors.Contains(err, errAuthentication)
```
I expected the bool to be `true`
4. What are the received results?
bool is `false`
5. What are the steps to reproduce the issue?
Use Errors.Contains(), to check if an error is wrapped within another error in Mainflux. It results in false in all cases.
7. In what environment did you encounter the issue?
Mainflux core
8. Additional information you deem important:

ianmuchyri:
closed with this pr #1891

PR #1891, opened 2023-08-18
1890 - Update error encoding

ianmuchyri:
### What does this do?
Updates how errors are encoded in the HTTP response.
#1890 
None
No
No


bootstrap/api/transport.go
```diff
@@ -302,14 +301,17 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 	}
 
 	if errorVal, ok := err.(errors.Error); ok {
-		w.Header().Set("Content-Type", contentType)
```
SammyOina, line 305: why is this removed

pkg/errors/sdk_errors.go
```diff
@@ -81,9 +102,18 @@ func CheckError(resp *http.Response, expectedStatusCodes ...int) SDKError {
...
+	if msg, ok := content[messageKey]; ok {
+		if msgError, ok := content[errorKey]; ok {
+			if m, ok := msg.(string); ok {
+				if e, ok := msgError.(string); ok {
+					if e != "" {
+						return NewSDKErrorWithStatus(Wrap(New(m), New(e)), resp.StatusCode)
+					}
+					return NewSDKErrorWithStatus(New(m), resp.StatusCode)
+				}
+				return NewSDKErrorWithStatus(fmt.Errorf("invalid err type: %v", msgError), resp.StatusCode)
```
SammyOina: extract to var

SammyOina (changes requested):

pkg/errors/sdk_errors.go
```diff
@@ -81,9 +102,18 @@ func CheckError(resp *http.Response, expectedStatusCodes ...int) SDKError {
 	var content map[string]interface{}
 	_ = json.Unmarshal(body, &content)
 
-	if msg, ok := content[errorKey]; ok {
-		if v, ok := msg.(string); ok {
-			return NewSDKErrorWithStatus(New(v), resp.StatusCode)
+	if msg, ok := content[messageKey]; ok {
```
dborovcanin: Since we are expecting a known format of the error, we should simplify this by unmarshalling the body to the expected struct  (rather than `map[string]interface{}`) and handling unmarshal errors, rather than using this unnecessarily complex approach.

dborovcanin (changes requested):

SammyOina (approved):

bootstrap/api/transport.go
```diff
@@ -303,13 +302,17 @@ func encodeError(_ context.Context, err error, w http.ResponseWriter) {
 
 	if errorVal, ok := err.(errors.Error); ok {
 		w.Header().Set("Content-Type", contentType)
-
-		errMsg := errorVal.Msg()
+		errorRes := apiutil.ErrorRes{
+			Err: "",
+			Msg: errorVal.Msg(),
+		}
```
dborovcanin: Can we simplify error inspection by implementing JSON Marshaller in our custom error (i.e. always do just `json.NewEncoder(w).Encode(err)` with no other inspections)?

dborovcanin (changes requested):

pkg/sdk/go/message_test.go
```diff
@@ -88,15 +90,15 @@ func TestSendMessage(t *testing.T) {
...
+			err:    errors.NewSDKErrorWithStatus(unexpectedJSONEnd, http.StatusUnauthorized),
 		},
 	}
 	for desc, tc := range cases {
 		err := mfsdk.SendMessage(tc.chanID, tc.msg, tc.auth)
 		if tc.err == nil {
 			assert.Nil(t, err, fmt.Sprintf("%s: got unexpected error: %s", desc, err))
 		} else {
-			assert.Equal(t, tc.err.Error(), err.Error(), fmt.Sprintf("%s: expected error %s, got %s", desc, err, tc.err))
+			assert.Equal(t, tc.err.Error(), err.Error(), fmt.Sprintf("%s: expected error %s, got %s", desc, tc.err, err))
```
dborovcanin, line 102: Use `switch-default` instead of `if-else`.

pkg/sdk/go/tokens_test.go
```diff
@@ -84,7 +84,7 @@ func TestIssueToken(t *testing.T) {
 			ok := repoCall.Parent.AssertCalled(t, "RetrieveByIdentity", mock.Anything, mock.Anything)
 			assert.True(t, ok, fmt.Sprintf("RetrieveByIdentity was not called on %s", tc.desc))
 		default:
-			assert.True(t, strings.Contains(err.Msg(), tc.err.Msg()), fmt.Sprintf("%s: expected error %s, got %s", tc.desc, tc.err, err))
+			assert.True(t, strings.Contains(err.Error(), tc.err.Msg()), fmt.Sprintf("%s: expected error %s, got %s", tc.desc, tc.err, err))
```
dborovcanin: You can use `assert.Contains` here to compare strings. However, we need to improve this check, since it does not include checking the status code. Probably the simplest way to do it would be to make it work with `errors.Contains`.

dborovcanin (changes requested):

pkg/errors/errors.go
```diff
@@ -49,6 +53,20 @@ func (ce *customError) Err() Error {
 	return ce.err
 }
 
+func (ce *customError) MarshalJSON() ([]byte, error) {
```
SammyOina, line 56: add comment on exported function

SammyOina (changes requested):

pkg/errors/errors.go
ianmuchyri: It's there: https://github.com/ianmuchyri/mainflux/blob/83b5da4b9e325d568ea6edae77fa20fd0d0194fd/pkg/errors/errors.go#L26

pkg/errors/errors.go
```diff
@@ -21,6 +22,9 @@ type Error interface {
 
 	// Err returns wrapped error
 	Err() Error
+
+	// MarshalJSON returns a marshaled error
```
SammyOina:
```suggestion
	// MarshalJSON returns a marshaled error.
```

SammyOina (changes requested):

Makefile
```diff
@@ -17,7 +17,7 @@ USER_REPO ?= $(shell git remote get-url origin | sed -e 's/.*\/\([^/]*\)\/\([^/]
 BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD  2>/dev/null || git describe --tags --abbrev=0  2>/dev/null )
 empty:=
 space:= $(empty) $(empty)
-DOCKER_PROJECT ?= $(shell echo $(subst $(space),,$(USER_REPO)_$(BRANCH)) | tr -c -s '[:alnum:][=-=]' '_' | tr '[:upper:]' '[:lower:]')
+DOCKER_PROJECT ?= $(shell echo $(subst $(space),,$(USER_REPO)_$(BRANCH)) | tr -c -s '[:alnum:][=-=]' '_')
```
dborovcanin: This is an unrelated change and should not be here. Please be careful when rebasing.

dborovcanin (changes requested):

dborovcanin (approved):

drasko (approved):
LGTM

PR #1892, opened 2023-08-21
NOISSUE - Change error description

WashingtonKK:

### What does this do?
Doesn't return specific error in case of invalid password. In the event of an invalid password, it returns a general error message so that it does not show whether the user exists or not.

### DIFF
--- a/pkg/errors/types.go
+++ b/pkg/errors/types.go
@@ -37,6 +37,6 @@ var (
 	// ErrWrongSecret indicates a wrong secret was provided.
 	ErrWrongSecret = New("wrong secret")
 
-	// ErrLogin indicates wrong login credentials
+	// ErrLogin indicates wrong login credentials.
 	ErrLogin = New("invalid user id or secret")
 )
--- a/pkg/errors/types.go
+++ b/pkg/errors/types.go
@@ -35,5 +35,5 @@ var (
 	ErrScanMetadata = New("failed to scan metadata in db")
 
 	// ErrWrongSecret indicates a wrong secret was provided.
-	ErrWrongSecret = New("wrong secret")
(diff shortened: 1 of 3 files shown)


pkg/errors/types.go
```diff
@@ -35,5 +35,5 @@ var (
 	ErrScanMetadata = New("failed to scan metadata in db")
 
 	// ErrWrongSecret indicates a wrong secret was provided.
-	ErrWrongSecret = New("wrong secret")
+	ErrWrongSecret = New("invalid user id or secret")
```
dborovcanin: Rename to `ErrLogin`.

users/hasher/hasher.go
```diff
@@ -13,7 +13,7 @@ const cost int = 10
 
 var (
 	errHashPassword    = errors.New("Generate hash from password failed")
-	errComparePassword = errors.New("Compare hash and password failed")
+	errComparePassword = errors.New("Failed to verify identity or secret")
```
dborovcanin: Make these errors lowercase. Also, this error is specific to the actual secret so it does not make sense to mention identity here.

dborovcanin (changes requested):

users/hasher/hasher.go
```diff
@@ -13,7 +13,7 @@ const cost int = 10
 
 var (
 	errHashPassword    = errors.New("Generate hash from password failed")
```
SammyOina:
```suggestion
	errHashPassword    = errors.New("generate hash from password failed")
```

SammyOina (changes requested):

SammyOina (approved):

dborovcanin (approved):

PR #1885, opened 2023-08-08
NOISSUE - Enable JetStream on NATS

rodneyosodo:
Signed-off-by: rodneyosodo <blackd0t@protonmail.com>

### What does this do?
Enables jestream on nats

### Which issue(s) does this PR fix/relate to?
No issue

### List any changes that modify/break current functionality
None

### Have you included tests for your changes?
Manually tested

### Did you document any new/modified functionality?
No

### Notes
This is the first of many PRs to move from Redis for streaming to Nats

PR #1893, opened 2023-08-23
NOISSUE - Use Nats JetStream As Internal Message Bus 

rodneyosodo:

### What does this do?
Moves the internal message bus implementation to use Nats JetStream.

No issue

Update the `NewPublisher` and `NewPubSub` function parameters.

I have added integration tests to ensure the new Nats JetStream implementation works correctly.

Not yet

### Notes
This change improves the performance and scalability of our internal message bus by switching to Nats JetStream. The new implementation maintains the same external interface, so there should be minimal disruption.

Nats provides 2 APIs to interact with Jestream. `jetstream` package is preferred compared to the `nats` package because,
> `jetstream` package is a new client API to interact with NATS JetStream, aiming to replace the JetStream client implementation from `nats` package.

To be merged after https://github.com/mainflux/mainflux/pull/1885


docker/.env
```diff
@@ -377,7 +379,7 @@ MF_POSTGRES_HOST=mainflux-postgres
 MF_POSTGRES_PORT=5432
 MF_POSTGRES_USER=mainflux
 MF_POSTGRES_PASS=mainflux
-MF_POSTGRES_NAME=mainflux
+MF_POSTGRES_NAME=messages
```
SammyOina, line 413: is this change required?

rodneyosodo: Yes.

drasko:
@dborovcanin and @arvindh123 to review

arvindh123 (commented):
LGTM

SammyOina (approved):

pkg/messaging/nats/pubsub.go
```diff
@@ -143,27 +169,35 @@ func (ps *pubsub) Unsubscribe(ctx context.Context, id, topic string) error {
 			return err
 		}
 	}
-	if err := current.Unsubscribe(); err != nil {
+
+	current.Stop()
+
+	if err := ps.stream.DeleteConsumer(ctx, id); err != nil {
```
dborovcanin: Unlike plain NATS, JS client is able to hold consumer (subscription) data. Do we need to hold our own list of subscriptions in that case?

dborovcanin (changes requested):

pkg/messaging/nats/pubsub.go
rodneyosodo: I have made the consumer durable meaning consumers can have subscriptions bound to the topic and resume until the consumer is explicitly deleted. We keep an internal list of subscriptions to track the `ConsumeContext` across consumers so that we can stop and delete the consumer upon unsubscribing

pkg/messaging/nats/pubsub.go
```diff
@@ -126,44 +116,42 @@ func (ps *pubsub) Unsubscribe(ctx context.Context, id, topic string) error {
...
+	// A durable name cannot contain whitespace, ., *, >, path separators (forward or backwards slash), and non-printable characters.
+	topic = strings.ReplaceAll(topic, " ", "_")
+	topic = strings.ReplaceAll(topic, ".", "_")
+	topic = strings.ReplaceAll(topic, "*", "_")
+	topic = strings.ReplaceAll(topic, ">", "_")
+	topic = strings.ReplaceAll(topic, "/", "_")
+	topic = strings.ReplaceAll(topic, "\\", "_")
+
+	return fmt.Sprintf("%s-%s", topic, id)
+}
```
dborovcanin: Simplify:
```go
func formatConsumerName(topic, id string) string {
	// A durable name cannot contain whitespace, ., *, >, path separators (forward or backwards slash), and non-printable characters.
	chars := []string{
		" ", "_",
		".", "_",
		"*", "_",
		">", "_",
		"/", "_",
		"\\", "_",
	}
	rplcr := strings.NewReplacer(chars...)
	topic = rplcr.Replace(topic)
	return fmt.Sprintf("%s-%s", topic, id)
}
```

dborovcanin (changes requested):

dborovcanin (approved):

drasko (approved):
LGTM

PR #1897, opened 2023-08-28
NOISSUE - Add Event Sourcing Package

rodneyosodo:
Signed-off-by: Rodney Osodo <socials@rodneyosodo.com>

### What does this do?
Adds a seperate package to handle event sourcing just like messaging

No issue

- move event sourcing publisher to one place
- move event sourcing subscriber to one place
- add handler function to services that consume event such as bootstrap

Yes

Not yet

### Notes
The new es handler looks like this 
```go
type Event interface {
	Encode() (map[string]interface{}, error)
}

type Publisher interface {
	Publish(ctx context.Context, event Event) error
	Close() error
}

type EventHandler interface {
	Handle(ctx context.Context, event Event) error
}

type Subscriber interface {
	Subscribe(ctx context.Context, handler EventHandler) error
	Close() error
}

```


drasko:
@dborovcanin and @arvindh123 to review

pkg/events/redis/publisher.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, err
+	}
+
+	es := &pubEventStore{
+		client:            redis.NewClient(opts),
+		unpublishedEvents: make(chan *redis.XAddArgs, events.MaxUnpublishedEvents),
+		stream:            stream,
+	}
+
+	go es.StartPublishingRoutine(ctx)
```
SammyOina: I don't think we need to have StartPublishingRoutine exported

pkg/events/events.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// Publisher specifies events publishing API.
+type Publisher interface {
+	// Publish publishes event to stream.
+	Publish(ctx context.Context, event Event) error
+
+	// StartPublishingRoutine starts a routine that checks for unpublished events
+	// and publishes them to stream.
+	StartPublishingRoutine(ctx context.Context)
```
SammyOina: this doesn't need to be part of the interface or exported

rodneyosodo: It is exported so that the different `es` should implement it. As we want the different es for example `Redis streams` or `NATS JS` to implement it as a functionality.

SammyOina: do both of them require such a routine with the same signature and some possible es we might use in the future

rodneyosodo: RabbitMQ and Redis don't support using buffers. Hence the implementation for Nats is ignore https://github.com/rodneyosodo/mainflux/blob/d140b10a4936bee0afbadd0c34486e70f5f4f153/pkg/events/nats/publisher.go#L77

SammyOina (approved):

pkg/events/events.go
rodneyosodo: The function only takes in context for cancellation purposes as the implementation might vary.

dborovcanin: Since it's purely related to the implementation, let's remove this method from the Publisher interface for the time being.

dborovcanin (approved):

drasko (approved):
LGTM

PR #1905, opened 2023-09-08
NOISSUE - Refactor `e2e` Testing Tool

rodneyosodo:
### What does this do?
This PR has been repurposed to refactoring the `e2e` testing tool

No issue
- There was an issue when trying to create more than 100 things/channels hence, we changed to batching in case of many entities past 100 for things and channels
- Fix sending of messages via ws
- Fix repetitive names by using a better library to generate random names

None

None

No

### Notes
There is an issue with ws connection using this method. When publishing messages sometimes some messages are dropped. This was temporarily solved by using one connection to publish but this is not ideal for our test case.


tools/e2e/e2e.go
```diff
@@ -67,8 +63,8 @@ func init() {
 // - Publish message from HTTP, MQTT, WS and CoAP Adapters.
 func Test(conf Config) {
 	sdkConf := sdk.Config{
-		ThingsURL:       fmt.Sprintf("http://%s", conf.Host),
-		UsersURL:        fmt.Sprintf("http://%s", conf.Host),
+		ThingsURL:       fmt.Sprintf("http://%s:9000", conf.Host),
```
SammyOina: extract to constant

SammyOina (changes requested):

drasko (approved):
LGTM

PR #1906, opened 2023-09-11
NOISSUE - Add Send Invitation To Register Users

rodneyosodo:

### What does this do?
Adds send invitation email for registering users

No issue

- Add `ownerID` if the token is either AccessToken or InvitationToken and there is no error while parsing the token
- Remove `email` from JWT Claims as they are not being used
- Change Reset Secret to Parse token with type ResetPassword i.e. AccessToken can't be used to reset secret
- Modified emailing template to accommodate both password Reset and Invitation hence the contents of both emails are defined in code level

Yes

Not yet

### Notes

There is a challenge with specifying more than one email template hence modifier template to accommodate 2 scenarios

#### Password Reset

image

#### Invitation Email

image


dborovcanin:
There is no need to integrate this into the Users service. It should be added as a separate `Invitation` service. Closing this one since we do not have an invitation model on the roadmap for an open-source solution anytime soon.

Issue #1758, opened 2023-04-02
Explore Grafana K6S tool for distributed performance testing


drasko:
**FEATURE REQUEST**

1. Is there an open issue addressing this request? If it does, please add a "+1" reaction to the
   existing issue, otherwise proceed to step 2.

2. Describe the feature you are requesting, as well as the possible use case(s) for it.

Research https://k6.io/ and it's use for MF benchmarking of all 4 protocols: HTTP, MQTT, CoAP and WS.

3. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).

Nice-to-have

PR #1912, opened 2023-10-04
MF - 1758 - Sync With Benchmark Testing

rodneyosodo:
### What does this do?
1. Add trace ratio for traces `MF_JAEGER_TRACE_RATIO` i.e. don't always trace every request. 1.0 means always sample while 0.0 means never sample. This has proven useful for benchmark testing as when the ratio is 1.0 the Jaeger container uses a lot of memory
2. Add Postgres max connection `MF_POSTGRES_MAX_CONNECTIONS`. This is to increase the connected clients to Postgres i.e. previously a max of 100 clients was not ideal for testing

Related to #1758 

None

No

No

### Notes
Linked with https://github.com/mainflux/benchmark/pull/10


SammyOina (approved):

docker/.env
```diff
@@ -60,10 +60,14 @@ MF_JAEGER_FRONTEND=16686
 MF_JAEGER_COLLECTOR=14268
 MF_JAEGER_CONFIGS=5778
 MF_JAEGER_URL=http://jaeger:14268/api/traces
+MF_JAEGER_TRACE_RATIO=1.0
 
 ## Call home
 MF_SEND_TELEMETRY=true
 
+## Postgres
+MF_POSTGRES_MAX_CONNECTIONS=100000
```
arvindh123: Why we should have 100,000 connections by default ?

arvindh123 (changes requested):
Expect `POSTGRES_MAX_CONNECTIONS` limit everything look good.

Some thing can we have like this below 

In .env
```diff
--  MF_POSTGRES_MAX_CONNECTIONS=100
```
In docker-compose.yaml
```diff
--       MF_POSTGRES_MAX_CONNECTIONS: ${MF_POSTGRES_MAX_CONNECTIONS}
++     MF_POSTGRES_MAX_CONNECTIONS: ${MF_POSTGRES_MAX_CONNECTIONS:-100}}
```

docker/.env
rodneyosodo: Reduced to 100 i.e default

arvindh123 (approved):
LGTM

dborovcanin (approved):

drasko (approved):
LGTM

PR #1923, opened 2023-10-16
NOISSUE:  Listing of shared things with users & Update SDK

arvindh123:

### What does this do?

- Provides endpoint and service functions to list shared things with users 
- Update SDK 
- Update CLI
- Add signature to previous commits


users/api/endpoints.go
```diff
@@ -102,14 +102,50 @@ func listClientsEndpoint(svc users.Service) endpoint.Endpoint {
...
+		}
+
+		return buildMembersResponse(page), nil
+	}
+}
+
+func listMembersByChannelEndpoint(svc users.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(listMembersByObjectReq)
+		// in spiceDB channel is group kind
```
drasko: Start comment with a capital letter. Also, capitalize SpiceDB. What s "group kind" - is kind some tem from SpiceDB?

drasko (changes requested):

users/api/endpoints.go
arvindh123: We not having any separate `type` for channel in spiceDB schema, Instead we using `group` type for storing channel relations.   
Here the function name is `listMembersByChannelEndpoint` , which does listing of users in a channel 
This function request we are passing `req.objectKind = "groups"` ,  

This function is related to channels, then Why we here passing `groups` instead of `channels` ? Like this kind of question will raise when someone see the code.     
Answer is same as said in the first, We are not having separate `type` for `channel` in spiceDB schema, Instead we using same `group` type for channel.

arvindh123: updated comment

pkg/sdk/go/groups.go
```diff
@@ -164,6 +145,64 @@ func (sdk mfSDK) DisableGroup(id, token string) (Group, errors.SDKError) {
...
+		return errors.NewSDKError(err)
+	}
+
+	url := fmt.Sprintf("%s/%s/%s/%s", sdk.usersURL, groupsEndpoint, groupID, usersEndpoint)
+
+	_, _, sdkerr := sdk.processRequest(http.MethodPost, url, token, data, nil, http.StatusOK)
+	return sdkerr
+}
+
+func (sdk mfSDK) RemoveUsersToGroup(groupID string, req UsersRelationRequest, token string) errors.SDKError {
```
dborovcanin: RemoveUserFromGroup

pkg/sdk/go/groups.go
```diff
@@ -164,6 +145,64 @@ func (sdk mfSDK) DisableGroup(id, token string) (Group, errors.SDKError) {
 	return sdk.changeGroupStatus(id, disableEndpoint, token)
 }
 
+func (sdk mfSDK) AddUsersToGroup(groupID string, req UsersRelationRequest, token string) errors.SDKError {
```
dborovcanin: AddUserToGroup (a single user)

dborovcanin (changes requested):

pkg/sdk/go/groups.go
arvindh123: Changed in all instance

pkg/sdk/go/groups.go
arvindh123: Changed in all instance

dborovcanin (dismissed):

dborovcanin (dismissed):

internal/groups/service.go
```diff
@@ -217,7 +217,9 @@ func (svc service) ListGroups(ctx context.Context, token string, memberKind, mem
 	}
 
 	if len(ids) <= 0 {
-		return groups.Page{}, errors.ErrNotFound
+		return groups.Page{
```
dborovcanin, line 220: Why do we even consider less than 0 for IDs length (i.e. why not `if len(ids) == 0`)?

things/service.go
```diff
@@ -133,27 +133,78 @@ func (svc service) ViewClient(ctx context.Context, token string, id string) (mfc
...
+			return mfclients.ClientsPage{}, err
+		}
+	default:
+		ids, err = svc.listClientIDs(ctx, userID, pm.Permission)
+		if err != nil {
+			return mfclients.ClientsPage{}, err
+		}
+	}
+
+	if len(ids) <= 0 {
```
dborovcanin: The same here.

dborovcanin (changes requested):

internal/groups/service.go
arvindh123: Since len() returns `int`

things/service.go
arvindh123: Since len() returns `int`

dborovcanin (dismissed):

drasko (approved):
LGTM

PR #2057, opened 2024-01-12
MG-1965 - Process Event Logs

### LINKED ISSUE #1965: Feature: Process Thing event log

I want to be able to see the history of actions on a single Thing.

A Thing admin or editor should be able to view the history of device changes, as well as device activity status (if the device is online or offline). Since we already have events for all Thing actions, we will need a service to process and permanently store those events.

Should-have

### Anything else?

_No response_

rodneyosodo:
# What type of PR is this?

This is a feature PR because it adds eventlogs service to store history of events

## What does this do?

Adds event logs service

- Related Issue #
- Resolves absmach/magistrala-ui-new#164 absmach/magistrala#1957


api/openapi/event-logs.yml
```diff
@@ -0,0 +1,278 @@
...
+      in: path
+      schema:
+        type: string
+        enum:
+          - user
+          - group
+          - domain
+          - thing
+          - platform
+          - thing
```
ianmuchyri:
```suggestion

```

ianmuchyri (changes requested):

api/openapi/event-logs.yml
rodneyosodo: Yes thing is a valid type

ianmuchyri: It's repeated

eventlogs/postgres/events.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (repo *repository) RetrieveAll(ctx context.Context, page eventlogs.Page) (eventlogs.EventsPage, error) {
+	query := pageQuery(page)
+
+	sq := "id, operation, occurred_at"
+	if page.WithPayload {
+		sq += ", payload"
+	}
+	q := fmt.Sprintf("SELECT %s FROM events %s ORDER BY occurred_at LIMIT :limit OFFSET :offset;", sq, query)
```
ianmuchyri: we should display events from the latest. Order by `DESC`

ianmuchyri (changes requested):

eventlogs/postgres/events.go
rodneyosodo: Added direction as a query parameter with the default being descending.

Makefile
```diff
@@ -5,7 +5,7 @@ MG_DOCKER_IMAGE_NAME_PREFIX ?= magistrala
 BUILD_DIR = build
 SERVICES = auth users things http coap ws lora influxdb-writer influxdb-reader mongodb-writer \
 	mongodb-reader cassandra-writer cassandra-reader postgres-writer postgres-reader timescale-writer timescale-reader cli \
-	bootstrap opcua twins mqtt provision certs smtp-notifier smpp-notifier invitations
+	bootstrap opcua twins mqtt provision certs smtp-notifier smpp-notifier invitations event-logs
```
dborovcanin: Rename to `activities` or similar alternative.

cmd/event-logs/main.go
```diff
@@ -0,0 +1,174 @@
...
+	subscriber, err := store.NewSubscriber(ctx, cfg.ESURL, logger)
+	if err != nil {
+		logger.Error("failed to create subscriber: %s", err)
+		exitCode = 1
+		return
+	}
+
+	logger.Info("Subscribed to Event Store")
+
+	if err := eventlogs.Start(ctx, svcName, subscriber, repo); err != nil {
```
dborovcanin: Add event processing to service, rather than using `repo` directly. Also, add logging to it.

eventlogs/events.go
```diff
@@ -0,0 +1,73 @@
...
+
+	return json.Marshal(a)
+}
+
+// Service provides access to the event log service.
+//
+//go:generate mockery --name Service --output=./mocks --filename service.go --quiet --note "Copyright (c) Abstract Machines"
+type Service interface {
+	// ReadAll retrieves all events from the database with the given page.
+	ReadAll(ctx context.Context, token string, page Page) (EventsPage, error)
```
dborovcanin: Add `Save` and use service instead of repo for consuming.

eventlogs/api/transport.go
```diff
@@ -0,0 +1,103 @@
...
+
+// MakeHandler returns a HTTP API handler with health check and metrics.
+func MakeHandler(svc eventlogs.Service, logger *slog.Logger, svcName, instanceID string) http.Handler {
+	opts := []kithttp.ServerOption{
+		kithttp.ServerErrorEncoder(apiutil.LoggingErrorEncoder(logger, api.EncodeError)),
+	}
+
+	mux := chi.NewRouter()
+
+	mux.Get("/events/{id}/{type}", otelhttp.NewHandler(kithttp.NewServer(
```
dborovcanin: ID and Type should be optional query params.

dborovcanin (changes requested):
Not a message consumer, but this is still a consumer, so let's move it there.

cli/activitylog.go
```diff
@@ -0,0 +1,50 @@
...
+		if err != nil {
+			logError(err)
+			return
+		}
+
+		logJSON(activities)
+	},
+}
+
+// NewActivitiesCmd returns invitations command.
```
WashingtonKK:
```suggestion
// NewActivitiesCmd returns activities command.
```

mqtt/events/events.go
WashingtonKK, line 1: Update also https://github.com/rodneyosodo/magistrala/blob/f34d4a24f611e6b99da06d37a0ced37cde4e06fc/internal/groups/events/events.go#L61-L69 .

WashingtonKK: The `unassignGroupEvent` lacks an id, how will you store it in the activities log?

docker/addons/activitylog/docker-compose.yml
WashingtonKK, line 1: Why is this in the addons folder yet in makefile you have included it as a core service?

Makefile
```diff
@@ -5,7 +5,7 @@ MG_DOCKER_IMAGE_NAME_PREFIX ?= magistrala
 BUILD_DIR = build
 SERVICES = auth users things http coap ws lora influxdb-writer influxdb-reader mongodb-writer \
 	mongodb-reader cassandra-writer cassandra-reader postgres-writer postgres-reader timescale-writer timescale-reader cli \
-	bootstrap opcua twins mqtt provision certs smtp-notifier smpp-notifier invitations
+	bootstrap opcua twins mqtt provision certs smtp-notifier smpp-notifier invitations activitylog
```
WashingtonKK, line 8:
```suggestion
	bootstrap opcua twins mqtt provision certs smtp-notifier smpp-notifier invitations
```

mqtt/events/events.go
rodneyosodo: reverting back

internal/groups/events/events.go
```diff
@@ -4,14 +4,13 @@
...
 import (
-	"encoding/json"
 	"time"
 
 	"github.com/absmach/magistrala/pkg/events"
 	groups "github.com/absmach/magistrala/pkg/groups"
 )
 
-const (
+var (
```
arvindh123, line 13: Should we use `var` here ?

activitylog/postgres/init.go
```diff
@@ -0,0 +1,32 @@
...
+				Id: "activities_01",
+				Up: []string{
+					`CREATE TABLE IF NOT EXISTS activities (
+						id 			UUID NOT NULL DEFAULT gen_random_uuid() PRIMARY KEY,
+						operation	VARCHAR NOT NULL CHECK (operation <> ''),
+						occurred_at	TIMESTAMP NOT NULL,
+						attributes	JSONB NOT NULL,
+						metadata	JSONB,
+						UNIQUE(operation, occurred_at, attributes)
+					)`,
```
arvindh123: We can have indexing here like `CREATE INDEX ON activities (operation, occurred_at);`

activitylog/consumer.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/absmach/magistrala/pkg/events"
+	"github.com/absmach/magistrala/pkg/events/store"
+)
+
+// Start method starts consuming messages received from Event store.
+func Start(ctx context.Context, consumer string, sub events.Subscriber, service Service) error {
+	subCfg := events.SubscriberConfig{
+		Consumer: consumer,
+		Stream:   store.StreamAllEvents,
+		Handler:  Handle(service),
```
arvindh123: Can we have something like `service.EventHandler`? 

And its definition will be like

```go
func (svc *service)EventHandle(ctx context.Context, event events.Event) error {
	data, err := event.Encode()
	if err != nil {
		return err
	}

	operation, ok := data["operation"].(string)
	if !ok {
		return errors.New("missing operation")
	}
	delete(data, "operation")

	if operation == "" {
		return errors.New("missing operation")
	}

	occurredAt, ok := data["occurred_at"].(float64)
	if !ok {
		return ErrMissingOccurredAt
	}
	delete(data, "occurred_at")

	if occurredAt == 0 {
		return ErrMissingOccurredAt
	}

	metadata, ok := data["metadata"].(map[string]interface{})
	if !ok {
		metadata = make(map[string]interface{})
	}
	delete(data, "metadata")

	if len(data) == 0 {
		return errors.New("missing attributes")
	}

	activity := Activity{
		Operation:  operation,
		OccurredAt: time.Unix(0, int64(occurredAt)),
		Attributes: data,
		Metadata:   metadata,
	}

	return svc.repository.Save(ctx, activity)
}

```

activitylog/activity.go
```diff
@@ -0,0 +1,165 @@
...
+	return json.Marshal(a)
+}
+
+// Service provides access to the activity log service.
+//
+//go:generate mockery --name Service --output=./mocks --filename service.go --quiet --note "Copyright (c) Abstract Machines"
+type Service interface {
+	// Save saves the activity to the database.
+	Save(ctx context.Context, activity Activity) error
+
```
arvindh123: As suggested before,  here can we have something like `EventHandler` https://github.com/absmach/magistrala/pull/2057#discussion_r1589139829

arvindh123 (changes requested):
User of another domain able to see the events all of all domain and non domain events. 

With the below request `user_2` whois in `domain_2` and not in `domain_1` able to see events of `domain_1`

```bash
curl --location 'localhost:9021/activities?with_attributes=true' \
--header 'Authorization: Bearer <TOKEN>'
```

We should have domain as sperate column in table, so that we can add indexing and also helps in filter based on domain.

And we need additional column to identify it is domain releated event or non domain event. 

We can reuse the domain column with const value non-domain-event to identify the non domain event 


Non domain event should able to access by SuperAdmins 

Domain related login events should able to access ony by Domain Admins or Editors 

Entity editor should able to see the entity activities and viewer and members should not able to see the activities of domain.

All the above consideration will lead to new challenge , List all events for particular domain will becomes a challenge, I suggestion something like , for now we can list events of particular entity id and  to list all events on particular domain, we can do it later after having local index caching of spicedb policy or after some other mechanism until we find.

activitylog/consumer.go
rodneyosodo: We would lose more information when logging and tracing since event has not yet been processed to an activity

activitylog/postgres/init.go
rodneyosodo: I have created the following index for testing purpose. 

```SQL
CREATE INDEX idx_activities_group_filter_sort ON activities(operation, (attributes->>'id'), (attributes->>'group_id'), occurred_at DESC);
```

And found out that creating an index actually hurts our total DB performance mostly on counting total. What do you think?

This is the trace report for groups given the command

```bash
curl -X GET -H "Authorization: Bearer $USER_TOKEN" "http://localhost:9021/activities?id=$GROUP_ID&entity_type=group&with_attributes=true"
```

image

For a different query which has not been index ie

```bash
curl -X GET -H "Authorization: Bearer $USER_TOKEN" "http://localhost:9021/activities?id=$THING_ID&entity_type=thing&with_attributes=true"
```

This is the trace report

image

rodneyosodo: Actually after indexing we have faster query times

internal/groups/events/events.go
rodneyosodo: We are using `var`

internal/apiutil/errors.go
```diff
@@ -173,4 +173,13 @@ var (
...
 	ErrMissingTo = errors.New("missing to time value")
+
+	// ErrInvalidEntity indicates invalid entity type.
+	ErrInvalidEntity = errors.New("invalid entity type")
+
+	// ErrMissingEntityType indicates missing entity type.
+	ErrMissingEntityType = errors.New("missing entity type")
+
+	// ErrInvalidTime indicates invalid time i.e not unix time.
+	ErrInvalidTime = errors.New("invalid time use unix time")
```
WashingtonKK: Reword this to something like: "invalid time format" or something that is clearer to understand.

WashingtonKK (commented):
A user in `domainA` can see activities of a separate `domainB` even though they are not part of the domain. Is this intended?
Can we have auth such that a user can only see the activities in their domain?
@rodneyosodo what do you think?

rodneyosodo:
> A user in `domainA` can see activities of a separate `domainB` even though they are not part of the domain. Is this intended? Can we have auth such that a user can only see the activities in their domain? @rodneyosodo what do you think?


They can't see the entity i.e. id or name or metadata. All authenticated users can only see the operation and when the event occurred. Sensitive information requires authorization. If you want to view entity events for a specific entity you need ViewPermission to be able to see the events. 

https://github.com/rodneyosodo/magistrala/blob/0a7e63801bc1649ac824635a85e5f2b490bb08ee/activitylog/service.go#L31C1-L50C2

activitylog/service.go
```diff
@@ -0,0 +1,84 @@
...
+		return svc.repository.RetrieveAll(ctx, page)
+	}
+
+	if err := svc.authorize(ctx, userID, domainID, auth.AdminPermission, auth.PlatformType, auth.MagistralaObject); err != nil {
+		page.WithAttributes = false
+		page.WithMetadata = false
+	}
+
+	return svc.repository.RetrieveAll(ctx, page)
+}
```
arvindh123: @rodneyosodo Do you think we should all activities if user is not super admin ? 
If yes, why we should all activites for non super admin, How it will be useful from them .

arvindh123 (changes requested):

activitylog/service.go
rodneyosodo: What is being retrieved doesn't contain sensitive information since with omit attributes and metadata we only retrieve the operation and when it occurred

arvindh123: I'm just thinking, How it will be usefull for user. It might usefull for case like how in `ecommerce` site they will show popup with hidden sensitive infromation of recently happened purchases.
But how it might be useful much in IoT case ?. 

So can we make it configurable in env with by default `false`? 

So if application want to show all events infromation, they can enable it by env variable

activitylog/api/endpoint_test.go
```diff
@@ -0,0 +1,282 @@
...
+			token:  validToken,
+			url:    "/user/",
+			status: http.StatusNotFound,
+			svcErr: nil,
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.desc, func(t *testing.T) {
+			repoCall := svc.On("RetrieveAll", mock.Anything, c.token, mock.Anything).Return(activitylog.ActivitiesPage{}, c.svcErr)
```
arvindh123: `svcCall` might be a better name

activitylog/activity.go
```diff
@@ -0,0 +1,162 @@
...
+func (e EntityType) Query() string {
+	switch e {
+	case UserEntity:
+		return "((operation LIKE 'user.%' AND attributes->>'id' = :entity_id) OR (attributes->>'user_id' = :entity_id))"
+	case GroupEntity:
+		return "((operation LIKE 'group.%' AND attributes->>'id' = :entity_id) OR (attributes->>'group_id' = :entity_id))"
+	case ThingEntity:
+		return "((operation LIKE 'thing.%' AND attributes->>'id' = :entity_id) OR (attributes->>'thing_id' = :entity_id))"
+	case ChannelEntity:
+		return "((operation LIKE 'channel.%' AND attributes->>'id' = :entity_id) OR (attributes->>'group_id' = :entity_id))"
```
arvindh123: @rodneyosodo 
I could not list `channels`create  event. 
I think this should be `group` or we should change in all events to `channel`

arvindh123 (changes requested):

activitylog/service.go
```diff
@@ -0,0 +1,68 @@
...
+
+	req := &magistrala.AuthorizeReq{
+		SubjectType: auth.UserType,
+		SubjectKind: auth.TokenKind,
+		Subject:     token,
+		Permission:  permission,
+		ObjectType:  objectType,
+		Object:      object,
+	}
+
```
arvindh123, line 58: This can be done this way, In this way you could not view user activities without domain token 

To view users there is no mandatory for domain token, both login token and domain token should work. 

But here with login token you could not view the user activities. It is because login token doesn't have subject, So the authorization fails.

activitylog/service.go
```diff
@@ -0,0 +1,68 @@
...
+	res, err := svc.auth.Authorize(ctx, req)
+	if err != nil {
+		return errors.Wrap(svcerr.ErrAuthorization, err)
+	}
+	if !res.GetAuthorized() {
+		return svcerr.ErrAuthorization
+	}
+
+	return nil
+}
```
arvindh123: This can be done this way, In this way you could not view user activities without domain token 

To view users there is no mandatory for domain token, both login token and domain token should work. 

But here with login token you could not view the user activities. It is because login token doesn't have subject, So the authorization fails. 


So I suggest something like below,
But drawback of this method is it requires two auth calls. 
This problem exits all over the project, So there is ticket to reslove this two auth call issue. 

```suggestion
func (svc *service) authorize(ctx context.Context, token, entityID, entityType string) error {
	res, err := svc.auth.Identify(ctx, &magistrala.IdentityReq{Token: token})
	if err != nil {
		return err
	}

	permission := auth.ViewPermission
	objectType := entityType
	object := entityID
	subject := res.GetId()
	// If the entity is a user, we need to check if the user is an admin
	if entityType == auth.UserType {
		permission = auth.AdminPermission
		objectType = auth.PlatformType
		object = auth.MagistralaObject
		subject = res.GetUserId()
	}

	req := &magistrala.AuthorizeReq{
		SubjectType: auth.UserType,
		SubjectKind: auth.UsersKind,
		Subject:     subject,
		Permission:  permission,
		ObjectType:  objectType,
		Object:      object,
	}

	authRes, err := svc.auth.Authorize(ctx, req)
	if err != nil {
		return errors.Wrap(svcerr.ErrAuthorization, err)
	}
	if !authRes.GetAuthorized() {
		return svcerr.ErrAuthorization
	}

	return nil
}
```

arvindh123 (changes requested):

cli/activitylog.go
```diff
@@ -0,0 +1,50 @@
...
+package cli
+
+import (
+	mgxsdk "github.com/absmach/magistrala/pkg/sdk/go"
+	"github.com/spf13/cobra"
+)
+
+var cmdActivity = cobra.Command{
+	Use:   "get <entity_id> <entity_type> <user_auth_token>",
+	Short: "Get activities",
```
arvindh123:
```suggestion
	Use:   "get <entity_type> <entity_id> <user_auth_token>",
```

arvindh123 (changes requested):

activitylog/service.go
rodneyosodo: Why should someone view activities without a domain login, this is because you can't view an entity without it too?

arvindh123: Users are not not owned by domain, they are owned by Platform
So `SuperAdmin` should able to see the users activity logs without login to domain. 
Like how we can able to list domain and list users without login to domain.

activitylog/service.go
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) Abstract Machines
```
dborovcanin: Add a doc.go explaining the purpose of the service.

activitylog/postgres/init.go
```diff
@@ -0,0 +1,36 @@
...
+
+func Migration() *migrate.MemoryMigrationSource {
+	return &migrate.MemoryMigrationSource{
+		Migrations: []*migrate.Migration{
+			{
+				Id: "activities_01",
+				Up: []string{
+					`CREATE TABLE IF NOT EXISTS activities (
+						id 			UUID NOT NULL DEFAULT gen_random_uuid() PRIMARY KEY,
+						operation	VARCHAR NOT NULL CHECK (operation <> ''),
```
dborovcanin: No need for this check, since we don't do that in the rest of the codebase. We will perform this check during validation.

dborovcanin (changes requested):
Please rename the service, Journal seems like a good name (Log is suboptimal, and Activities and Events could be easily misinterpreted).

arvindh123 (dismissed):

journal/service.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/absmach/magistrala"
+	"github.com/absmach/magistrala/auth"
+	"github.com/absmach/magistrala/pkg/errors"
+	svcerr "github.com/absmach/magistrala/pkg/errors/service"
+)
+
+type service struct {
+	auth       magistrala.AuthServiceClient
+	repository Repository
```
dborovcanin, line 18: Let's use a UUID provider like in other services.

dborovcanin (changes requested):

dborovcanin (approved):

PR #2083, opened 2024-02-15
NOISSUE - Fix Bootstrap thing creation flow

arvindh123:
# What type of PR is this?
This is a bug fix because it fixes the following issue:  
Modify Bootstrap thing creation flow

## What does this do?
Modify Bootstrap thing creation flow

N/A

N/A

### DIFF
--- a/bootstrap/service.go
+++ b/bootstrap/service.go
@@ -104,19 +104,21 @@ type ConfigReader interface {
 }
 
 type bootstrapService struct {
-	auth    magistrala.AuthServiceClient
-	configs ConfigRepository
-	sdk     mgsdk.SDK
-	encKey  []byte
+	auth       magistrala.AuthServiceClient
+	configs    ConfigRepository
+	sdk        mgsdk.SDK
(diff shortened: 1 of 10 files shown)


bootstrap/service.go
```diff
@@ -152,8 +154,10 @@ func (bs bootstrapService) Add(ctx context.Context, token string, cfg Config) (C
 
 	saved, err := bs.configs.Save(ctx, cfg, toConnect)
 	if err != nil {
+		// If id is empty, then a new thing has been created function - bs.thing(id, token)
```
dborovcanin, line 157: // If the ID is empty, a new thing has been created by `bs.thing(id, token)`, so if the bootstrap config creation returns an error, try to delete the newly created thing.

dborovcanin (changes requested):

bootstrap/service.go
arvindh123: Changed

dborovcanin (approved):

PR #2084, opened 2024-02-19
NOISSUE - Vault operations with app role authentication 

arvindh123:
# What type of PR is this?

- Related Issue #
- Resolves #


docker/addons/vault/vault-env-copy.sh
```diff
@@ -0,0 +1,21 @@
...
+
+write_env() {
+    sed -i "s,MG_VAULT_UNSEAL_KEY_1=.*,MG_VAULT_UNSEAL_KEY_1=$(awk -F ": " '$1 == "Unseal Key 1" {print $2}' data/secrets)," $MAGISTRALA_DIR/docker/.env
+    sed -i "s,MG_VAULT_UNSEAL_KEY_2=.*,MG_VAULT_UNSEAL_KEY_2=$(awk -F ": " '$1 == "Unseal Key 2" {print $2}' data/secrets)," $MAGISTRALA_DIR/docker/.env
+    sed -i "s,MG_VAULT_UNSEAL_KEY_3=.*,MG_VAULT_UNSEAL_KEY_3=$(awk -F ": " '$1 == "Unseal Key 3" {print $2}' data/secrets)," $MAGISTRALA_DIR/docker/.env
+    sed -i "s,MG_VAULT_TOKEN=.*,MG_VAULT_TOKEN=$(awk -F ": " '$1 == "Initial Root Token" {print $2}' data/secrets)," $MAGISTRALA_DIR/docker/.env
+}
+
+
+
```
dborovcanin: Remove multiple empty lines.

docker/addons/vault/vault-env-copy.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/bash
```
dborovcanin: Consider renaming all `copy` files (maybe to `vault_copy_certs` and `vault_copy_env` - also, use underscore for the sake of consistency) because the `copy` suffix can be confused with accidentally copied files.

docker/addons/vault/vault-set-pki.sh
```diff
@@ -20,90 +20,145 @@ vault() {
...
 
+vaultConfigPKIClusterPath() {
+    vault write -namespace=${MG_VAULT_NAMESPACE} -address=${MG_VAULT_ADDR} ${MG_VAULT_PKI_PATH}/config/cluster aia_path=${MG_VAULT_PKI_CLUSTER_AIA_PATH} path=${MG_VAULT_PKI_CLUSTER_PATH}
+}
+
+vaultConfigPKICrl() {
+    vault write -namespace=${MG_VAULT_NAMESPACE} -address=${MG_VAULT_ADDR} ${MG_VAULT_PKI_PATH}/config/crl expiry="5m"  ocsp_disable=false ocsp_expiry=0 auto_rebuild=true auto_rebuild_grace_period="2m" enable_delta=true delta_rebuild_interval="1m"
+}
+
+
```
dborovcanin: Remove multiple empty lines.

docker/nginx/nginx-x509.conf
```diff
@@ -45,6 +45,15 @@ http {
         listen [::]:80 default_server;
         listen 443 ssl http2 default_server;
         listen [::]:443 ssl http2 default_server;
+
+        set $dynamic_server_name $MG_NGINX_SERVER_NAME;
+
+        if ($dynamic_server_name = '') {
+            set $dynamic_server_name "localhost";
+        }
```
dborovcanin, line 53: If we set it by default to `localhost`, there is no need for this check.

dborovcanin (changes requested):

docker/nginx/nginx-x509.conf
arvindh123: If the env variable `$MG_NGINX_SERVER_NAME` is empty, In that case `$dynamic_server_name` should not be empty, So it was set to localhost

arvindh123: Changed approach

docker/addons/vault/vault-set-pki.sh
arvindh123: done

docker/addons/vault/vault-env-copy.sh
arvindh123: file name replace - with underscore

docker/addons/vault/vault-env-copy.sh
arvindh123: done

dborovcanin (approved):

PR #2085, opened 2024-02-20
NOISSUE - Update timescale reader

Musilah:
# What type of PR is this?
This a feature because it updates the Timescale reader.

## What does this do?
This PR updates the Timescale Reader to be able to use aggregation queries for timeseries data to be used in dashboards charts. 

NOISSUE on Magistrala

No

No


docker/docker-compose.yml
```diff
@@ -741,7 +741,7 @@ services:
       POSTGRES_DB: ${MG_UI_DB_NAME}
       MG_POSTGRES_MAX_CONNECTIONS: ${MG_POSTGRES_MAX_CONNECTIONS}
     ports:
-      - 6007:5432
+      - 6042:5432
```
dborovcanin: Where does port `6042` come from?

pkg/sdk/go/message.go
```diff
@@ -69,3 +71,135 @@ func (sdk *mgSDK) SetContentType(ct ContentType) errors.SDKError {
...
+func (sdk mgSDK) withMessageQueryParams(baseURL, endpoint string, mpm MessagePageMeta) (string, error) {
+	q, err := mpm.mquery()
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("%s/%s?%s", baseURL, endpoint, q), nil
+}
+
+func (mpm MessagePageMeta) mquery() (string, error) {
```
dborovcanin: This looks like it can be handled better with a map and JSON serialization of MessagePageMeta.

dborovcanin (changes requested):

docker/docker-compose.yml
Musilah: It's the correct UI-db port.

dborovcanin: The correct UI port is default Postgres port at `5432`. The host port in compose is really any port we chose, and the reason we have this mapping in the first place is to expose UI DB so we can debug and deploy UI locally. We used `6xxx` ports for those purposes, but `6042` seems out of place.

ianmuchyri:
```
{
    "offset": 0,
    "limit": 10,
    "format": "messages",
    "aggregation": "SUM",
    "interval": "1 minute",
    "total": 100,
    "messages": [
        {
            "time": 1708687680,
            "value": 18474
        },
        {
            "time": 1708687620,
            "value": 27894
        },
        {
            "time": 1708687560,
            "value": 3737
        }
    ]
}
```
The total count is not correct

ianmuchyri:
Also for me to query for messages with the current implementation, I have to include aggregation and interval or it will not work. However, they are supposed to be conditional filters and the querying should work even without having to add them as filter parameters. Check on how to solve this.

image

pkg/sdk/go/message.go
```diff
@@ -69,3 +71,135 @@ func (sdk *mgSDK) SetContentType(ct ContentType) errors.SDKError {
...
+		q.Add("total", strconv.FormatUint(mpm.Total, 10))
+	}
+	if mpm.Aggregation != "" {
+		q.Add("aggregation", mpm.Aggregation)
+	}
+	if mpm.Interval != "" {
+		q.Add("interval", mpm.Interval)
+	}
+	return q.Encode(), nil
+}
```
dborovcanin: @Musilah Something like this:
```suggestion
func (mpm MessagePageMeta) q() (string, error) {
	b, err := json.Marshal(mpm)
	if err != nil {
		return "", err
	}
	q := map[string]interface{}{}
	if err := json.Unmarshal(b, &q); err != nil {
		return "", err
	}
	ret := url.Values{}
	for k, v := range q {
		switch t := v.(type) {
		case string:
			ret.Add(k, t)
		case float64:
			ret.Add(k, strconv.FormatFloat(t, 'f', -1, 64))
		case uint64:
			ret.Add(k, strconv.FormatUint(t, 10))
		case int64:
			ret.Add(k, strconv.FormatInt(t, 10))
		case json.Number:
			ret.Add(k, t.String())
		case bool:
			ret.Add(k, strconv.FormatBool(t))
		}
	}
	return ret.Encode(), nil
}
```
However, have on mind that things are a little bit more complex than this. For example, BoolValue and `omitempty` won't work well (it won't work in the current implementation, either) since it will result in `false`. You will need to play with pointers to make it work, i.e. nil value indicating that there is no interest in that in query.

pkg/sdk/go/sdk.go
```diff
@@ -69,6 +69,20 @@ var (
...
+	Subtopic    string  `json:"subtopic,omitempty"`
+	Publisher   string  `json:"publisher,omitempty"`
+	Comparator  string  `json:"comparator,omitempty"`
+	BoolValue   bool    `json:"vb,omitempty"`
+	StringValue string  `json:"vs,omitempty"`
+	DataValue   string  `json:"vd,omitempty"`
+	From        float64 `json:"from,omitempty"`
+	To          float64 `json:"to,omitempty"`
+	Aggregation string  `json:"aggregation,omitempty"`
+	Interval    string  `json:"interval,omitempty"`
```
ianmuchyri, line 83: add value, protocol, and subtopic to the pagemetadata

pkg/sdk/go/sdk.go
```diff
@@ -69,6 +69,20 @@ var (
 	ErrInvalidJWT = errors.New("invalid JWT")
 )
 
+type MessagePageMeta struct {
```
ianmuchyri:
```suggestion
type MessagePageMetadata struct {
```

readers/timescale/messages.go
```diff
@@ -37,7 +37,37 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
...
+        :limit OFFSET :offset;
+    `,
+			rpm.Interval,
+			rpm.Aggregation,
+			format,
+			fmtCondition(chanID, rpm),
+			order,
+		)
+	}
+	fmt.Println("query", q)
```
ianmuchyri: remove these logs

ianmuchyri (changes requested):

docker/.env
```diff
@@ -260,9 +260,9 @@ MG_UI_CONTENT_TYPE=application/senml+json
 MG_UI_INSTANCE_ID=
 MG_UI_DB_HOST=ui-db
 MG_UI_DB_PORT=5432
-MG_UI_DB_USER=magistrala
-MG_UI_DB_PASS=magistrala
-MG_UI_DB_NAME=ui
+MG_UI_DB_USER=magistrala-ui
```
rodneyosodo: Revert this

pkg/sdk/go/sdk.go
```diff
@@ -69,6 +69,22 @@ var (
...
+	Subtopic    string  `json:"subtopic,omitempty"`
+	Publisher   string  `json:"publisher,omitempty"`
+	Comparator  string  `json:"comparator,omitempty"`
+	BoolValue   *bool   `json:"vb,omitempty"`
+	StringValue string  `json:"vs,omitempty"`
+	DataValue   string  `json:"vd,omitempty"`
+	From        float64 `json:"from,omitempty"`
+	To          float64 `json:"to,omitempty"`
+	Aggregation string  `json:"aggregation,omitempty"`
+	Interval    string  `json:"interval,omitempty"`
```
rodneyosodo, line 83: Interval is covered by `to` and `from` values i.e` interval = to - from`

pkg/sdk/go/sdk.go
```diff
@@ -834,7 +850,7 @@ type SDK interface {
 	//  }
 	//  msgs, _ := sdk.ReadMessages(pm,"channelID", "token")
 	//  fmt.Println(msgs)
-	ReadMessages(pm PageMetadata, chanID, token string) (MessagesPage, errors.SDKError)
+	ReadMessages(pm MessagePageMeta, chanID, token string) (MessagesPage, errors.SDKError)
```
rodneyosodo: Update godocs

readers/messages.go
```diff
@@ -3,7 +3,9 @@
 
 package readers
 
-import "errors"
+import (
+	"errors"
```
rodneyosodo: inline

readers/api/transport.go
```diff
@@ -140,6 +142,15 @@ func decodeList(_ context.Context, r *http.Request) (interface{}, error) {
 	if err != nil {
 		return nil, errors.Wrap(apiutil.ErrValidation, err)
 	}
```
rodneyosodo, line 145: Add check if aggregation if valid then there need to be a non empty to and from values

readers/timescale/messages.go
```diff
@@ -36,8 +36,37 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
 		order = "created"
 		format = rpm.Format
 	}
-
-	q := fmt.Sprintf(`SELECT * FROM %s WHERE %s ORDER BY %s DESC LIMIT :limit OFFSET :offset;`, format, fmtCondition(chanID, rpm), order)
+	var q string
+	// Construct the base query without time_bucket and aggregation
+	// q := fmt.Sprintf(`SELECT * FROM %s WHERE %s ORDER BY %s ASC LIMIT :limit OFFSET :offset;`, format, fmtCondition(chanID, rpm), order)
```
rodneyosodo: Remove commented code

readers/timescale/messages.go
```diff
@@ -36,8 +36,37 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
...
+
+	// If aggregation is provided, add time_bucket and aggregation to the query
+	if rpm.Aggregation != "" {
+		q = fmt.Sprintf(`
+			SELECT
```
rodneyosodo: Why not something like for finding max value
```SQL
SELECT MAX(*) FROM %s WHERE %s
```

readers/timescale/messages.go
```diff
@@ -93,17 +122,36 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
...
+				WHERE 
+					%s 
+				GROUP BY 
+					time
+			) AS subquery;
+		`, rpm.Interval, format, fmtCondition(chanID, rpm))
+	} else {
+		countQuery = fmt.Sprintf(`SELECT COUNT(*) FROM %s WHERE %s;`, format, fmtCondition(chanID, rpm))
+	}
+	fmt.Println("countquery:", countQuery)
```
rodneyosodo: Remove debug statemements

readers/timescale/messages.go
```diff
@@ -93,17 +122,36 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
...
+	fmt.Println("query:", q)
 
-	q = fmt.Sprintf(`SELECT COUNT(*) FROM %s WHERE %s;`, format, fmtCondition(chanID, rpm))
-	rows, err = tr.db.NamedQuery(q, params)
```
rodneyosodo, line 98: Revert this

rodneyosodo (changes requested):

pkg/sdk/go/sdk.go
ianmuchyri: This is not the same. The interval here refers to the time_bucket width: https://docs.timescale.com/api/latest/hyperfunctions/time_bucket/#required-arguments-for-interval-time-inputs

readers/api/transport.go
ianmuchyri: the non-empty should be interval in this case

readers/timescale/messages_test.go
```diff
@@ -514,7 +514,7 @@ func TestReadSenml(t *testing.T) {
 		result, err := reader.ReadAll(tc.chanID, tc.pageMeta)
 		assert.Nil(t, err, fmt.Sprintf("%s: expected no error got %s", tc.desc, err))
 		assert.ElementsMatch(t, tc.page.Messages, result.Messages, fmt.Sprintf("%s: expected %v got %v", tc.desc, tc.page.Messages, result.Messages))
-		assert.Equal(t, tc.page.Total, result.Total, fmt.Sprintf("%s: expected %v got %v", tc.desc, tc.page.Total, result.Total))
+		// assert.Equal(t, tc.page.Total, result.Total, fmt.Sprintf("%s: expected %v got %v", tc.desc, tc.page.Total, result.Total))
```
ianmuchyri: remove comments

readers/timescale/messages_test.go
```diff
@@ -652,7 +652,7 @@ func TestReadJSON(t *testing.T) {
 		result, err := reader.ReadAll(tc.chanID, tc.pageMeta)
 		assert.Nil(t, err, fmt.Sprintf("%s: expected no error got %s", desc, err))
 		assert.ElementsMatch(t, tc.page.Messages, result.Messages, fmt.Sprintf("%s: got incorrect list of json Messages from ReadAll()", desc))
-		assert.Equal(t, tc.page.Total, result.Total, fmt.Sprintf("%s: expected %v got %v", desc, tc.page.Total, result.Total))
+		// assert.Equal(t, tc.page.Total, result.Total, fmt.Sprintf("%s: expected %v got %v", desc, tc.page.Total, result.Total))
```
ianmuchyri: remove comments

ianmuchyri (changes requested):

readers/timescale/messages.go
```diff
@@ -36,8 +36,34 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
 		order = "created"
 		format = rpm.Format
 	}
-
-	q := fmt.Sprintf(`SELECT * FROM %s WHERE %s ORDER BY %s DESC LIMIT :limit OFFSET :offset;`, format, fmtCondition(chanID, rpm), order)
+	var q string
+	// If aggregation is provided, add time_bucket and aggregation to the query
+	if rpm.Aggregation != "" {
```
dborovcanin, line 44: Use switch-case instead of if-else.

readers/timescale/messages.go
```diff
@@ -93,17 +119,46 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
...
+			) AS subquery;`,
+			rpm.Interval,
+			rpm.Aggregation,
+			format,
+			fmtCondition(chanID, rpm),
+			order,
+		)
+	} else {
+		countQuery = fmt.Sprintf(`SELECT COUNT(*) FROM %s WHERE %s;`, format, fmtCondition(chanID, rpm))
+	}
```
dborovcanin: DRY. Also, offset and limit should not be used when calculating the total.

dborovcanin (changes requested):

readers/timescale/messages.go
```diff
@@ -93,17 +124,34 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
...
+			SELECT COUNT(*) 
+			FROM (
+				SELECT 
+				EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time))) AS time, 
+				%s(value) AS value 
+				%s
+			) AS subquery;`,
+			rpm.Interval,
+			rpm.Aggregation,
+			baseQuery,
```
ianmuchyri: The base query has a `order by` section that is not required in a count query

readers/timescale/messages.go
```diff
@@ -36,8 +36,39 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
...
+			:limit 
+			OFFSET 
+			:offset;`,
+			rpm.Interval,
+			rpm.Aggregation,
+			baseQuery,
+		)
+	default:
+		// Construct the base query without time_bucket and aggregation
+		q = fmt.Sprintf(`SELECT * FROM %s WHERE %s ORDER BY %s DESC LIMIT :limit OFFSET :offset;`, format, fmtCondition(chanID, rpm), order)
```
ianmuchyri: add base query to this query too

ianmuchyri (changes requested):

ianmuchyri (dismissed):

docker/.env
```diff
@@ -261,8 +261,8 @@ MG_UI_INSTANCE_ID=
 MG_UI_DB_HOST=ui-db
 MG_UI_DB_PORT=5432
 MG_UI_DB_USER=magistrala
-MG_UI_DB_PASS=magistrala
-MG_UI_DB_NAME=ui
+MG_UI_DB_PASS=magistrala-ui
```
rodneyosodo: revert this

docker/.env
```diff
@@ -261,8 +261,8 @@ MG_UI_INSTANCE_ID=
...
+MG_UI_DB_NAME=dashboards
```
rodneyosodo: Same as this

readers/api/transport.go
rodneyosodo: if the interval is empty and aggregation is not it should return an error for example. I want to get the maximum value of data values over a 1hr period in the last 24 hours.

readers/timescale/messages.go
```diff
@@ -36,8 +36,39 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
 		order = "created"
 		format = rpm.Format
 	}
-
-	q := fmt.Sprintf(`SELECT * FROM %s WHERE %s ORDER BY %s DESC LIMIT :limit OFFSET :offset;`, format, fmtCondition(chanID, rpm), order)
+	baseQuery := fmt.Sprintf(`FROM
```
rodneyosodo: Put it on one line

rodneyosodo (changes requested):

pkg/sdk/go/message.go
```diff
@@ -23,14 +25,14 @@ func (sdk mgSDK) SendMessage(chanName, msg, key string) errors.SDKError {
 		subtopicPart = fmt.Sprintf("/%s", strings.ReplaceAll(chanNameParts[1], ".", "/"))
 	}
 
-	url := fmt.Sprintf("%s/channels/%s/messages%s", sdk.httpAdapterURL, chanID, subtopicPart)
+	reqUrl := fmt.Sprintf("%s/channels/%s/messages%s", sdk.httpAdapterURL, chanID, subtopicPart)
```
rodneyosodo:
```suggestion
	reqURL := fmt.Sprintf("%s/channels/%s/messages%s", sdk.httpAdapterURL, chanID, subtopicPart)
```

readers/api/transport.go
```diff
@@ -141,6 +143,20 @@ func decodeList(_ context.Context, r *http.Request) (interface{}, error) {
...
+	if err != nil {
+		return nil, errors.Wrap(apiutil.ErrValidation, err)
+	}
+
+	interval, err := apiutil.ReadStringQuery(r, intervalKey, "")
+	if err != nil {
+		return nil, errors.Wrap(apiutil.ErrValidation, err)
+	}
+
+	if aggregation == "" || interval == "" {
```
rodneyosodo: This should be done in `requests.go`

readers/timescale/messages.go
```diff
@@ -31,13 +31,37 @@ func New(db *sqlx.DB) readers.MessageRepository {
...
 		format = rpm.Format
 	}
 
-	q := fmt.Sprintf(`SELECT * FROM %s WHERE %s ORDER BY %s DESC LIMIT :limit OFFSET :offset;`, format, fmtCondition(chanID, rpm), order)
+	var q string
+	// If aggregation is provided, add time_bucket and aggregation to the query
+	switch {
+	case rpm.Aggregation != "":
+		q = fmt.Sprintf(`
+			SELECT
```
rodneyosodo: make it inline

rodneyosodo (changes requested):

felixgateru (commented):
The epoch time attached to an aggregation query such as MAX and MIN for a value is not the time in which the MAX or MIN value occured.
Screenshot from 2024-02-29 11-31-49
Which is different from when the value occured:
image
It makes sense for aggregation to do this but there is a concern that the reason you are doing aggregation is to not only get the value but also the exact time within your desired interval in which the max occured.

internal/apiutil/errors.go
```diff
@@ -164,4 +164,7 @@ var (
 
 	// ErrRollbackTx indicates failed to rollback transaction.
 	ErrRollbackTx = errors.New("failed to rollback transaction")
+
+	// ErrAggregation indicates missing aggregation and interval.
+	ErrAggregation = errors.New("fill in aggregation, interval, to and from fields")
```
rodneyosodo: Create 4 errors one for missing aggregation another one for interval and the other for to and from values

```go
ErrMissingAggregation = errors.New("missing aggregation")
ErrMissingInterval    = errors.New("missing interval")
ErrMissingTo          = errors.New("missing to time")
ErrMissingFrom        = errors.New("missing from time")
```

readers/timescale/messages.go
```diff
@@ -93,17 +102,23 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
...
+		countQuery = fmt.Sprintf(`SELECT COUNT(*) FROM (SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time))) AS time, %s(value) AS value %s) AS subquery;`, rpm.Interval, rpm.Aggregation, baseQuery)
+	default:
+		countQuery = fmt.Sprintf(`SELECT COUNT(*) FROM %s WHERE %s;`, format, fmtCondition(chanID, rpm))
+	}
+	countRows, err := tr.db.NamedQuery(countQuery, params)
 	if err != nil {
 		return readers.MessagesPage{}, errors.Wrap(readers.ErrReadMessages, err)
 	}
-	defer rows.Close()
+	defer countRows.Close()
```
rodneyosodo: revert to rows

readers/api/requests.go
```diff
@@ -39,5 +39,10 @@ func (req listMessagesReq) validate() error {
 		return apiutil.ErrInvalidComparator
 	}
 
+	if (req.pageMeta.Aggregation != "" && req.pageMeta.Interval == "" && req.pageMeta.To == 0 && req.pageMeta.From == 0) ||
```
rodneyosodo: add a check for valid aggregation and interval values

readers/timescale/messages.go
```diff
@@ -31,13 +31,22 @@ func New(db *sqlx.DB) readers.MessageRepository {
 func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (readers.MessagesPage, error) {
 	order := "time"
 	format := defTable
+	baseQuery := fmt.Sprintf(`FROM %s WHERE %s GROUP BY 1`, format, fmtCondition(rpm))
```
rodneyosodo:
```suggestion
	baseQuery := fmt.Sprintf(`FROM %s WHERE %s GROUP BY time`, format, fmtCondition(rpm))
```

readers/timescale/messages.go
```diff
@@ -31,13 +31,22 @@ func New(db *sqlx.DB) readers.MessageRepository {
...
 		order = "created"
 		format = rpm.Format
 	}
 
-	q := fmt.Sprintf(`SELECT * FROM %s WHERE %s ORDER BY %s DESC LIMIT :limit OFFSET :offset;`, format, fmtCondition(chanID, rpm), order)
+	var q string
+	// If aggregation is provided, add time_bucket and aggregation to the query
+	switch {
+	case rpm.Aggregation != "":
+		q = fmt.Sprintf(`SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time))) AS time, %s(value) AS value %s ORDER BY %s DESC LIMIT :limit OFFSET :offset;`, rpm.Interval, rpm.Aggregation, baseQuery, order)
```
rodneyosodo:
```suggestion
		q = fmt.Sprintf(`SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000))) AS time, %s(value) AS value %s ORDER BY %s DESC LIMIT :limit OFFSET :offset;`, rpm.Interval, rpm.Aggregation, baseQuery, order)
```

readers/timescale/messages.go
```diff
@@ -93,17 +102,23 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
...
 		}
 	}
-
-	q = fmt.Sprintf(`SELECT COUNT(*) FROM %s WHERE %s;`, format, fmtCondition(chanID, rpm))
-	rows, err = tr.db.NamedQuery(q, params)
+	// countQuery is a string variable that holds the SQL query for counting total messages.
+	var countQuery string
+	switch {
+	case rpm.Aggregation != "":
+		countQuery = fmt.Sprintf(`SELECT COUNT(*) FROM (SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time))) AS time, %s(value) AS value %s) AS subquery;`, rpm.Interval, rpm.Aggregation, baseQuery)
```
rodneyosodo:
```suggestion
		countQuery = fmt.Sprintf(`SELECT COUNT(*) FROM (SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000))) AS time, %s(value) AS value %s) AS subquery;`, rpm.Interval, rpm.Aggregation, baseQuery)
```

rodneyosodo (changes requested):

internal/apiutil/errors.go
Musilah: its a single condition- I don't think it needs multiple errors when one serves as well.

rodneyosodo: You are combining multiple errors into one

rodneyosodo (changes requested):

rodneyosodo (changes requested):
Also update api documentation to include the 2 filters

readers/api/requests.go
```diff
@@ -39,5 +41,47 @@ func (req listMessagesReq) validate() error {
 		return apiutil.ErrInvalidComparator
 	}
 
+	validAggregations := map[string]bool{
```
ianmuchyri: we have 5 aggregation functions: https://www.simplilearn.com/tutorials/sql-tutorial/sql-aggregate-functions#:~:text=An%20aggregate%20function%20in%20SQL%20performs%20a%20calculation%20on%20multiple,except%20for%20the%20count%20function.

readers/api/requests.go
```diff
@@ -39,5 +41,47 @@ func (req listMessagesReq) validate() error {
...
+		"MAX": true,
+		"MIN": true,
+		"AVG": true,
+		"SUM": true,
+	}
+
+	_, aggregationValid := validAggregations[req.pageMeta.Aggregation]
```
ianmuchyri:
```suggestion
aggregationValid := validAggregations[req.pageMeta.Aggregation]
```

ianmuchyri (changes requested):

ianmuchyri (dismissed):

api/openapi/readers.yml
```diff
@@ -234,6 +236,34 @@ components:
       schema:
         type: number
       required: false
+    Aggregation:
+      name: aggregation
+      description: Aggregation function.
+      in: query
+      schema:
+        type: string
+        default: none
```
rodneyosodo: This is not a valid openapi file. Use swagger-editor to cross check.

api/openapi/readers.yml
```diff
@@ -234,6 +236,34 @@ components:
...
+          - avg
+          - count
+      required: false
+    Interval:
+      name: interval
+      description: Aggregation interval.
+      in: query
+      schema:
+        type: string
+        default: none
```
rodneyosodo: This is not the default parameter

readers/api/requests.go
```diff
@@ -39,5 +41,53 @@ func (req listMessagesReq) validate() error {
...
+		"COUNT": true,
+		"max":   true,
+		"min":   true,
+		"avg":   true,
+		"sum":   true,
+		"count": true,
+	}
+
+	aggregationValid := validAggregations[req.pageMeta.Aggregation]
+
```
rodneyosodo: Let's keep this simple to something like this

```go
	validAggregations := []string{"MAX", "MIN", "AVG", "SUM", "COUNT"}

	if req.pageMeta.Aggregation != "" {
		if !slices.Contains(validAggregations, strings.ToUpper(req.pageMeta.Aggregation)) {
			return apiutil.ErrInvalidAggregation
		}
	}
```
You can change  `validAggregations` to a global internal constant in the API package

readers/api/transport.go
```diff
@@ -141,6 +143,16 @@ func decodeList(_ context.Context, r *http.Request) (interface{}, error) {
 		return nil, errors.Wrap(apiutil.ErrValidation, err)
 	}
 
+	aggregation, err := apiutil.ReadStringQuery(r, aggregationKey, "")
+	if err != nil {
+		return nil, errors.Wrap(apiutil.ErrValidation, err)
+	}
+
+	interval, err := apiutil.ReadStringQuery(r, intervalKey, "")
```
rodneyosodo: Since interval and aggregation are tightly coupled we can have something like this, where we have a default interval value if the aggregation is present. This can be any value based on the generic usecase.
```go
	var interval string
	if aggregation != "" {
		interval, err = apiutil.ReadStringQuery(r, intervalKey, "1 day")
		if err != nil {
			return nil, errors.Wrap(apiutil.ErrValidation, err)
		}
	}
```

This will make validations simpler since if aggregation is present the interval definitely has a value.

```go
	if req.pageMeta.Aggregation != "" && req.pageMeta.To == 0 {
		return apiutil.ErrMissingTo
	}
```

readers/api/requests.go
```diff
@@ -39,5 +41,53 @@ func (req listMessagesReq) validate() error {
...
+	_, err := time.ParseDuration(req.pageMeta.Interval)
```
rodneyosodo: Why not something simpler like this

```go
	if _, err := time.ParseDuration(req.pageMeta.Interval); req.pageMeta.Interval != "" && err != nil {
		return apiutil.ErrInvalidInterval
	}
```

readers/timescale/messages.go
rodneyosodo: Why did you revert back?

readers/timescale/messages.go
```diff
@@ -31,13 +31,22 @@ func New(db *sqlx.DB) readers.MessageRepository {
...
 	format := defTable
+	baseQuery := fmt.Sprintf(`FROM %s WHERE %s GROUP BY 1`, format, fmtCondition(rpm))
 
 	if rpm.Format != "" && rpm.Format != defTable {
 		order = "created"
 		format = rpm.Format
 	}
 
-	q := fmt.Sprintf(`SELECT * FROM %s WHERE %s ORDER BY %s DESC LIMIT :limit OFFSET :offset;`, format, fmtCondition(chanID, rpm), order)
+	var q string
```
rodneyosodo: This can be made simpler 
```go
	q := fmt.Sprintf(`SELECT * FROM %s WHERE %s ORDER BY %s DESC LIMIT :limit OFFSET :offset;`, format, fmtCondition(rpm), order)
	totalQuery := fmt.Sprintf(`SELECT COUNT(*) FROM %s WHERE %s;`, format, fmtCondition(rpm))
	// If aggregation is provided, add time_bucket and aggregation to the query
	if rpm.Aggregation != "" {
		q = fmt.Sprintf(`SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000))) AS time, %s(value) AS value FROM %s WHERE %s GROUP BY time ORDER BY time DESC LIMIT :limit OFFSET :offset;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
		totalQuery = fmt.Sprintf(`SELECT COUNT(*) FROM (SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000))) AS time, %s(value) AS value FROM %s WHERE %s GROUP BY time) AS subquery;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
	}
```

rodneyosodo (changes requested):

readers/api/transport.go
rodneyosodo: @ianmuchyri what would be the ideal interval?

ianmuchyri: Lets go with the minimum we allow as the default, hence we can go with `1s`

readers/timescale/messages.go
rodneyosodo: Convert it to a single if statement

ianmuchyri (approved):

dborovcanin (approved):

PR #2087, opened 2024-02-20
NOISSUE - Add property based testing to users service

rodneyosodo:
# What type of PR is this?

This is a feature PR as it adds property-based testing to users API. In addition it is a refactor to error handling on the API

## What does this do?

Adds [schemathesis](https://github.com/schemathesis/schemathesis#getting-started), property-based testing to users service

No issue

No

### DIFF
--- a/.github/workflows/api-tests.yml
+++ b/.github/workflows/api-tests.yml
@@ -1,0 +1,54 @@
+# Copyright (c) Abstract Machines
+# SPDX-License-Identifier: Apache-2.0
+
+name: Property Based Tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+env:
+  TOKEN_URL: http://localhost:9002/users/tokens/issue
+  USER_IDENTITY: admin@example.com
+  USER_SECRET: 12345678
+
+jobs:
+  api-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+
+      - name: Install Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: 1.21.x
+          cache-dependency-path: "go.sum"
+
+      - name: Build images
+        run: make all -j $(nproc) && make dockers_dev -j $(nproc)
+
+      - name: Start containers
+        run: make run up args="-d" && sleep 10
+
+      - name: Set access token
+        run: |
+          export USER_TOKEN=$(curl -sSX POST $TOKEN_URL -H "Content-Type: application/json" -d "{\"identity\": \"$USER_IDENTITY\",\"secret\": \"$USER_SECRET\"}" | jq -r .access_token)
+          echo "USER_TOKEN=$USER_TOKEN" >> $GITHUB_ENV
+
+      - name: Run Users API tests
+        uses: schemathesis/action@v1
+        with:
+          schema: api/openapi/users.yml
+          base-url: http://localhost:9002
+          checks: all
+          report: false
+          args: '--header "Authorization: Bearer ${{ env.USER_TOKEN }}" --contrib-unique-data --contrib-openapi-formats-uuid --hypothesis-suppress-health-check=filter_too_much --stateful=links'
+
+      - name: Stop containers
+        if: always()
+        run: make run down args="-v"
--- a/.github/workflows/api-tests.yml
+++ b/.github/workflows/api-tests.yml
@@ -47,7 +47,7 @@ jobs:
           base-url: http://localhost:9002
           checks: all
           report: false
-          args: '--header "Authorization: Bearer ${{ env.USER_TOKEN }}" --contrib-unique-data --contrib-openapi-formats-uuid --hypothesis-suppress-health-check=filter_too_much --stateful=links'
+          args: '--header "Authorization: Bearer ${{ env.USER_TOKEN }}" --contrib-openapi-formats-uuid --hypothesis-suppress-health-check=filter_too_much --stateful=links'
 
       - name: Stop containers
         if: always()
--- a/.github/workflows/api-tests.yml
+++ b/.github/workflows/api-tests.yml
@@ -12,9 +12,11 @@ on:
       - main
 
 env:
-  TOKEN_URL: http://localhost:9002/users/tokens/issue
+  TOKENS_URL: http://localhost:9002/users/tokens/issue
+  DOMAINS_URL: http://localhost:8189/domains
   USER_IDENTITY: admin@example.com
   USER_SECRET: 12345678
+  DOMAIN_NAME: demo-test
 
 jobs:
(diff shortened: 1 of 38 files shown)


.github/workflows/api-tests.yml
```diff
@@ -0,0 +1,54 @@
...
+          echo "USER_TOKEN=$USER_TOKEN" >> $GITHUB_ENV
+
+      - name: Run Users API tests
+        uses: schemathesis/action@v1
+        with:
+          schema: api/openapi/users.yml
+          base-url: http://localhost:9002
+          checks: all
+          report: false
+          args: '--header "Authorization: Bearer ${{ env.USER_TOKEN }}" --contrib-unique-data --contrib-openapi-formats-uuid --hypothesis-suppress-health-check=filter_too_much --stateful=links'
```
Stranger6667: The `--contrib-unique-data` option is deprecated and going to be removed in the next major version of Schemathesis. I suggest avoiding it as it leads to unstable tests

drasko: @rodneyosodo can you please inspect this and see if we can remove `--contrib-unique-data` and what implications it will make (i.e. how we can find a future-proof approach)

drasko (changes requested):

api/openapi/users.yml
```diff
@@ -305,15 +362,20 @@ paths:
...
+        "415":
+          description: Missing or invalid content type.
+        "422":
+          description: Database can't process request.
         "500":
           $ref: "#/components/responses/ServiceError"
 
   /password/reset-request:
     post:
+      operationId: requestPasswordReset
```
arvindh123, line 378: `/password` is comes under the `/users` route in present `main` branch

users/api/clients.go
```diff
@@ -139,6 +126,22 @@ func clientsHandler(svc users.Service, r *chi.Mux, logger *slog.Logger, pr *rege
...
+		), "password_reset_req").ServeHTTP)
+
+		r.Put("/reset", otelhttp.NewHandler(kithttp.NewServer(
+			passwordResetEndpoint(svc),
+			decodePasswordReset,
+			api.EncodeResponse,
+			opts...,
+		), "password_reset").ServeHTTP)
+	})
+
```
arvindh123, line 144: Why we are going to change these routes for password?

arvindh123 (changes requested):
LGTM 

At [here](https://github.com/rodneyosodo/magistrala/blob/5e139ad19603ac4efa3d4b9d3f2bc148a8184683/users/postgres/clients.go#L78-L94) in users repository , we are using svcerr, instead of svcerr, does the repoerr will fit there ?

Makefile
```diff
@@ -129,6 +129,16 @@ test: mocks
...
 
+test_api:
+	@which st > /dev/null || (echo "schemathesis not found, please install it from https://github.com/schemathesis/schemathesis#getting-started" && exit 1)
+	st run api/openapi/users.yml \
+	--checks all \
+	--base-url http://localhost:9002 \
+	--header "Authorization: Bearer $(USER_TOKEN)" \
+	--contrib-openapi-formats-uuid \
+	--hypothesis-suppress-health-check=filter_too_much \
+	--stateful=links
```
arvindh123, line 144: It will be better if we impove the this command, like If USER_TOKEN is empty throw error or add mandatory required arguments username, password and domain to this command to get user_token

users/api/clients.go
rodneyosodo: From SDK it is under` /password` that is why I made the change to API docs and code base https://github.com/absmach/magistrala/blob/364749f28182d4e4a7a317e605f38c5c2b6aa222/pkg/sdk/go/users.go#L204

arvindh123 (dismissed):

arvindh123:
Got the below warnings from schemathesis in local and in CI,
Is it possible to provision Magistrala with domain and groups ? , so that schemathesis could test all the cases. 

```
WARNINGS:
  - Most of the responses from `GET /groups/{groupID}` have a 403 status code. Did you specify proper API credentials?
  - Most of the responses from `PUT /groups/{groupID}` have a 403 status code. Did you specify proper API credentials?
  - Most of the responses from `DELETE /groups/{groupID}` have a 403 status code. Did you specify proper API credentials?
  - Most of the responses from `POST /groups/{groupID}/enable` have a 403 status code. Did you specify proper API credentials?
  - Most of the responses from `POST /groups/{groupID}/disable` have a 403 status code. Did you specify proper API credentials?
  ```

File: internal/groups/service.go [hunk 5]
```diff
@@ -696,7 +694,7 @@ func (svc service) changeGroupStatus(ctx context.Context, token string, group gr
 		return groups.Group{}, err
 	}
 	if dbGroup.Status == group.Status {
-		return groups.Group{}, mgclients.ErrStatusAlreadyAssigned
+		return groups.Group{}, errors.ErrStatusAlreadyAssigned
```
arvindh123, line 697: @rodneyosodo, No need to change the following comments in this PR. 

We need to have some kind of error layers. 

Idea is to prevent or avoid sending of Reposistory errors in API response 

One of the way which I have in my mind to implement this idea is, reposistory should always return erorrs from pacakge `repoerr`
service should return errors from package `svcerr`, 

In API error encoder, We need to check for api errors and service errors and send only api errors and service errors in response and don't want to send repo errors in response. 

Example: 
for GET request of `things/<thingID>`
In response If entity is not found, then i should get error message of `svcerr.ErrNotFound`, not from repository layer, error return by postgres library.

In logs all the errors should be present. So if there we can get the more error details from service logs.

users/api/clients.go
rodneyosodo: We agreed and reverted to using non users prefix on password endpoint

Makefile
rodneyosodo: Updated with [`0f27a04` (#2087)](https://github.com/absmach/magistrala/pull/2087/commits/0f27a04f67d4f2202a948a7c8748c13fa67ffef7)

arvindh123 (approved):

rodneyosodo:
> Is it possible to provision Magistrala with domain and groups ? , so that schemathesis could test all the cases.

This is completely fine. Most of the endpoints schemathesis is probing are not valid hence we get an authorization error because the id is not for a valid entity

internal/groups/service.go
rodneyosodo: This can be done with something like this https://github.com/absmach/magistrala-ui/pull/240 whereby all errors from the repository layer must be wrapped with the service layer unless specified otherwise. But it is highly intrusive.

arvindh123: Created a issue [here](https://github.com/absmach/magistrala/issues/2117) for this comment

dborovcanin (approved):

dborovcanin (approved):

PR #2106, opened 2024-03-06
NOISSUE - Remove OAuth2.0 tokens from Magistrala token

rodneyosodo:
# What type of PR is this?

This is a refactor PR because it refactors how OAuth2.0 works in relation to magistrala

## What does this do?

Removes OAuth2.0 tokens from magistrala token

No issue

Yes, I have included tests for my changes and tested them manually.

Yes, I have added comments to the code.

### Notes

Demo https://jam.dev/c/84106a0d-f6a1-43b4-92af-23d7db977041

#### Explanation

From https://datatracker.ietf.org/doc/html/rfc6749#section-1.4
> Access tokens are credentials used to access protected resources.  An access token is a string representing an authorization issued to the client.

We only use this token once when accessing identity provider to fetch user details during registration or signup


auth/jwt/tokenizer.go
```diff
@@ -84,10 +84,18 @@ func (tok *tokenizer) Issue(key auth.Key) (string, error) {
 			return "", errors.Wrap(svcerr.ErrAuthentication, errInvalidProvider)
 		}
 		builder.Claim(oauthProviderField, provider.Name())
-		builder.Claim(provider.Name(), map[string]interface{}{
-			oauthAccessTokenField:  key.OAuth.AccessToken,
-			oauthRefreshTokenField: key.OAuth.RefreshToken,
-		})
+
+		// when issuing an access token, the refresh token is not present and vice versa
```
dborovcanin: Make this comment a full sentence.

auth/jwt/tokenizer.go
```diff
@@ -184,31 +192,24 @@ func parseOAuthToken(ctx context.Context, provider oauth2.Provider, token jwt.To
...
-			return auth.Key{}, errors.Wrap(ErrParseToken, fmt.Errorf("invalid access token claim for %s token", provider.Name()))
-		}
-		refreshToken, ok := claims[oauthRefreshTokenField].(string)
-		if !ok {
-			return auth.Key{}, errors.Wrap(ErrParseToken, fmt.Errorf("invalid refresh token claim for %s token", provider.Name()))
-		}
+
+		// access token and refresh token are not mandatory either of them can be present
+		accessToken, _ := claims[oauthAccessTokenField].(string)
+		refreshToken, _ := claims[oauthRefreshTokenField].(string)
```
dborovcanin: Fix according to [the comment](https://github.com/absmach/magistrala/pull/2106/commits/9ada86bd5ebcea212a71ada6d1d44d577275abda#r1516018971). Use only a single token (access or refresh):
```go
if ok {
		claims, ok := oauthToken.(pair)
		if !ok {
			return auth.Key{}, errors.Wrap(ErrParseToken, fmt.Errorf("invalid claims for %s token", provider.Name()))
		}

		// // access token and refresh token are not mandatory either of them can be present
		// accessToken, _ := claims[oauthAccessTokenField].(string)
		// refreshToken, _ := claims[oauthRefreshTokenField].(string)\
		switch claims.key {
		case oauthAccessTokenField:
			if err := provider.Validate(ctx, claims.value)
```
Also, fix the messy switch, since both cases do almost the same. Validation should inspect the error, since fallback to refresh may not make sense in some scenarios (revoked access, for example).

dborovcanin (changes requested):

auth/jwt/tokenizer.go
```diff
@@ -84,10 +84,18 @@ func (tok *tokenizer) Issue(key auth.Key) (string, error) {
...
 		}
 		builder.Claim(oauthProviderField, provider.Name())
-		builder.Claim(provider.Name(), map[string]interface{}{
-			oauthAccessTokenField:  key.OAuth.AccessToken,
-			oauthRefreshTokenField: key.OAuth.RefreshToken,
-		})
+
+		// when issuing an access token, the refresh token is not present and vice versa
+		if key.OAuth.AccessToken != "" && key.OAuth.RefreshToken == "" {
+			builder.Claim(provider.Name(), map[string]interface{}{
```
dborovcanin: Why don't we make this a `map[string]string` or a struct:
```go
type pair struct {
	key   string
	value string
}
```
So we have:
```go
		if key.OAuth.AccessToken != "" && key.OAuth.RefreshToken == "" {
			builder.Claim(provider.Name(), pair{
				key:   oauthAccessTokenField,
				value: key.OAuth.AccessToken,
			})
		}
		if key.OAuth.AccessToken == "" && key.OAuth.RefreshToken != "" {
			builder.Claim(provider.Name(), pair{
				key:   oauthRefreshTokenField,
				value: key.OAuth.RefreshToken,
			})
		}
```

dborovcanin (changes requested):
Simplify the Tokenizer `Parse` method. Extract validation, key parsing, and Oauth handling to three separate functions or methods.

users/service.go
```diff
@@ -587,11 +586,8 @@ func (svc service) OAuthCallback(ctx context.Context, provider string, state mgo
 			return &magistrala.Token{}, err
 		}
 		claims := &magistrala.IssueReq{
-			UserId:            rclient.ID,
-			Type:              0,
-			OauthProvider:     provider,
-			OauthAccessToken:  token.AccessToken,
-			OauthRefreshToken: token.RefreshToken,
+			UserId: rclient.ID,
+			Type:   0,
```
dborovcanin: Use constant/enum instead of literal.

dborovcanin (changes requested):

dborovcanin (approved):

PR #2109, opened 2024-03-08
NOISSUE - Add Secret Validation on Registration

rodneyosodo:
# What type of PR is this?

This is a bug fix because it fixes missing secret validation on registering users.

## What does this do?

Add secret validation on the service layer when registering users. Updated HTTP status codes and adjusted test cases for issue token and client creation validation with missing either identity or secret.

image

No documentation was needed.

### DIFF
--- a/users/api/requests.go
+++ b/users/api/requests.go
@@ -4,8 +4,6 @@
 package api
 
 import (
-	"regexp"
-
 	"github.com/absmach/magistrala/internal/api"
 	"github.com/absmach/magistrala/internal/apiutil"
 	mgclients "github.com/absmach/magistrala/pkg/clients"
@@ -18,7 +16,7 @@ type createClientReq struct {
 	token  string
 }
 
-func (req createClientReq) validate(passRegex *regexp.Regexp) error {
+func (req createClientReq) validate() error {
 	if len(req.client.Name) > api.MaxNameSize {
 		return apiutil.ErrNameSize
 	}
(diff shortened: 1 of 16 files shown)


users/service.go
```diff
@@ -84,6 +84,9 @@ func (svc service) RegisterClient(ctx context.Context, token string, cli mgclien
 	}
 
 	if cli.Credentials.Secret != "" {
+		if !svc.passRegex.MatchString(cli.Credentials.Secret) {
```
dborovcanin: Let's move this to API layer (validate request) and use `apiutil.ErrValidation`.

dborovcanin (changes requested):

users/service.go
rodneyosodo: It is repeated both in the API layer and the service layer. In the API layer, we check if it is empty, and then in the service layer, we check depending on the regex pattern.

dborovcanin: Let's group validation in the API layer.

dborovcanin (changes requested):

users/api/requests.go
```diff
@@ -16,10 +18,19 @@ type createClientReq struct {
 	token  string
 }
 
-func (req createClientReq) validate() error {
+func (req createClientReq) validate(passRegex *regexp.Regexp) error {
```
dborovcanin: Let's use package-level variable for regex that we set in `MakeHandler` and use it directly, rather than passing it as a parameter.

dborovcanin (approved):

PR #2115, opened 2024-03-14
NOISSUE - Convert Payload Time To UnixNano

### LINKED ISSUE #2113: Bug: Have a fixed unix timeformat precision in SenML

Send messages in different unix precisions and be able to process the time correctly

I should be able to accept different precision levels since different devices produce different time precision levels. But the time stored in the database should be in a fixed precision level i.e unixNano precision.

Data is stored in the database in different precision levels depending on the precision of the time sent in the message payload.

### Steps To Reproduce

1. Run latest magistrala
2. send messages in SenML format with the message time with different unix precisions. i.e unix() -10, unixNano() -19, unixMilli() - 13
3. Try carrying out SQL queries on the data, such as aggregation, and filter with to and from, they will be affected due to the precision level being different.

latest magistrala

_No response_

rodneyosodo:
# What type of PR is this?

This is a feature PR that adds the functionality to always convert absolute time to UNIX nano

## What does this do?

In the transformer, it ensures that absolute time consumed by consumers is always unixnano

- Resolves #2113 

### DIFF
--- a/pkg/transformers/senml/transformer.go
+++ b/pkg/transformers/senml/transformer.go
@@ -15,6 +15,8 @@ const (
 	JSON = "application/senml+json"
 	// CBOR represents SenML in CBOR format content type.
 	CBOR = "application/senml+cbor"
+
+	maxRelativeTime = 1 << 28
 )
 
 var (
@@ -64,10 +66,10 @@ func (t transformer) Transform(msg *messaging.Message) (interface{}, error) {
 
 		// If time is below 2**28 it is relative to the current time
 		// https://datatracker.ietf.org/doc/html/rfc8428#section-4.5.3
-		if t > 1<<28 {
+		if t >= maxRelativeTime {
 			t = transformers.ToUnixNano(t)
 		}
-		if v.UpdateTime > 1<<28 {
+		if v.UpdateTime >= maxRelativeTime {
 			v.UpdateTime = transformers.ToUnixNano(v.UpdateTime)
 		}
 
--- a/pkg/transformers/senml/transformer.go
+++ b/pkg/transformers/senml/transformer.go
@@ -59,8 +59,16 @@ func (t transformer) Transform(msg *messaging.Message) (interface{}, error) {
 		// Use reception timestamp if SenML messsage Time is missing
 		t := v.Time
 		if t == 0 {
-			// Convert the Unix timestamp in nanoseconds to float64
-			t = float64(msg.GetCreated()) / float64(1e9)
+			t = float64(msg.GetCreated())
+		}
+
+		// If time is below 2**28 it is relative to the current time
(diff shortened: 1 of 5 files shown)


pkg/transformers/senml/transformer.go
```diff
@@ -59,8 +59,16 @@ func (t transformer) Transform(msg *messaging.Message) (interface{}, error) {
...
 		t := v.Time
 		if t == 0 {
-			// Convert the Unix timestamp in nanoseconds to float64
-			t = float64(msg.GetCreated()) / float64(1e9)
+			t = float64(msg.GetCreated())
+		}
+
+		// If time is below 2**28 it is relative to the current time
+		// https://datatracker.ietf.org/doc/html/rfc8428#section-4.5.3
+		if t > 1<<28 {
```
dborovcanin: Extract `2**28` to constant.

pkg/transformers/transformer.go
```diff
@@ -10,3 +10,28 @@ type Transformer interface {
...
+	uint64 | int64 | float64
+}
+
+// ToUnixNano converts time to UnixNano time format.
+func ToUnixNano[N number](t N) N {
+	if t == 0 {
+		return 0
+	}
+	// Check if the value is in nanoseconds
+	if t > 1e18 {
```
dborovcanin: Should this be `>=`? We need either `>=` or `<=` in next cases to cover _equal_ scenario.

dborovcanin: By the way, it's perfectly fine to use literals here over constants. If nothing, it makes it more readable since you access values directly,

dborovcanin (changes requested):

pkg/transformers/transformer.go
```diff
@@ -10,3 +10,28 @@ type Transformer interface {
...
+	if t >= 1e15 && t < 1e18 {
+		return t * 1e3
+	}
+	// Check if the value is in microseconds
+	if t >= 1e12 && t < 1e15 {
+		return t * 1e6
+	}
+	// Assume it's in seconds (Unix time)
+	return t * 1e9
+}
```
dborovcanin: You can simplify with:
```suggestion
// ToUnixNano converts time to UnixNano time format.
func ToUnixNano[N number](t N) N {
	switch {
	case t == 0, t >= 1e18: // nanoseconds or 0
		return 0
	case 1e15 <= t && t < 1e18: // milliseconds
		return t * 1e3
	case 1e12 <= t && t < 1e15: // microseconds
		return t * 1e6
	default:
		return t * 1e9 // seconds (Unix time)
	}
}
```

dborovcanin (changes requested):

dborovcanin (approved):

PR #2119, opened 2024-03-19
MG-2117 - Remove repository errors from API layer

### LINKED ISSUE #1998: Refactoring Error Handling & Return Varaibles
**ENHANCEMENT**
   - code refactor
  
### Suggestion 1:
In a substantial portion of the codebase, functions tend to return the outcome of a call directly, neglecting the handling of potential errors stemming from these calls.
### Example:
```go
func (svc Service) Function1(param1 string) (client, error) {
   return svc.anotherService.Call(param1)
}
```

### Desired Solution:
Refactor these functions to incorporate error handling:
```go
func (svc Service) Function1(param1 string) (client, error) {
   client, err := svc.anotherService.Call(param1)
   if err != nil {
      return client{}, errors.Wrap(svcerr.CallErr, err) 
   }
   return client, nil 
}
```
### Suggestion 2:
Function with return struct/variable and error should return empty struct/variable on error return 

### Not looks function
```go
func (svc service) ServiceCall(param1 string) (client, error) {
    cli, err := svc.anotherServiceCall(param1)

    if err != nil {
        return cli, errors.Wrap(svcerr.RepoCall, err)
    }

    return cli, nil
}
```
### Looks good function
```go
func (svc service) ServiceCall(param1 string) (client, error) {
    cli, err := svc.anotherServiceCall(param1)

    if err != nil {
        return client{}, errors.Wrap(svcerr.RepoCall, err)
    }

    return cli, nil
}
```
must-have

### LINKED ISSUE #2117: Feature: Remove repository error in API response

I'm worried about getting Postgres error message in API response 

We need to have some kind of error layers.

Idea is to prevent or avoid sending of Reposistory errors in API response

One of the way which I have in my mind to implement this idea is, reposistory should always return erorrs from pacakge repoerr
service should return errors from package svcerr,

In API error encoder, We need to check for api errors and service errors and send only api errors and service errors in response and don't want to send repo errors in response.

Example:
for GET request of things/<thingID>
In response If entity is not found, then i should get error message of svcerr.ErrNotFound, not from repository layer, error return by postgres library. 

In logs all the errors should be present. So if there we can get the more error details from service logs.

We need to modify the api.EncodeError to something like below, so that actual error can be avoided to sending back to API response

```go 
func EncodeError(_ context.Context, err error, w http.ResponseWriter) {
	var wrapper error
	if errors.Contains(err, apiutil.ErrValidation) {
		wrapper, err = errors.Unwrap(err)
	}

	w.Header().Set("Content-Type", ContentType)

	switch {
	case errors.Contains(err, svcerr.ErrMalformedEntity):
		err = svcerr.ErrMalformedEntity

	case errors.Contains(err, errors.ErrMalformedEntity):
		err = errors.ErrMalformedEntity

	case errors.Contains(err, apiutil.ErrMissingID):
		err = apiutil.ErrMissingID

	case errors.Contains(err, apiutil.ErrEmptyList):
		err = apiutil.ErrEmptyList

	case errors.Contains(err, apiutil.ErrMissingMemberType):
		err = apiutil.ErrMissingMemberType

	case errors.Contains(err, apiutil.ErrMissingMemberKind):
		err = apiutil.ErrMissingMemberKind

	case errors.Contains(err, apiutil.ErrLimitSize):
		err = apiutil.ErrLimitSize

	case errors.Contains(err, apiutil.ErrBearerKey):
		err = apiutil.ErrBearerKey

	case errors.Contains(err, apiutil.ErrNameSize):
		err = apiutil.ErrNameSize

	case errors.Contains(err, svcerr.ErrInvalidStatus):
		err = svcerr.ErrInvalidStatus

	case errors.Contains(err, apiutil.ErrInvalidIDFormat):
		err = apiutil.ErrInvalidIDFormat

	case errors.Contains(err, apiutil.ErrInvalidQueryParams):
		err = apiutil.ErrInvalidQueryParams

	case errors.Contains(err, apiutil.ErrMissingRelation):
		err = apiutil.ErrMissingRelation

	case errors.Contains(err, apiutil.ErrValidation):
		err = apiutil.ErrValidation

	case errors.Contains(err, apiutil.ErrMissingIdentity):
		err = apiutil.ErrMissingIdentity

	case errors.Contains(err, apiutil.ErrMissingSecret):
		err = apiutil.ErrMissingSecret

	case errors.Contains(err, apiutil.ErrMissingPass):
		err = apiutil.ErrMissingPass

	case errors.Contains(err, apiutil.ErrMissingConfPass):
		err = apiutil.ErrMissingConfPass

	case errors.Contains(err, apiutil.ErrPasswordFormat):
		err = apiutil.ErrPasswordFormat

	case errors.Contains(err, svcerr.ErrAuthentication):
		err = svcerr.ErrAuthentication

	case errors.Contains(err, apiutil.ErrBearerToken):
		err = apiutil.ErrBearerToken

	case errors.Contains(err, svcerr.ErrNotFound):
		err = svcerr.ErrNotFound
		w.WriteHeader(http.StatusNotFound)

	case errors.Contains(err, postgres.ErrMemberAlreadyAssigned):
		err = postgres.ErrMemberAlreadyAssigned
		w.WriteHeader(http.StatusConflict)

	case errors.Contains(err, svcerr.ErrConflict):
		err = svcerr.ErrConflict
		w.WriteHeader(http.StatusConflict)

	case errors.Contains(err, svcerr.ErrAuthorization):
		err = svcerr.ErrAuthorization

	case errors.Contains(err, svcerr.ErrDomainAuthorization):
		err = svcerr.ErrDomainAuthorization

	case errors.Contains(err, apiutil.ErrUnsupportedContentType):
		err = apiutil.ErrUnsupportedContentType
		w.WriteHeader(http.StatusUnsupportedMediaType)

	case errors.Contains(err, svcerr.ErrCreateEntity):
		err = svcerr.ErrCreateEntity

	case errors.Contains(err, svcerr.ErrUpdateEntity):
		err = svcerr.ErrUpdateEntity

	case errors.Contains(err, svcerr.ErrViewEntity):
		err = svcerr.ErrViewEntity

	case errors.Contains(err, svcerr.ErrRemoveEntity):
		err = svcerr.ErrRemoveEntity

	default:
	}

	if wrapper != nil {
		err = errors.Wrap(wrapper, err)
	}

		if err := json.NewEncoder(w).Encode(errorVal); err != nil {
		}
	}
}```
https://github.com/absmach/magistrala/pull/2087#discussion_r1525990456

Must-have

### Anything else?

_No response_

### PR DESCRIPTION
**Pull request by WashingtonKK**
# What type of PR is this?

This is a feature that removes removes repository errors from the api response and only the service error and api errors are sent as part of the response. All errors are logged and can be evaluated from the service logs.

## What does this do?

Ensures only errors from the service and api are sent as part of the API response, and leaves repo errors at service level.

- Resolves #2117
- Related to #1998 

No

No

### DIFF
--- a/api/openapi/bootstrap.yml
+++ b/api/openapi/bootstrap.yml
@@ -96,6 +96,8 @@ paths:
       responses:
         "200":
           $ref: "#/components/responses/ConfigRes"
+        "400":
+          description: Missing or invalid config.
         "401":
           description: Missing or invalid access token provided.
         "404":
--- a/auth/service.go
+++ b/auth/service.go
@@ -139,7 +139,7 @@ func (svc service) Identify(ctx context.Context, token string) (Key, error) {
 	key, err := svc.tokenizer.Parse(token)
 	if errors.Contains(err, ErrExpiry) {
 		err = svc.keys.Remove(ctx, key.Issuer, key.ID)
-		return Key{}, errors.Wrap(ErrKeyExpired, err)
+		return Key{}, errors.Wrap(svcerr.ErrAuthentication, errors.Wrap(ErrKeyExpired, err))
 	}
 	if err != nil {
 		return Key{}, errors.Wrap(svcerr.ErrAuthentication, errors.Wrap(errIdentify, err))
@@ -186,14 +186,7 @@ func (svc service) Authorize(ctx context.Context, pr PolicyReq) error {
 func (svc service) checkPolicy(ctx context.Context, pr PolicyReq) error {
 	// Domain status is required for if user sent authorization request on things, channels, groups and domains
 	if pr.SubjectType == UserType && (pr.ObjectType == GroupType || pr.ObjectType == ThingType || pr.ObjectType == DomainType) {
-		domainID := pr.Domain
-		if domainID == "" {
-			if pr.ObjectType != DomainType {
-				return svcerr.ErrDomainAuthorization
-			}
-			domainID = pr.Object
-		}
-		if err := svc.checkDomain(ctx, pr.SubjectType, pr.Subject, domainID); err != nil {
+		if err := svc.checkDomain(ctx, pr.SubjectType, pr.Subject, pr.Domain); err != nil {
 			return err
 		}
 	}
@@ -204,6 +197,16 @@ func (svc service) checkPolicy(ctx context.Context, pr PolicyReq) error {
 }
 
 func (svc service) checkDomain(ctx context.Context, subjectType, subject, domainID string) error {
+	if err := svc.agent.CheckPolicy(ctx, PolicyReq{
+		Subject:     subject,
+		SubjectType: subjectType,
+		Permission:  MembershipPermission,
+		Object:      domainID,
+		ObjectType:  DomainType,
+	}); err != nil {
+		return svcerr.ErrDomainAuthorization
+	}
+
 	d, err := svc.domains.RetrieveByID(ctx, domainID)
 	if err != nil {
 		return errors.Wrap(svcerr.ErrViewEntity, err)
@@ -531,7 +534,7 @@ func (svc service) CreateDomain(ctx context.Context, token string, d Domain) (do
 
 	domainID, err := svc.idProvider.ID()
 	if err != nil {
-		return Domain{}, err
+		return Domain{}, errors.Wrap(svcerr.ErrCreateEntity, err)
 	}
 	d.ID = domainID
 
@@ -580,7 +583,7 @@ func (svc service) RetrieveDomain(ctx context.Context, token, id string) (Domain
 func (svc service) RetrieveDomainPermissions(ctx context.Context, token, id string) (Permissions, error) {
 	res, err := svc.Identify(ctx, token)
 	if err != nil {
-		return []string{}, errors.Wrap(svcerr.ErrAuthentication, err)
+		return []string{}, err
 	}
 
 	if err := svc.Authorize(ctx, PolicyReq{
@@ -591,7 +594,7 @@ func (svc service) RetrieveDomainPermissions(ctx context.Context, token, id stri
 		ObjectType:  DomainType,
 		Permission:  MembershipPermission,
 	}); err != nil {
-		return []string{}, errors.Wrap(svcerr.ErrAuthorization, err)
+		return []string{}, err
 	}
 
 	lp, err := svc.ListPermissions(ctx, PolicyReq{
@@ -609,7 +612,7 @@ func (svc service) RetrieveDomainPermissions(ctx context.Context, token, id stri
 func (svc service) UpdateDomain(ctx context.Context, token, id string, d DomainReq) (Domain, error) {
 	key, err := svc.Identify(ctx, token)
 	if err != nil {
-		return Domain{}, errors.Wrap(svcerr.ErrAuthentication, err)
+		return Domain{}, err
 	}
 	if err := svc.Authorize(ctx, PolicyReq{
 		Subject:     key.Subject,
@@ -619,7 +622,7 @@ func (svc service) UpdateDomain(ctx context.Context, token, id string, d DomainR
 		ObjectType:  DomainType,
 		Permission:  EditPermission,
 	}); err != nil {
-		return Domain{}, errors.Wrap(svcerr.ErrAuthorization, err)
+		return Domain{}, err
 	}
 
 	dom, err := svc.domains.Update(ctx, id, key.User, d)
@@ -642,7 +645,7 @@ func (svc service) ChangeDomainStatus(ctx context.Context, token, id string, d D
 		ObjectType:  DomainType,
 		Permission:  AdminPermission,
 	}); err != nil {
-		return Domain{}, errors.Wrap(svcerr.ErrAuthorization, err)
+		return Domain{}, err
 	}
 
 	dom, err := svc.domains.Update(ctx, id, key.User, d)
@@ -765,7 +768,7 @@ func (svc service) UnassignUsers(ctx context.Context, token, id string, userIds 
 	for _, rel := range []string{MemberRelation, ViewerRelation, EditorRelation} {
 		// Remove only non-admins.
 		if err := svc.removeDomainPolicies(ctx, id, rel, userIds...); err != nil {
-			return errors.Wrap(errRemovePolicies, err)
+			return err
 		}
 	}
 
--- a/auth/service.go
+++ b/auth/service.go
@@ -186,7 +186,14 @@ func (svc service) Authorize(ctx context.Context, pr PolicyReq) error {
 func (svc service) checkPolicy(ctx context.Context, pr PolicyReq) error {
 	// Domain status is required for if user sent authorization request on things, channels, groups and domains
 	if pr.SubjectType == UserType && (pr.ObjectType == GroupType || pr.ObjectType == ThingType || pr.ObjectType == DomainType) {
-		if err := svc.checkDomain(ctx, pr.SubjectType, pr.Subject, pr.Domain); err != nil {
+		domainID := pr.Domain
+		if domainID == "" {
+			if pr.ObjectType != DomainType {
+				return svcerr.ErrDomainAuthorization
+			}
+			domainID = pr.Object
+		}
+		if err := svc.checkDomain(ctx, pr.SubjectType, pr.Subject, domainID); err != nil {
 			return err
 		}
 	}
--- a/auth/service_test.go
+++ b/auth/service_test.go
@@ -215,6 +215,7 @@ func TestIssue(t *testing.T) {
 				SubjectType: auth.UserType,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
+				Object:      groupName,
 			},
 			checkPolicyErr:       repoerr.ErrNotFound,
 			retrieveByIDResponse: auth.Domain{},
@@ -253,6 +254,7 @@ func TestIssue(t *testing.T) {
 				Permission:  auth.MembershipPermission,
 			},
 			checkPolicyErr:       svcerr.ErrAuthorization,
+			checkPolicyErr1:      svcerr.ErrAuthorization,
 			retrieveByIDResponse: auth.Domain{Status: auth.EnabledStatus},
 			err:                  svcerr.ErrAuthorization,
 		},
@@ -288,7 +290,7 @@ func TestIssue(t *testing.T) {
 				Permission:  auth.MembershipPermission,
 			},
 			checkPolicyErr:       svcerr.ErrAuthorization,
-			checkPolicyErr1:      nil,
+			checkPolicyErr1:      svcerr.ErrAuthorization,
 			retrieveByIDResponse: auth.Domain{Status: auth.EnabledStatus},
 			err:                  svcerr.ErrAuthorization,
 		},
@@ -443,6 +445,7 @@ func TestIssue(t *testing.T) {
 				Subject:     "mgx_test@example.com",
 				SubjectType: auth.UserType,
 				SubjectKind: "",
+				Object:      groupName,
 				ObjectKind:  "",
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
@@ -522,6 +525,7 @@ func TestIssue(t *testing.T) {
 			},
 			checkPolicyRequest1: auth.PolicyReq{
 				SubjectType: auth.UserType,
+				Object:      groupName,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
@@ -876,8 +880,9 @@ func TestAuthorize(t *testing.T) {
 			checkPolicyReq: auth.PolicyReq{
 				Subject:     id,
 				SubjectType: auth.UserType,
+				Object:      validID,
 				ObjectType:  auth.DomainType,
-				Permission:  auth.AdminPermission,
+				Permission:  auth.MembershipPermission,
 			},
 			checkPolicyReq1: auth.PolicyReq{
 				Subject:     id,
@@ -890,8 +895,9 @@ func TestAuthorize(t *testing.T) {
 			checkPolicyReq2: auth.PolicyReq{
 				Subject:     id,
 				SubjectType: auth.UserType,
+				Object:      validID,
 				ObjectType:  auth.DomainType,
-				Permission:  auth.MembershipPermission,
+				Permission:  auth.AdminPermission,
 			},
 
 			retrieveDomainRes: auth.Domain{
@@ -928,6 +934,7 @@ func TestAuthorize(t *testing.T) {
 			checkPolicyReq2: auth.PolicyReq{
 				Subject:     id,
 				SubjectType: auth.UserType,
+				Object:      validID,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
@@ -968,6 +975,7 @@ func TestAuthorize(t *testing.T) {
 			checkPolicyReq2: auth.PolicyReq{
 				Subject:     id,
 				SubjectType: auth.UserType,
+				Object:      validID,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
@@ -1007,6 +1015,7 @@ func TestAuthorize(t *testing.T) {
 			checkPolicyReq2: auth.PolicyReq{
 				Subject:     id,
 				SubjectType: auth.UserType,
+				Object:      validID,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
@@ -1047,6 +1056,7 @@ func TestAuthorize(t *testing.T) {
 			checkPolicyReq2: auth.PolicyReq{
 				Subject:     id,
 				SubjectType: auth.UserType,
+				Object:      validID,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
@@ -1080,6 +1090,7 @@ func TestAuthorize(t *testing.T) {
 			checkPolicyReq2: auth.PolicyReq{
 				Subject:     id,
 				SubjectType: auth.UserType,
+				Object:      validID,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
@@ -1104,6 +1115,7 @@ func TestAuthorize(t *testing.T) {
 			checkPolicyReq2: auth.PolicyReq{
 				Subject:     id,
 				SubjectType: auth.UserType,
+				Object:      validID,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
@@ -1128,6 +1140,7 @@ func TestAuthorize(t *testing.T) {
 			checkPolicyReq2: auth.PolicyReq{
 				Subject:     id,
 				SubjectType: auth.UserType,
+				Object:      validID,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
@@ -1153,6 +1166,7 @@ func TestAuthorize(t *testing.T) {
 			checkPolicyReq2: auth.PolicyReq{
 				Subject:     id,
 				SubjectType: auth.UserType,
+				Object:      validID,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
@@ -1177,6 +1191,7 @@ func TestAuthorize(t *testing.T) {
 			checkPolicyReq2: auth.PolicyReq{
 				Subject:     id,
 				SubjectType: auth.UserType,
+				Object:      validID,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
@@ -1184,6 +1199,7 @@ func TestAuthorize(t *testing.T) {
 		},
 	}
 	for _, tc := range cases {
+		fmt.Println(tc.desc)
 		repoCall := prepo.On("CheckPolicy", mock.Anything, tc.checkPolicyReq).Return(tc.checkPolicyErr)
 		repoCall1 := drepo.On("RetrieveByID", mock.Anything, mock.Anything).Return(tc.retrieveDomainRes, nil)
 		repoCall2 := prepo.On("CheckPolicy", mock.Anything, tc.checkPolicyReq1).Return(tc.checkPolicyErr1)
@@ -1197,7 +1213,7 @@ func TestAuthorize(t *testing.T) {
 		repoCall3.Unset()
 		repoCall4.Unset()
 	}
-
+	fmt.Println("Cases 1 end")
 	cases2 := []struct {
 		desc      string
 		policyReq auth.PolicyReq
--- a/auth/service_test.go
+++ b/auth/service_test.go
@@ -133,19 +133,19 @@ func TestIssue(t *testing.T) {
 	}
 
 	cases2 := []struct {
-		desc                 string
-		key                  auth.Key
-		saveResponse         auth.Key
-		retrieveByIDResponse auth.Domain
-		token                string
-		saveErr              error
-		checkPolicyRequest   auth.PolicyReq
-		checkPolicyRequest1  auth.PolicyReq
-		checkPolicyRequest2  auth.PolicyReq
-		checkPolicyErr       error
-		checkPolicyErr1      error
-		retreiveByIDErr      error
-		err                  error
+		desc                   string
+		key                    auth.Key
+		saveResponse           auth.Key
+		retrieveByIDResponse   auth.Domain
+		token                  string
+		saveErr                error
+		checkPolicyReq3uest     auth.PolicyReq
+		checkPlatformPolicyReq auth.PolicyReq
+		checkDomainPolicyReq   auth.PolicyReq
+		checkPolicyErr         error
+		checkPolicyErr1        error
+		retreiveByIDErr        error
+		err                    error
 	}{
 		{
 			desc: "issue login key",
@@ -153,16 +153,13 @@ func TestIssue(t *testing.T) {
 				Type:     auth.AccessKey,
 				IssuedAt: time.Now(),
 			},
-			checkPolicyRequest: auth.PolicyReq{
-				Subject:     "",
+			checkPolicyReq3uest: auth.PolicyReq{
 				SubjectType: auth.UserType,
-				SubjectKind: "",
 				Object:      auth.MagistralaObject,
-				ObjectKind:  "",
 				ObjectType:  auth.PlatformType,
 				Permission:  auth.AdminPermission,
 			},
-			checkPolicyRequest1: auth.PolicyReq{
+			checkDomainPolicyReq: auth.PolicyReq{
 				SubjectType: auth.UserType,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
@@ -177,16 +174,13 @@ func TestIssue(t *testing.T) {
 				IssuedAt: time.Now(),
 				Domain:   groupName,
 			},
-			checkPolicyRequest: auth.PolicyReq{
-				Subject:     "",
+			checkPolicyReq3uest: auth.PolicyReq{
 				SubjectType: auth.UserType,
-				SubjectKind: "",
 				Object:      auth.MagistralaObject,
-				ObjectKind:  "",
 				ObjectType:  auth.PlatformType,
 				Permission:  auth.AdminPermission,
 			},
-			checkPolicyRequest1: auth.PolicyReq{
+			checkDomainPolicyReq: auth.PolicyReq{
 				SubjectType: auth.UserType,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
@@ -202,16 +196,13 @@ func TestIssue(t *testing.T) {
 				Domain:   groupName,
 			},
 			token: accessToken,
-			checkPolicyRequest: auth.PolicyReq{
-				Subject:     "",
+			checkPolicyReq3uest: auth.PolicyReq{
 				SubjectType: auth.UserType,
-				SubjectKind: "",
 				Object:      auth.MagistralaObject,
-				ObjectKind:  "",
 				ObjectType:  auth.PlatformType,
 				Permission:  auth.AdminPermission,
 			},
-			checkPolicyRequest1: auth.PolicyReq{
+			checkPlatformPolicyReq: auth.PolicyReq{
 				SubjectType: auth.UserType,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
@@ -230,25 +221,19 @@ func TestIssue(t *testing.T) {
 				Domain:   groupName,
 			},
 			token: accessToken,
-			checkPolicyRequest: auth.PolicyReq{
-				Subject:     "",
+			checkPolicyReq3uest: auth.PolicyReq{
 				SubjectType: auth.UserType,
-				SubjectKind: "",
 				Object:      auth.MagistralaObject,
-				ObjectKind:  "",
 				ObjectType:  auth.PlatformType,
 				Permission:  auth.AdminPermission,
 			},
-			checkPolicyRequest1: auth.PolicyReq{
-				Subject:     "",
+			checkPlatformPolicyReq: auth.PolicyReq{
 				SubjectType: auth.UserType,
-				SubjectKind: "",
 				Object:      groupName,
-				ObjectKind:  "",
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
-			checkPolicyRequest2: auth.PolicyReq{
+			checkDomainPolicyReq: auth.PolicyReq{
 				SubjectType: auth.UserType,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
@@ -266,25 +251,19 @@ func TestIssue(t *testing.T) {
 				Domain:   groupName,
 			},
 			token: accessToken,
-			checkPolicyRequest: auth.PolicyReq{
-				Subject:     "",
+			checkPolicyReq3uest: auth.PolicyReq{
 				SubjectType: auth.UserType,
-				SubjectKind: "",
 				Object:      auth.MagistralaObject,
-				ObjectKind:  "",
 				ObjectType:  auth.PlatformType,
 				Permission:  auth.AdminPermission,
 			},
-			checkPolicyRequest1: auth.PolicyReq{
-				Subject:     "",
+			checkPlatformPolicyReq: auth.PolicyReq{
 				SubjectType: auth.UserType,
-				SubjectKind: "",
 				Object:      groupName,
-				ObjectKind:  "",
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
-			checkPolicyRequest2: auth.PolicyReq{
+			checkDomainPolicyReq: auth.PolicyReq{
 				SubjectType: auth.UserType,
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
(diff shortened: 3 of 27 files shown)

### CODE REVIEW

File: users/service.go [hunk 1]
```diff
@@ -64,10 +44,10 @@ func (svc service) RegisterClient(ctx context.Context, token string, cli mgclien
 	if !svc.selfRegister {
 		userID, err := svc.Identify(ctx, token)
 		if err != nil {
-			return mgclients.Client{}, err
+			return mgclients.Client{}, errors.Wrap(svcerr.ErrAuthorization, err)
```
arvindh123: In Line 45 `svc.Identify(ctx, token)` already wraps `svcerr.ErrAuthorization`, so need to wrap it again here 
If the functions call is goes out the package then wrapping is required, function call is within the package, the called function should wrap right  errors

users/service.go
```diff
@@ -64,10 +44,10 @@ func (svc service) RegisterClient(ctx context.Context, token string, cli mgclien
...
 		}
 		if err := svc.checkSuperAdmin(ctx, userID); err != nil {
-			return mgclients.Client{}, err
+			return mgclients.Client{}, errors.Wrap(svcerr.ErrAuthorization, err)
```
arvindh123: `svc.checkSuperAdmin(ctx, userID)` function wraps `svcerr.ErrAuthorization`, so need to wrap it again here 
If the functions call is goes out the package then wrapping is required, function call is within the package, the called function should wrap right  errors

users/service.go
```diff
@@ -94,7 +74,7 @@ func (svc service) RegisterClient(ctx context.Context, token string, cli mgclien
 	cli.CreatedAt = time.Now()
 
 	if err := svc.addClientPolicy(ctx, cli.ID, cli.Role); err != nil {
-		return mgclients.Client{}, err
+		return mgclients.Client{}, errors.Wrap(svcerr.ErrInvalidPolicy, err)
```
arvindh123: same here

users/service.go
```diff
@@ -517,7 +513,7 @@ func (svc service) retrieveObjectUsersPermissions(ctx context.Context, domainID,
 	userID := auth.EncodeDomainUserID(domainID, client.ID)
 	permissions, err := svc.listObjectUserPermission(ctx, userID, objectType, objectID)
 	if err != nil {
-		return err
+		return errors.Wrap(svcerr.ErrFailedPermissionsList, err)
```
arvindh123: From user prespective , this can be a internal error.
@WashingtonKK What do you think? Please give your comments

WashingtonKK: Yes, it needs to be an internal error but we still need to notify the user of an error.

We can change it to an errAuthorization instead.
What do you think about this @arvindh123 ?

users/service.go
```diff
@@ -581,9 +577,9 @@ func (svc service) OAuthCallback(ctx context.Context, state mgoauth2.State, clie
 		rclient, err := svc.clients.RetrieveByIdentity(ctx, client.Credentials.Identity)
 		if err != nil {
 			if errors.Contains(err, repoerr.ErrNotFound) {
-				return &magistrala.Token{}, errors.New("user not signed up")
+				return &magistrala.Token{}, errUserNotSignedUp
```
arvindh123: This one could not be a internal error,we can simply return  ErrAuthorization
@WashingtonKK Please give your comments

WashingtonKK: Here it is better to notify that the user is not signed up so that the user can sign up.

WashingtonKK: I think it is better to return errUserNotSignedUp here @arvindh123

users/service.go
arvindh123: I see  `svc.listObjectUserPermission` already wrap with error `svcerr.ErrFailedPermissionsList`, we can make `svcerr.ErrFailedPermissionsList` as internal error and return `svcerr.ErrAuthroization` or something like denied

users/service.go
arvindh123: This endpoint presents a potential risk of user data leakage, as unauthorized individuals could potentially scrape data by sending requests to it. If they encounter an authorization error, they can confirm whether a user's email is registered within the system. Conversely, if the endpoint returns "UserNotSignedUp," they can deduce that the user is not registered.

users/service.go
WashingtonKK: This make sense, okay

users/service.go
WashingtonKK: Okay, let me wrap the error with errAuthorization so that userNotSignedUp can remain internal

auth/service.go
```diff
@@ -656,7 +656,7 @@ func (svc service) ChangeDomainStatus(ctx context.Context, token, id string, d D
 		ObjectType:  DomainType,
 		Permission:  AdminPermission,
 	}); err != nil {
-		return Domain{}, errors.Wrap(svcerr.ErrAuthorization, err)
+		return Domain{}, err
```
arvindh123, line 659: `svc.Authorize` go#L181-L182  this function belongs to Auth service and It is exported function, 
So this function will return error with svcerr wrapped. 
Reference: https://github.com/WashingtonKK/magistrala/blob/81328c217c29766674966f63cb4b834fed586294/auth/service.
So here there is no need to warp again here `svcerr`

auth/service.go
```diff
@@ -633,7 +633,7 @@ func (svc service) UpdateDomain(ctx context.Context, token, id string, d DomainR
 		ObjectType:  DomainType,
 		Permission:  EditPermission,
 	}); err != nil {
-		return Domain{}, errors.Wrap(svcerr.ErrAuthorization, err)
+		return Domain{}, err
```
arvindh123, line 636: `svc.Authorize`  this function belongs to Auth service and It is exported function, 
So this function will return error with svcerr wrapped.   
Reference: https://github.com/WashingtonKK/magistrala/blob/81328c217c29766674966f63cb4b834fed586294/auth/service.go#L181-L182     
So here there is no need to warp again here `svcerr`

auth/service.go
```diff
@@ -605,7 +605,7 @@ func (svc service) RetrieveDomainPermissions(ctx context.Context, token, id stri
 		ObjectType:  DomainType,
 		Permission:  MembershipPermission,
 	}); err != nil {
-		return []string{}, errors.Wrap(svcerr.ErrAuthorization, err)
+		return []string{}, err
```
arvindh123, line 608: `svc.Authorize`  this function belongs to Auth service and It is exported function, 
So this function will return error with svcerr wrapped.   
Reference: https://github.com/WashingtonKK/magistrala/blob/81328c217c29766674966f63cb4b834fed586294/auth/service.go#L181-L182     
So here there is no need to warp again here `svcerr`

things/service.go
```diff
@@ -125,7 +119,7 @@ func (svc service) CreateThings(ctx context.Context, token string, cls ...mgclie
 		})
 	}
 	if _, err := svc.auth.AddPolicies(ctx, &policies); err != nil {
-		return nil, errors.Wrap(errAddPolicies, err)
+		return nil, errors.Wrap(svcerr.ErrAddPolicies, err)
```
arvindh123: I think, this can be an internal error. The user won't doesn't need information about backend errors.

users/service.go
```diff
@@ -704,10 +701,10 @@ func (svc service) updateClientPolicy(ctx context.Context, userID string, role m
 			Object:      auth.MagistralaObject,
 		})
 		if err != nil {
-			return errors.Wrap(errDeletePolicies, err)
+			return errors.Wrap(svcerr.ErrDeletePolicies, err)
```
arvindh123: Same here, I think for user information about policy error might not be useful.

auth/service_test.go
```diff
@@ -1975,7 +1975,7 @@ func TestChangeDomainStatus(t *testing.T) {
 				Status: &disabledStatus,
 			},
 			retreieveByIDErr: repoerr.ErrNotFound,
-			err:              svcerr.ErrAuthorization,
+			err:              svcerr.ErrNotFound,
```
arvindh123, line 2059: @WashingtonKK, Why we are changing this error type?

arvindh123: If the domain is empty or not avialable with MG, then it should return Authorization Error, 
We should not return 404 , So I'm suggesting following changes in two functions 


1.) 

https://github.com/washingtonkk/magistrala/blob/b50c99851ec8b3bbb23b26476896c1aad8f12fde/auth/service.go#L187-L205
Suggestion  for the above function 
```go
func (svc service) checkPolicy(ctx context.Context, pr PolicyReq) error {
	// Domain status is required for if user sent authorization request on things, channels, groups and domains
	if pr.SubjectType == UserType && (pr.ObjectType == GroupType || pr.ObjectType == ThingType || pr.ObjectType == DomainType) {
		if err := svc.checkDomain(ctx, pr.SubjectType, pr.Subject, pr.Domain); err != nil {
			return err
		}
	}
	if err := svc.agent.CheckPolicy(ctx, pr); err != nil {
	}
	return nil
}
```

2.) 
https://github.com/washingtonkk/magistrala/blob/b50c99851ec8b3bbb23b26476896c1aad8f12fde/auth/service.go#L207-L240
Suggestion for above function 

```
func (svc service) checkDomain(ctx context.Context, subjectType, subject, domainID string) error {
	if err := svc.agent.CheckPolicy(ctx, PolicyReq{
		Subject:     subject,
		SubjectType: subjectType,
		Permission:  MembershipPermission,
		Object:      domainID,
		ObjectType:  DomainType,
	}); err != nil {
		return svcerr.ErrDomainAuthorization
	}

	d, err := svc.domains.RetrieveByID(ctx, domainID)
	if err != nil {
		return errors.Wrap(errors.ErrUnidentified, err)
	}

	switch d.Status {
	case EnabledStatus:
	case DisabledStatus:
		if err := svc.agent.CheckPolicy(ctx, PolicyReq{
			Subject:     subject,
			SubjectType: subjectType,
			Permission:  AdminPermission,
			Object:      domainID,
			ObjectType:  DomainType,
		}); err != nil {
			return svcerr.ErrDomainAuthorization
		}
	case FreezeStatus:
		if err := svc.agent.CheckPolicy(ctx, PolicyReq{
			Subject:     subject,
			SubjectType: subjectType,
			Permission:  AdminPermission,
			Object:      MagistralaObject,
			ObjectType:  PlatformType,
		}); err != nil {
			return svcerr.ErrDomainAuthorization
		}
	default:
		return svcerr.ErrDomainAuthorization
	}

	return nil
}
```

WashingtonKK: This makes sense @arvindh123, it is a great idea. Let me implement it

arvindh123: Thank for considering suggestions @WashingtonKK

arvindh123:
@WashingtonKK 

In `users/service.go` repoerr are returned by service function , 
Could you please change to respective service error and handle them  as you did for other service

things/service.go
```diff
@@ -125,7 +119,7 @@ func (svc service) CreateThings(ctx context.Context, token string, cls ...mgclie
 		})
 	}
 	if _, err := svc.auth.AddPolicies(ctx, &policies); err != nil {
-		return nil, errors.Wrap(errAddPolicies, err)
+		return nil, errors.Wrap(svcerr.ErrCreateEntity, err)
```
rodneyosodo, line 122: Also add `svcerr.ErrAddPolicies`

things/service.go
```diff
@@ -411,10 +405,10 @@ func (svc service) Share(ctx context.Context, token, id, relation string, userid
 	}
 	res, err := svc.auth.AddPolicies(ctx, &policies)
 	if err != nil {
-		return errors.Wrap(errAddPolicies, err)
+		return errors.Wrap(svcerr.ErrUpdateEntity, err)
```
rodneyosodo, line 408: same

things/service.go
```diff
@@ -411,10 +405,10 @@ func (svc service) Share(ctx context.Context, token, id, relation string, userid
...
 	}
 	if !res.Added {
-		return err
+		return errors.Wrap(svcerr.ErrUpdateEntity, err)
```
rodneyosodo, line 411: Don't wrap the error since it has already been handled above just return `svcerr.ErrAddPolicies`

things/service.go
```diff
@@ -440,7 +434,7 @@ func (svc service) Unshare(ctx context.Context, token, id, relation string, user
 	}
 	res, err := svc.auth.DeletePolicies(ctx, &policies)
 	if err != nil {
-		return errors.Wrap(errRemovePolicies, err)
+		return errors.Wrap(svcerr.ErrUpdateEntity, err)
```
rodneyosodo, line 437: use `svcerr.ErrDeletePolicies` instead

things/service_test.go
```diff
@@ -43,8 +43,7 @@ var (
 	invalid           = "invalid"
 	validID           = "d4ebb847-5d0e-4e46-bdd9-b6aceaaa3a22"
 	wrongID           = testsutil.GenerateUUID(&testing.T{})
-	errAddPolicies    = errors.New("failed to add policies")
-	errRemovePolicies = errors.New("failed to remove the policies")
+	errRemovePolicies = errors.New("failed to delete policies")
```
rodneyosodo, line 46: no need for this use errors from `svcerr`

users/service.go
```diff
@@ -19,6 +19,25 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+var (
```
rodneyosodo, line 22: No need for this errors use errors from `svcerr`

rodneyosodo (changes requested):

things/service.go
WashingtonKK: We do not want ```svcerr.ErrAddPolicies``` to be sent to a user, that is why we send ```svcerr.ErrCreateEntity``` which is what the user will see. errAddPolicies is left as an internal error which will be logged but we do not want it to be sent along with the response

things/service.go
WashingtonKK: We do not want ```svcerr.ErrAddPolicies``` to be sent to a user, that is why we send ```svcerr.ErrCreateEntity``` which is what the user will see ```errAddPolicies``` is left as an internal error which will be logged but we do not want it to be sent along with the response

things/service.go
WashingtonKK: We do not want ```svcerr.ErrAddPolicies``` to be sent to a user, that is why we send ```svcerr.ErrCreateEntity``` which is what the user will see ```errAddPolicies``` is left as an internal error which will be logged but we do not want it to be sent along with the response

things/service_test.go
WashingtonKK: We want these to be internal errors, as mentioned above. So we define them within the service so that they can be logged. Errors in the svcerr package are what we want to be added to the request body.

things/service.go
WashingtonKK: We do not want ```svcerr.ErrDeletePolicies``` to be sent to a user, that is why we send ```svcerr.ErrUpdateEntity``` which is what the user will see.

users/service.go
WashingtonKK: We need them locally and not on svcerr as described in the comments above.

users/service.go
```diff
@@ -19,6 +19,25 @@ import (
...
+	// ErrRecoveryToken indicates error in generating password recovery token.
+	ErrRecoveryToken = errors.New("failed to generate password recovery token")
+
+	// ErrFailedPolicyUpdate indicates a failure to update user policy.
+	ErrFailedPolicyUpdate = errors.New("failed to update user policy")
```
rodneyosodo: No need to export them if they are not used outside the package.

users/service.go
```diff
@@ -288,7 +307,7 @@ func (svc service) ResetSecret(ctx context.Context, resetToken, secret string) e
 	}
 	secret, err = svc.hasher.Hash(secret)
 	if err != nil {
-		return err
+		return errors.Wrap(svcerr.ErrAuthorization, err)
```
rodneyosodo: This is not an authorization error

users/service.go
```diff
@@ -399,10 +418,10 @@ func (svc service) DisableClient(ctx context.Context, token, id string) (mgclien
 func (svc service) changeClientStatus(ctx context.Context, token string, client mgclients.Client) (mgclients.Client, error) {
 	tokenUserID, err := svc.Identify(ctx, token)
 	if err != nil {
-		return mgclients.Client{}, err
+		return mgclients.Client{}, errors.Wrap(svcerr.ErrAuthorization, err)
```
rodneyosodo:
```suggestion
		return mgclients.Client{}, errors.Wrap(svcerr.ErrAuthentication, err)
```

users/service.go
```diff
@@ -399,10 +418,10 @@ func (svc service) DisableClient(ctx context.Context, token, id string) (mgclien
...
 	}
 	if err := svc.checkSuperAdmin(ctx, tokenUserID); err != nil {
-		return mgclients.Client{}, err
+		return mgclients.Client{}, errors.Wrap(svcerr.ErrAuthentication, err)
```
rodneyosodo:
```suggestion
		return mgclients.Client{}, errors.Wrap(svcerr.ErrAuthorization, err)
```

users/service.go
```diff
@@ -562,9 +581,9 @@ func (svc service) OAuthCallback(ctx context.Context, state mgoauth2.State, clie
 		rclient, err := svc.clients.RetrieveByIdentity(ctx, client.Credentials.Identity)
 		if err != nil {
 			if errors.Contains(err, repoerr.ErrNotFound) {
-				return &magistrala.Token{}, errors.New("user not signed up")
+				return &magistrala.Token{}, errors.Wrap(svcerr.ErrAuthorization, errUserNotSignedUp)
```
rodneyosodo: This is not an authorization error

users/service.go
```diff
@@ -562,9 +581,9 @@ func (svc service) OAuthCallback(ctx context.Context, state mgoauth2.State, clie
...
 			}
-			return &magistrala.Token{}, err
+			return &magistrala.Token{}, errors.Wrap(svcerr.ErrAuthorization, err)
```
rodneyosodo: same

users/service.go
```diff
@@ -575,9 +594,9 @@ func (svc service) OAuthCallback(ctx context.Context, state mgoauth2.State, clie
 		rclient, err := svc.RegisterClient(ctx, "", client)
 		if err != nil {
 			if errors.Contains(err, repoerr.ErrConflict) {
-				return &magistrala.Token{}, errors.New("user already exists")
+				return &magistrala.Token{}, errors.Wrap(svcerr.ErrConflict, err)
```
rodneyosodo: same

users/service.go
```diff
@@ -376,7 +395,7 @@ func (svc service) EnableClient(ctx context.Context, token, id string) (mgclient
 	}
 	client, err := svc.changeClientStatus(ctx, token, client)
 	if err != nil {
-		return mgclients.Client{}, errors.Wrap(mgclients.ErrEnableClient, err)
+		return mgclients.Client{}, err
```
rodneyosodo: revert to give more context to the error

certs/service.go
```diff
@@ -154,7 +153,7 @@ func (cs *certsService) ListCerts(ctx context.Context, token, thingID string, of
 
 	cp, err := cs.certsRepo.RetrieveByThing(ctx, u.GetId(), thingID, offset, limit)
 	if err != nil {
-		return Page{}, errors.Wrap(repoerr.ErrNotFound, err)
+		return Page{}, errors.Wrap(svcerr.ErrNotFound, err)
```
rodneyosodo: Should this be an `ErrViewEntity` or a `ErrNotFound`

rodneyosodo (changes requested):

users/service.go
```diff
@@ -260,15 +274,15 @@ func (svc service) UpdateClientIdentity(ctx context.Context, token, clientID, id
 func (svc service) GenerateResetToken(ctx context.Context, email, host string) error {
 	client, err := svc.clients.RetrieveByIdentity(ctx, email)
 	if err != nil || client.Credentials.Identity == "" {
-		return repoerr.ErrNotFound
+		return svcerr.ErrNotFound
```
arvindh123: This can lead to data leak, Any one can check user exists in platform or not, by requesting reset token   
@dborovcanin & @rodneyosodo  What do you think about this approach

api/openapi/bootstrap.yml
```diff
@@ -96,6 +96,8 @@ paths:
       responses:
         "200":
           $ref: "#/components/responses/ConfigRes"
+        "400":
+          $description: Missing or invalid config.
         "401":
```
arvindh123, line 101:
```suggestion
          description: Missing or invalid config.
```

arvindh123 (changes requested):
https://github.com/absmach/magistrala/pull/2119#discussion_r1549247412

Second part is not addressed

users/service.go
rodneyosodo: Refactor this to something like
```go
client, err := svc.clients.RetrieveByIdentity(ctx, email)
if err != nil {
	return errors.Wrap(svcerr.ErrViewEntity, err)
}
```

This also `ResetSecret` where we check that identity is not empty

users/service.go
```diff
@@ -284,11 +298,11 @@ func (svc service) ResetSecret(ctx context.Context, resetToken, secret string) e
 		return errors.Wrap(svcerr.ErrViewEntity, err)
 	}
 	if c.Credentials.Identity == "" {
-		return repoerr.ErrNotFound
+		return svcerr.ErrNotFound
```
rodneyosodo: Remove this check no need to check `c.Credentials.Identity` is empty as it is always present

users/service.go
```diff
@@ -575,9 +589,9 @@ func (svc service) OAuthCallback(ctx context.Context, state mgoauth2.State, clie
 		rclient, err := svc.RegisterClient(ctx, "", client)
 		if err != nil {
 			if errors.Contains(err, repoerr.ErrConflict) {
-				return &magistrala.Token{}, errors.New("user already exists")
+				return &magistrala.Token{}, errors.Wrap(svcerr.ErrConflict, err)
```
rodneyosodo: I don't think we need to wrap with `svcerr.ErrConflict` since it contains `repoerr.ErrConflict` and they have the same error message

users/service.go
```diff
@@ -19,6 +19,20 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+var (
+
+	// ErrFailedUpdateRole indicates a failure to update user role.
+	ErrFailedUpdateRole = errors.New("failed to update user role")
```
rodneyosodo: No need to export this if it is not being used outside this package

users/service.go
```diff
@@ -19,6 +19,20 @@ import (
...
+
+	// ErrIssueToken indicates a failure to issue token.
+	ErrIssueToken = errors.New("failed to issue token")
```
rodneyosodo: same

pkg/errors/service/types.go
```diff
@@ -55,18 +55,23 @@ var (
...
-	ErrIssueToken = errors.New("failed to issue token")
-
 	// ErrPasswordFormat indicates weak password.
 	ErrPasswordFormat = errors.New("password does not meet the requirements")
 
 	// ErrFailedUpdateRole indicates a failure to update user role.
 	ErrFailedUpdateRole = errors.New("failed to update user role")
+
+	// ErrFailedPermissionsList indicates a failure to list permissions.
+	ErrFailedPermissionsList = errors.New("failed to list permissions")
```
rodneyosodo, line 65: Add new line after

certs/service.go
```diff
@@ -186,7 +185,7 @@ func (cs *certsService) ViewCert(ctx context.Context, token, serialID string) (C
 
 	cert, err := cs.certsRepo.RetrieveBySerial(ctx, u.GetId(), serialID)
 	if err != nil {
-		return Cert{}, errors.Wrap(repoerr.ErrNotFound, err)
+		return Cert{}, errors.Wrap(svcerr.ErrNotFound, err)
```
rodneyosodo: This should be. Also check other cases where we wrap `ErrNotFound` when we fail to view an entity from the DB.
```suggestion
		return Cert{}, errors.Wrap(svcerr.ErrViewEntity, err)
```

rodneyosodo (changes requested):

users/service.go
WashingtonKK: I don't understand, `repoerr.ErrConflict` has message `entity already exists`.

rodneyosodo: `svcerr.ErrConflict` and `repoerr.ErrConflict` have the same error message the difference is one should be used at service layer while the other should be used in repository

users/service.go
```diff
@@ -283,12 +291,10 @@ func (svc service) ResetSecret(ctx context.Context, resetToken, secret string) e
...
 		return errors.Wrap(svcerr.ErrViewEntity, err)
 	}
-	if c.Credentials.Identity == "" {
-		return repoerr.ErrNotFound
-	}
+
 	secret, err = svc.hasher.Hash(secret)
 	if err != nil {
-		return err
+		return errors.Wrap(svcerr.ErrUpdateEntity, err)
```
rodneyosodo:
```suggestion
		return errors.Wrap(svcerr.ErrMalformedEntity, err)
```

rodneyosodo (changes requested):

users/service.go
```diff
@@ -259,16 +267,16 @@ func (svc service) UpdateClientIdentity(ctx context.Context, token, clientID, id
 
 func (svc service) GenerateResetToken(ctx context.Context, email, host string) error {
 	client, err := svc.clients.RetrieveByIdentity(ctx, email)
-	if err != nil || client.Credentials.Identity == "" {
-		return repoerr.ErrNotFound
+	if err != nil {
+		return errors.Wrap(svcerr.ErrViewEntity, err)
 	}
```
arvindh123, line 277: @rodneyosodo  @WashingtonKK @dborovcanin 

What do you think about this suggestion ? 
```suggestion
	if err != nil {
		if errors.Contains(repoerr.ErrNotFound, err) {
			return nil
		} else {
			return errors.Wrap(svcerr.ErrViewEntity, err) // it should be an internal error
		}
	}
```

WashingtonKK: Why would we return nil when the error contains `ErrNotFound` ? Shouldn't we return `svcerr.ErrNotFound` in this case?

users/service.go
```diff
@@ -399,10 +405,10 @@ func (svc service) DisableClient(ctx context.Context, token, id string) (mgclien
 func (svc service) changeClientStatus(ctx context.Context, token string, client mgclients.Client) (mgclients.Client, error) {
 	tokenUserID, err := svc.Identify(ctx, token)
 	if err != nil {
-		return mgclients.Client{}, err
+		return mgclients.Client{}, errors.Wrap(svcerr.ErrAuthentication, err)
 	}
 	if err := svc.checkSuperAdmin(ctx, tokenUserID); err != nil {
-		return mgclients.Client{}, err
+		return mgclients.Client{}, errors.Wrap(svcerr.ErrAuthorization, err)
```
arvindh123: @WashingtonKK 
`svc.Identify` already returns the `svcerr.ErrAuthentication` and `svc.checkSuperAdmin`, So no need to wrap here again. 
We are calling here function which belongs same service layer.

users/service.go
```diff
@@ -577,7 +583,7 @@ func (svc service) OAuthCallback(ctx context.Context, state mgoauth2.State, clie
 			if errors.Contains(err, repoerr.ErrConflict) {
 				return &magistrala.Token{}, errors.New("user already exists")
```
arvindh123: Wrapping  this error with `svcerr.ErrConflict` will return conflict status code in HTTP response

users/service.go
```diff
@@ -315,11 +321,11 @@ func (svc service) UpdateClientSecret(ctx context.Context, token, oldSecret, new
 		return mgclients.Client{}, errors.Wrap(svcerr.ErrViewEntity, err)
 	}
 	if _, err := svc.IssueToken(ctx, dbClient.Credentials.Identity, oldSecret, ""); err != nil {
-		return mgclients.Client{}, errors.Wrap(svcerr.ErrIssueToken, err)
+		return mgclients.Client{}, errors.Wrap(errIssueToken, err)
```
arvindh123: In `svc.IssueToken` gRPC calls error is not handled. 
Instead of wrapping error here, we can wrap and return correct error from `svc.IssueToken`, since `svc.IssueToken` is present in same service layer.

users/service.go
```diff
@@ -577,7 +583,7 @@ func (svc service) OAuthCallback(ctx context.Context, state mgoauth2.State, clie
 			if errors.Contains(err, repoerr.ErrConflict) {
 				return &magistrala.Token{}, errors.New("user already exists")
 			}
-			return &magistrala.Token{}, err
+			return &magistrala.Token{}, errors.Wrap(svcerr.ErrCreateEntity, err)
```
arvindh123: This wrap can be remove since `svc.RegisterClient` is present in same user service layer.

arvindh123 (commented):
For the below request,  I get error `{"error":"","message":"view entity failed"}`

```bash
curl --location 'http://localhost/users/tokens/issue' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--data '{
        "identity": "some-random-string",
        "secret": "12345678"
}'
```

If we try to update tag of non existance user id , 
Example
```bash
curl --location --request PATCH 'http://localhost/users/unknown-user-id/tags' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'Authorization: Bearer <Token>' \
--data '{
  "tags": [
    "tag1", "tag2"
  ]
}'
```

we are getting  `{   "error": "",    "message": "update entity failed" }` in response 

Same kind of result get for below request too.
curl --location --request PATCH 'http://localhost/users/03cb42ba-4592-4f53-97c9-a776ee31a06a11/role'

arvindh123:
> For the below request, I get error `{"error":"","message":"view entity failed"}`
> 
> ```shell
> curl --location 'http://localhost/users/tokens/issue' \
> --header 'Content-Type: application/json' \
> --header 'Accept: application/json' \
> --data '{
>         "identity": "some-random-string",
>         "secret": "12345678"
> }'
> ```

Lets create a separate ticket for this problem.

arvindh123:
For the change secret requets, If we send wrong password  like below
```
curl --location --request PATCH 'http://localhost/users/secret' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'Authorization: Bearer <TOKEN>' \
--data '{
  "old_secret": "wrong-old-password",
  "new_secret": "new-password"
}'
```
we get response like below 

```json
{
    "error": "",
    "message": "failed to issue token"
}
```

in Users logs we get proper errror which should sent back to user 

JSON Log 
```json
{
    "time": "2024-04-24T11:57:37.002060394Z",
    "level": "WARN",
    "msg": "Update user secret failed to complete successfully",
    "duration": "76.890158ms",
    "user": {
        "id": "",
        "name": ""
    },
    "error": {
        "error": "invalid user id or secret",
        "message": "failed to issue token"
    }
}
```

arvindh123:
Issue token doesn't return proper error 

**What were you trying to achieve?**

Login to Magistrala with unknow / invalid user identity

```bash
curl --location 'http://localhost/users/tokens/issue' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--data '{
        "identity": "unknown-users-indentity",
        "secret": "12345678"
}'
```
**What are the expected results?**

HTTP/1.1 401 Unauthorized
```json
{
    "error": "",
    "message": "invalid user id or secret"
}
```

**What are the received results?**
HTTP/1.1 400 Bad Request

```json
{
    "error": "",
    "message": "view entity failed"
}
```

arvindh123 (changes requested):
Issue token doesn't return proper error 

**What were you trying to achieve?**

Login to Magistrala with unknow / invalid user identity

```bash
curl --location 'http://localhost/users/tokens/issue' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--data '{
        "identity": "unknown-users-indentity",
        "secret": "12345678"
}'
```
**What are the expected results?**

HTTP/1.1 401 Unauthorized
```json
{
    "error": "",
    "message": "invalid user id or secret"
}
```

**What are the received results?**
HTTP/1.1 400 Bad Request

```json
{
    "error": "",
    "message": "entity not found"
}
```

users/service.go
```diff
@@ -85,15 +93,15 @@ func (svc service) RegisterClient(ctx context.Context, token string, cli mgclien
...
+		return mgclients.Client{}, errors.Wrap(svcerr.ErrCreateEntity, err)
 	}
 	return client, nil
 }
 
 func (svc service) IssueToken(ctx context.Context, identity, secret, domainID string) (*magistrala.Token, error) {
 	dbUser, err := svc.clients.RetrieveByIdentity(ctx, identity)
 	if err != nil {
-		return &magistrala.Token{}, errors.Wrap(svcerr.ErrViewEntity, err)
+		return &magistrala.Token{}, err
```
arvindh123: Please Wrap this error with `svcerr.ErrAuthentication`, most of the times, error here  will be repoerr.ErrNotFound, So i suggest to wrap with `svcerr.ErrAuthentication`.

users/service.go
arvindh123: @rodneyosodo @WashingtonKK No need to wrap here , because it is already wrapped in `svc.Identify` , which is present in the same layer.

users/service.go
```diff
@@ -619,7 +631,7 @@ func (svc service) addClientPolicy(ctx context.Context, userID string, role mgcl
 	}
 	resp, err := svc.auth.AddPolicies(ctx, &policies)
 	if err != nil {
-		return errors.Wrap(svcerr.ErrAddPolicies, err)
+		return errors.Wrap(errAddPolicies, err)
```
arvindh123: Can we have some thing similar to what we have in remove policy 
```suggestion
		return errors.Wrap(svcerr.ErrAddPolicies, err)
```

users/service.go
```diff
@@ -685,10 +697,10 @@ func (svc service) updateClientPolicy(ctx context.Context, userID string, role m
 			Object:      auth.MagistralaObject,
 		})
 		if err != nil {
-			return errors.Wrap(svcerr.ErrDeletePolicies, err)
+			return errors.Wrap(svcerr.ErrUpdateEntity, err)
 		}
 		if !resp.Deleted {
```
arvindh123: Please revert back them to original, We can handle errors in ticket https://github.com/absmach/magistrala/issues/2202

auth/service.go
```diff
@@ -186,14 +186,7 @@ func (svc service) Authorize(ctx context.Context, pr PolicyReq) error {
...
 	if pr.SubjectType == UserType && (pr.ObjectType == GroupType || pr.ObjectType == ThingType || pr.ObjectType == DomainType) {
-		domainID := pr.Domain
-		if domainID == "" {
-			if pr.ObjectType != DomainType {
-				return svcerr.ErrDomainAuthorization
-			}
-			domainID = pr.Object
-		}
-		if err := svc.checkDomain(ctx, pr.SubjectType, pr.Subject, domainID); err != nil {
+		if err := svc.checkDomain(ctx, pr.SubjectType, pr.Subject, pr.Domain); err != nil {
```
arvindh123: @WashingtonKK , 
Please revert back to original, My suggestion doesn't works properly for domain related operations 
It fails at domain authroization

arvindh123 (changes requested):
@WashingtonKK , 
Please revert back to original, My suggestion doesn't works properly for domain related operations 
It fails at domain authroization

auth/service.go
```diff
@@ -204,6 +204,23 @@ func (svc service) checkPolicy(ctx context.Context, pr PolicyReq) error {
 }
 
 func (svc service) checkDomain(ctx context.Context, subjectType, subject, domainID string) error {
+	fmt.Println(PolicyReq{
+		Subject:     subject,
+		SubjectType: subjectType,
+		Permission:  MembershipPermission,
+		Object:      domainID,
+		ObjectType:  DomainType,
+	})
```
arvindh123: Please remove this debug lines
```suggestion
```

auth/service_test.go
```diff
@@ -231,9 +248,15 @@ func TestIssue(t *testing.T) {
 				ObjectType:  auth.DomainType,
 				Permission:  auth.MembershipPermission,
 			},
+			checkPolicyRequest2: auth.PolicyReq{
```
dborovcanin: Rename `checkPolicyRequest` fields to add more context. For example `checkPolicyReq` and `checkPolicyWithStatusReq`.

dborovcanin (changes requested):

internal/api/common.go
```diff
@@ -70,6 +70,8 @@ const (
 	DescDir      = "desc"
 )
 
+var errAddBootstrap = errors.New("failed to add bootstrap configuration")
```
arvindh123: @WashingtonKK 
Instead of using `errAddBootstrap`in bootstrap,  can we use `svcerr.CreateEntity` ?

WashingtonKK: Okay @arvindh123 . Although I think `errAddBootstrap` is a better option since `ErrCreateEntity` does not fully indicate the nature of the error and may be confusing to the user. What do you think about this?

arvindh123: It looks odd for me, The only un exported variable in that file.  And we are duplicating the same variable in two files

WashingtonKK: The error was un exported in the bootstrap service, let me place it in `svcerr` package so that we will not have this odd looking error.

arvindh123 (dismissed):
LGTM

users/service.go
```diff
@@ -562,9 +573,9 @@ func (svc service) OAuthCallback(ctx context.Context, state mgoauth2.State, clie
 		rclient, err := svc.clients.RetrieveByIdentity(ctx, client.Credentials.Identity)
 		if err != nil {
 			if errors.Contains(err, repoerr.ErrNotFound) {
-				return &magistrala.Token{}, errors.New("user not signed up")
+				return &magistrala.Token{}, errors.Wrap(svcerr.ErrNotFound, errUserNotSignedUp)
 			}
-			return &magistrala.Token{}, err
+			return &magistrala.Token{}, errors.Wrap(svcerr.ErrNotFound, err)
```
rodneyosodo: This should be wrapped with ViewEntity
```suggestion
			return &magistrala.Token{}, errors.Wrap(svcerr.ErrViewEntity, err)
```

rodneyosodo (changes requested):

rodneyosodo (approved):

arvindh123 (approved):

dborovcanin (approved):

PR #2122, opened 2024-03-20
MG - 1529 - User Removal

### LINKED ISSUE #1529: Add user removal and/or blocking feature
**FEATURE REQUEST**

There is issue #458.

User removal is a natural feature that we are missing. By initial design, we avoided it due to the complexity of the cascade effect caused by the ownership relation: since only owners could manipulate their Things and Channels, it would imply that removing removes all of its assets in a cascade way. Now, when we introduced administrators and detach access control from ownership, user removal should be as simple as setting a `blocked` flag or physically removing it. After that, the administrator can take care of all the orphaned assets - remove them or assign them to someone else.
We also need to introduce refresh tokens and make access tokens last shorter to reduce the possible attack time frame.
We need to define what removing the user exactly means and do we support user removal or only block it, but either way, it should implement at least one of those two options.

Comments are welcome. @mainflux/maintainers 

This is a must-have.

rodneyosodo:
# What type of PR is this?

This is a feature because it adds user removal

## What does this do?

This PR adds a new method to delete a user with a given ID in the repository package. The implementation of this method uses an SQL query to delete the client from the database.

Additionally, an extra method has been added to the gRPC auth service to handle entity deletion. This is useful because it reduces network trips back and forth from the client microservices to the auth microservice. Test cases have been added for the delete method to ensure its functionality. 

- Resolves #1529 


users/service.go
```diff
@@ -439,6 +439,33 @@ func (svc service) changeClientStatus(ctx context.Context, token string, client
...
 }
 
+func (svc service) DeleteClient(ctx context.Context, token, id string) error {
+	res, err := svc.identify(ctx, token)
+	if err != nil {
+		return err
+	}
+	if err := svc.checkSuperAdmin(ctx, res.GetId()); err != nil {
+		return err
+	}
```
arvindh123: @rodneyosodo We are planning like, only super admins can delete users

auth/service.go
```diff
@@ -956,3 +961,122 @@ func DecodeDomainUserID(domainUserID string) (string, string) {
...
+			return err
+		}
+
+		// Remove policy of users
+		req.SubjectType = UserType
+		if err := svc.DeletePolicy(ctx, req); err != nil {
+			return err
+		}
+
+		return nil
```
arvindh123: For `svc.DeletePolicy` only object Type is required field. 
So the below snippet will delete all type of subjects related to thing object
```suggestion
		// Remove all policies related to thing
		req := PolicyReq{
			Object:     id,
			ObjectType: ThingType,
		}

		if err := svc.DeletePolicy(ctx, req); err != nil {
			return err
		}
		return nil
```

auth/service.go
```diff
@@ -956,3 +961,122 @@ func DecodeDomainUserID(domainUserID string) (string, string) {
...
+		for _, domain := range domainsPage.Domains {
+			policy := PolicyReq{
+				Subject:     EncodeDomainUserID(domain.ID, id),
+				SubjectType: UserType,
+				Object:      domain.ID,
+				ObjectType:  DomainType,
+			}
+			if err := svc.agent.DeletePolicy(ctx, policy); err != nil {
+				return err
+			}
```
arvindh123: This below snippet will delete all Thing, Group, Channel relation related to user id within the domain
```suggestion
			policy := PolicyReq{
				Subject:     EncodeDomainUserID(domain.ID, id),
				SubjectType: UserType,
				ObjectType:  ThingType,
			}
			if err := svc.agent.DeletePolicy(ctx, policy); err != nil {
				return err
			}
			policy = PolicyReq{
				Subject:     EncodeDomainUserID(domain.ID, id),
				SubjectType: UserType,
				ObjectType:  GroupType,
			}
			if err := svc.agent.DeletePolicy(ctx, policy); err != nil {
				return err
			}

			policy = PolicyReq{
				Subject:     EncodeDomainUserID(domain.ID, id),
				SubjectType: UserType,
				Object:      domain.ID,
				ObjectType:  DomainType,
			}
			if err := svc.agent.DeletePolicy(ctx, policy); err != nil {
				return err
			}
```

auth/service.go
```diff
@@ -956,3 +961,122 @@ func DecodeDomainUserID(domainUserID string) (string, string) {
...
+			return err
+		}
+		req.ObjectType = GroupType
+		if err := svc.agent.DeletePolicy(ctx, req); err != nil {
+			return err
+		}
+		req.ObjectType = DomainType
+		if err := svc.agent.DeletePolicy(ctx, req); err != nil {
+			return err
+		}
```
arvindh123: There is no need to remove policy between user_id and things/groups, because,There will be no subject just with user id in spicedb. 
User's Subject id in spicesdb will be always combination <domain_id>_<user_id>

```suggestion
		req := PolicyReq{
			Subject:     id,
			SubjectType: UserType,
			ObjectType:  PlatformType,
		    req.Object = MagistralaObject
		}
		if err := svc.agent.DeletePolicy(ctx, req); err != nil {
			return err
		}
```

arvindh123 (changes requested):
Hello @rodneyosodo left some comments for delete policies in auth service.

users/service.go
rodneyosodo: Yes. It is only super admins who can delete users

auth/service.go
rodneyosodo: If we leave it spiceDB will see it as a valid policy and will not clean up

auth/service.go
rodneyosodo: This doesn't work. I get 
```json
{"time":"2024-03-21T08:36:41.624438869Z","level":"WARN","msg":"Delete entity policies failed to complete successfully","duration":"284.946µs","entity_type":"thing","id":"feae8203-452b-4f7c-a003-76b475e43b56","error":{"error":"malformed entity specification","message":"failed to remove the policies"}}
```

users/service.go
rodneyosodo: Should I also add that a user can be able to delete themselves?

arvindh123: We can add that later when required

arvindh123:
@rodneyosodo  @dborovcanin 

For Delete users , Instead of deleting the user policy & removing from repo on Delete request,

I propose some thing like below:

On Delete request , change the `state` of user to `delete` in repo for the user. 
Then we can delete the user with a `background event handler / task`  written in GO , not using postgres extension, these handler will remove policies and user from repo 

This gives advantage like,
- If user present in large domains and owns large entities , if we do on Delete request it will take time return response , the requesting doing the delete process
- During the delete , if any error happens, remain policy stay on spiceDB, So still user have access to them.

We can even have configuration for background user removal handler to start the removal operation after certain period.  Example, background handler will remove after 30 days when user status changes to delete 


We will have new challenges on this approach, 
When user `state` changes to `delete`, If user signup again back to platform before the background user removal handler starts its operation to remove the user, then user might not allowed to signup. 
Because in users Database we have unique contraint for email and username. 

To overcome this issue, when user while chaning the user state delete, at the same time, we can make the `email` and `username` field in database empty and move the existing value to some other column like `email_before_delete`    and `username_before_delete` . This helps to keep the record of username and email.

auth/service.go
```diff
@@ -1030,3 +1035,132 @@ func DecodeDomainUserID(domainUserID string) (string, string) {
...
+
+		// Remove policy of users
+		req.SubjectType = UserType
+		if err := svc.DeletePolicy(ctx, req); err != nil {
+			return err
+		}
+
+		return nil
```
arvindh123:
```suggestion
	case ThingType:
		// Remove policy of groups
		req := PolicyReq{
			Object:     id,
			ObjectType: ThingType,
		}

		if err := svc.DeletePolicy(ctx, req); err != nil {
			return err
		}
		return nil
```

The above suggestin will work , if we have changes in svc.agent.DeletePolicy show below 

```
func (pa *policyAgent) DeletePolicy(ctx context.Context, pr auth.PolicyReq) error {
	req := &v1.DeleteRelationshipsRequest{
		RelationshipFilter: &v1.RelationshipFilter{
			ResourceType:       pr.ObjectType,
			OptionalResourceId: pr.Object,
		},
	}

	if pr.Relation != "" {
		req.RelationshipFilter.OptionalRelation = pr.Relation
	}

	if pr.SubjectType != "" {
		req.RelationshipFilter.OptionalSubjectFilter = &v1.SubjectFilter{
			SubjectType: pr.SubjectType,
		}
		if pr.Subject != "" {
			req.RelationshipFilter.OptionalSubjectFilter.OptionalSubjectId = pr.Subject
		}
		if pr.SubjectRelation != "" {
			req.RelationshipFilter.OptionalSubjectFilter.OptionalRelation = &v1.SubjectFilter_RelationFilter{
				Relation: pr.SubjectRelation,
			}
		}
	}

	if _, err := pa.permissionClient.DeleteRelationships(ctx, req); err != nil {
		return errors.Wrap(errRemovePolicies, handleSpicedbError(err))
	}
	return nil
}
```

auth/service.go
```diff
@@ -1030,3 +1035,132 @@ func DecodeDomainUserID(domainUserID string) (string, string) {
...
+		req.ObjectType = GroupType
+		if err := svc.DeletePolicy(ctx, req); err != nil {
+			return err
+		}
+
+		// Remove policy of users
+		req.SubjectType = UserType
+		if err := svc.DeletePolicy(ctx, req); err != nil {
+			return err
+		}
```
arvindh123:
```suggestion
	case GroupType:
		// Remove policies where group is subject and object is another child group.
		req := PolicyReq{
			SubjectType: GroupType,
			Subject:     id,
			ObjectType:  GroupType,
		}
		if err := svc.DeletePolicy(ctx, req); err != nil {
			return err
		}

		// Remove policies where group is subject and object is thing.
		req.ObjectType = ThingType
		if err := svc.DeletePolicy(ctx, req); err != nil {
			return err
		}

		// Remove all policies where group is object.
		req = PolicyReq{
			Object:     id,
			ObjectType: GroupType,
		}
		if err := svc.DeletePolicy(ctx, req); err != nil {
			return err
		}
```

The above suggestin will work , if we have changes in svc.agent.DeletePolicy show below 

```
func (pa *policyAgent) DeletePolicy(ctx context.Context, pr auth.PolicyReq) error {
	req := &v1.DeleteRelationshipsRequest{
		RelationshipFilter: &v1.RelationshipFilter{
			ResourceType:       pr.ObjectType,
			OptionalResourceId: pr.Object,
		},
	}

	if pr.Relation != "" {
		req.RelationshipFilter.OptionalRelation = pr.Relation
	}

	if pr.SubjectType != "" {
		req.RelationshipFilter.OptionalSubjectFilter = &v1.SubjectFilter{
			SubjectType: pr.SubjectType,
		}
		if pr.Subject != "" {
			req.RelationshipFilter.OptionalSubjectFilter.OptionalSubjectId = pr.Subject
		}
		if pr.SubjectRelation != "" {
			req.RelationshipFilter.OptionalSubjectFilter.OptionalRelation = &v1.SubjectFilter_RelationFilter{
				Relation: pr.SubjectRelation,
			}
		}
	}

	if _, err := pa.permissionClient.DeleteRelationships(ctx, req); err != nil {
		return errors.Wrap(errRemovePolicies, handleSpicedbError(err))
	}
	return nil
}
```

arvindh123 (changes requested):

pkg/clients/status.go
```diff
@@ -19,6 +19,8 @@ const (
 	EnabledStatus Status = iota
 	// DisabledStatus represents disabled Client.
 	DisabledStatus
+	// DeletedStatus represents a client that will be deleted.
+	DeletedStatus
```
arvindh123, line 23: `DeleteStatus` can be highest status , If we want to hide the Deleted users from List with status=all `http://localhost/users?status=all`, 
If DeleteStatus becmes higher status, then superadmin can get deleted user by explicitly like `status=deleted`

arvindh123 (changes requested):

pkg/clients/status.go
rodneyosodo: Highest status meaning 0 or n?

arvindh123: `n`

auth/service.go
rodneyosodo: This automatically deletes the other group entities

auth/domains.go
```diff
@@ -199,4 +199,7 @@ type DomainsRepository interface {
 
 	// CheckPolicy check policy in domains database.
 	CheckPolicy(ctx context.Context, pc Policy) error
+
+	// DeleteUserPolicies deletes user policies from domains database.
+	DeleteUserPolicies(ctx context.Context, id string) (err error)
```
dborovcanin, line 204: Ranme to `DeletePolicies`.

users/delete_handler.go
```diff
@@ -0,0 +1,102 @@
...
+func (h *handler) routine(ctx context.Context) {
+	pm := mgclients.Page{Limit: defLimit, Offset: 0, Status: mgclients.DeletedStatus}
+
+	dbUsers, err := h.clients.RetrieveAll(ctx, pm)
+	if err != nil {
+		h.logger.Error("failed to retrieve users", slog.Any("error", err))
+		return
+	}
+
+	if dbUsers.Total > defLimit {
```
dborovcanin: Let's fetch and remove in batches, rather than fetching all first then removing, just to avoid edge cases where we slice of removed users are huge.

users/delete_handler.go
```diff
@@ -0,0 +1,102 @@
...
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				handler.routine(ctx)
+			}
+		}
+	}()
+}
+
+func (h *handler) routine(ctx context.Context) {
```
dborovcanin: Rename to `handle`.

users/delete_handler.go
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) Abstract Machines
+// SPDX-License-Identifier: Apache-2.0
```
dborovcanin, line 2: Add a description of this cron job explaining what it does and why. Keep in mind that scaling Users svc will result in scaling this cron, which is not ideal, and we will probably extract this to a separate process in the future.

dborovcanin (changes requested):

auth/domains.go
rodneyosodo: This is different from `DeletePolicies` will delete based on `subject_type`, `subject_id`, `subject_relation`, `object_type` and `object_id` while `DeleteUserPolicies` will delete based on `subject_id` which is `user_id`

dborovcanin (approved):

PR #2127, opened 2024-03-25
MG-1962 - Add single endpoint to change status of domain

### LINKED ISSUE #1962: Feature: Single endpoint to change status of domain

There are three different endpoint  (`/enabled` , `/disable`, `/freeze`) to change status of domain.

@dborovcanin recommendations :

We should probably do the following:
- ~separate enable and disable and freeze from soft delete~
- add delete endpoint - delete operation can't be undone
- add change-status endpoint for all other changeable statuses.

Must-have

### Anything else?

_No response_

janael-pinheiro:
# What type of PR is this?

This is a feature because it adds the following functionality: adds a single endpoint to change domain status and an endpoint to remove domain.

## What does this do?
removes the `/domains/{domainID}/enabled`, `/domains/{domainID}/disabled`, and `/domains/{domainID}/freezed` endpoints. Adds the `/domains/{domainID}/status` (HTTP PUT) endpoint that concentrates domain status change operations. Additionally, it adds the `/domains/{domainID}` (HTTP DELETE) endpoint to remove a specific domain.

- Related Issue #1962 
- Resolves #1962

Yes, I have included tests for my changes. I added tests to evaluate the behavior of the new endpoints ( `/domains/{domainID}/status` and `/domains/{domainID}`).

Yes, I have updated the documentation for the new feature. Yes, I updated the `api/openapi/auth.yml` file

### DIFF
--- a/api/openapi/auth.yml
+++ b/api/openapi/auth.yml
@@ -175,15 +175,18 @@ paths:
           description: Database can't process request.
         "500":
           $ref: "#/components/responses/ServiceError"
-  /domains/{domainID}/enable:
-    post:
-      summary: Enables a domain
+
+  /domains/{domainID}/status:
+    put:
+      summary: changes the domain status
       description: |
-        Enables a specific domain that is identified by the domain ID.
+        Changes a specific domain status that is identified by the domain ID.
       tags:
         - Domains
       parameters:
         - $ref: "#/components/parameters/DomainID"
+      requestBody:
+        $ref: "#/components/requestBodies/DomainStatusUpdateReq"
       security:
         - bearerAuth: []
       responses:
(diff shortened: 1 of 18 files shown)


dborovcanin (commented):
Please sign-off your commits (check DCO report) and fix CI.

auth/api/http/domains/requests.go
```diff
@@ -107,12 +108,17 @@ func (req listDomainsReq) validate() error {
 	return nil
 }
 
-type enableDomainReq struct {
+type statusDomainReq struct {
 	token    string
```
WashingtonKK, line 111:
```
type statusDomainReq struct {
	token    string
	domainID string
	Status   auth.Status `json:"status,omitempty"`
}
```

auth/api/http/domains/requests.go
```diff
@@ -121,32 +127,18 @@ func (req enableDomainReq) validate() error {
...
-type disableDomainReq struct {
-	token    string
-	domainID string
-}
-
-func (req disableDomainReq) validate() error {
-	if req.token == "" {
-		return apiutil.ErrBearerToken
+	if _, ok := validStatuses[req.Status]; !ok {
+		return svcerr.ErrInvalidStatus
```
WashingtonKK: Return `apiutil.ErrMissingID`
Service errors are only returned on the service layer.

auth/service.go
```diff
@@ -560,6 +560,24 @@ func (svc service) CreateDomain(ctx context.Context, token string, d Domain) (do
...
+		Subject:     key.Subject,
+		SubjectType: UserType,
+		SubjectKind: UsersKind,
+		ObjectType:  DomainType,
+		Permission:  AdminPermission,
+		Object:      d.ID,
+	}); err != nil {
+		return errors.Wrap(svcerr.ErrAuthorization, err)
+	}
+	return svc.domains.Delete(ctx, d.ID)
```
WashingtonKK:
```go
if err := svc.domains.Delete; err != nil {
    return errors.Wrap(svcerr.ErrFailedDelete, err)
}
```

We should not return the repo error directly since it may leak some db info which is not desirable.

WashingtonKK (changes requested):

drasko:
@dborovcanin and @arvindh123 please review this one

api/openapi/auth.yml
```diff
@@ -175,74 +175,23 @@ paths:
...
-          description: A non-existent entity request.
-        "422":
-          description: Database can't process request.
-        "500":
-          $ref: "#/components/responses/ServiceError"
 
-  /domains/{domainID}/disable:
-    post:
-      summary: Disable a domain
+  /domains/{domainID}/status:
```
drasko, line 179: Why is important to have one endpoint? It looks more self-documenting to me that way, than having sing `status` endpoint - which usually prompt me that it is status indicator (i.e. something I would use with `GET` to fetch the current status), and not a control switch.

auth/api/logging.go
```diff
@@ -349,6 +349,25 @@ func (lm *loggingMiddleware) CreateDomain(ctx context.Context, token string, d a
 	return lm.svc.CreateDomain(ctx, token, d)
 }
 
+func (lm *loggingMiddleware) DeleteDomain(ctx context.Context, token string, d auth.Domain) (err error) {
```
drasko, line 352: Probably deletion was avoided for now because of cascade consequences - @dborovcanin must confirm. But probably you'll need to remove all resources related to this domain (things, channels, ...) once you decide to remove a specific domain.

dborovcanin: @drasko This is correct, it requires a careful analysis due to the cascade nature of removal.

dborovcanin:

dborovcanin:
After [this comment](https://github.com/absmach/magistrala/pull/2127#discussion_r1542169662) and an internal discussion between the maintainer and due to the inactivity of this pull request, we decided to close this PR and linked issue, since we will keep the current approach. Both approaches diverge from the usual REST principle, the current one (with 3 endpoints) being less prone to errors and a little more verbose. @janael-pinheiro Thanks for the contribution and effort you put into this pull request.

PR #2128, opened 2024-03-27
MG-2136 - Generate mocks with mockery for Bootstrap service 

### LINKED ISSUE #1993: Feature: Generate Mocks Per Service

The problem we are currently having is that we write mocks for interfaces ourselves which might be cumbersome and had to maintain especially for SDK.

Instead of writing down each mock for individual interfaces use https://github.com/vektra/mockery or https://github.com/uber-go/mock to generate mocks for interfaces. This will be instrumental in testing. This is quite useful for services depending on the SDK for example, bootstrap and certs service. This approach simplifies the testing process as initially, one would need to generate the things services, then the things server and finally use the things server URL on the SDK. This effective means that when testing other services you inherently test services that depend on the SDK which is not ideal. With SDK mocks you can mock the SDK rather than mocking dependent services.

Nice-to-have

### Anything else?

Initial work is done at https://github.com/absmach/magistrala-old/pull/127

## Procedure

- [x] auth
- [x] bootstrap https://github.com/absmach/magistrala/issues/2136
- [x] http - https://github.com/absmach/magistrala/issues/2170
- [x] invitations - https://github.com/absmach/magistrala/issues/2143
- [x] mqtt - https://github.com/absmach/magistrala/issues/2161
- [x] pkg
- [x] provision
- [x] things
- [x] users - https://github.com/absmach/magistrala/issues/2164

- [x] certs - https://github.com/absmach/magistrala/issues/2137
- [x] consumers - https://github.com/absmach/magistrala/issues/2147
- [x] lora - https://github.com/absmach/magistrala/issues/2145
- [x] readers - https://github.com/absmach/magistrala/issues/2166
- [x] twins - https://github.com/absmach/magistrala/issues/2173

Service/modules which have no mocks before
- ws
- opcua
- coap
- logger
- internal

nyagamunene:
# What type of PR is this?
This is a feature because its add the following functionality: generates mocks for bootstrap.

## What does this do?
It generates the mocks for bootstrap service.

- Related to #2136

Yes, I have included tests.

No, I have not updated the documentation.


WashingtonKK:
Please sign off your commits, check DCO report [here](https://github.com/absmach/magistrala/pull/2128/checks?check_run_id=23168073503) and fix.

WashingtonKK (changes requested):
We do not need mocks for the ws and coap services since they are not used. Only replace manually written mocks with generated mocks.
An example is bootstrap service, we have manually written mocks for the service but we need to have them generated instead.

WashingtonKK:
Also update the mockery version used to use [the latest version](https://github.com/vektra/mockery/releases/tag/v2.42.1)

bootstrap/configs.go
```diff
@@ -60,6 +60,8 @@ type ConfigsPage struct {
 }
 
 // ConfigRepository specifies a Config persistence API.
+//
+//go:generate mockery --name ConfigRepository --output=./mocks --filename configs.go --quiet --note "Copyright (c) Abstract Machines"
```
rodneyosodo, line 64: Run 
```bash
make mocks
```

to generate the mocks

rodneyosodo (changes requested):

bootstrap/configs.go
nyagamunene: Okay

PR #2130, opened 2024-04-02
MG-2069 - Remove relation requirement from entity unassignment

### LINKED ISSUE #2069: Feature:  Relationship might not be required for unassigning User for Entity

Before removing user from entity , user relationship need to know.

In SpiceDB , with Delete Relationship Filter function, might helps to remove user to entity without knowing the relationship. 

Must-have

### Anything else?

This Issue can be considered after analyzing the spiceDB  Delete Relationship Filter function and testing with current implementation. 

Challenges : 
In unassign request,  we can provide array of users_ids. 
User with `Editor` relation should not able to  remove users id with administrator
So for each user_id in arrary we should check relation with entities. 
If user_id have administrator relation and unassing request by editor user, then it should be denied

WashingtonKK:
# What type of PR is this?

This is a refactor because it changes the following functionality: Unassign entity.

## What does this do?

Removes the relation requirement from unassigning an entity. When the relation is provided, only the user with that relation is removed, when not provided - all relations are removed.

- Related Issue #2069
- Resolves #2069

No

### Notes

Signed-off-by: WashingtonKK <washingtonkigan@gmail.com>

### DIFF
--- a/auth/postgres/domains.go
+++ b/auth/postgres/domains.go
@@ -378,20 +378,11 @@ func (repo domainRepo) DeletePolicies(ctx context.Context, pcs ...auth.Policy) (
 			AND subject_relation = :subject_relation
 			AND object_type = :object_type
 			AND object_id = :object_id
-		;`
+		`
 		if pc.Relation != "" {
-			q = `
-			DELETE FROM
-				policies
-			WHERE
-				subject_type = :subject_type
-				AND subject_id = :subject_id
-				AND subject_relation = :subject_relation
-				AND relation = :relation
-				AND object_type = :object_type
-				AND object_id = :object_id
-			;`
+			q = fmt.Sprintf("%s AND relation = :relation", q)
 		}
+		q = fmt.Sprintf("%s;", q)
 
 		dbpc := toDBPolicy(pc)
 		row, err := tx.NamedQuery(q, dbpc)
--- a/auth/postgres/domains.go
+++ b/auth/postgres/domains.go
@@ -370,6 +370,17 @@ func (repo domainRepo) DeletePolicies(ctx context.Context, pcs ...auth.Policy) (
 
 	for _, pc := range pcs {
 		q := `
+		DELETE FROM
+			policies
+		WHERE
+			subject_type = :subject_type
+			AND subject_id = :subject_id
+			AND subject_relation = :subject_relation
+			AND object_type = :object_type
+			AND object_id = :object_id
+		;`
+		if pc.Relation != "" {
+			q = `
 			DELETE FROM
 				policies
 			WHERE
@@ -380,6 +391,7 @@ func (repo domainRepo) DeletePolicies(ctx context.Context, pcs ...auth.Policy) (
 				AND object_type = :object_type
 				AND object_id = :object_id
 			;`
+		}
 
 		dbpc := toDBPolicy(pc)
 		row, err := tx.NamedQuery(q, dbpc)
--- a/auth/service.go
+++ b/auth/service.go
@@ -718,87 +718,76 @@ func (svc service) AssignUsers(ctx context.Context, token, id string, userIds []
 }
 
 func (svc service) UnassignUsers(ctx context.Context, token, id string, userIds []string, relation string) error {
-	if err := svc.Authorize(ctx, PolicyReq{
+	pr := PolicyReq{
 		Subject:     token,
 		SubjectType: UserType,
 		SubjectKind: TokenKind,
 		Object:      id,
 		ObjectType:  DomainType,
 		Permission:  SharePermission,
-	}); err != nil {
+	}
+	if err := svc.Authorize(ctx, pr); err != nil {
 		return err
 	}
 
 	if relation != "" {
-		if err := svc.Authorize(ctx, PolicyReq{
-			Subject:     token,
-			SubjectType: UserType,
-			SubjectKind: TokenKind,
-			Object:      id,
-			ObjectType:  DomainType,
-			Permission:  SwitchToPermission(relation),
-		}); err != nil {
+		pr.Permission = SwitchToPermission(relation)
+		if err := svc.Authorize(ctx, pr); err != nil {
 			return err
 		}
 
 		if err := svc.removeDomainPolicies(ctx, id, relation, userIds...); err != nil {
 			return errors.Wrap(errRemovePolicies, err)
 		}
-	} else {
-		if err := svc.Authorize(ctx, PolicyReq{
-			Subject:     token,
-			SubjectType: UserType,
-			SubjectKind: TokenKind,
-			Object:      id,
-			ObjectType:  DomainType,
-			Permission:  AdminPermission,
-		}); err != nil {
-			// User is not admin.
-			var ids []string
-			for _, userID := range userIds {
-				if err := svc.Authorize(ctx, PolicyReq{
-					Subject:     userID,
-					SubjectType: UserType,
-					SubjectKind: UsersKind,
-					Permission:  AdminPermission,
-					Object:      id,
-					ObjectType:  DomainType,
-				}); err != nil {
-					// Append all non-admins to ids.
-					ids = append(ids, userID)
-				}
-			}
-
-			// Remove only non-admins.
-			if err := svc.removeDomainPolicies(ctx, id, MemberRelation, ids...); err != nil {
-				return errors.Wrap(errRemovePolicies, err)
-			}
-
-			if err := svc.removeDomainPolicies(ctx, id, ViewerRelation, ids...); err != nil {
-				return errors.Wrap(errRemovePolicies, err)
-			}
-
-			if err := svc.removeDomainPolicies(ctx, id, EditorRelation, ids...); err != nil {
-				return errors.Wrap(errRemovePolicies, err)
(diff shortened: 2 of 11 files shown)


things/api/http/requests.go
```diff
@@ -245,10 +245,6 @@ func (req unassignUsersRequest) validate() error {
 		return apiutil.ErrBearerToken
 	}
```
arvindh123, line 246: Please remove Relation file from `unassignUsersReq`

auth/spicedb/policies.go
```diff
@@ -132,24 +132,13 @@ func (pa *policyAgent) AddPolicy(ctx context.Context, pr auth.PolicyReq) error {
...
-	}
-	_, err := pa.permissionClient.WriteRelationships(ctx, &v1.WriteRelationshipsRequest{Updates: updates})
-	if err != nil {
-		return errors.Wrap(errRemovePolicies, handleSpicedbError(err))
+		err := pa.DeletePolicy(ctx, pr)
+		if err != nil {
+			return err
+		}
 	}
+
```
arvindh123: @WashingtonKK 
`DeletePolicies` is different from `DeletePolicy`
`DeletePolicies` supports filteing of options, But DeletePolicy doesn't support filtering options

auth/service.go
```diff
@@ -730,15 +730,17 @@ func (svc service) UnassignUsers(ctx context.Context, token, id string, userIds
...
+			Subject:     token,
+			SubjectType: UserType,
+			SubjectKind: TokenKind,
+			Object:      id,
+			ObjectType:  DomainType,
+			Permission:  SwitchToPermission(relation),
+		}); err != nil {
+			return err
+		}
 	}
```
arvindh123: We should not allow editor to remove administrator. 
This check will be bypassed if relation is empty. 
We need to think this approch in different way

auth/service.go
```diff
@@ -729,20 +729,78 @@ func (svc service) UnassignUsers(ctx context.Context, token, id string, userIds
...
+		}); err != nil {
+			return err
+		}
 
-	if err := svc.removeDomainPolicies(ctx, id, relation, userIds...); err != nil {
-		return errors.Wrap(errRemovePolicies, err)
+		if err := svc.removeDomainPolicies(ctx, id, relation, userIds...); err != nil {
+			return errors.Wrap(errRemovePolicies, err)
+		}
+	} else {
```
dborovcanin: Avoid using `if-else` to make it consistent with the rest of the codebase. Also, try to simplify with something like:
```go
pr := PolicyReq{
		Subject:     token,
		SubjectType: UserType,
		SubjectKind: TokenKind,
		Object:      id,
		ObjectType:  DomainType,
		Permission:  SharePermission,
	}
	if err := svc.Authorize(ctx, pr); err != nil {
		return err
	}

	if relation != "" {
		pr.Permission = SwitchToPermission(relation)
		if err := svc.Authorize(ctx, pr); err != nil {
			return err
		}

		if err := svc.removeDomainPolicies(ctx, id, relation, userIds...); err != nil {
			return errors.Wrap(errRemovePolicies, err)
		}
		return nil
	}
	pr.Permission = AdminPermission
	if err := svc.Authorize(ctx, pr); err != nil {
		// User is not admin.
		var ids []string
		for _, userID := range userIds {
			if err := svc.Authorize(ctx, PolicyReq{
				Subject:     userID,
				SubjectType: UserType,
				SubjectKind: UsersKind,
				Permission:  AdminPermission,
				Object:      id,
				ObjectType:  DomainType,
			}); err != nil {
				// Append all non-admins to ids.
				ids = append(ids, userID)
			}
		}

		// Remove only non-admins.
		if err := svc.removeDomainPolicies(ctx, id, MemberRelation, ids...); err != nil {
			return errors.Wrap(errRemovePolicies, err)
		}

		if err := svc.removeDomainPolicies(ctx, id, ViewerRelation, ids...); err != nil {
			return errors.Wrap(errRemovePolicies, err)
		}

		if err := svc.removeDomainPolicies(ctx, id, EditorRelation, ids...); err != nil {
			return errors.Wrap(errRemovePolicies, err)
		}
	}

	// If user is admin, remove all policies from all users.
	if err := svc.removeDomainPolicies(ctx, id, MemberRelation, userIds...); err != nil {
		return errors.Wrap(errRemovePolicies, err)
	}

	if err := svc.removeDomainPolicies(ctx, id, ViewerRelation, userIds...); err != nil {
		return errors.Wrap(errRemovePolicies, err)
	}

	if err := svc.removeDomainPolicies(ctx, id, EditorRelation, userIds...); err != nil {
		return errors.Wrap(errRemovePolicies, err)
	}

	if err := svc.removeDomainPolicies(ctx, id, AdministratorRelation, userIds...); err != nil {
		return errors.Wrap(errRemovePolicies, err)
	}

	return nil
```

dborovcanin (changes requested):

auth/postgres/domains.go
```diff
@@ -368,6 +368,17 @@ func (repo domainRepo) DeletePolicies(ctx context.Context, pcs ...auth.Policy) (
 
 	for _, pc := range pcs {
 		q := `
+		DELETE FROM
```
rodneyosodo, line 373: Add tests for this

rodneyosodo (changes requested):

auth/postgres/domains.go
```diff
@@ -370,6 +370,17 @@ func (repo domainRepo) DeletePolicies(ctx context.Context, pcs ...auth.Policy) (
...
+			policies
+		WHERE
```
rodneyosodo, line 375: Try and make this section DRY. You can use this as an example https://github.com/rodneyosodo/magistrala/blob/9f2fa81532db77ce790db3e5a938e7188065f73c/activitylog/postgres/events.go#L46C1-L50C130

rodneyosodo (changes requested):

rodneyosodo (dismissed):
LGTM

auth/service.go
```diff
@@ -718,31 +718,78 @@ func (svc service) AssignUsers(ctx context.Context, token, id string, userIds []
...
+			return errors.Wrap(errRemovePolicies, err)
+		}
+
+		if err := svc.removeDomainPolicies(ctx, id, ViewerRelation, ids...); err != nil {
+			return errors.Wrap(errRemovePolicies, err)
+		}
+
+		if err := svc.removeDomainPolicies(ctx, id, EditorRelation, ids...); err != nil {
+			return errors.Wrap(errRemovePolicies, err)
+		}
```
dborovcanin: You can simplify with something like:
```suggestion
for _, rel := range []string{MemberRelation, ViewerRelation, EditorRelation} {
			// Remove only non-admins.
			if err := svc.removeDomainPolicies(ctx, id, rel, ids...); err != nil {
				return errors.Wrap(errRemovePolicies, err)
			}
		}
```

auth/service.go
```diff
@@ -718,31 +718,78 @@ func (svc service) AssignUsers(ctx context.Context, token, id string, userIds []
...
 		return errors.Wrap(errRemovePolicies, err)
 	}
+
+	if err := svc.removeDomainPolicies(ctx, id, EditorRelation, userIds...); err != nil {
+		return errors.Wrap(errRemovePolicies, err)
+	}
+
+	if err := svc.removeDomainPolicies(ctx, id, AdministratorRelation, userIds...); err != nil {
+		return errors.Wrap(errRemovePolicies, err)
+	}
```
dborovcanin: Same here.

dborovcanin (changes requested):
This starts to be a little cryptic and complex. We must simplify, but let's not do that in this PR.

rodneyosodo (dismissed):

auth/service.go
```diff
@@ -718,31 +718,63 @@ func (svc service) AssignUsers(ctx context.Context, token, id string, userIds []
...
+				ids = append(ids, userID)
+			}
+		}
+
+		for _, rel := range []string{MemberRelation, ViewerRelation, EditorRelation} {
+			// Remove only non-admins.
+			if err := svc.removeDomainPolicies(ctx, id, rel, ids...); err != nil {
+				return errors.Wrap(errRemovePolicies, err)
+			}
+		}
```
dborovcanin: You can replace this whole section, since you have the loop in the block below:
```suggestion
	userIds = ids
```

dborovcanin (changes requested):

dborovcanin (approved):

auth/postgres/domains.go
```diff
@@ -370,16 +370,19 @@ func (repo domainRepo) DeletePolicies(ctx context.Context, pcs ...auth.Policy) (
...
+			subject_type = :subject_type
+			AND subject_id = :subject_id
+			AND subject_relation = :subject_relation
+			AND object_type = :object_type
+			AND object_id = :object_id
+		`
+		if pc.Relation != "" {
+			q = fmt.Sprintf("%s AND relation = :relation", q)
```
arvindh123, line 383: @WashingtonKK  There will be only one relation in the Domain. 
So no need to  consider `relation` here

dborovcanin: @WashingtonKK Please check this one and send a follow-up PR for it.

PR #2191, opened 2024-04-19
MG-2092 - Remove unused field in token object

### LINKED ISSUE #2092: Feature: Remove unused Fields in the Token object

When a user is logged in but not to a specified domain, the token has the domain field as empty which is a bug on the UI when we are checking for token claims["domain"] it will be ok showing that the token has a domain field.

When the user is not logged in to a domain, the token produced during login should not contain the domain field.

Should-have

### Anything else?

_No response_

JeffMboya:
### What type of PR is this?
This is a bug fix because it fixes the issue where an unused field in the token object is causing a bug on the UI.

### What does this do?
This PR removes the unused domain field from the token object when a user is not logged in to a specific domain. This prevents the UI from incorrectly indicating that the token has a domain field when it does not.

Resolves #2092 

Yes

### Did you document any new/modified feature?
No


rodneyosodo (changes requested):
Remove commits not related to this PR

rodneyosodo (changes requested):
Try and test this PR. I have loggedin without domain login and still there are unused fields

```bash
USER_TOKEN=$(magistrala-cli users token admin@example.com 12345678 | jq -r .access_token)
```

```bash
jwt decode $USER_TOKEN


Token header
------------
{
  "typ": "JWT",
  "alg": "HS512"
}

Token claims
------------
{
  "exp": 1713534942,
  "iat": 1713531342,
  "iss": "magistrala.auth",
  "sub": "",
  "type": 0,
  "user": "bcc7b52f-70f8-43b9-8a60-ef9549b76db1"
}
```

Subject is unused

auth/jwt/token_test.go
```diff
@@ -76,8 +76,20 @@ func TestIssue(t *testing.T) {
...
+				ID:        testsutil.GenerateUUID(t),
+				Type:      auth.AccessKey,
+				Subject:   "",
+				User:      testsutil.GenerateUUID(t),
+				Domain:    "",
+				IssuedAt:  time.Now().Add(-10 * time.Second).Round(time.Second),
+				ExpiresAt: time.Now().Add(10 * time.Minute).Round(time.Second),
+			},
+			err: nil,
+		},
```
rodneyosodo, line 119: Add the following case:
- issue token without domain
- issue token without subject

auth/jwt/token_test.go
```diff
@@ -148,6 +164,12 @@ func TestParse(t *testing.T) {
 			token: newToken(issuerName, key()),
 			err:   authjwt.ErrJSONHandle,
 		},
+		{
```
rodneyosodo, line 207: Add the following cases
- parse token with an empty domain
- parse token with empty subject

rodneyosodo (changes requested):

auth/jwt/token_test.go
```diff
@@ -106,6 +142,19 @@ func TestParse(t *testing.T) {
 
 	inValidToken := newToken("invalid", key())
 
+	createKeyAndToken := func(domain, subject string) (auth.Key, string) {
```
rodneyosodo: No need for this function refactor `newToken` to provide a token when the subject or domain is empty

rodneyosodo (changes requested):

auth/jwt/token_test.go
```diff
@@ -106,6 +146,11 @@ func TestParse(t *testing.T) {
 
 	inValidToken := newToken("invalid", key())
 
+	emptyKey := key()
+	emptyKey.Domain = ""
+	emptyKey.Subject = ""
+	emptyToken, err := tokenizer.Issue(emptyKey)
+
```
rodneyosodo, line 164: Create 3 different keys:
- empty domain
- empty subject
- both empty

rodneyosodo (changes requested):

rodneyosodo (approved):
LGTM

dborovcanin (approved):

Issue #2142, opened 2024-04-08
Consume Things connect event in Bootstrap


JeffMboya:
### Is your feature request related to a problem? Please describe.

Currently, Bootstrap ```events.go``` handles ```removeEvent```, ```updateChannelEvent```, and ```disconnectEvent``` in the ```consumer``` package. However, we do not have a mechanism to consume a ```connectEvent``` when a Thing connect event

### Describe the feature you are requesting, as well as the possible use case(s) for it.

Addition of a ```connectEvent``` struct in the ```consumer``` package. This struct would contain the ```thingID``` and ```channelID```, similar to the ```disconnectEvent``` shown below. This event would be triggered whenever a Thing connects to the system.
![image](https://github.com/absmach/magistrala/assets/44696487/d7b47a33-30c3-4355-9c4d-5726346fd3d2)

### Indicate the importance of this feature to you.

Must-have

### Anything else?

Here's a rough idea of what the independent `connectEvent` struct might look like:

```go
type connectEvent struct {
	thingID   string
	channelID string
}
```

Alternatively, define a connection event that consumes both ```Thing connect``` and ```Thing disconnect``` as below:
```go
type connectionEvent struct {
	chanID  string
	thingID string
}

func decodeConnectionThing(event map[string]interface{}) connectionThingEvent {
	return connectionThingEvent{
		chanID:  read(event, "chan_id", ""),
		thingID: read(event, "thing_id", ""),
	}
func (es *eventHandler) Handle(ctx context.Context, event events.Event) error {
	msg, err := event.Encode()
	if err != nil {
		return err
	}
	switch msg["operation"] {
	case thingConnect:
		tce := decodeConnectionThing(msg)
		err = es.svc.ConnectThing(ctx, tce.chanID, tce.thingID)
	case thingDisconnect:
		tde := decodeConnectionThing(msg)
		err = es.svc.DisconnectThing(ctx, tde.chanID, tde.thingID)
	}
	if err != nil && err != errMetadataType {
		return err
	}

	return nil
}
```
This feature request is based on the current code in the `consumer` package:

dborovcanin:
This is related to #1955.

PR #2199, opened 2024-04-25
MG-1955 - Update Bootstrap service access control

### LINKED ISSUE #1955: Update Bootstrap service access control

List Bootstrap configurations 

- I should only be able to view bootstrap configurations of the things belonging to the domain I am currently logged into. 

- Also, as a member of a domain, I should be able to see the configurations for the things in the domain, whether I am the one who created the configuration or not, as long as I have the correct permissions to view the configuration for the thing.

- I can see all my bootstrap configurations regardless of the domain they belong to.
- I am not able to see the configurations for things that belong to a domain I am a member of even if I have the necessary permissions in that domain. I am only able to see the configurations I create.

### Steps To Reproduce

1. log in
2. Create domain (`domain 1`)
3. Log into `domain 1`
4. Create thing (`thing 1`)
5. Create channel (`channel 1`) (optional) 
6. Create a new bootstrap configuration for `thing 1`
7. Create another domain
8. log into `domain 2`
9. list `bootstraps`

1. Create a new user (`user 1`)
2. Add `user 1` to `domain 1`
3. log in to `domain 1` as `user 1`
4. list `bootstraps`

Bootstrap service

_No response_

JeffMboya:
# What type of PR is this?
This PR is a bug fix.

## What does this do?
This PR ensures that users can only view bootstrap configurations of the things within their current domain. It also allows domain members with appropriate permissions to view configurations for things within the domain, regardless of who created the configuration.

- Resolves #1955 
- Related to #2142

Yes, tests have been included in this PR.

No new features were documented in this PR.


bootstrap/postgres/init.go
```diff
@@ -13,38 +13,38 @@ func Migration() *migrate.MemoryMigrationSource {
 				Id: "configs_1",
 				Up: []string{
 					`CREATE TABLE IF NOT EXISTS configs (
-						mainflux_thing TEXT UNIQUE NOT NULL,
```
rodneyosodo: Use migrations

rodneyosodo (changes requested):

bootstrap/configs.go
```diff
@@ -17,7 +17,7 @@ import (
 // MGChannels is a list of Magistrala Channels corresponding Magistrala Thing connects to.
 type Config struct {
 	ThingID     string    `json:"thing_id"`
-	Owner       string    `json:"owner,omitempty"`
+	Owner       string    `json:"domain_id,omitempty"`
```
rodneyosodo:
```suggestion
	DomainID       string    `json:"domain_id,omitempty"`
```

Replace this everywhere

bootstrap/configs.go
```diff
@@ -82,23 +82,23 @@ type ConfigRepository interface {
 	// to indicate operation failure.
 	Update(ctx context.Context, cfg Config) error
 
-	// UpdateCerts updates and returns an existing Config certificate and owner.
+	// UpdateCerts updates and returns an existing Config certificate and domainID.
 	// A non-nil error is returned to indicate operation failure.
-	UpdateCert(ctx context.Context, owner, thingID, clientCert, clientKey, caCert string) (Config, error)
+	UpdateCert(ctx context.Context, domain_id, thingID, clientCert, clientKey, caCert string) (Config, error)
```
rodneyosodo: use camelcase for go naming. Have a look at https://go.dev/doc/effective_go

```suggestion
	UpdateCert(ctx context.Context, domainID, thingID, clientCert, clientKey, caCert string) (Config, error)
```

bootstrap/postgres/configs.go
```diff
@@ -483,27 +483,27 @@ func (cr configRepository) retrieveAll(owner string, filter bootstrap.Filter) (s
...
 
-func (cr configRepository) rollback(content string, tx *sqlx.Tx) {
+func (cr configRepository) rollback(_ string, tx *sqlx.Tx) {
 	if err := tx.Rollback(); err != nil {
 		cr.log.Error(fmt.Sprintf("Failed to rollback due to %s", err))
 	}
 }
 
-func insertChannels(_ context.Context, owner string, channels []bootstrap.Channel, tx *sqlx.Tx) error {
+func insertChannels(_ context.Context, domain_id string, channels []bootstrap.Channel, tx *sqlx.Tx) error {
```
rodneyosodo: If context is unused remove it

bootstrap/postgres/init.go
```diff
@@ -83,6 +83,15 @@ func Migration() *migrate.MemoryMigrationSource {
...
 			},
+			{
+				Id: "configs_5",
+				Up: []string{
+					`ALTER TABLE IF EXISTS configs RENAME COLUMN owner TO domain_id`,
+					`ALTER TABLE IF EXISTS channels RENAME COLUMN owner TO domain_id`,
+					`ALTER TABLE IF EXISTS connections RENAME COLUMN channel_owner TO channel_domain_id`,
+					`ALTER TABLE IF EXISTS connections RENAME COLUMN config_owner TO config_domain_id`,
+				},
+			},
```
rodneyosodo, line 93: I should be able top create a bootstrap config with the same name on different domains

bootstrap/service.go
```diff
@@ -378,13 +445,37 @@ func (bs bootstrapService) DisconnectThingHandler(ctx context.Context, channelID
...
 	ctx, cancel := context.WithTimeout(ctx, time.Second)
 	defer cancel()
 
 	res, err := bs.auth.Identify(ctx, &magistrala.IdentityReq{Token: token})
 	if err != nil {
-		return "", errors.Wrap(svcerr.ErrAuthentication, err)
+		return nil, svcerr.ErrAuthentication
+	}
+	if res.GetId() == "" || res.GetDomainId() == "" {
+		return nil, svcerr.ErrDomainAuthorization
```
rodneyosodo:
```suggestion
		return nil, svcerr.ErrAuthentication
```

rodneyosodo (changes requested):

bootstrap/service.go
```diff
@@ -117,26 +118,42 @@ type bootstrapService struct {
...
+	// If domain is disabled , then this authorization will fail for all non-admin domain users.
+	if _, err := bs.authorize(ctx, "", auth.UserType, auth.UsersKind, user.GetId(), auth.MembershipPermission, auth.DomainType, user.GetDomainId()); err != nil {
+		return Config{}, err
+	}
+
+	for _, channel := range cfg.Channels {
+		if channel.ID == "" || channel.ID == "invalid" {
+			return Config{}, svcerr.ErrMalformedEntity
+		}
+	}
```
arvindh123: This can be moved to API layer

bootstrap/service.go
```diff
@@ -176,57 +193,94 @@ func (bs bootstrapService) Add(ctx context.Context, token string, cfg Config) (C
...
 		return Config{}, errors.Wrap(svcerr.ErrViewEntity, err)
 	}
+
+	// Check if the thing is part of the same domain as the user.
+	thing, err := bs.sdk.Thing(cfg.ThingID, token)
+	if err != nil {
+		return Config{}, errors.Wrap(svcerr.ErrViewEntity, err)
+	}
+
+	if thing.DomainID != user.GetDomainId() {
```
arvindh123: How this case will be possible?   
Is there way to add to a thing which is not belongs same domain of user?

bootstrap/service.go
```diff
@@ -307,12 +370,16 @@ func (bs bootstrapService) Bootstrap(ctx context.Context, externalKey, externalI
 }
 
 func (bs bootstrapService) ChangeState(ctx context.Context, token, id string, state State) error {
-	owner, err := bs.identify(ctx, token)
+	user, err := bs.identify(ctx, token)
 	if err != nil {
 		return errors.Wrap(svcerr.ErrAuthentication, err)
 	}
+	_, err = bs.authorize(ctx, "", auth.UserType, auth.UsersKind, user.GetId(), auth.EditPermission, auth.DomainType, user.GetDomainId())
```
arvindh123: @JeffMboya Why we are checking the user have edit access to Domain ?

arvindh123 (changes requested):

bootstrap/api/endpoint.go
```diff
@@ -22,6 +22,9 @@ func addEndpoint(svc bootstrap.Service) endpoint.Endpoint {
 		channels := []bootstrap.Channel{}
 		for _, c := range req.Channels {
 			channels = append(channels, bootstrap.Channel{ID: c})
+			if c == "" || c == "invalid" {
```
rodneyosodo: Is there a point that c == "invalid" this should not be the case ?

bootstrap/service.go
```diff
@@ -380,13 +397,36 @@ func (bs bootstrapService) DisconnectThingHandler(ctx context.Context, channelID
...
+		Permission:  perm,
+		ObjectType:  objType,
+		Object:      obj,
+	}
+	res, err := bs.auth.Authorize(ctx, req)
+	if err != nil {
+		return "", errors.Wrap(svcerr.ErrAuthorization, err)
+	}
+	if !res.GetAuthorized() {
+		return "", errors.Wrap(svcerr.ErrAuthorization, err)
```
rodneyosodo, line 509:
```suggestion
		return "", svcerr.ErrAuthorization
```

bootstrap/postgres/configs.go
```diff
@@ -483,27 +483,27 @@ func (cr configRepository) retrieveAll(owner string, filter bootstrap.Filter) (s
 	return fmt.Sprintf(template, f), params
 }
 
-func (cr configRepository) rollback(content string, tx *sqlx.Tx) {
+func (cr configRepository) rollback(_ string, tx *sqlx.Tx) {
```
rodneyosodo: remove the first argument if it is not being used

bootstrap/postgres/configs.go
```diff
@@ -537,21 +537,21 @@ func insertConnections(_ context.Context, cfg bootstrap.Config, connections []st
 	return err
 }
 
-func updateConnections(_ context.Context, owner, id string, connections []string, tx *sqlx.Tx) error {
+func updateConnections(_ context.Context, domainID, id string, connections []string, tx *sqlx.Tx) error {
```
rodneyosodo: same

bootstrap/service.go
```diff
@@ -380,13 +397,36 @@ func (bs bootstrapService) DisconnectThingHandler(ctx context.Context, channelID
...
-		return "", errors.Wrap(svcerr.ErrAuthentication, err)
+		return nil, svcerr.ErrAuthentication
+	}
+	if res.GetId() == "" || res.GetDomainId() == "" {
+		return nil, svcerr.ErrAuthentication
+	}
+	return res, nil
+}
+
+func (bs bootstrapService) authorize(ctx context.Context, domainID, subjType, subjKind, subj, perm, objType, obj string) (string, error) {
```
rodneyosodo: If authorize is being called with the same parameters we can reduce the number of parameters and use constants for example 
```diff
- Permission: perm,
+ Permission: auth.MembershipPermission,
```

rodneyosodo (changes requested):

bootstrap/postgres/init.go
JeffMboya: You should be able to, with the current codebase just as it is

bootstrap/api/requests_test.go
```diff
@@ -13,41 +13,67 @@ import (
 )
 
 func TestAddReqValidation(t *testing.T) {
+	const (
```
rodneyosodo: Define this at the top of the file and use `testsutil.GenerateUUID(&testing.T{})` to generate UUIDs

bootstrap/api/requests_test.go
```diff
@@ -13,41 +13,67 @@ import (
...
+			externalKey: "",
+			channels:    []string{channel1, channel2},
+			err:         apiutil.ErrMissingID,
+		},
+		{
+			desc:        "empty channel",
+			token:       "token",
+			externalID:  "external-id",
+			externalKey: "external-key",
+			channels:    []string{""},
```
rodneyosodo:
```suggestion
			channels:    []string{},
```

bootstrap/api/requests_test.go
```diff
@@ -13,41 +13,67 @@ import (
 )
 
 func TestAddReqValidation(t *testing.T) {
+	const (
+		channel1 = "e4d2aaaf-23f5-45e3-a713-f6cbdc2c9d9d"
+		channel2 = "77bc4ac6-42f2-4385-af1d-c60408197940"
+	)
+
```
rodneyosodo, line 20: Add a valid request

bootstrap/service.go
```diff
@@ -176,57 +181,61 @@ func (bs bootstrapService) Add(ctx context.Context, token string, cfg Config) (C
...
 
 func (bs bootstrapService) View(ctx context.Context, token, id string) (Config, error) {
-	owner, err := bs.identify(ctx, token)
+	user, err := bs.identify(ctx, token)
 	if err != nil {
 		return Config{}, errors.Wrap(svcerr.ErrAuthentication, err)
 	}
-	cfg, err := bs.configs.RetrieveByID(ctx, owner, id)
+
+	// Retrieve the bootstrap configuration.
```
rodneyosodo: remove comment

rodneyosodo: Same as things service we need to authorize the users has `ViewPermission` to the thing if the boostrap config has thingID

rodneyosodo: This also applies to update operations `EditPermission`

rodneyosodo (changes requested):

bootstrap/service.go
```diff
@@ -176,57 +181,79 @@ func (bs bootstrapService) Add(ctx context.Context, token string, cfg Config) (C
...
 func (bs bootstrapService) Update(ctx context.Context, token string, cfg Config) error {
-	owner, err := bs.identify(ctx, token)
+	user, err := bs.identify(ctx, token)
 	if err != nil {
 		return errors.Wrap(svcerr.ErrAuthentication, err)
 	}
 
-	cfg.Owner = owner
+	cfg.DomainID = user.GetDomainId()
+	_, err = bs.authorize(ctx, "", auth.TokenKind, token, auth.EditPermission, auth.DomainType, cfg.DomainID)
```
rodneyosodo: Since we have `userID` from identify we can user `usersKind` and `userID`

rodneyosodo (changes requested):

bootstrap/api/requests.go
```diff
@@ -36,6 +36,16 @@ func (req addReq) validate() error {
 		return apiutil.ErrBearerKey
 	}
 
+	if len(req.Channels) == 0 {
+		return apiutil.ErrMissingID
```
rodneyosodo:
```suggestion
		return apiutil.ErrEmptyList
```

bootstrap/service.go
```diff
@@ -176,57 +181,76 @@ func (bs bootstrapService) Add(ctx context.Context, token string, cfg Config) (C
 }
 
 func (bs bootstrapService) View(ctx context.Context, token, id string) (Config, error) {
-	owner, err := bs.identify(ctx, token)
+	user, err := bs.identify(ctx, token)
 	if err != nil {
 		return Config{}, errors.Wrap(svcerr.ErrAuthentication, err)
 	}
-	cfg, err := bs.configs.RetrieveByID(ctx, owner, id)
+	_, err = bs.authorize(ctx, user.GetDomainId(), auth.UsersKind, user.GetId(), auth.ViewPermission, auth.ThingType, id)
```
rodneyosodo: We should authorize if the bootstrap has a thingID

bootstrap/service.go
```diff
@@ -176,57 +181,76 @@ func (bs bootstrapService) Add(ctx context.Context, token string, cfg Config) (C
...
 	return cfg, nil
 }
 
 func (bs bootstrapService) Update(ctx context.Context, token string, cfg Config) error {
-	owner, err := bs.identify(ctx, token)
+	user, err := bs.identify(ctx, token)
 	if err != nil {
 		return errors.Wrap(svcerr.ErrAuthentication, err)
 	}
+	_, err = bs.authorize(ctx, user.GetDomainId(), auth.UsersKind, user.GetId(), auth.EditPermission, auth.ThingType, cfg.ThingID)
```
rodneyosodo: What if `ThingID` is empty?

bootstrap/service.go
```diff
@@ -176,57 +181,76 @@ func (bs bootstrapService) Add(ctx context.Context, token string, cfg Config) (C
...
 }
 
 func (bs bootstrapService) UpdateCert(ctx context.Context, token, thingID, clientCert, clientKey, caCert string) (Config, error) {
-	owner, err := bs.identify(ctx, token)
+	user, err := bs.identify(ctx, token)
 	if err != nil {
 		return Config{}, errors.Wrap(svcerr.ErrAuthentication, err)
 	}
-	cfg, err := bs.configs.UpdateCert(ctx, owner, thingID, clientCert, clientKey, caCert)
+	_, err = bs.authorize(ctx, user.GetDomainId(), auth.UsersKind, user.GetId(), auth.EditPermission, auth.ThingType, thingID)
```
rodneyosodo: same

bootstrap/service.go
```diff
@@ -262,26 +286,47 @@ func (bs bootstrapService) UpdateConnections(ctx context.Context, token, id stri
...
 	if err != nil {
 		return ConfigsPage{}, errors.Wrap(svcerr.ErrAuthentication, err)
 	}
-	return bs.configs.RetrieveAll(ctx, owner, filter, offset, limit), nil
+
+	configsPage := bs.configs.RetrieveAll(ctx, user.GetDomainId(), filter, offset, limit)
+
+	j := 0
+	for _, config := range configsPage.Configs {
+		_, err := bs.authorize(ctx, user.GetDomainId(), auth.UsersKind, user.GetId(), auth.ViewPermission, auth.ThingType, config.ThingID)
```
rodneyosodo: This doesn't seem good since we will make n calls to auth depending on the list of configs. You can check something like this https://github.com/absmach/magistrala/pull/2168. So if they aren't domain admins they get basic information. @arvindh123 what do you think?

rodneyosodo (changes requested):

bootstrap/service.go
arvindh123: If `ThingID` is empty then authorization will fails

bootstrap/service.go
arvindh123: Yes, This should be like, first we need to get all `thingID`s from spiceDB for user and then pass the `thingID`s to config repo to retrieve the bootstrap config

bootstrap/service.go
arvindh123: In additon, In API layer we validate the ThingID , If ThingID is empty, API will retun error

arvindh123 (changes requested):

bootstrap/api/requests.go
rodneyosodo: This has not been addressed

bootstrap/configs.go
```diff
@@ -69,11 +69,11 @@ type ConfigRepository interface {
 
 	// RetrieveByID retrieves the Config having the provided identifier, that is owned
 	// by the specified user.
-	RetrieveByID(ctx context.Context, owner, id string) (Config, error)
+	RetrieveByID(ctx context.Context, domainID, id string) (Config, error)
 
 	// RetrieveAll retrieves a subset of Configs that are owned
 	// by the specific user, with given filter parameters.
-	RetrieveAll(ctx context.Context, owner string, filter Filter, offset, limit uint64) ConfigsPage
+	RetrieveAll(ctx context.Context, domainID, thingID string, filter Filter, offset, limit uint64) ConfigsPage
```
rodneyosodo: This should be
```suggestion
	RetrieveAll(ctx context.Context, domainID string, thingIDs []string, filter Filter, offset, limit uint64) ConfigsPage
```

Where we can retrieve configs for a specific domain or configs for a list of things.

bootstrap/service.go
rodneyosodo: inline

bootstrap/service.go
rodneyosodo: inline

bootstrap/service.go
```diff
@@ -262,26 +286,112 @@ func (bs bootstrapService) UpdateConnections(ctx context.Context, token, id stri
...
 	if err != nil {
 		return ConfigsPage{}, errors.Wrap(svcerr.ErrAuthentication, err)
 	}
-	return bs.configs.RetrieveAll(ctx, owner, filter, offset, limit), nil
+
+	if err := bs.checkSuperAdmin(ctx, user.GetId()); err == nil {
+		return bs.configs.RetrieveAll(ctx, user.GetDomainId(), "", filter, offset, limit), nil
+	}
+
+	// Check user is admin of domain, if yes then show listing on domain context.
```
rodneyosodo: Remove comment

bootstrap/service.go
```diff
@@ -262,26 +286,112 @@ func (bs bootstrapService) UpdateConnections(ctx context.Context, token, id stri
...
+	_, authErr := bs.authorize(ctx, "", auth.UsersKind, user.GetId(), auth.AdminPermission, auth.DomainType, user.GetDomainId())
+	if authErr == nil {
+		return bs.configs.RetrieveAll(ctx, user.GetDomainId(), "", filter, offset, limit), nil
+	}
+
+	rtids, err := bs.listClientIDs(ctx, user.GetId(), auth.ViewPermission)
```
rodneyosodo:
```suggestion
	rtids, err := bs.listClientIDs(ctx, user.GetId())
```
Since we are using it at one place

bootstrap/service.go
```diff
@@ -262,26 +286,112 @@ func (bs bootstrapService) UpdateConnections(ctx context.Context, token, id stri
...
+	if err != nil {
+		return ConfigsPage{}, errors.Wrap(svcerr.ErrNotFound, err)
+	}
+
+	thingIDs, err := bs.filterAllowedThingIDs(ctx, user.GetId(), auth.ViewPermission, rtids)
```
rodneyosodo: same

bootstrap/service.go
```diff
@@ -262,26 +286,112 @@ func (bs bootstrapService) UpdateConnections(ctx context.Context, token, id stri
...
+	if err != nil {
+		return ConfigsPage{}, errors.Wrap(svcerr.ErrNotFound, err)
+	}
+
+	var configs []Config
+	for _, thingID := range thingIDs {
+		configPage := bs.configs.RetrieveAll(ctx, user.GetDomainId(), thingID, filter, offset, limit)
```
rodneyosodo: This should not be in a for loop

```suggestion
	configPage := bs.configs.RetrieveAll(ctx, user.GetDomainId(), thingIDs, filter, offset, limit)
```

rodneyosodo (changes requested):

bootstrap/postgres/configs.go
```diff
@@ -460,50 +460,58 @@ func (cr configRepository) DisconnectThing(ctx context.Context, channelID, thing
...
-	queries := []string{""}
-	// Since owner is the first param, start from 2.
-	counter := 2
+func (cr configRepository) retrieveAll(domainID string, thingIDs []string, filter bootstrap.Filter) (string, []interface{}) {
+	params := []interface{}{}
+	queries := []string{}
+
+	if domainID != "" {
+		params = append(params, domainID)
+		queries = append(queries, fmt.Sprintf("domain_id = $%d", len(params)))
```
rodneyosodo: Why not?

```suggestion
		queries = append(queries, fmt.Sprintf("domain_id = %s", domainID))
```

bootstrap/service.go
```diff
@@ -117,26 +118,30 @@ type bootstrapService struct {
...
 	}
 }
 
 func (bs bootstrapService) Add(ctx context.Context, token string, cfg Config) (Config, error) {
-	owner, err := bs.identify(ctx, token)
+	user, err := bs.identify(ctx, token)
 	if err != nil {
 		return Config{}, errors.Wrap(svcerr.ErrAuthentication, err)
 	}
+	// If domain is disabled , then this authorization will fail for all non-admin domain users.
```
rodneyosodo: Remove this comment

bootstrap/service.go
```diff
@@ -262,26 +281,100 @@ func (bs bootstrapService) UpdateConnections(ctx context.Context, token, id stri
...
 	if err != nil {
 		return ConfigsPage{}, errors.Wrap(svcerr.ErrAuthentication, err)
 	}
-	return bs.configs.RetrieveAll(ctx, owner, filter, offset, limit), nil
+
+	if err := bs.checkSuperAdmin(ctx, user.GetId()); err == nil {
+		return bs.configs.RetrieveAll(ctx, user.GetDomainId(), []string{}, filter, offset, limit), nil
+	}
+
+	_, authErr := bs.authorize(ctx, "", auth.UsersKind, user.GetId(), auth.AdminPermission, auth.DomainType, user.GetDomainId())
```
rodneyosodo:
```suggestion
	if _, err := bs.authorize(ctx, "", auth.UsersKind, user.GetId(), auth.AdminPermission, auth.DomainType, user.GetDomainId()); err != nil {
```

bootstrap/service.go
```diff
@@ -380,13 +473,36 @@ func (bs bootstrapService) DisconnectThingHandler(ctx context.Context, channelID
...
 
-func (bs bootstrapService) identify(ctx context.Context, token string) (string, error) {
+func (bs bootstrapService) identify(ctx context.Context, token string) (*magistrala.IdentityRes, error) {
 	ctx, cancel := context.WithTimeout(ctx, time.Second)
 	defer cancel()
 
 	res, err := bs.auth.Identify(ctx, &magistrala.IdentityReq{Token: token})
 	if err != nil {
-		return "", errors.Wrap(svcerr.ErrAuthentication, err)
+		return nil, svcerr.ErrAuthentication
```
rodneyosodo: wrap the error
```suggestion
		return nil, errors.Wrap(svcerr.ErrAuthentication, err)
```

bootstrap/service.go
```diff
@@ -380,13 +473,36 @@ func (bs bootstrapService) DisconnectThingHandler(ctx context.Context, channelID
...
+		SubjectType: auth.UserType,
+		SubjectKind: subjKind,
+		Subject:     subj,
+		Permission:  perm,
+		ObjectType:  objType,
+		Object:      obj,
+	}
+	res, err := bs.auth.Authorize(ctx, req)
+	if err != nil {
+		return "", svcerr.ErrAuthorization
```
rodneyosodo: same
```suggestion
		return "", errors.Wrap(svcerr.ErrAuthorization, err)
```

rodneyosodo (changes requested):

bootstrap/postgres/configs.go
JeffMboya: I think doing so would be risking SQL injection attacks

bootstrap/service.go
```diff
@@ -262,26 +280,99 @@ func (bs bootstrapService) UpdateConnections(ctx context.Context, token, id stri
...
+	})
+	if err != nil {
+		return nil, errors.Wrap(svcerr.ErrNotFound, err)
+	}
+	return tids.Policies, nil
+}
+
+func (bs bootstrapService) filterAllowedThingIDs(ctx context.Context, userID string, thingIDs []string) ([]string, error) {
+	var ids []string
+	tids, err := bs.auth.ListAllObjects(ctx, &magistrala.ListObjectsReq{
```
arvindh123: `bs.auth.ListAllObjects` is already called in `listClientIDs` , Why it is called again here. 
Please explain the reason

arvindh123 (changes requested):

bootstrap/events/producer/streams_test.go
```diff
@@ -107,44 +109,55 @@ func TestAdd(t *testing.T) {
...
 	}
 
 	lastID := "0"
 	for _, tc := range cases {
-		repoCall := auth.On("Identify", mock.Anything, &magistrala.IdentityReq{Token: tc.token}).Return(&magistrala.IdentityRes{Id: validID}, nil)
-		repoCall1 := sdk.On("Thing", mock.Anything, mock.Anything).Return(mgsdk.Thing{ID: tc.config.ThingID, Credentials: mgsdk.Credentials{Secret: tc.config.ThingKey}}, errors.NewSDKError(tc.err))
-		repoCall2 := sdk.On("Channel", mock.Anything, mock.Anything).Return(toChannel(tc.config.Channels[0]), errors.NewSDKError(tc.err))
-		repoCall3 := boot.On("ListExisting", context.Background(), mock.Anything, mock.Anything).Return(tc.config.Channels, tc.err)
-		repoCall4 := boot.On("Save", context.Background(), mock.Anything, mock.Anything).Return(mock.Anything, tc.err)
+		authCall := auth.On("Identify", mock.Anything, &magistrala.IdentityReq{Token: tc.token}).Return(&magistrala.IdentityRes{Id: validID, DomainId: domainID}, tc.identifyErr)
```
rodneyosodo: `validID` and `domainID` are fixed and for all requests `identifyErr` is nil?

bootstrap/events/producer/streams_test.go
```diff
@@ -107,44 +109,55 @@ func TestAdd(t *testing.T) {
...
+		authCall1 := auth.On("Authorize", context.Background(), mock.Anything).Return(tc.authResponse, tc.authorizeErr)
```
rodneyosodo, line 208: All the times `authorizeErr` is nil

bootstrap/events/producer/streams_test.go
```diff
@@ -107,44 +109,55 @@ func TestAdd(t *testing.T) {
...
+		sdkCall := sdk.On("Thing", tc.config.ThingID, tc.token).Return(mgsdk.Thing{ID: tc.config.ThingID, Credentials: mgsdk.Credentials{Secret: tc.config.ThingKey}}, errors.NewSDKError(tc.authorizeErr))
```
rodneyosodo: The error here shouldn't be `authorizeErr`

bootstrap/events/producer/streams_test.go
```diff
@@ -107,44 +109,55 @@ func TestAdd(t *testing.T) {
...
+		sdkCall1 := sdk.On("Channel", channel.ID, tc.token).Return(toChannel(tc.config.Channels[0]), errors.NewSDKError(tc.channelErr))
+		svcCall := boot.On("ListExisting", context.Background(), domainID, mock.Anything).Return(tc.config.Channels, tc.listErr)
```
rodneyosodo: This is a repoCall

bootstrap/events/producer/streams_test.go
```diff
@@ -107,44 +109,55 @@ func TestAdd(t *testing.T) {
...
+		svcCall1 := boot.On("Save", context.Background(), mock.Anything, mock.Anything).Return(mock.Anything, tc.saveErr)
```
rodneyosodo: Same to this

bootstrap/events/producer/streams_test.go
```diff
@@ -162,41 +175,52 @@ func TestAdd(t *testing.T) {
...
+		authCall.Unset()
+		authCall1.Unset()
+		sdkCall.Unset()
+		sdkCall1.Unset()
+		svcCall.Unset()
+		svcCall1.Unset()
 	}
 }
 
 func TestView(t *testing.T) {
```
rodneyosodo, line 238: What are we testing here, because we don't check the emitted event is what we expect?

bootstrap/events/producer/streams_test.go
```diff
@@ -212,19 +236,24 @@ func TestUpdate(t *testing.T) {
...
-	repoCall3 := boot.On("ListExisting", context.Background(), mock.Anything, mock.Anything).Return(c.Channels, nil)
-	repoCall4 := boot.On("Save", context.Background(), mock.Anything, mock.Anything).Return(mock.Anything, nil)
+
+	authCall := auth.On("Identify", mock.Anything, &magistrala.IdentityReq{Token: validToken}).Return(&magistrala.IdentityRes{Id: validID, DomainId: domainID}, nil)
+	authCall1 := auth.On("Authorize", context.Background(), mock.Anything).Return(&magistrala.AuthorizeRes{Authorized: true}, nil)
+	sdkCall := sdk.On("Thing", c.ThingID, validToken).Return(mgsdk.Thing{ID: c.ThingID, Credentials: mgsdk.Credentials{Secret: c.ThingKey}}, nil)
+	sdkCall1 := sdk.On("Channel", ch.ID, validToken).Return(toChannel(c.Channels[0]), nil)
+	svcCall := boot.On("ListExisting", context.Background(), domainID, mock.Anything).Return(c.Channels, nil)
+	svcCall1 := boot.On("Save", context.Background(), mock.Anything, mock.Anything).Return(mock.Anything, nil)
+
```
rodneyosodo, line 356: There is no need to simulate `Add` operation because we are not persisting the value. This applies to everywhere

bootstrap/events/producer/streams_test.go
```diff
@@ -241,17 +270,22 @@ func TestUpdate(t *testing.T) {
...
-		event  map[string]interface{}
+		desc         string
+		config       bootstrap.Config
+		token        string
+		authResponse *magistrala.AuthorizeRes
+		authorizeErr error
+		identifyErr  error
+		updateErr    error
+		err          error
+		event        map[string]interface{}
```
rodneyosodo, line 377: Add the following cases this applies to other places too
- update with failed to identify a user
- update with failed to authorize user
- update with failed to fetch thing
- update with failed to fetch channel
- update with failed to fetch config
- update with failed to list existing
- update with failed to update connections

bootstrap/events/producer/streams_test.go
```diff
@@ -545,40 +605,47 @@ func TestUpdateCert(t *testing.T) {
...
 
-		repoCall.Unset()
-		repoCall1.Unset()
+		authCall.Unset()
+		authCall1.Unset()
+		svcCall.Unset()
 	}
 }
 
 func TestList(t *testing.T) {
```
rodneyosodo, line 774: What are we testing here since we are not checking the emitted event

rodneyosodo (changes requested):

bootstrap/postgres/configs.go
```diff
@@ -460,50 +460,58 @@ func (cr configRepository) DisconnectThing(ctx context.Context, channelID, thing
...
-	return fmt.Sprintf(template, f), params
+	return "WHERE " + f, params
 }
 
-func (cr configRepository) rollback(content string, tx *sqlx.Tx) {
+func (cr configRepository) rollback(tx *sqlx.Tx) {
 	if err := tx.Rollback(); err != nil {
 		cr.log.Error(fmt.Sprintf("Failed to rollback due to %s", err))
 	}
 }
```
arvindh123: Wrap and return the error , Don't login the error here.

bootstrap/postgres/configs.go
```diff
@@ -64,36 +64,36 @@ func (cr configRepository) Save(ctx context.Context, cfg bootstrap.Config, chsCo
...
 
 	if err := insertConnections(ctx, cfg, chsConnIDs, tx); err != nil {
-		cr.rollback("Failed to insert connections", tx)
+		cr.rollback(tx)
 		return "", errors.Wrap(errSaveConnections, err)
 	}
 
 	if err := tx.Commit(); err != nil {
-		cr.rollback("Failed to commit Config save", tx)
+		cr.rollback(tx)
```
arvindh123: Please call `cr.rollback ` inside `defer` instead of mutiple `cr.rollback`

bootstrap/postgres/configs.go
```diff
@@ -460,50 +460,58 @@ func (cr configRepository) DisconnectThing(ctx context.Context, channelID, thing
...
 
 	f := strings.Join(queries, " AND ")
 
-	return fmt.Sprintf(template, f), params
+	return "WHERE " + f, params
 }
 
-func (cr configRepository) rollback(content string, tx *sqlx.Tx) {
+func (cr configRepository) rollback(tx *sqlx.Tx) {
 	if err := tx.Rollback(); err != nil {
```
arvindh123, line 516: Please provide `content` string, This helps to identity at which step error has occured during debug.

```suggestion
func (cr configRepository) rollback(content string, tx *sqlx.Tx) {
	if err := tx.Rollback(); err != nil {
```

arvindh123 (changes requested):

bootstrap/service.go
```diff
@@ -262,26 +307,90 @@ func (bs bootstrapService) UpdateConnections(ctx context.Context, token, id stri
...
+	var ids []string
+	for _, thingID := range thingIDs {
+		for _, tid := range thingIDs {
+			if thingID == tid {
+				ids = append(ids, thingID)
+			}
+		}
+	}
+	return ids, nil
+}
```
arvindh123: This function work doesn't make sense. 
Please verify 
Do we need filter Allowd ThingsIDs ?

arvindh123 (changes requested):

bootstrap/service.go
JeffMboya: @arvindh123 on second check, since `bs.auth.ListAllObjects` is already called, we don't need to `filterAllowedThingsIDs`

bootstrap/postgres/configs.go
```diff
@@ -451,59 +460,86 @@ func (cr configRepository) RemoveChannel(ctx context.Context, id string) error {
...
 	q := `UPDATE configs SET state = $1 WHERE EXISTS (
 		SELECT 1 FROM connections WHERE config_id = $2 AND channel_id = $3)`
-	if _, err := cr.db.ExecContext(ctx, q, bootstrap.Inactive, thingID, channelID); err != nil {
+	result, err := cr.db.ExecContext(ctx, q, bootstrap.Inactive, thingID, channelID)
+	if err != nil {
 		return errors.Wrap(errDisconnectThing, err)
 	}
+	if rows, _ := result.RowsAffected(); rows == 0 {
+		return repoerr.ErrNotFound
+	}
```
arvindh123: Please explain why we need this error.

bootstrap/postgres/configs.go
```diff
@@ -451,59 +460,86 @@ func (cr configRepository) RemoveChannel(ctx context.Context, id string) error {
...
-	// Since owner is the first param, start from 2.
-	counter := 2
+func (cr configRepository) retrieveAll(domainID string, thingIDs []string, filter bootstrap.Filter) (string, []interface{}) {
+	params := []interface{}{}
+	queries := []string{}
+
+	if domainID != "" {
+		params = append(params, domainID)
+		queries = append(queries, fmt.Sprintf("domain_id = $%d", len(params)))
+	}
```
arvindh123: Why we made domainID as optional here ? 
In any case domainID might be empty?

bootstrap/postgres/init.go
```diff
@@ -83,6 +83,16 @@ func Migration() *migrate.MemoryMigrationSource {
 					`ALTER TABLE IF EXISTS channels RENAME COLUMN mainflux_channel TO magistrala_channel`,
 				},
 			},
+			{
+				Id: "configs_5",
+				Up: []string{
+					`ALTER TABLE IF EXISTS configs RENAME COLUMN owner TO domain_id`,
+					`ALTER TABLE IF EXISTS channels RENAME COLUMN owner TO domain_id`,
+					`ALTER TABLE IF EXISTS connections RENAME COLUMN channel_owner TO channel_domain_id`,
+					`ALTER TABLE IF EXISTS connections RENAME COLUMN config_owner TO config_domain_id`,
```
arvindh123: There is no need for these two  `channel_domain_id` and `config_domain_id`, because both chanenl and domain should be in same domain. 
We should impose this validation, while bootstrap creation.  
We will get the channel and thing domain id while retrieving their information form things service. 
There we can validate both thing and channel are belong to same domain.

bootstrap/postgres/configs.go
```diff
@@ -47,53 +48,57 @@ func NewConfigRepository(db postgres.Database, log *slog.Logger) bootstrap.Confi
...
 	if err := insertConnections(ctx, cfg, chsConnIDs, tx); err != nil {
-		cr.rollback("Failed to insert connections", tx)
-		return "", errors.Wrap(errSaveConnections, err)
-	}
-
-	if err := tx.Commit(); err != nil {
-		cr.rollback("Failed to commit Config save", tx)
+		err = errors.Wrap(errSaveConnections, err)
 		return "", err
 	}
```
arvindh123: What do you think about the below suggestion
```suggestion
	if err := insertConnections(ctx, cfg, chsConnIDs, tx); err != nil {
		return "", errors.Wrap(errSaveConnections, err)
	}

```

bootstrap/postgres/configs.go
```diff
@@ -47,53 +48,57 @@ func NewConfigRepository(db postgres.Database, log *slog.Logger) bootstrap.Confi
...
+		return "", err
 	}
 
-	if err := insertChannels(ctx, cfg.Owner, cfg.Channels, tx); err != nil {
-		cr.rollback("Failed to insert Channels", tx)
-		return "", errors.Wrap(errSaveChannels, err)
+	if err := insertChannels(cfg.DomainID, cfg.Channels, tx); err != nil {
+		err = errors.Wrap(errSaveChannels, err)
+		return "", err
 	}
```
arvindh123: What do you think about below suggestion
```suggestion
	if err := insertChannels(cfg.DomainID, cfg.Channels, tx); err != nil {
		return "", err = errors.Wrap(errSaveChannels, err)
	}
```

bootstrap/postgres/configs.go
```diff
@@ -47,53 +48,57 @@ func NewConfigRepository(db postgres.Database, log *slog.Logger) bootstrap.Confi
...
 
 	dbcfg := toDBConfig(cfg)
+	defer func() {
+		if err != nil {
+			err = cr.rollback("Save method", err, tx)
+		} else {
+			if commitErr := tx.Commit(); commitErr != nil {
+				err = commitErr
+			}
```
arvindh123: Why you have  move the `tx.Commit()` to defer statement ? .

arvindh123 (changes requested):
Please change/remove  the owner at here https://github.com/jeffmboya/magistrala/blob/MG-1955/bootstrap/events/producer/streams_test.go#L398 and https://github.com/jeffmboya/magistrala/blob/MG-1955/bootstrap/service.go#L86

Check and modify this comment if needed at here https://github.com/jeffmboya/magistrala/blob/MG-1955/bootstrap/service.go#L90

bootstrap/postgres/configs.go
JeffMboya: The error was initially added to inform the caller that the specified `thingID` and `channelID` pair do not exist in Magistrala-platform. However, we have previously agreed that when disconnecting, we don't need to care about the existence of this `thingID` and `channelID`  pair. If they don't exist, it means they are already disconnected. Soooo I'll fix this to return something like a `HTTP status 204` to indicate successful disconnection, whether or not the connection existed.

bootstrap/postgres/configs.go
JeffMboya: I made `domainID` optional to allow for more flexible querying.

bootstrap/postgres/configs.go
JeffMboya: I think it is more compact

bootstrap/postgres/configs.go
JeffMboya: I think it is more compact

bootstrap/postgres/configs.go
arvindh123: But If domainID becomes empty, then it should not return any config, 
having a test case for this will be good.

bootstrap/postgres/init.go
```diff
@@ -83,6 +83,14 @@ func Migration() *migrate.MemoryMigrationSource {
 					`ALTER TABLE IF EXISTS channels RENAME COLUMN mainflux_channel TO magistrala_channel`,
 				},
 			},
+			{
+				Id: "configs_5",
+				Up: []string{
+					`ALTER TABLE IF EXISTS configs RENAME COLUMN owner TO domain_id`,
+					`ALTER TABLE IF EXISTS channels RENAME COLUMN owner TO domain_id`,
+					`ALTER TABLE IF EXISTS configs ADD CONSTRAINT configs_name_domain_id_key UNIQUE (name, domain_id)`,
```
arvindh123, line 91: There is no need for these two `channel_domain_id` and `config_domain_id`, because both chanenl and thing should be in same domain. 
We should impose this validation, while bootstrap creation.
We will get the channel and thing domain id while retrieving their information form things service. 
There we can validate both thing and channel are belong to same domain.

Here please add new column `domain_id` and remove  `channel_domain_id` , `config_domain_id`

arvindh123 (changes requested):

bootstrap/postgres/configs.go
```diff
@@ -472,62 +475,65 @@ func (cr configRepository) ConnectThing(ctx context.Context, channelID, thingID
...
 }
 
-func (cr configRepository) retrieveAll(owner string, filter bootstrap.Filter) (string, []interface{}) {
-	template := `WHERE owner = $1 %s`
-	params := []interface{}{owner}
-	// One empty string so that strings Join works if only one filter is applied.
-	queries := []string{""}
-	// Since owner is the first param, start from 2.
-	counter := 2
+func (cr configRepository) retrieveAll(domainID string, thingIDs []string, filter bootstrap.Filter) (string, []interface{}) {
```
arvindh123: Do you think , here need for `retrieveAll` as method of `configRepository` struct ? 
And additonally, this function can be renamed to `buildQueryParams` or `buildRetreiveQueryParams`

bootstrap/service.go
```diff
@@ -158,8 +162,14 @@ func (bs bootstrapService) Add(ctx context.Context, token string, cfg Config) (C
 		return Config{}, errors.Wrap(errThingNotFound, err)
 	}
 
+	for _, channel := range cfg.Channels {
+		if channel.DomainID != mgThing.DomainID {
+			return Config{}, svcerr.ErrAddPolicies
+		}
```
arvindh123:
```suggestion
			return Config{}, errors.Wrap(svcerr.ErrMalformedEntity, errNotInSameDomain)
```

arvindh123 (changes requested):

bootstrap/postgres/configs.go
JeffMboya: I think it should remain within the struct for consistency and separation of concerns. The name `retrieveAll` implies that the method is responsible for retrieving data from the database, which is misleading as it just builds the query parameters. So yes, a name like `buildRetrieveQueryParams` more accurately reflects its function.

JeffMboya: I think it should remain within the struct for consistency and separation of concerns. The name `retrieveAll` implies that the method is responsible for retrieving data from the database, which is misleading as it just builds the query parameters. So yes, a name like `buildRetrieveQueryParams` more accurately reflects its function.

bootstrap/service.go
```diff
@@ -158,8 +165,14 @@ func (bs bootstrapService) Add(ctx context.Context, token string, cfg Config) (C
 		return Config{}, errors.Wrap(errThingNotFound, err)
 	}
 
+	for _, channel := range cfg.Channels {
+		if channel.DomainID != mgThing.DomainID {
+			return Config{}, errors.Wrap(svcerr.ErrMalformedEntity, errNotInSameDomain)
```
arvindh123, line 170: I could not add config  and getting error , Because thing domain id is empty. 
This happens because in thing service , the json have field name `domain` but `sdk.Thing` expects field with name `domain_id`
To fix this please change here to `domain_id` https://github.com/absmach/magistrala/blob/main/pkg/clients/clients.go#L45
Or update the sdk.thing domain field json tag to `domain`

arvindh123 (changes requested):

bootstrap/service.go
```diff
@@ -123,26 +127,29 @@ type bootstrapService struct {
...
 	}
+	if _, err := bs.authorize(ctx, "", auth.UsersKind, user.GetId(), auth.MembershipPermission, auth.DomainType, user.GetDomainId()); err != nil {
+		return Config{}, err
+	}
 
 	toConnect := bs.toIDList(cfg.Channels)
 
 	// Check if channels exist. This is the way to prevent fetching channels that already exist.
-	existing, err := bs.configs.ListExisting(ctx, owner, toConnect)
+	existing, err := bs.configs.ListExisting(ctx, user.GetDomainId(), toConnect)
```
arvindh123, line 152: @JeffMboya Please check thing domain id and exitisting channel domain id are equal

arvindh123 (changes requested):

arvindh123 (approved):
LGTM

arvindh123 (dismissed):

dborovcanin (approved):

PR #2234, opened 2024-05-14
MG-2201 - Fail role update if user does not exist

### LINKED ISSUE #2201: Feature: Fail role update if user does not exist

During update user role for non existence user id, doesn't fails in policy layer, Instead it fails in repo layer after adding policy in spiceDB them it rollback the policy. 

Instead of checking the user existence in repo layer, we should check user id is belongs to Magistrala Platform as member in policy layer( spiceDB).

Should-have

### Anything else?

_No response_

JeffMboya:
# What type of PR is this?

This is a feature because it adds the following functionality: client check check in policy layer when updating client role.

## What does this do?

This PR modifies the update client role functionality to check for client existence in the policy layer (spiceDB) instead of the repo layer. This prevents the system from adding a policy for a non-existent user and then having to roll it back.

- Resolves #2201 

No

No


auth/api/grpc/server.go
```diff
@@ -53,12 +53,13 @@ var (
 
 type grpcServer struct {
 	magistrala.UnimplementedAuthServiceServer
-	issue           kitgrpc.Handler
-	refresh         kitgrpc.Handler
-	identify        kitgrpc.Handler
-	authorize       kitgrpc.Handler
-	addPolicy       kitgrpc.Handler
-	addPolicies     kitgrpc.Handler
+	issue       kitgrpc.Handler
```
rodneyosodo: Revert

rodneyosodo (changes requested):
Since we check if the user is SuperAdmin first before updating the role, then they are certainly in the platform @arvindh123 what do you think?

arvindh123 (approved):

arvindh123 (commented):
> Since we check if the user is SuperAdmin first before updating the role, then they are certainly in the platform @arvindh123 what do you think? 

@rodneyosodo 
In previous part of the code, we are checking the Request User is SuperAdmin or Normal User. 
we are not checking the role change user ID .
It is like super admin can make other user as super admin or normal user
In role change API endpoint, SuperAdmin will be sending the user id of the other user to change the role. 
So we should check the user id given by super admin in request , is exists in platform or not.

users/service_test.go
```diff
@@ -1075,7 +1075,7 @@ func TestUpdateClientRole(t *testing.T) {
 			updateRoleResponse:   mgclients.Client{},
 			token:                validToken,
 			deletePolicyErr:      svcerr.ErrMalformedEntity,
-			err:                  svcerr.ErrDeletePolicies,
```
arvindh123: @JeffMboya 
Instead of this changing this test response, Please add new test case for invalid / non existance user

arvindh123 (changes requested):

arvindh123 (dismissed):

users/service.go
```diff
@@ -694,6 +694,19 @@ func (svc service) addClientPolicyRollback(ctx context.Context, userID string, r
 }
 
 func (svc service) updateClientPolicy(ctx context.Context, userID string, role mgclients.Role) error {
+	res, err := svc.auth.Authorize(ctx, &magistrala.AuthorizeReq{
```
rodneyosodo: This authorization should happen once not during rolling back. This is because when we roll back we are sure the user is in the platform

```go
client := mgclients.Client{
	ID:        cli.ID,
	Role:      cli.Role,
	UpdatedAt: time.Now(),
	UpdatedBy: tokenUserID,
}

if _, err := svc.authorize(ctx, auth.UserType, auth.UsersKind, tokenUserID, auth.MembershipPermission, auth.PlatformType, auth.MagistralaObject); err != nil {
	return mgclients.Client{}, err
}

if err := svc.updateClientPolicy(ctx, cli.ID, cli.Role); err != nil {
	return mgclients.Client{}, err
}
```

rodneyosodo (changes requested):

users/service.go
JeffMboya: @rodneyosodo, since we want to verify if the user whose role the super admin wants to update is on the platform, I think the subject should be `client.ID`.

rodneyosodo: Update accordingly

users/service_test.go
```diff
@@ -1101,6 +1103,14 @@ func TestUpdateClientRole(t *testing.T) {
 			updateRoleErr:        svcerr.ErrAuthentication,
 			err:                  svcerr.ErrAuthentication,
 		},
+		{
```
rodneyosodo, line 1161: Add missing test case. You can run the following command and look at the HTML output for the missing case
```bash
cd users && go test -v --race -covermode=atomic -coverprofile=cover.out . && go tool cover -html cover.out -o cover.html
```

rodneyosodo (changes requested):

users/service.go
JeffMboya: Okay

users/service_test.go
```diff
@@ -1101,6 +1103,22 @@ func TestUpdateClientRole(t *testing.T) {
...
+			err:               svcerr.ErrAuthorization,
+		},
+		{
+			desc:             "Update client role for non-existent user",
+			client:           mgclients.Client{},
+			token:            validToken,
+			identifyResponse: &magistrala.IdentityRes{},
+			identifyErr:      svcerr.ErrAuthorization,
+			err:              svcerr.ErrAuthorization,
+		},
```
WashingtonKK: These test cases are not correct since the mock call for authorize returns `tc.authorizeErr` yet you have only provided `err` meaning that authorize will not return an error when it should.

WashingtonKK:
```suggestion
			desc:              "update client role with failed MembershipPermission authorization",
			client:            client,
			identifyResponse:  &magistrala.IdentityRes{UserId: client.ID},
			authorizeResponse: &magistrala.AuthorizeRes{Authorized: false},
			token:             validToken,
			err:               svcerr.ErrAuthorization,
            authorizeErr:      svcerr.ErrAuthorization,
		},
		{
			desc:             "Update client role for non-existent user",
			client:           mgclients.Client{},
			token:            validToken,
			identifyResponse: &magistrala.IdentityRes{},
			identifyErr:      svcerr.ErrAuthorization,
			err:              svcerr.ErrAuthorization,
            authorizeErr:      svcerr.ErrAuthorization,
		},
```

users/service_test.go
```diff
@@ -1101,6 +1103,22 @@ func TestUpdateClientRole(t *testing.T) {
 			updateRoleErr:        svcerr.ErrAuthentication,
 			err:                  svcerr.ErrAuthentication,
 		},
+		{
+			desc:              "update client role with failed MembershipPermission authorization",
```
WashingtonKK, line 1162: This test does not cover your changes for a failed membership permission. It fails at `svc.checkSuperAdmin` when checking for admin permissions at `svc.auth.Authorize`.

Update the test and add a mock call which will pass in this case, and cover your changes.

users/service_test.go
```diff
@@ -1101,6 +1103,22 @@ func TestUpdateClientRole(t *testing.T) {
...
+			authorizeResponse: &magistrala.AuthorizeRes{Authorized: false},
+			token:             validToken,
+			err:               svcerr.ErrAuthorization,
+		},
+		{
+			desc:             "Update client role for non-existent user",
+			client:           mgclients.Client{},
+			token:            validToken,
+			identifyResponse: &magistrala.IdentityRes{},
+			identifyErr:      svcerr.ErrAuthorization,
```
WashingtonKK: If `svc.Identify` fails, it means that the token provided is not valid, or the user who is trying to update the role does not exist and not that the client is non-existent.

Effectively this means this test will still also not cover your changes for this PR.

WashingtonKK (changes requested):
Please update tests.

WashingtonKK (dismissed):
LGTM

arvindh123 (approved):

rodneyosodo (approved):

rodneyosodo (dismissed):

users/service.go
```diff
@@ -558,7 +563,7 @@ func (svc *service) checkSuperAdmin(ctx context.Context, adminID string) error {
 		if err := svc.clients.CheckSuperAdmin(ctx, adminID); err != nil {
 			return errors.Wrap(svcerr.ErrAuthorization, err)
 		}
-		return errors.Wrap(svcerr.ErrAuthorization, err)
+		return nil
```
arvindh123: @JeffMboya   Could please explain the reason for this change. 
This should not be like this, If the user is not SuperAdmin in policies, then it should send Authorization error

arvindh123 (changes requested):

arvindh123 (dismissed):

felixgateru (dismissed):

rodneyosodo (dismissed):

dborovcanin (commented):
@JeffMboya Please resolve conflicts.

arvindh123 (approved):

rodneyosodo (approved):

dborovcanin (approved):

PR #2263, opened 2024-05-29
NOISSUE: Fix timescale reader

Musilah:
# What type of PR is this?
This PR fixes a slight bug with the reader. To assist the charts development this adds new columns in the return statement of the aggregation query.

## What does this do?
This add a protocol and publisher column to the return statement that will be used by the React Chart. These are needed for the timeseries charts as well as the pie charts. 

None.

No.

No.

### DIFF
--- a/readers/timescale/messages.go
+++ b/readers/timescale/messages.go
@@ -42,8 +42,8 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
 
 	// If aggregation is provided, add time_bucket and aggregation to the query
 	if rpm.Aggregation != "" {
-		q = fmt.Sprintf(`SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) *1000000 AS time, %s(value) AS value FROM %s WHERE %s GROUP BY 1 ORDER BY time DESC LIMIT :limit OFFSET :offset;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
-		totalQuery = fmt.Sprintf(`SELECT COUNT(*) FROM (SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) AS time, %s(value) AS value FROM %s WHERE %s GROUP BY 1) AS subquery;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
+		q = fmt.Sprintf(`SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) *1000000 AS time,publisher,protocol, %s(value) AS value FROM %s WHERE %s GROUP BY 1, publisher, protocol ORDER BY time DESC LIMIT :limit OFFSET :offset;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
+		totalQuery = fmt.Sprintf(`SELECT COUNT(*) FROM (SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) AS time,publisher,protocol, %s(value) AS value FROM %s WHERE %s GROUP BY 1, publisher, protocol) AS subquery;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
 	}
 
 	params := map[string]interface{}{
--- a/readers/timescale/messages.go
+++ b/readers/timescale/messages.go
@@ -43,7 +43,7 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
 	// If aggregation is provided, add time_bucket and aggregation to the query
 	if rpm.Aggregation != "" {
 		q = fmt.Sprintf(`SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) *1000000 AS time,publisher,protocol, %s(value) AS value FROM %s WHERE %s GROUP BY 1, publisher, protocol ORDER BY time DESC LIMIT :limit OFFSET :offset;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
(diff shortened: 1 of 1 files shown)


readers/timescale/messages.go
```diff
@@ -42,8 +42,8 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
...
 	if rpm.Aggregation != "" {
-		q = fmt.Sprintf(`SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) *1000000 AS time, %s(value) AS value FROM %s WHERE %s GROUP BY 1 ORDER BY time DESC LIMIT :limit OFFSET :offset;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
-		totalQuery = fmt.Sprintf(`SELECT COUNT(*) FROM (SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) AS time, %s(value) AS value FROM %s WHERE %s GROUP BY 1) AS subquery;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
+		q = fmt.Sprintf(`SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) *1000000 AS time,publisher,protocol, %s(value) AS value FROM %s WHERE %s GROUP BY 1, publisher, protocol ORDER BY time DESC LIMIT :limit OFFSET :offset;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
+		totalQuery = fmt.Sprintf(`SELECT COUNT(*) FROM (SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) AS time,publisher,protocol, %s(value) AS value FROM %s WHERE %s GROUP BY 1, publisher, protocol) AS subquery;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
```
arvindh123: Do we need `GROUP BY` for `COUNT` ?

arvindh123 (changes requested):

readers/timescale/messages.go
```diff
@@ -42,8 +42,8 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
 
 	// If aggregation is provided, add time_bucket and aggregation to the query
 	if rpm.Aggregation != "" {
-		q = fmt.Sprintf(`SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) *1000000 AS time, %s(value) AS value FROM %s WHERE %s GROUP BY 1 ORDER BY time DESC LIMIT :limit OFFSET :offset;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
-		totalQuery = fmt.Sprintf(`SELECT COUNT(*) FROM (SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) AS time, %s(value) AS value FROM %s WHERE %s GROUP BY 1) AS subquery;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
+		q = fmt.Sprintf(`SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) *1000000 AS time,publisher,protocol, %s(value) AS value FROM %s WHERE %s GROUP BY 1, publisher, protocol ORDER BY time DESC LIMIT :limit OFFSET :offset;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
```
rodneyosodo: Is there a reason we are not returning the other fields like subtopic, name etc.?

rodneyosodo (changes requested):

readers/timescale/messages.go
Musilah: Yes, we do need the `GROUP BY` for the query to be complete. Without it the request returns an error. However we do not need the 1 that follows the `GROUP BY`.

ianmuchyri: we don't need to return the other values in count since they are not being utilized. For the `totalQuery`, let it remain as it was before

ianmuchyri (dismissed):

readers/timescale/messages.go
```diff
@@ -42,7 +42,7 @@ func (tr timescaleRepository) ReadAll(chanID string, rpm readers.PageMetadata) (
 
 	// If aggregation is provided, add time_bucket and aggregation to the query
 	if rpm.Aggregation != "" {
-		q = fmt.Sprintf(`SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) *1000000 AS time, %s(value) AS value FROM %s WHERE %s GROUP BY 1 ORDER BY time DESC LIMIT :limit OFFSET :offset;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
+		q = fmt.Sprintf(`SELECT EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) *1000000 AS time, publisher, protocol, subtopic, name, unit, %s(value) AS value FROM %s WHERE %s GROUP BY 1, publisher, protocol, subtopic, name, unit ORDER BY time DESC LIMIT :limit OFFSET :offset;`, rpm.Interval, rpm.Aggregation, format, fmtCondition(rpm))
```
arvindh123: In PostgreSQL, the `GROUP BY 1` clause is used to refer to the first column in the SELECT list, which in this case it is the `time` column derived from `EXTRACT(epoch FROM time_bucket('%s', to_timestamp(time/1000000))) *1000000`.

Do we need group remove `BY 1`,  i.e group `BY time` ?
I don't think we need.

arvindh123 (changes requested):

readers/timescale/messages.go
Musilah: https://docs.timescale.com/api/latest/hyperfunctions/time_bucket/

From this documentation we need the `GROUP BY TIME` to ensure the timebuckets work as expected.

arvindh123: Okay I agree, In that case can we have like `GROUP BY time` instead of `1`, Precisely by column name ?

arvindh123 (approved):

arvindh123 (approved):

dborovcanin (approved):

PR #2274, opened 2024-06-05
MG-2236 - Add Constraints Service Provider

WashingtonKK:
# What type of PR is this?

This is a feature because it adds a constraint provider to magistrala.

## What does this do?

This introduces a constraint provider to magistrala that reads constraints from a toml file and uses this to limit the entities created.

- Resolves absmach/amdm#98 


arvindh123:
```
|-- constraints
|   |-- config
|   |   |-- configprovider.go
|--constraints.go
```
@WashingtonKK   What  do you think about this folder structure, inside `pkg`  dir 
But again it like similar to repo folder structure

arvindh123:
> ```
> |-- constraints
> |   |-- config
> |   |   |-- configprovider.go
> |--constraints.go
> ```
> 
> @WashingtonKK What do you think about this folder structure, inside `pkg` dir But again it like similar to repo folder structure

In future we can add other providers 

like example 

```
|-- constraints
|   |-- config
|   |   |-- configprovider.go
|   |-- strip
|   |   |-- stripprovider.go
|--constraints.go
```

WashingtonKK:
> > ```
> > |-- constraints
> > |   |-- config
> > |   |   |-- configprovider.go
> > |--constraints.go
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @WashingtonKK What do you think about this folder structure, inside `pkg` dir But again it like similar to repo folder structure
> 
> In future we can add other providers
> 
> like example
> 
> ```
> |-- constraints
> |   |-- config
> |   |   |-- configprovider.go
> |   |-- strip
> |   |   |-- stripprovider.go
> |--constraints.go
> ```

@arvindh123 I think this is a good idea, since it will make it easy for us to add more providers. I will update it to that

constraints_config.toml
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) Abstract Machines
+# SPDX-License-Identifier: Apache-2.0
+
+users = "10"
+things = "10"
+domains = "17"
+channels = "10"
+groups = "10"
+msg_rate_limit = "10"
```
arvindh123: It will great if we have something like 
```yaml
[things]
  [things.limits]
    create = 10
    update = 10


[users]
  [users.limits]
    create = 10
    update = 10
```

constraints.go
```diff
@@ -0,0 +1,19 @@
...
+	Groups       uint32 `toml:"groups"`
+	Channels     uint32 `toml:"channels"`
+	MsgRateLimit uint32 `toml:"msg_rate_limit"`
+}
+
+// ConstraintsProvider specifies an API for obtaining entity constraints.
+type ConstraintsProvider interface {
+	// Constraints returns constraints for the entities.
+	Constraints() (Constraints, error)
+}
```
arvindh123:
```suggestion
// Copyright (c) Abstract Machines
// SPDX-License-Identifier: Apache-2.0

package magistrala

type Operation int

const (
	Create Operation = iota
)

// Constraints specifies an API for obtaining entity constraints.
type Constraints interface {
	// Constraints returns constraints for the entities.
	CheckLimits(operation Operation) error
}
```

something like this expected

pkg/constraints/constraints.go
```diff
@@ -0,0 +1,25 @@
...
+	return &constraintProvider{}
+}
+
+func (c *constraintProvider) Constraints() (magistrala.Constraints, error) {
+	consts, err := config.ParseConstraints()
+	if err != nil {
+		return magistrala.Constraints{}, err
+	}
+	return consts, nil
+}
```
arvindh123: Remove this file

constraints.go
arvindh123:
```suggestion
// Copyright (c) Abstract Machines
// SPDX-License-Identifier: Apache-2.0

package magistrala

type Operation int

const (
	Create Operation = iota
)

// Constraints specifies an API for obtaining entity constraints.
type Constraints interface {
	// Constraints returns constraints for the entities.
	CheckLimits(operation Operation, currentValue int64) error
}
```

pkg/constraints/config/configProvider.go
```diff
@@ -0,0 +1,120 @@
...
+	return c, nil
+}
+
+func convertStringToUint32(str string) (uint32, error) {
+	parsedValue, err := strconv.ParseUint(str, 10, 32)
+	if err != nil {
+		return 0, err
+	}
+	return uint32(parsedValue), nil
+}
```
arvindh123:
```suggestion
// Copyright (c) Abstract Machines
// SPDX-License-Identifier: Apache-2.0

package config

import (
	"fmt"

	"github.com/BurntSushi/toml"
	"github.com/absmach/magistrala"
)

const (
	defaultConstraintPath = "./constraints_config.toml"
	filePermission        = 0o644
)

type Limits struct {
	Create int64 `toml:"create"`
	Update int64 `toml:"update"`
}

type Constraints struct {
	Limits Limits `toml:"limits"`
}

type tomlConfig struct {
	Services map[string]Constraints
}

// Attempts to read constraints from the default path, if the file does not exist,
// it will be created with the default constraints.
func New(serviceName string) (magistrala.Constraints, error) {
	return read(defaultConstraintPath, serviceName)
}

func read(file, serviceName string) (magistrala.Constraints, error) {
	var tc tomlConfig
	if _, err := toml.DecodeFile(file, &tc.Services); err != nil {
		return nil, fmt.Errorf("Error reading config file: %s", err)
	}

	svcConstraint, exists := tc.Services[serviceName]
	if !exists {
		return nil, fmt.Errorf("Error: section [%s] not found", serviceName)
	}
	return svcConstraint, nil
}

func (c Constraints) CheckLimits(operation magistrala.Operation, currentValue int64) error {
	return nil
}
```

arvindh123 (changes requested):
Please try the suggetion

constraints.go
```diff
@@ -0,0 +1,15 @@
...
+type Operation int
+
+const (
+	Create Operation = iota
+)
+
+// Constraints specifies an API for obtaining entity constraints.
+type Constraints interface {
+	// Constraints returns constraints for the entities.
+	CheckLimits(operation Operation, currentValue uint64) error
```
dborovcanin, line 14: Is there any other type of limit that will not be `uint64`? I suspect at least float64 is a better fit, but maybe we should consider `any` as well.

constraints.go
```diff
@@ -0,0 +1,15 @@
...
+package magistrala
+
+type Operation int
+
+const (
+	Create Operation = iota
+)
+
+// Constraints specifies an API for obtaining entity constraints.
+type Constraints interface {
```
dborovcanin, line 12: Rename to `ConstraintProvider`.

dborovcanin (changes requested):
Interface definitions and configs should probably not be in the root, let's find them a better package.

things/service.go
```diff
@@ -68,6 +70,16 @@ func (svc service) CreateThings(ctx context.Context, token string, cls ...mgclie
 	if _, err := svc.authorize(ctx, "", auth.UserType, auth.UsersKind, user.GetId(), auth.CreatePermission, auth.DomainType, user.GetDomainId()); err != nil {
 		return []mgclients.Client{}, err
 	}
+	ths, err := svc.clients.RetrieveAll(ctx, mgclients.Page{Role: mgclients.AllRole})
+	if err != nil {
+		return []mgclients.Client{}, errors.Wrap(svcerr.ErrViewEntity, err)
+	}
```
arvindh123, line 76: This should be moved to inside the constraint provider `CheckLimits`

We should have generic repo function in all the service to retrieve `Count` of entitt

constraints_config.toml
```diff
@@ -0,0 +1,27 @@
...
+
+[channels]
+[channels.limits]
+create = 10
+update = 10
+
+[groups]
+[groups.limits]
+create = 10
+update = 10
```
arvindh123, line 27: May be we should switch to YAML file

constraints.go
```diff
@@ -0,0 +1,15 @@
...
+	// Constraints returns constraints for the entities.
+	CheckLimits(operation Operation, currentValue uint64) error
+}
```
arvindh123, line 15: We should think about domains in constraint provider instead of passing current value. 
Current value should be retrieved by Check function itself
```suggestion
	Check(operation Operation, domainID string) error

```

arvindh123 (changes requested):

constraints_config.toml
arvindh123: YAML file should have default config, which apply for all domain, 
And it should have domain specific constraint, it will be applied for specific domain, 
If domain specific constraint is not available then it should consider for default config

dborovcanin:
We will be using middlewares for this, to keep Service relatively independent. Also, some constraints may refer to certain types of limits that fit better to API (filtering by API key, advanced rate-limiting...).

PR #2288, opened 2024-06-13
MG-2287 - Improve Search For Users

### LINKED ISSUE #1976: Bug: New user to the platform sees all users

Bootstrap magistrala i.e add random users, things and channels then create a new user and try to view the index page using the UI

To be able not to see any users since I have logged in on a new domain i.e I should not be able to see the users

A new user on the platform can see all users in other domains. Is this the intended behaviour? If we are to have a multi-tenant platform, then this is not the intended behaviour. A user should only be able to see users in his/her domain.

### Steps To Reproduce

1. Start magistrala

	```bash
	make run
	```

2. Run the following command to create different users at least twice 

	```bash
	go run tools/e2e/cmd/main.go -n 10 -N 1
	```
3. Create a new user

	```bash
	magistrala-cli users create demo demo@demo.mail 12345678
	```

4. Login the user to UI and view the dashboard

Git commit - https://github.com/absmach/magistrala/commit/492ce14baff1d40ef4ff567e6955690219212e44
Docker Compose version 2.23.3
Docker version

> 	Client:
> 	 Version:           24.0.7
> 	 API version:       1.43
> 	 OS/Arch:           linux/amd64
> 	
> 	Server:
> 	 Engine:
> 	  Version:          24.0.7
> 	  API version:      1.43 (minimum version 1.12)

image

### LINKED ISSUE #2287: Improve search for Users, Things, Channels, and  Groups

nyagamunene:
# What type of PR is this?
This is a feature because it improve search for Users.

## What does this do?
This enables back-end support for filtering by name.

- Related Issue #2287 
- Resolves #2287
- Resolves #1976.


dborovcanin (changes requested):
Let's make the search case insensitive.

pkg/clients/postgres/clients.go
```diff
@@ -500,16 +500,19 @@ func PageQuery(pm clients.Page) (string, error) {
...
 
-func constructSearchQuery(pm clients.Page) (string, string) {
+func ConstructSearchQuery(pm clients.Page) (string, string) {
 	var query []string
 	var emq string
 	var tq string
 
 	if pm.Name != "" {
-		query = append(query, "name ~ :name")
+		query = append(query, "name ~* :name")
```
dborovcanin: Use LIKE ili ILIKE instead of regexp. It's a better fit and faster.

pkg/clients/postgres/clients.go
```diff
@@ -500,16 +500,19 @@ func PageQuery(pm clients.Page) (string, error) {
...
 	}
 	if pm.Identity != "" {
-		query = append(query, "identity ~ :identity")
+		query = append(query, "identity ~* :identity")
+	}
+	if pm.Tag != "" {
+		query = append(query, "id ~* :tag")
```
dborovcanin: Same.

dborovcanin (changes requested):

cli/users.go
```diff
@@ -460,6 +463,46 @@ var cmdUsers = []cobra.Command{
 				return
 			}
 
+			logJSON(users)
+		},
+	},
+	{
+		Use:   "search <query> <user_auth_token>",
+		Short: "Search users",
+		Long: "Search users by name, id or identity\n" +
```
dborovcanin: Use `username` instead of `identity` here.

users/service_test.go
```diff
@@ -2167,6 +2167,50 @@ func TestListMembers(t *testing.T) {
...
+			token: validToken,
+			page:  mgclients.Page{Offset: 0, Name: "clientname", Limit: 100},
+			response: mgclients.ClientsPage{
+				Page:    mgclients.Page{Total: 1, Offset: 0, Limit: 100},
+				Clients: []mgclients.Client{client},
+			},
+			identifyResp: &magistrala.IdentityRes{UserId: client.ID},
+		},
+		{
+			desc:        "search clients with invalid token",
```
dborovcanin, line 2195: Add more search criteria scenarios.

dborovcanin (changes requested):

pkg/sdk/go/users.go
```diff
@@ -321,6 +321,25 @@ func (sdk mgSDK) ListUserThings(userID string, pm PageMetadata, token string) (T
 	return tp, nil
 }
 
+func (sdk mgSDK) SearchUsers(pm PageMetadata, token string) (UsersPage, errors.SDKError) {
```
SammyOina: how does this method differe from the `Users` method. since both call the same repository method, `retrieve basic info` and both have the same function signature, it is redundant

SammyOina (changes requested):

pkg/sdk/go/users.go
nyagamunene: I don't think so. It is calling search endpoint which calls the `searchBasicInfo` method

SammyOina: ok, but the methods `RetrieveAllBasicInfo` and `SearchBasicInfo` are functionally simillar why do they have to be distinct functions

nyagamunene: There is a slight difference on how we construct the query. Also the `RetrieveAllBasicInfo` is supposed to be remove later from the conversation we had with @arvindh123. Initially I was to rename it to `searchBasicInfo` but it was agreed to keep it as it is and create the method seperately `searchBasicInfo`

dborovcanin: It is not the same, but we should `RetrieveBasicInfo` in `ListClients` in case the user is not an admin. @nyagamunene

SammyOina (approved):

dborovcanin (approved):

PR #2297, opened 2024-06-18
MG-2296 - Domain creation allowed with nil alias

### LINKED ISSUE #2296: Bug: domain creation allowed with nil alias

Create `domain` with NULL alias

Domain should not create

Domain created

### Steps To Reproduce

REST API To create domain with null domain

Docker

_No response_

JeffMboya:
# What type of PR is this?
This is a bug fix

## What does this do?
This PR modifies the `alias` field in the `auth` table to be NOT NULL and/or empty and UNIQUE. This prevents the creation of a domain with a NULL or empty alias, which was previously allowed.

- Resolves #2296 

### DIFF
--- a/auth/postgres/init.go
+++ b/auth/postgres/init.go
@@ -51,6 +51,12 @@ func Migration() *migrate.MemoryMigrationSource {
 					`DROP TABLE IF EXISTS keys`,
 				},
 			},
+			{
+				Id: "auth_2",
+				Up: []string{
+					`ALTER TABLE domains ALTER COLUMN alias SET NOT NULL`,
+				},
+			},
 		},
 	}
 }
--- a/auth/postgres/init.go
+++ b/auth/postgres/init.go
@@ -30,7 +30,7 @@ func Migration() *migrate.MemoryMigrationSource {
                         name        VARCHAR(254),
                         tags        TEXT[],
                         metadata    JSONB,
-                        alias       VARCHAR(254) NOT NULL UNIQUE CHECK (alias <> ''),
+                        alias       VARCHAR(254) NULL UNIQUE,
                         created_at  TIMESTAMP,
                         updated_at  TIMESTAMP,
                         updated_by  VARCHAR(254),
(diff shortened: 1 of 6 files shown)


auth/postgres/init.go
```diff
@@ -30,7 +30,7 @@ func Migration() *migrate.MemoryMigrationSource {
                         name        VARCHAR(254),
                         tags        TEXT[],
                         metadata    JSONB,
-                        alias       VARCHAR(254) NULL UNIQUE,
+                        alias       VARCHAR(254) NOT NULL UNIQUE,
```
rodneyosodo: It should also be not empty

rodneyosodo (changes requested):

auth/postgres/init.go
```diff
@@ -51,6 +51,14 @@ func Migration() *migrate.MemoryMigrationSource {
 					`DROP TABLE IF EXISTS keys`,
 				},
 			},
+			{
+				Id: "auth_2",
+				Up: []string{
+					`ALTER TABLE domains 
+                     ALTER COLUMN alias SET NOT NULL, 
+                     ADD CONSTRAINT alias_not_empty CHECK (alias <> '');`,
```
dborovcanin: There is a need for this constraint. It makes sense, but for consistency with the rest of the codebase, let's have this check on API or svc layers.

dborovcanin (changes requested):

auth/postgres/init.go
arvindh123: Additonal to this constraint, we can add the this check in API Layer, where we do data validation.

arvindh123 (changes requested):

auth/postgres/init.go
dborovcanin: Since you have added API validation, please remove it from the DB for the sake of consistency with the rest of the codebase.

JeffMboya: Okay

arvindh123 (approved):

dborovcanin (approved):

PR #2305, opened 2024-06-23
MG-2287 - Improve Search For Things

### LINKED ISSUE #2287: Improve search for Users, Things, Channels, and  Groups

nyagamunene:
# What type of PR is this?
This is a feature because it improve search for Things

## What does this do?
This enables back-end support for filtering by name, ID and tags.

- Related Issue #2287 
- Resolves #2287 


pkg/clients/postgres/clients.go
```diff
@@ -190,7 +190,7 @@ func (repo *Repository) RetrieveAll(ctx context.Context, pm clients.Page) (clien
 	return page, nil
 }
 
-func (repo *Repository) RetrieveAllBasicInfo(ctx context.Context, pm clients.Page) (clients.ClientsPage, error) {
+func (repo *Repository) SearchBasicInfo(ctx context.Context, pm clients.Page) (clients.ClientsPage, error) {
```
arvindh123: I think this change is not required, since we have `SearchBasicInfo` function in things postgres

pkg/clients/clients.go
```diff
@@ -78,8 +78,8 @@ type Repository interface {
 	// RetrieveAll retrieves all clients.
 	RetrieveAll(ctx context.Context, pm Page) (ClientsPage, error)
 
-	// RetrieveAllBasicInfo list all clients only with basic information.
-	RetrieveAllBasicInfo(ctx context.Context, pm Page) (ClientsPage, error)
+	// SearchBasicInfo list all clients only with basic information.
+	SearchBasicInfo(ctx context.Context, pm Page) (ClientsPage, error)
```
arvindh123: I think this change is not required, since we have SearchBasicInfo function in things postgres

arvindh123 (changes requested):

pkg/clients/postgres/clients.go
nyagamunene: From the comment left by @dborovcanin [here](https://github.com/absmach/magistrala/issues/2287#issue-2349057815), I understood that the renaming is necessary to avoid confusion between retrieve and search.

arvindh123: Yeah, I agree 
But the RetrieveAllBasicInfo will be no more needed , it will be replaced by search.

nyagamunene: Okay then

things/postgres/clients.go
```diff
@@ -6,7 +6,9 @@ package postgres
 import (
 	"context"
 	"fmt"
+	"strings"
 
+	"github.com/absmach/magistrala/internal/api"
```
dborovcanin: Avoid importing API to repo even if it's internal.

things/postgres/clients.go
```diff
@@ -33,6 +35,12 @@ type Repository interface {
 
 	// RetrieveBySecret retrieves a client based on the secret (key).
 	RetrieveBySecret(ctx context.Context, key string) (mgclients.Client, error)
+
+	// Delete deletes client with given id
+	Delete(ctx context.Context, id string) error
```
dborovcanin: This change does not belong to this PR.

things/postgres/clients.go
```diff
@@ -121,3 +129,96 @@ func (repo clientRepo) RetrieveBySecret(ctx context.Context, key string) (mgclie
...
+func constructSearchQuery(pm mgclients.Page) (string, string) {
+	var query []string
+	var emq string
+	var tq string
+
+	if pm.Name != "" {
+		query = append(query, "name ~* :name")
+	}
+	if pm.Identity != "" {
+		query = append(query, "id ~* :identity")
```
dborovcanin: There is no need to use regular expresions, since we do not have an advanced query. LIKE or ILIKE should be a better fit and faster.

dborovcanin (changes requested):

things/service.go
```diff
@@ -530,6 +530,30 @@ func (svc service) ListClientsByGroup(ctx context.Context, token, groupID string
 	}, nil
 }
 
+func (svc service) SearchThings(ctx context.Context, token string, pm mgclients.Page) (mgclients.ClientsPage, error) {
+	res, err := svc.identify(ctx, token)
+	if err != nil {
+		return mgclients.ClientsPage{}, err
+	}
+
+	cp, err := svc.clients.SearchBasicInfo(ctx, pm)
```
arvindh123: @nyagamunene 

Instead of `SearchBasicInfo` can rename to `SearchClients` ?

things/service.go
```diff
@@ -181,7 +181,7 @@ func (svc service) ListClients(ctx context.Context, token, reqUserID string, pm
 
 	pm.IDs = ids
 
-	tp, err := svc.clients.RetrieveAllByIDs(ctx, pm)
+	tp, err := svc.SearchThings(ctx, token, pm)
```
arvindh123: Why here we are calling `SearchThing` function
And not calling repo function `svc.clients.SearchBasicInfo(ctx, pm)`
`

arvindh123 (changes requested):

things/service.go
nyagamunene: In the SearchThings were are checking if the user has authority to view the thing so it is separated from `listclients` when retrieving the basic info

things/service.go
nyagamunene: We can rename it to `SearchClients` no problem

things/service.go
arvindh123: Authorization are done on above switch case.

things/service.go
```diff
@@ -530,6 +530,30 @@ func (svc service) ListClientsByGroup(ctx context.Context, token, groupID string
...
+		return mgclients.ClientsPage{}, err
+	}
+
+	things := mgclients.ClientsPage{}
+	for _, val := range cp.Clients {
+		if _, err := svc.authorize(ctx, res.GetDomainId(), auth.UserType, auth.UsersKind, res.GetId(), auth.ViewPermission, auth.ThingType, val.ID); err == nil {
+			things.Clients = append(things.Clients, val)
+			things.Page.Total += 1
+		}
+	}
```
arvindh123: This will not return proper page items.

arvindh123: I think we don't need this `SearchThings` function in service layer

arvindh123 (changes requested):

things/service.go
nyagamunene: I will remove the `SearchThings` function then.

nyagamunene: The authorization is needed because the `member` should not be able to `view` things created by other members

arvindh123: Yeah, In `ListThings`  it is handled.

pkg/clients/postgres/clients.go
```diff
@@ -537,6 +519,12 @@ func ConstructSearchQuery(pm clients.Page) (string, string) {
 	if pm.Id != "" {
 		query = append(query, "id ILIKE '%' || :id || '%'")
 	}
+	if pm.Tag != "" {
+		query = append(query, "EXISTS (SELECT 1 FROM unnest(tags) AS tag WHERE tag ILIKE '%' || :tag || '%')")
+	}
+	if len(pm.IDs) != 0 {
+		query = append(query, fmt.Sprintf("id IN ('%s')", strings.Join(pm.IDs, "','")))
+	}
```
arvindh123: Something like this will make code reusable and make simpler.
```suggestion
func buildQueryConditions(pm clients.Page) ([]string) {
	var query []string

	if pm.Name != "" {
		query = append(query, "name ILIKE '%' || :name || '%'")
	}
	if pm.Identity != "" {
		query = append(query, "identity ILIKE '%' || :identity || '%'")
	}
	if pm.Id != "" {
		query = append(query, "id ILIKE '%' || :id || '%'")
	}
	if pm.Tag != "" {
		query = append(query, "EXISTS (SELECT 1 FROM unnest(tags) AS tag WHERE tag ILIKE '%' || :tag || '%')")
	}
	if len(pm.IDs) != 0 {
		query = append(query, fmt.Sprintf("id IN ('%s')", strings.Join(pm.IDs, "','")))
	}

	

	return query
}

func applyOrdering(emq string, pm clients.Page) string {
	switch pm.Order {
	case "name", "identity", "created_at", "updated_at":
		emq = fmt.Sprintf("%s ORDER BY %s", emq, pm.Order)
		if pm.Dir == api.AscDir || pm.Dir == api.DescDir {
			emq = fmt.Sprintf("%s %s", emq, pm.Dir)
		}
	}
	return emq
}

func PageQuery(pm clients.Page) (string, error) {
	mq, _, err := postgres.CreateMetadataQuery("", pm.Metadata)
	if err != nil {
		return "", errors.Wrap(errors.ErrMalformedEntity, err)
	}

	query := buildQueryConditions(pm)
	if mq != "" {
		query = append(query, mq)
	}

	// Additional conditions specific to PageQuery
	if pm.Status != clients.AllStatus {
		query = append(query, "c.status = :status")
	}
	if pm.Domain != "" {
		query = append(query, "c.domain_id = :domain_id")
	}
	if pm.Role != clients.AllRole {
		query = append(query, "c.role = :role")
	}

	if len(query) > 0 {
		emq = fmt.Sprintf("WHERE %s", strings.Join(query, " AND "))
	}

	var emq string
	if len(query) > 0 {
		emq = fmt.Sprintf("WHERE %s", strings.Join(query, " AND "))
	}
	emq = applyOrdering(emq, pm)
	return emq, nil
}

func ConstructSearchQuery(pm clients.Page) (string) {
	query := buildQueryConditions(pm)
	var emq string
	if len(query) > 0 {
		emq = fmt.Sprintf("WHERE %s", strings.Join(query, " AND "))
	}
	emq = applyOrdering(emq, pm)
	return emq, tq
}
```

arvindh123 (changes requested):

things/api/http/endpoints_test.go
```diff
@@ -364,280 +364,6 @@ func TestCreateThings(t *testing.T) {
 	}
 }
 
-func TestListThings(t *testing.T) {
```
dborovcanin: Why is this test removed?

dborovcanin (changes requested):

pkg/clients/postgres/clients.go
```diff
@@ -487,47 +469,38 @@ func PageQuery(pm clients.Page) (string, error) {
...
+	var emq string
 	if len(query) > 0 {
 		emq = fmt.Sprintf("WHERE %s", strings.Join(query, " AND "))
 	}
 	return emq, nil
 }
 
-func ConstructSearchQuery(pm clients.Page) (string, string) {
-	var query []string
+func constructSearchQuery(pm clients.Page) (string, string) {
```
dborovcanin: This function is used only once. Remove it and put the code directly to the `SearchClients`.

pkg/clients/postgres/clients.go
```diff
@@ -537,20 +510,26 @@ func ConstructSearchQuery(pm clients.Page) (string, string) {
 	if pm.Id != "" {
 		query = append(query, "id ILIKE '%' || :id || '%'")
 	}
-
-	if len(query) > 0 {
-		emq = fmt.Sprintf("WHERE %s", strings.Join(query, " AND "))
+	if pm.Tag != "" {
+		query = append(query, "EXISTS (SELECT 1 FROM unnest(tags) AS tag WHERE tag ILIKE '%' || :tag || '%')")
+	}
+	if len(query) == 0 && len(pm.IDs) != 0 {
```
dborovcanin: We are treating IDs differently, i.e. taking it into account only if other criteria is not there, making `buildQueryConditions` mask the logic of processing IDs. Why?

dborovcanin (changes requested):

pkg/clients/postgres/clients.go
nyagamunene: This is allows `searchClients` to retrieve basic info by id or search. So if either `name`, `tag`, `identify`, or `id` is set we don't retrieve by `id`

pkg/clients/postgres/clients.go
nyagamunene: Okay

pkg/clients/postgres/clients.go
dborovcanin: Why don't we still combine? If you want only IDs, you send only IDs in the query.

nyagamunene: We can but to get you properly, you are suggesting we combine both `searchClients` and `retrieveAllByIds` right?

dborovcanin: Eventually, that would be the outcome here, yes. I.e. Search users from this pool of IDs with this criteria. It's unlikely to need such a thing, but it makes `buildQueryConditions` consistent.

arvindh123: @dborovcanin  @nyagamunene 
For search with ID, it will work even the ID matches partially. 
So we need list of things to which user have access.
As you know, Things service could not have information of user accessiable thing, so it will retreive from spiceDB for all the IDs to which user have access. 

Once we find way to list things efficently with access control, then we can remove this IDs from entities

pkg/clients/postgres/clients.go
```diff
@@ -487,47 +475,27 @@ func PageQuery(pm clients.Page) (string, error) {
...
 		query = append(query, "c.status = :status")
 	}
 	if pm.Domain != "" {
 		query = append(query, "c.domain_id = :domain_id")
 	}
-
-	if pm.Role != clients.AllRole {
-		query = append(query, "c.role = :role")
-	}
+	var emq string
```
arvindh123: At present `RetrieveAll` and `RetrieveByIds` doesn't have ordering . 
I think it is good to have ordering in all list and search function.

arvindh123 (changes requested):

pkg/clients/postgres/clients.go
nyagamunene: I can add that

users/service.go
```diff
@@ -193,10 +194,11 @@ func (svc service) ListClients(ctx context.Context, token string, pm mgclients.P
 		Offset:   pm.Offset,
 		Limit:    pm.Limit,
 		Name:     pm.Name,
+		Id:       pm.Id,
 		Identity: pm.Identity,
 		Role:     mgclients.UserRole,
 	}
-	pg, err := svc.clients.SearchBasicInfo(ctx, p)
+	pg, err := svc.clients.SearchClients(ctx, p)
```
arvindh123: Why do we need to pulicate the mgclients.Page, instead we can pass variable pm directly to SearchClients
```suggestion
    pg, err := svc.clients.SearchClients(ctx, pm)
```

arvindh123 (changes requested):

pkg/clients/postgres/clients.go
```diff
@@ -487,47 +477,46 @@ func PageQuery(pm clients.Page) (string, error) {
...
+	case "name", "identity", "created_at", "updated_at":
+		emq = fmt.Sprintf("%s ORDER BY %s", emq, pm.Order)
+		if pm.Dir == api.AscDir || pm.Dir == api.DescDir {
+			emq = fmt.Sprintf("%s %s", emq, pm.Dir)
+		}
+	}
+	return emq
+}
 
+func buildSearch(query []string, pm clients.Page) string {
```
dborovcanin: Remove this function and streamline `PageQuery` to something like: 
```suggestion
func PageQuery(pm clients.Page) (string, error) {
	mq, _, err := postgres.CreateMetadataQuery("", pm.Metadata)
	if err != nil {
		return "", errors.Wrap(errors.ErrMalformedEntity, err)
	}

	var query []string

	if pm.Name != "" {
		query = append(query, "name ILIKE '%' || :name || '%'")
	}
	if pm.Identity != "" {
		query = append(query, "identity ILIKE '%' || :identity || '%'")
	}
	if pm.Id != "" {
		query = append(query, "id ILIKE '%' || :id || '%'")
	}
	if pm.Tag != "" {
		query = append(query, "EXISTS (SELECT 1 FROM unnest(tags) AS tag WHERE tag ILIKE '%' || :tag || '%')")
	}
	if pm.Role != clients.AllRole {
		query = append(query, "c.role = :role")
	}
	// If there are search params presents, use search and ignore other options.
	// Always combine role with search params, so len(query) > 1.
	if len(query) > 1 {
		return fmt.Sprintf("WHERE %s", strings.Join(query, " AND ")), nil
	}

	if mq != "" {
		query = append(query, mq)
	}
	if len(pm.IDs) != 0 {
		query = append(query, fmt.Sprintf("id IN ('%s')", strings.Join(pm.IDs, "','")))
	}
	if pm.Status != clients.AllStatus {
		query = append(query, "c.status = :status")
	}
	if pm.Domain != "" {
		query = append(query, "c.domain_id = :domain_id")
	}

	var emq string
	if len(query) > 0 {
		emq = fmt.Sprintf("WHERE %s", strings.Join(query, " AND "))
	}
	return emq, nil
}
```
Don't forget comments, it's important because the function is a little cryptic. We must improve, I have opened a ticket for this: #2326.

dborovcanin (changes requested):

dborovcanin (approved):

PR #2315, opened 2024-06-25
MG-2287 - Improve Search For Channels and Groups

### LINKED ISSUE #2287: Improve search for Users, Things, Channels, and  Groups

nyagamunene:
# What type of PR is this?
This is a feature because it improve search for Channels and Groups.

## What does this do?
This enables back-end support for filtering by name, ID and tags.

- Related Issue #2287 
- Resolves #2287 


pkg/groups/page.go
```diff
@@ -12,6 +12,7 @@ type PageMeta struct {
 	Limit    uint64           `json:"limit"`
 	Name     string           `json:"name,omitempty"`
 	DomainID string           `json:"domain_id,omitempty"`
+	SearchID string           `json:"search_id,omitempty"`
```
arvindh123: In Client we have `ID`, So lets here Rename it to `ID`

arvindh123 (changes requested):

arvindh123 (approved):

arvindh123 (approved):

dborovcanin (approved):

PR #2324, opened 2024-07-02
MG-2193 - Generate mocks for proto interface

### LINKED ISSUE #2193: Generate mocks for proto interfaces
https://github.com/absmach/magistrala/blob/49028e29507f2b92d235bcd53af652e17213573e/auth/mocks/auth_client.go#L15-L22

//go:generate mockery --name=AuthServiceClient --output=./mocks --filename auth_client.go --quiet

JeffMboya:
# What type of PR is this?

This is a feature

## What does this do?

This PR adds mocks for proto interface.

- Resolves #2193 

Yes

No

### Notes


dborovcanin (changes requested):
Don't edit generated files, but use config or CLI commands to generate mock for generated interfaces.

.github/workflows/tests.yml
```diff
@@ -27,6 +27,7 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           version: v1.59.1
+          args: --timeout 3m
```
dborovcanin: Since the config is moved from the root, we must pass the config location so golangci-lint runs correctly.

dborovcanin (changes requested):

dborovcanin (approved):

PR #2331, opened 2024-07-05
MG-2330 - Fix non admins search with identity

### LINKED ISSUE #2330: Users are able to search other users via identity

I am searching users as a non-admin user.

I use  URL in form of `http://localhost:{{users_port}}/users?identity=absmach` and I am able to see a single user. Since user identity is actually users' emails, this means that we leak that information. While I can't see the full address, I still can use it as filter criteria.

I receive a limited user scope response, but still can filter users by the criteria.

### Steps To Reproduce

Run composition and provision a couple of users. Try to do search in the form provided above.

I'm using the latest MG and the latest Docker.

N/A

nyagamunene:
# What type of PR is this?
This is a bug fix because it fixes the following issue: #2330 

## What does this do?
It restricts normal users from searching with `identity`.

- Related Issue #2330
- Resolves #2330

No.

No

### DIFF
--- a/pkg/clients/postgres/clients.go
+++ b/pkg/clients/postgres/clients.go
@@ -482,7 +482,7 @@ func PageQuery(pm clients.Page) (string, error) {
 	if pm.Name != "" {
 		query = append(query, "name ILIKE '%' || :name || '%'")
 	}
-	if pm.Identity != "" && pm.Role != clients.UserRole {
+	if pm.Identity != "" {
 		query = append(query, "identity ILIKE '%' || :identity || '%'")
 	}
 	if pm.Id != "" {
--- a/pkg/clients/postgres/clients.go
+++ b/pkg/clients/postgres/clients.go
@@ -482,7 +482,7 @@ func PageQuery(pm clients.Page) (string, error) {
 	if pm.Name != "" {
 		query = append(query, "name ILIKE '%' || :name || '%'")
 	}
-	if pm.Identity != "" {
+	if pm.Identity != "" && pm.Role != clients.UserRole {
 		query = append(query, "identity ILIKE '%' || :identity || '%'")
 	}
 	if pm.Id != "" {
--- a/users/api/requests.go
+++ b/users/api/requests.go
@@ -99,6 +99,10 @@ func (req searchClientsReq) validate() error {
 		return apiutil.ErrBearerToken
 	}
 
+	if req.Name == "" && req.Identity == "" && req.Id == "" {
+		return apiutil.ErrEmptySearchQuery
+	}
+
 	return nil
 }
 
--- a/users/api/requests.go
+++ b/users/api/requests.go
@@ -90,7 +90,6 @@ func (req listClientsReq) validate() error {
 }
 
 type searchClientsReq struct {
-	mgclients.Page
 	token    string
 	Offset   uint64
 	Limit    uint64
--- a/users/api/requests.go
+++ b/users/api/requests.go
@@ -93,7 +93,6 @@ type searchClientsReq struct {
 	token    string
 	Offset   uint64
 	Limit    uint64
-	Identity string
 	Name     string
 	Id       string
 	Order    string
@@ -105,7 +104,7 @@ func (req searchClientsReq) validate() error {
 		return apiutil.ErrBearerToken
 	}
 
-	if req.Name == "" && req.Identity == "" && req.Id == "" {
+	if req.Name == "" && req.Id == "" {
 		return apiutil.ErrEmptySearchQuery
 	}
 
--- a/users/api/requests.go
+++ b/users/api/requests.go
@@ -91,7 +91,14 @@ func (req listClientsReq) validate() error {
 
 type searchClientsReq struct {
 	mgclients.Page
-	token string
+	token    string
+	Offset   uint64
+	Limit    uint64
+	Identity string
+	Name     string
+	Id       string
+	Order    string
+	Dir      string
 }
 
 func (req searchClientsReq) validate() error {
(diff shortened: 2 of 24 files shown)


pkg/clients/postgres/clients.go
```diff
@@ -482,7 +482,7 @@ func PageQuery(pm clients.Page) (string, error) {
 	if pm.Name != "" {
 		query = append(query, "name ILIKE '%' || :name || '%'")
 	}
-	if pm.Identity != "" {
+	if pm.Identity != "" && pm.Role == clients.AdminRole {
```
arvindh123: Instead of chaning here in service layer, If  request user is not user admin , then create new pagemeta with limited (required) fields like only with id, name or may be you can add  tag

image

arvindh123 (changes requested):

pkg/clients/postgres/clients.go
nyagamunene: I have implemented as per your suggestion but the superadmin role is not set to `all` anywhere so I have left that line of code.

users/service.go
```diff
@@ -189,7 +189,21 @@ func (svc service) ListClients(ctx context.Context, token string, pm mgclients.P
 		return pg, err
 	}
 
-	pg, err := svc.clients.SearchClients(ctx, pm)
+	// Return empty page if user is not super admin
+	// And search criteria by identity
+	if pm.Identity != "" {
+		return mgclients.ClientsPage{}, nil
+	}
+
```
arvindh123: We can remove this lines, since we are creating new page in below line with omitting the identity from actual page request 
```suggestion
```

arvindh123 (changes requested):
Please test the suggestion and let us know

users/service.go
nyagamunene: By removing that line the result is not an empty `clientspage` since it is as if we are passing a search query with empty parameters which returns all the clients.

arvindh123: Then we should check for other query params like `tag` and `metadata`
So code becomes like

```go
// Return empty page if user is not super admin
// And search criteria by identity
if pm.Identity != "" {
	return mgclients.ClientsPage{}, nil
}

// Return empty page if user is not super admin
// And search criteria by identity
if pm.Tag != "" { 
    return mgclients.ClientsPage{}, nil
}

// Return empty page if user is not super admin
// And search criteria by identity

if pm.Metadata != "" { 
    return mgclients.ClientsPage{}, nil
}
``` 
This will looks like validation of input. 

So @dborovcanin how we can proceed ? 

At present known options 
- Validation of inputs in service for non-admins.
- Having seperate serach endpoint.

If you have any other idea , please let us know here

dborovcanin: Let's switch to the search endpoint.

users/api/requests.go
```diff
@@ -89,6 +89,29 @@ func (req listClientsReq) validate() error {
 	return nil
 }
 
+type searchClientsReq struct {
+	token    string
+	Offset   uint64
+	Limit    uint64
+	Identity string
+	Name     string
```
arvindh123: `Identity` is not required

users/service.go
```diff
@@ -189,7 +189,15 @@ func (svc service) ListClients(ctx context.Context, token string, pm mgclients.P
...
 	}
 
-	pg, err := svc.clients.SearchClients(ctx, pm)
+	page := mgclients.Page{
+		Offset: pm.Offset,
+		Limit:  pm.Limit,
+		Name:   pm.Name,
+		Id:     pm.Id,
+		Role:   mgclients.UserRole,
+	}
```
arvindh123: If user is non super admin then don't allow to list the users

users/service.go
```diff
@@ -200,6 +208,31 @@ func (svc service) ListClients(ctx context.Context, token string, pm mgclients.P
...
+	if err != nil {
+		return mgclients.ClientsPage{}, err
+	}
+	if err := svc.checkSuperAdmin(ctx, userID); err == nil {
+		pg, err := svc.clients.SearchClients(ctx, pm)
+		if err != nil {
+			return mgclients.ClientsPage{}, errors.Wrap(svcerr.ErrViewEntity, err)
+		}
+		return pg, err
+	}
```
arvindh123: No need for super admin check here 
we can support only search by `Name` i.e `Username` or `ID` and we should not support search by `Identity`.

arvindh123 (changes requested):

dborovcanin (commented):
@arvindh123 Please review.

users/service.go
```diff
@@ -180,24 +180,38 @@ func (svc service) ListClients(ctx context.Context, token string, pm mgclients.P
...
-	if err := svc.checkSuperAdmin(ctx, userID); err == nil {
-		pm.Role = mgclients.AllRole
-		pg, err := svc.clients.RetrieveAll(ctx, pm)
-		if err != nil {
-			return mgclients.ClientsPage{}, errors.Wrap(svcerr.ErrViewEntity, err)
-		}
-		return pg, err
+	if err := svc.checkSuperAdmin(ctx, userID); err != nil {
+		return mgclients.ClientsPage{}, errors.Wrap(svcerr.ErrViewEntity, err)
 	}
```
arvindh123, line 185: `svc.checkSuperAdmin` have already wrapped with service error 
No need for wrap with error again here.
```suggestion
		return mgclients.ClientsPage{}, err
```

arvindh123 (changes requested):

arvindh123 (commented):
LGTM

dborovcanin (approved):

PR #2367, opened 2024-07-31
MG-1969 - Check parent group domain while creating/assigning groups

### LINKED ISSUE #1969: Check parent group domain while creating/assigning groups

While creating a group with parent group or assigning a group to a parent group, check parent group and group are in same domain

Must-have

nyagamunene:
# What type of PR is this?
This is a feature because it adds the following functionality: it check if parent and child groups are in the same domain before creating or assigning the child. 

## What does this do?
It implements a grpc call to check if parent is in the same domain as the child.

- Related Issue #1969 
- Resolves #1969 

Yes

N/A

### DIFF
--- a/internal/groups/service.go
+++ b/internal/groups/service.go
@@ -56,6 +56,17 @@ func (svc service) CreateGroup(ctx context.Context, token, kind string, g groups
 		return groups.Group{}, svcerr.ErrInvalidStatus
 	}
 
+	if g.Parent != "" {
+		x,y := svc.auth.Authorize(ctx, &magistrala.AuthorizeReq{
+			SubjectType: auth.GroupType,
+			SubjectKind: auth.GroupsKind,
+			Subject:     g.Parent,
+			Permission:  auth.MembershipPermission,
+			Object:      res.GetDomainId(),
+			ObjectType:  auth.DomainType,	
+		})
+		fmt.Printf("Response is %+v and error is %+v\n", x, y)
+	}
 	g.ID = groupID
 	g.CreatedAt = time.Now()
 	g.Domain = res.GetDomainId()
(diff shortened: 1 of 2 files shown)


internal/groups/service.go
```diff
@@ -64,6 +65,18 @@ func (svc service) CreateGroup(ctx context.Context, token, kind string, g groups
 		if err != nil {
 			return groups.Group{}, errors.Wrap(errParentUnAuthz, err)
 		}
+		allowedGrps, err := svc.auth.ListAllObjects(ctx, &magistrala.ListObjectsReq{
+			SubjectType: auth.DomainType,
+			Subject:     res.GetDomainId(),
+			Permission:  auth.DomainRelation,
+			ObjectType:  auth.GroupType,
+		})
```
dborovcanin, line 73: This will be extremely inefficient, especially since we already have domain ID in the group itself. We only need to be able to extract the domain from the request. There are two ways of doing it: updating API to use domain ID or enabling domain extraction from the token (using Auth or direct JWT inspection but without validation). I will get back to you soon with the path forward once the team discusses it internally.

dborovcanin (changes requested):

internal/groups/service.go
nyagamunene: @dborovcanin  @arvindh123 Without this PR,  you can't add a parent group to a child group without them being in the same domain because of ` groupPreConditions` in the `auth/spicedb/policies`. The challenge is that the error logging is inadequate, only showing "failed to add policies" without specifying the cause of the failure.

dborovcanin: If we add a domain ID as the prefix to this API, we could simply check if the parent group exists and is in the same domain using Repository, rather than calling Auth, right?

nyagamunene: Yes. We will not need call Auth if we implement it this way.

dborovcanin:

arvindh123:
@nyagamunene  If SpiceDB prevents the creation of group with parent group id of different domain. 
Then this PR is no need.

nyagamunene:
> @nyagamunene If SpiceDB prevents the creation of group with parent group id of different domain. Then this PR is no need.

I'll close this PR for now, but we still need to improve the error handling so that the error messages clearly specify the cause of the failure.

Issue #2028, opened 2023-10-17
Auth client that wraps authn and authz 


arvindh123:
- Extract all the common things (kinds, types, relations, permissions, gRPC client...) in a lib.  The library could encapsulate gRPC call directly to the SpiceDB (no auth svc before) in some cases (things, where you don't need to process token; but you need access to Redis).

- [x] #2403
  
- [x] #2404 In PolicyCRUD wrapper, replace the Auth service and make it direct communication with SpiceDB. 
  
- [x] #2410  Create wrappers for Authn/Authz, and then move Auth to middleware. 

- [x] Move Auth to middleware in Users Service .

- [x] Move Auth to middleware in Things Service .
  
- [ ] The Authn/Authz wrapper should use JWT for user authentication. For thing authentication, it should verify with the things service via gRPC.

dborovcanin:
This is related to #2264. In #2264 the idea is to create separate gRPC APIs for policy CRUD, AuthN, and AuthZ, and in this one, the idea is to have the abstraction for those clients so that we can either call Auth service gRPC APIs or SpiceDB directly.
@arvindh123 Please doublecheck if we need all 3, or if we can have only CRUD and Auth APIs.

arvindh123:
After this PR https://github.com/absmach/magistrala/pull/2392 merged 
We need to have kind of client on top AuthService and PolicyService gRPC Client 

And I'm think of like having Authentication and Authorization in single place, i.e in AuthService for both users and things. 
In this method,  AuthService will communicate with users and things via gRPC to verify user password or thing key (Authentication) and then proceed with spiceDB for Authorization.

arvindh123:
@dborovcanin's idea , Services like Users, Things, Bootstrap, etc.. will import the "Policy Service" and "Auth service" as library in their code.  Service don't know about how these libraries are connected to "Policy Service" or "Auth Service". 

If we want to achieve this, one of the way is , we need add wrap above the existing gRPC client. 

But I'm not sure about approach for "Policy Service" , Because the Policy Service library can either directly connect to SpiceDB or it can connect to Auth Service , where auth service act like proxy to SpiceDB

arvindh123:
Create wrapper for PolicyCURD alone, at present wrapper can communicate to Auth service.  

In next PR replace Auth service in wrapper to SpiceDB.  

Same way we need wrap for  Authn/Authz  and then Move Auth to middleware, It should impalement service by service , (Individual service in one PR ) 

[...]

PR #2411, opened 2024-09-03
MG-2410 - Add wrapper for authN/authZ

### LINKED ISSUE #2410: Feature: Create wrapper for Authn/Authz

Related to issue https://github.com/absmach/magistrala/issues/2028

A wrapper for AuthN/Z which can facilitate moving auth to middleware.

Must-have

### Anything else?

_No response_

felixgateru:
# What type of PR is this?
This is a feature as it adds a wrapper for authn/authz that can be used by services instead of communication directly with the auth service.

## What does this do?
This pr adds the following files:

- `pkg/auth/auth.go`
- `pkg/auth/doc.go`
- `internal/auth/auth.go`
- `internal/auth/doc.go`

- Related Issue #2028
- Resolves #2410

No,

Yes, incode documentation for new functions has been done.


internal/auth/wrapper.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"context"
+
+	"github.com/absmach/magistrala"
+	grpcclient "github.com/absmach/magistrala/auth/api/grpc"
+	"github.com/absmach/magistrala/pkg/auth"
+)
+
+type wrapper struct {
+	auth grpcclient.AuthServiceClient
```
dborovcanin: Since `grpcclient.AuthServiceClient` is already a wrapper of https://github.com/absmach/magistrala/blob/main/auth/api/grpc/client.go#L34, there is no point in in wrapping it twice. Let's move the implementation here, and use `pkg.AuthWrapper` as a dependency instead of the gRPC client from the Auth service.

pkg/auth/auth.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+type AuthorizeRes struct {
+	Authorized bool
+	ID         string
+}
+
+// AuthWrapper specifies an API for interacting with authentication and authorization for magistrala services.
+// Acts a wrapper for services that communicates with the magistrala auth service.
+type AuthWrapper interface {
```
dborovcanin: Rename to `Auth` or `AuthClient`.

dborovcanin (changes requested):

dborovcanin (approved):

Issue #2415, opened 2024-09-05
Feature: Move auth to middleware in things service


felixgateru:
### Is your feature request related to a problem? Please describe.

Related to issue https://github.com/absmach/magistrala/issues/2028

### Describe the feature you are requesting, as well as the possible use case(s) for it.

Move existing auth implementation to middleware for the things service

### Indicate the importance of this feature to you.

Must-have

dborovcanin:
It was resolved by #2424.

Issue #2414, opened 2024-09-05
Feature: Move auth to middleware in users service


felixgateru:
### Is your feature request related to a problem? Please describe.

Related to issue #2028 

### Describe the feature you are requesting, as well as the possible use case(s) for it.

Move exixting auth implementation to middleware for the users service

### Indicate the importance of this feature to you.

Must-have

dborovcanin:
It was resolved by #2416.

PR #2434, opened 2024-09-20
MG2415 - Extract auth in groups to api middleware

felixgateru:
# What type of PR is this?
This is feature as it extracts existing auth into api middleware that can be used to wrap existing routes and endpoints.

## What does this do?

- This pr unifies the api middleware for users and things.
- This pr fixes respective api tests.

- Related Issue #2415 
- Related Issue #2414

Yes, all relevant tests have been updated.

No


internal/groups/service.go
```diff
@@ -54,14 +53,14 @@ func (svc service) CreateGroup(ctx context.Context, session auth.Session, kind s
 	g.CreatedAt = time.Now()
 	g.Domain = session.DomainID
 
-	policies, err := svc.addGroupPolicy(ctx, session.DomainUserID, session.DomainID, g.ID, g.Parent, kind)
+	policyList, err := svc.addGroupPolicy(ctx, session.DomainUserID, session.DomainID, g.ID, g.Parent, kind)
```
dborovcanin, line 56: Don't rename unless necessary. The PR is already large enough.

internal/auth/client.go
```diff
@@ -209,3 +209,14 @@ func decodeError(err error) error {
 	}
 	return err
 }
+
+type identityRes struct {
+	id       string
```
dborovcanin: Change ID to `subject` or whatever it represents; just `id` is misleading since this is not the response ID.

internal/groups/api/authrequests.go
```diff
@@ -0,0 +1,274 @@
...
+			Permission:  policies.EditPermission,
+			ObjectType:  policies.GroupType,
+			Object:      req.Group.Parent,
+		})
+	}
+
+	return prs, nil
+}
+
+func ViewGroupAuthReq(_ context.Context, request interface{}) ([]*magistrala.AuthorizeReq, error) {
```
dborovcanin: Why are we using slices with only a single element in these methods?

dborovcanin (changes requested):

internal/groups/api/authrequests.go
felixgateru: There are some methods that have more than one authorize request

internal/groups/service.go
felixgateru: I got a linter error for import shadow. Is this enough to rename them or should I just revert?

users/service.go
```diff
@@ -21,18 +21,15 @@ import (
...
 )
 
 type service struct {
-	clients      postgres.Repository
-	idProvider   magistrala.IDProvider
-	auth         auth.AuthClient
-	policies     policies.PolicyClient
-	hasher       Hasher
-	email        Emailer
-	selfRegister bool
```
arvindh123, line 35: Why this is removed ?

arvindh123, line 35: We can have selfRegister in service layer. 
With session we can determin the login and authentication state

users/api/clients.go
```diff
@@ -35,7 +35,7 @@ func clientsHandler(svc users.Service, authClient auth.AuthClient, selfRegister
 	opts := []kithttp.ServerOption{
 		kithttp.ServerErrorEncoder(apiutil.LoggingErrorEncoder(logger, api.EncodeError)),
 	}
-	checkSuperAdminMiddleware := checkSuperAdminMiddleware(authClient)
+	checkSuperAdminMiddleware := api.CheckSuperAdminMiddleware(authClient)
 
 	r.Route("/users", func(r chi.Router) {
 		switch selfRegister {
```
arvindh123, line 40: We can have self register in service layer, and with session we can detect the state authentication or login state.

users/api/groups.go
```diff
@@ -189,21 +201,15 @@ func assignUsersEndpoint(svc groups.Service, authClient auth.AuthClient) endpoin
 	}
 }
 
-func unassignUsersEndpoint(svc groups.Service, authClient auth.AuthClient) endpoint.Endpoint {
+func unassignUsersEndpoint(svc groups.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
-		req := request.(unassignUsersReq)
+		req := request.(assignUsersReq)
```
arvindh123: I think, it is good to have individual req for each endpoint.

users/api/groups.go
```diff
@@ -28,121 +30,137 @@ func groupsHandler(svc groups.Service, authClient auth.AuthClient, r *chi.Mux, l
...
-		), "update_group").ServeHTTP)
-
-		r.Get("/", otelhttp.NewHandler(kithttp.NewServer(
-			gapi.ListGroupsEndpoint(svc, authClient, "groups", "users"),
+	checkSuperAdminMiddleware := api.CheckSuperAdminMiddleware(authClient)
+	r.Group(func(r chi.Router) {
+		r.Use(api.IdentifyMiddleware(authClient))
+
+		r.Route("/groups", func(r chi.Router) {
+			authzMiddleware := api.AuthorizeMiddleware(authClient, gapi.CreateGroupAuthReq)
```
arvindh123: Why don't we have Authorization as Service middleware, Instead of having In API?

arvindh123: Since any how for each service function, we need to have its own authorization function. 
So having in Service middleware like logging , metering will make ensure that all the function have authorization function.

arvindh123 (changes requested):

users/api/groups.go
dborovcanin: Because that will alter the service interface. I.e. for auth, we need only a token; for other service functions - we need a session (user ID, domain ID, object ID, whatever), so we would have to use them exclusively (token only for auth middleware, session for everything else) while keeping them both in the interface - resulting in the poor design.

internal/groups/service.go
dborovcanin: Keep the renamed version.

internal/api/auth.go
```diff
@@ -49,44 +48,46 @@ func identifyMiddleware(authClient auth.AuthClient) func(http.Handler) http.Hand
...
+			for _, pr := range prs {
+				res, err := authClient.Authorize(ctx, pr)
+				if err != nil || !res.Authorized {
+					return nil, errors.Wrap(svcerr.ErrAuthorization, err)
+				}
 			}
 			return next(ctx, request)
 		}
 	}
 }
```
dborovcanin: Is this method used?

internal/api/auth.go
```diff
@@ -49,44 +48,46 @@ func identifyMiddleware(authClient auth.AuthClient) func(http.Handler) http.Hand
...
 
-func checkSuperAdminMiddleware(authClient auth.AuthClient) endpoint.Middleware {
+func CheckSuperAdminMiddleware(authClient auth.AuthClient) endpoint.Middleware {
```
dborovcanin: Same, are we using this?

internal/api/auth.go
```diff
@@ -19,26 +18,26 @@ import (
...
 type sessionKeyType string
 
-const sessionKey = sessionKeyType("session")
+const SessionKey = sessionKeyType("session")
 
-type authEndpointFunc func(context.Context, interface{}) (*magistrala.AuthorizeReq, error)
+type authEndpointFunc func(context.Context, interface{}) ([]*magistrala.AuthorizeReq, error)
 
-func identifyMiddleware(authClient auth.AuthClient) func(http.Handler) http.Handler {
+func IdentifyMiddleware(authClient auth.AuthClient, ) func(http.Handler) http.Handler {
```
dborovcanin: Rename to AuthenticateMiddleware.

users/middleware/authorization.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func (am *authorizationMiddleware) RegisterClient(ctx context.Context, session auth.Session, client clients.Client, selfRegister bool) (clients.Client, error) {
+	if selfRegister {
+		if err := am.checkSuperAdmin(ctx, session.UserID); err == nil {
+			session.SuperAdmin = true
+		}
+	}
+
+	return am.svc.RegisterClient(ctx, session, client, selfRegister) ///// fgsf
```
dborovcanin: What's this comment?

dborovcanin (changes requested):

dborovcanin (approved):

PR #2409, opened 2024-09-03
MG-2456 - Refactor auth

arvindh123:
# What type of PR is this?

This is a feature because it adds the following functionality mentioned in following tickets 
- [ ] https://github.com/absmach/magistrala/issues/2456
- [x] https://github.com/absmach/magistrala/issues/2214
- [ ] https://github.com/absmach/magistrala/issues/2053

## What does this do?
It adds the features mentioned in this ticket https://github.com/absmach/magistrala/issues/2456

## Which issue(s) does this PR fix/relate to?

- Related Issue https://github.com/absmach/magistrala/issues/2456
- Resolves partially #2456
- Resolves partially #2214
- Resolves partially #2053

## Have you included tests for your changes?
Yes , Tested manually 

## Did you document any new/modified feature?
No

PR #2442, opened 2024-09-24
MG-2441 - Add domain ID to API

### LINKED ISSUE #2441: Feature: Add Domain to API

When using PAT this is the best approach to get Domain ID for authorization. Also this approach will be the most efficient for adding domains to events.

For example, when sharing a thing:
`http://localhost:9000/domains{{DOMAINID}}/things/{{THINGID}}/share`

Must-have

### Anything else?

_No response_

nyagamunene:
# What type of PR is this?
This is a feature because it adds the following functionality: it add domain id when passing requests in the API layer.

## What does this do?
It add domain id when passing requests in the API layer for users and things.

- Related Issue #2441 
- Resolves #2441 

Yes.

Yes


dborovcanin:
@nyagamunene Please make sure this is ready for review today. Also, please sign off your commits.

nyagamunene:
> @nyagamunene Please make sure this is ready for review today. Also, please sign off your commits.

No problem

dborovcanin (changes requested):
@nyagamunene Please rebase. This is to be merged before #2409.

users/api/clients.go
```diff
@@ -85,6 +85,20 @@ func clientsHandler(svc users.Service, authn mgauthn.Authentication, tokenClient
...
+				api.EncodeResponse,
+				opts...,
+			), "list_clients").ServeHTTP)
+
+			r.Get("/search", otelhttp.NewHandler(kithttp.NewServer(
+				searchClientsEndpoint(svc),
+				decodeSearchClients,
+				api.EncodeResponse,
+				opts...,
+			), "search_clients").ServeHTTP)
```
dborovcanin: Remove duplicate handlers.

dborovcanin (changes requested):
Overall, looks good. Fix CI and comments and it's ready for another review. I'll also check if there is a better way to handle things in Bootstrap.

arvindh123 (changes requested):
For Things, groups ,channels  
DomainID should be automatically decoded from URL , 
We can do that in Middleware and add to authn.Session

nyagamunene:
> For Things, groups ,channels DomainID should be automatically decoded from URL , We can do that in Middleware and add to authn.Session

Since in this PR we are just adding the domainID to URL should I make this change in the other PR?

users/api/requests.go
```diff
@@ -89,9 +89,13 @@ type listMembersByObjectReq struct {
 	mgclients.Page
 	objectKind string
 	objectID   string
+	domainID   string
 }
 
 func (req listMembersByObjectReq) validate() error {
+	if req.domainID == "" {
```
arvindh123: @nyagamunene  We have changes in  request in this PR , It is not relevant to this PR, please remove it

arvindh123 (changes requested):

dborovcanin:
> > For Things, groups ,channels DomainID should be automatically decoded from URL , We can do that in Middleware and add to authn.Session
> 
> Since in this PR we are just adding the domainID to URL should I make this change in the other PR?

Since we are already breaking API, you can do it here. In the next one, you can remove the domain from the token, just like planned.

cmd/things/main.go
```diff
@@ -231,7 +231,6 @@ func main() {
 	g.Go(func() error {
 		return server.StopSignalHandler(ctx, cancel, logger, svcName, httpSvc)
 	})
```
arvindh123: Please undo this empty changes

consumers/notifiers/api/requests.go
```diff
@@ -51,5 +51,6 @@ func (req listSubsReq) validate() error {
 	if req.token == "" {
 		return apiutil.ErrBearerToken
 	}
+
```
arvindh123: Please remove this empty changes , Is this due to linter?

pkg/authn/authn.go
```diff
@@ -14,6 +14,11 @@ type Session struct {
 	SuperAdmin   bool
 }
 
+func (s *Session) UpdateSession(DomainID string) {
```
arvindh123: No need of this changes function, 

Instead check and invalidate authn session to empty object , if authn session domain id is not equal to domainID 

We could not have these logic for all endpoints, it is needed only required for domains, things, groups and channels 

For users this logic is no needed logic.

certs/api/requests.go
```diff
@@ -39,22 +44,30 @@ func (req addCertsReq) validate() error {
...
+	thingID  string
+	domainID string
+	pm       certs.PageMetadata
 }
 
 func (req *listReq) validate() error {
 	if req.pm.Limit > maxLimitSize {
 		return apiutil.ErrLimitSize
 	}
+	if req.domainID == "" {
```
arvindh123: Certs service doesn't have authn middleware ? Do we need token here?

bootstrap/api/requests.go
```diff
@@ -28,6 +29,10 @@ func (req addReq) validate() error {
 		return apiutil.ErrBearerToken
 	}
 
+	if req.domainID == "" {
```
arvindh123: Bootstrap service doesn't have authn middleware ? Do we need token here?

arvindh123 (changes requested):

certs/api/requests.go
nyagamunene: > Certs service doesn't have authn middleware ? Do we need token here?

DomainID is needed for issuing and revoking cert

internal/groups/api/requests.go
```diff
@@ -160,5 +160,6 @@ func (req listMembersReq) validate() error {
 	if req.groupID == "" {
 		return apiutil.ErrMissingID
 	}
+
```
arvindh123: remove this line

arvindh123 (commented):
LGTM

users/api/clients.go
```diff
@@ -164,7 +171,7 @@ func clientsHandler(svc users.Service, authn mgauthn.Authentication, tokenClient
 		// SpiceDB provides list of user ids in given user_group_id
 		// and users service can access spiceDB and get the user list with user_group_id.
 		// Request to get list of users present in the user_group_id {groupID}
-		r.Get("/groups/{groupID}/users", otelhttp.NewHandler(kithttp.NewServer(
+		r.Get("/domains/{domainID}/groups/{groupID}/users", otelhttp.NewHandler(kithttp.NewServer(
```
dborovcanin: There is no need for `/domains` prefix here, use `/{domainID}/groups/...`. I.e. unlike other entities, let's not prefix domain ID.

dborovcanin: Same for other endpoints.

dborovcanin (changes requested):

dborovcanin (approved):

PR #2444, opened 2024-09-27
MG-2443 - Move Auth to the middleware

dborovcanin:
# What type of PR is this?

This enhancement moves the Authorization layer to the middleware and Authentication to the API layer.
## What does this do?

We are moving the authN layer to API and the authZ layer to service middleware. We are not moving everything to API middleware just yet because Auth often has much to do with business logic and is tightly coupled.
## Which issue(s) does this PR fix/relate to?

Resolves #2443.

## Have you included tests for your changes?
Yes.

## Did you document any new/modified feature?
N/A

### Notes
N/A

PR #2453, opened 2024-10-07
NOISSUE - Refactor auth policy service to domains service

felixgateru:
# What type of PR is this?
This is a refactor as it changes the functionality of the auth grpc server

## What does this do?
This pr:
 - renames `grpcclient` package to `authclient`
 - refactors `PolicyService` in auth proto file to `DomainsService`

- Related Issue #2444 

Yes, relevant tests have been updated.

NO


arvindh123:
### Lets have two interface in policies

1.)     `Manager` : Manger will have all CURD and list  operations over policies , So in present PR you can rename `PolicyClient`  to `Manager`
 and then rename file `pkg/policies/policies.go`  to `pkg/policies/manager.go`
 
Then move `internal/policies/client.go` to `pkg/policies/spicedb/manager.go`


2.)     Then add another new interface called `Evaluator` or `Checker`
This do only check policy
For this you can move `PolicyAgent` interface in auth to `pkg/policies/Checker.go` and rename to `Evaluator` or `Checker`
Then move `auth/spicedb/policies.go`  to  `pkg/policies/spicedb/checker.go` or `pkg/policies/spicedb/evaluator.go`
Then move `internal/policies/client.go` to `pkg/policies/spicedb/`

### Summary
`pkg/policies/manager.go` and `pkg/policies/evaluator.go` have interface to manager and verify policies respectively
`pkg/policies/spciedb/` contains implementation of `Manager` and `Evaluator`  for `SpiceDB
`


Please give your comments on this approach

arvindh123:
In the above approach, Policy Evaluator is moved out of the Authorization service and kept command. 
So it can be used by both Auth service and Things service

arvindh123:
`pkg/auth` can be splited into two  `pkg/authn` and `pkg/authz`

`pkg/authn` should have a file name authn.go and it should a interface `Authn` 


Interface should have their own types in function parameters 

Example: 

`Authn` should have `Authenticate` , `Issue`, `Refresh` function

Each function should have function parameters as same package type , ie new type belongs to package `pkg/auth`
Like below 
```diff
type SessionInfo struct { 
  Id string
  UserId string
  DomainId string
}
- Identify(ctx context.Context, in *magistrala.IdentityReq, opts ...grpc.CallOption) (*magistrala.IdentityRes, error)
+ Authenticate(ctx context.Context, token string ) (SessionInfo, error)
```

Then a new folder with name + path : `pkg/authn/authsvcclient` which contains implementation a file `authn.go` using auth service gRPC client

same way `pkg/authz` should a file name authz.go and  it should have interface `Authz` 
`Authz` interface should have only `Authorize` function, and its parameters should be of same package type 

Then a new folder with name `pkg/authz/authsvcclient` which contains implementation a file `authz.go` using auth service gRPC client

In future there will folder with name ``pkg/authz/spicedbclient` which will directly communicate to spicedb for authorization.
It will be using `policiy.Evaluvator`

The folder structure will be like below : 

```md
pkg
├── authz
│   ├── authz.go
│   ├── authsvcclient
│   │     ├── authz.go
│   ├── spicedbclient
│          ├── authz.go
├── authn
│   ├── authn.go
│   ├── authsvcclient
│   │     ├── authn.go
│   ├── asymmetricjwtjwks
│          ├── authn.go
```
`internal/auth/client.go` this should be moved back to `auth/api/grpc/client.go`

dborovcanin:
> `pkg/auth` can be splited into two `pkg/authn` and `pkg/authz`
> 
> `pkg/authn` should have a file name authn.go and it should a interface `Authn`
> 
> Interface should have their own types in function parameters
> 
> Example:
> 
> `Authn` should have `Authenticate` , `Issue`, `Refresh` function
> 
> Each function should have function parameters as same package type , ie new type belongs to package `pkg/auth` Like below
> 
> ```diff
> type SessionInfo struct { 
>   Id string
>   UserId string
>   DomainId string
> }
> - Identify(ctx context.Context, in *magistrala.IdentityReq, opts ...grpc.CallOption) (*magistrala.IdentityRes, error)
> + Authenticate(ctx context.Context, token string ) (SessionInfo, error)
> ```
> 
> Then a new folder with name + path : `pkg/authn/authsvcclient` which contains implementation a file `authn.go` using auth service gRPC client
> 
> same way `pkg/authz` should a file name authz.go and it should have interface `Authz` `Authz` interface should have only `Authorize` function, and its parameters should be of same package type
> 
> Then a new folder with name `pkg/authz/authsvcclient` which contains implementation a file `authz.go` using auth service gRPC client
> 
> In future there will folder with name ``pkg/authz/spicedbclient`which will directly communicate to spicedb for authorization. It will be using`policiy.Evaluvator`
> 
> The folder structure will be like below :
> 
> ```
> pkg
> ├── authz
> │   ├── authz.go
> │   ├── authsvcclient
> │   │     ├── authz.go
> │   ├── spicedbclient
> │          ├── authz.go
> ├── authn
> │   ├── authn.go
> │   ├── authsvcclient
> │   │     ├── authn.go
> │   ├── asymmetricjwtjwks
> │          ├── authn.go
> ```
> 
> `internal/auth/client.go` this should be moved back to `auth/api/grpc/client.go`

This is correct. Let's keep a simpler package names like:
> ```
> pkg
> ├── authz
> │   ├── authz.go
> │   ├── authsvc
> │   │     ├── authz.go
> │   ├── spicedb
> │          ├── authz.go
> ├── authn
> │   ├── authn.go
> │   ├── authsvc
> │   │     ├── authn.go
> │   ├── jwks
> │          ├── authn.go
> ```

dborovcanin (changes requested):
@felixgateru @arvindh123 I have rebased the target branch to match `main`, so there are some conflicts. Please resolve them (you can merge instead of rebase if you prefer).

arvindh123 (approved):
LGTM

auth.proto
```diff
@@ -37,11 +41,11 @@ message Token {
     string accessType = 3;
 }
 
-message IdentityReq {
+message AuthenticateReq {
```
dborovcanin: Let's rename to `AuthNReq`, `AuthNResp`. Do the same for `AuthZReq` and `AuthZResp`.

pkg/authz/authz.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Abstract Machines
+// SPDX-License-Identifier: Apache-2.0
+
+package authz
+
+import "context"
+
+type PolicyReq struct {
+	// Domain contains the domain ID.
+	Domain string `json:"domain,omitempty"`
```
dborovcanin, line 10: We have exactly the same structure in `pkg/policies/manager.go`. Overall, it looks like we have too many abstractions and you got tangled in them. Please simplify.

dborovcanin (changes requested):

pkg/authz/authz.go
arvindh123: Yes @dborovcanin, because  `pkg/authz` is a library, The library should have its own type and it should be independent of underlying implementation, So only it is repeating.

> It is for decoupling purpose. 


If we have abstract struct, then both library and underlying clients will be using same struct. 

`pkg/auth/PolicyReq`  struct is used in all service ,
This library `pkg/authz` is used by all the service 

In future, if we change underlying client, 
then we don't need to change in all service , because `pkg/auth/PolicyReq`  be same and it belongs to `pkg/authz` library

arvindh123: Lets take this case 

https://github.com/absmach/magistrala/pull/2453#discussion_r1799589120
> Let's rename to AuthNReq, AuthNResp. Do the same for AuthZReq and AuthZResp.

Now we need to make change only in `pkg/authz` and `pkg/authn` not in all service

dborovcanin: OK, let's rename Manager to Service (it's policy CRUD service) and `PolicyReq` to `Policy` in `pkg/policies/manager.go`. Also, consider moving that to `internal` rather than `pkg` because Policy CRUD service is meant to be internal (the client, on the other hand, could be public).

dborovcanin: Let's go with only name changes, but keep it in `pkg` for the time being. If it turns out we don't need it shared, which is desirable, we'll move it to `internal`.

dborovcanin (approved):

arvindh123 (approved):
LGTM

PR #2506, opened 2024-11-05
NOISSUE: Get group role actions from spicedb schema & List groups by user, clients, channels

arvindh123:
# What type of PR is this?

- Related Issue #
- Resolves #


groups/postgres/groups.go
```diff
@@ -576,6 +576,7 @@ func (repo groupRepository) RetrieveChildrenGroups(ctx context.Context, domainID
 		query = query + " AND " + levelCondition
 	}
 
+	fmt.Println(startLevel, endLevel)
```
felixgateru: Remove debug log

groups/postgres/groups.go
```diff
@@ -516,6 +520,239 @@ func (repo groupRepository) Delete(ctx context.Context, groupID string) error {
...
 
+func (repo groupRepository) RetrieveAllParentGroups(ctx context.Context, domainID, userID, groupID string, pm mggroups.PageMeta) (mggroups.Page, error) {
+	cGroup, err := repo.RetrieveByID(ctx, groupID)
+	if err != nil {
+		return mggroups.Page{}, err
+	}
+
+	query := buildQuery(pm)
+
+	levelCondition := fmt.Sprintf(" @> '%s' ", cGroup.Path)
```
felixgateru: The left side of this query will be empty.

felixgateru (changes requested):

felixgateru (approved):

groups/builtinroles.go
```diff
@@ -0,0 +1,7 @@
+package groups
+
+import "github.com/absmach/magistrala/pkg/roles"
+
+const (
+	BuiltInRoleAdmin roles.BuiltInRoleName = "admin"
+)
```
dborovcanin: Inline `const Admin roles.BuiltInRoleName = "admin"` (also, Admin name is enough - there is no need to to put type in the constant name).

groups/status.go
```diff
@@ -3,7 +3,12 @@
...
-import svcerr "github.com/absmach/magistrala/pkg/errors/service"
+import (
+	"encoding/json"
+	"strings"
+
+	svcerr "github.com/absmach/magistrala/pkg/errors/service"
+)
 
 // Status represents Group status.
 type Status uint8
```
dborovcanin, line 14: We are repeating Status across services. Since all the entities have the same status options, please extract to the common package so we don't reimplement per-service.

dborovcanin (changes requested):

groups/status.go
arvindh123: @dborovcanin I agree, Can I do it in different PR. 
Because it touches all the service ?

dborovcanin: We can do this in a separate PR.

dborovcanin (approved):

Issue #2457, opened 2024-10-08
Improve tests for new Auth


dborovcanin:
New Auth will be implemented as per this ticket https://github.com/absmach/magistrala/issues/2456
New Auth breaks most of the unit test. These test should be rewritten.

dborovcanin:
@arvindh123 Feel free to add a proper description and explain.

dborovcanin:
Also, perform manual testing of new Auth changes.

PR #2544, opened 2024-11-22
MG-2457 - Update readers tests

felixgateru:
# What type of PR is this?
This is a bug fix as it fixes failing tests in readers service.

## What does this do?
This pr updates the tests in readers with changes to auth.

- Related Issue #2457

Yes the changes are to the test files

No,

### Notes
None


readers/api/requests.go
```diff
@@ -20,17 +20,13 @@ type listMessagesReq struct {
 	chanID   string
 	token    string
 	key      string
-	domainID string
```
arvindh123, line 23: I think, we should have domainID for readers too. 
@dborovcanin  What do you think ?

arvindh123 (approved):

readers/api/requests.go
dborovcanin: Yes, we should. @felixgateru

dborovcanin: @felixgateru Let's use a different PR for that, I'll merge this one.

dborovcanin (approved):

PR #2612, opened 2024-12-20
SMQ-2566 - User role ID instead of name in the API

arvindh123:
# What type of PR is this?
Change from Role Name to Role ID

- Related Issue #
- Resolves #

### DIFF
--- a/api/http/util/errors.go
+++ b/api/http/util/errors.go
@@ -78,10 +78,10 @@ var (
 	// ErrEmptyList indicates that entity data is empty.
 	ErrEmptyList = errors.New("empty list provided")
 
-	// ErrMissingRoleName indicates that role name are empty.
+	// ErrMissingRoleName indicates that role name is empty.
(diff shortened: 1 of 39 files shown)


api/http/util/errors.go
```diff
@@ -81,6 +81,9 @@ var (
 	// ErrMissingRoleName indicates that role name are empty.
 	ErrMissingRoleName = errors.New("empty role name")
 
+	// ErrMissingRoleID indicates that role name are empty.
```
felixgateru:
```suggestion
	// ErrMissingRoleID indicates that role id is empty.
```

felixgateru (changes requested):

felixgateru (commented):
We will need to update the api docs for this change

arvindh123:
This PR is depend on  https://github.com/absmach/supermq/pull/2585

dborovcanin (approved):

Issue #2546, opened 2024-11-22
Improve journaling with clients telemetry


dborovcanin:
### Is your feature request related to a problem? Please describe.

I want the ability to see some basic telemetry related to my domain.

### Describe the feature you are requesting, as well as the possible use case(s) for it.

Here is an example of telemetry I want to check:
- number of connections per client
- number of subscriptions per client
- number of inbound and outbound messages per client
- total number of all the above per domain
- last seen per client

We will use Journal service, but not as a consumer/collector, but as an aggregator and API provider for this information, and have a background process to sync with the actual collector.

This is an umbrella issue we'll handle in the following way:
- [x] #2648
- [ ] #2649

### Indicate the importance of this feature to you.

Must-have

felixgateru:
NATS monitors our subscriptions differently for the adapters. For COAP individual subscriptions are made to NATS but for MQTT a single susbcription for all the subjects is made. This means we will not be able to directly get the number of active subscriptions from NATS.

dborovcanin:
We have entity-level journals. For Domain-level, we can create a new issue. Closing.

PR #2659, opened 2025-01-21
SMQ-2546 - Add events to adapters

felixgateru:
# What type of PR is this?
This is a feature as it adds events to `coap` and `ws` and updates events in `mqtt`.

## What does this do?
This pr  adds `publish`, `susbscribe` and `unsubscribe` events to the adapters.

- Related Issue #2546 

No

Yes, in code docuementation is included

### Notes
None


ws/handler.go
```diff
@@ -233,6 +255,31 @@ func (h *handler) Unsubscribe(ctx context.Context, topics *[]string) error {
...
+		return errMissingTopicSub
+	}
+
+	var token string
+	switch {
+	case strings.HasPrefix(string(s.Password), "Client"):
+		token = strings.ReplaceAll(string(s.Password), "Client ", "")
+	default:
+		token = string(s.Password)
+	}
```
dborovcanin: Use `extractClientSecret` instead.

dborovcanin (changes requested):

ws/events/stream.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/absmach/supermq/pkg/events"
+	"github.com/absmach/supermq/pkg/events/store"
+)
+
+const streamID = "supermq.websocket"
+
+//go:generate mockery --name EventStore --output=../mocks --filename events.go --quiet --note "Copyright (c) Abstract Machines"
+type EventStore interface {
+	Publish(ctx context.Context, clientID, channelID, topic string) error
```
dborovcanin: Use `subtopic` here, just like in other methods.

http/handler.go
```diff
@@ -188,6 +196,10 @@ func (h *handler) Publish(ctx context.Context, topic *string, payload *[]byte) e
 		return errors.Wrap(errFailedPublishToMsgBroker, err)
 	}
 
+	if err := h.es.Publish(ctx, clientID, msg.Channel, msg.Subtopic); err != nil {
+		return errors.Wrap(errFailedPublishEvent, err)
+	}
```
dborovcanin: Event store should be wrapper around interfaces. I.e. it should be transparent from the adapter implementation because we are using middlewares. Use same approach we have for all the other services.

dborovcanin (changes requested):

dborovcanin (approved):

PR #2695, opened 2025-02-10
NOISSUE - Use RabbitMQ as MQTT broker

rodneyosodo:
# What type of PR is this?

This is a feature because it adds the following functionality: Add RabbitMQ as a MQTT broker.

## What does this do?

The changes update the MQTT adapter and Docker configurations across the project. In the MQTT adapter, new configuration fields, a connection interceptor, and enhanced publisher/forwarder logic have been introduced, including authentication and QoS support. Docker-related files have been modified to integrate RabbitMQ as the MQTT broker, with updated environment variables, deployment documentation, and RabbitMQ-specific settings. Additional improvements include new tracing middleware for MQTT forwarding, updated dependency versions, and comprehensive tests for the MQTT pub/sub functionality using Docker.

- Related Issue #2681 

Yes, I have included tests for my changes. I have tested MQTT pubsub, COAP pubsub and HTTP pub. WS will be fixed in another PR


dborovcanin:
@rodneyosodo Please rebase.

cmd/mqtt/main.go
```diff
@@ -185,7 +225,10 @@ func main() {
 		go chc.CallHome(ctx)
 	}
 
-	var interceptor session.Interceptor
+	interceptor := interceptor{
+		username: cfg.MQTTTargetUsername,
+		password: cfg.MQTTTargetPassword,
+	}
```
dborovcanin, line 231: Why do we need this interceptor to set username and password?

dborovcanin (changes requested):

cmd/mqtt/main.go
rodneyosodo: RabbitMQ runs in two modes, with and without authentication. For our case devices, i.e, clients always have authentication (thingID and thingKey). So we either add the RabbitMQ username and password when sending the request or remove the username and password when forwarding the request to RabbitMQ

dborovcanin:
@felixgateru Please review and test this one.

felixgateru (dismissed):

cmd/mqtt/main.go
dborovcanin: In that case, let's remove Auth on RabbitMQ, since we will be using mGate for that anyways.

dborovcanin: Since authN can't be turned off (i.e. we need at least the default user), let's keep this approach (see [https://www.rabbitmq.com/docs/mqtt#anonymous-connections](https://www.rabbitmq.com/docs/mqtt#anonymous-connections))

dborovcanin (approved):

nintran52:
I’m wondering why we’re not using VerneMQ as the MQTT broker, considering that it’s a native MQTT solution. @dborovcanin @arvindh123  @rodneyosodo

PR #2696, opened 2025-02-10
SMQ-2689 - Allow user defined ids for domains

### LINKED ISSUE #2689: Feature: Allow Users to Create Custom IDs for Domains

No

Allow Users to Create Custom IDs for Domains. 
Constraints for  custom ID
1. The ID must be exactly 36 characters long, same length of UUID.
2. It should consist of alphanumeric characters (letters and numbers), including undersocre (_).
          2.1 Dashes or underscores cannot appear at the beginning or end of the domainname(e.g.,domainname_, _domainname  are invalid).
          2.2  No consecutive underscores (e.g., domain__name is invalid).
          2.3 No spaces or other special characters (e.g., @, !, #).
3. Start and end: The domainame should start and end with an alphanumeric character (letters or numbers).
4. The ID should be case-insensitive (e.g., "domainID" and "domainid" should be treated the same).
5. The ID must be unique within the system.

Must-have

### Anything else?

_No response_

felixgateru:
# What type of PR is this?
This is a feature as it allows user defined ids for domains

## What does this do?
Creates provision to allow for custom ids for domains during creation.

- Resolves #2689

Yes

No

### DIFF
--- a/api/http/common.go
+++ b/api/http/common.go
@@ -7,6 +7,8 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+	"regexp"
+	"strings"
 
 	"github.com/absmach/supermq"
 	apiutil "github.com/absmach/supermq/api/http/util"
@@ -87,11 +89,25 @@ const (
 	MaxIDSize    = 36
 )
 
+var nameRegExp = regexp.MustCompile(`^[a-z0-9][a-z0-9_-]{34}[a-z0-9]$`)
+
 // ValidateUUID validates UUID format.
 func ValidateUUID(extID string) (err error) {
 	id, err := uuid.FromString(extID)
 	if id.String() != extID || err != nil {
 		return apiutil.ErrInvalidIDFormat
+	}
+
+	return nil
+}
+
+// ValidateName validates name format.
+func ValidateName(id string) error {
+	if !nameRegExp.MatchString(id) {
+		return apiutil.ErrInvalidNameFormat
+	}
+	if strings.Contains(id, "__") {
+		return apiutil.ErrInvalidNameFormat
 	}
 
 	return nil
--- a/api/http/common.go
+++ b/api/http/common.go
@@ -89,7 +89,10 @@ const (
 	MaxIDSize    = 36
 )
 
-var nameRegExp = regexp.MustCompile(`^[a-z0-9][a-z0-9_-]{34}[a-z0-9]$`)
+var (
+	nameRegExp                = regexp.MustCompile(`^[a-z0-9][a-z0-9_-]{34}[a-z0-9]$`)
+	errDoubleDashOrUnderscore = errors.New("name containing double underscores or double dashes not allowed")
+)
 
 // ValidateUUID validates UUID format.
 func ValidateUUID(extID string) (err error) {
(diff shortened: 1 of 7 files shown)


domains/service.go
```diff
@@ -176,3 +188,15 @@ func (svc service) ListDomains(ctx context.Context, session authn.Session, p Pag
 	}
 	return dp, nil
 }
+
+func validateID(id string) error {
```
arvindh123: @felixgateru  
Lets move this validation to `api/http/common.go` and rename the function to `ValidateName`
For Now lets use ValidateUUID in transport layer for ID 

So the plan for now is , we provide two option for ID, 
- We provide UUID 
Or 
- Bring your own UUID 

We will introduce another field called name/alias where user can put any string with the constraint mentioned in the ticket https://github.com/absmach/supermq/issues/2689#issue-2841428635

arvindh123 (changes requested):

arvindh123 (dismissed):

api/http/common.go
```diff
@@ -96,6 +100,18 @@ func ValidateUUID(extID string) (err error) {
...
 }
 
+// ValidateName validates name format.
+func ValidateName(id string) error {
+	if !nameRegExp.MatchString(id) {
+		return apiutil.ErrInvalidNameFormat
+	}
+	if strings.Contains(id, "__") {
+		return apiutil.ErrInvalidNameFormat
+	}
```
dborovcanin: Why do we have a special case here?

dborovcanin (changes requested):

api/http/common.go
felixgateru: So that we dont allow a name with consecutive underscores

dborovcanin: Let's add a comment explaining this (it can look very ambiguous in certain fonts) and add a check for double dash `--` for the same reason.

dborovcanin (approved):

api/http/common.go
```diff
@@ -96,6 +100,19 @@ func ValidateUUID(extID string) (err error) {
...
 }
 
+// ValidateName validates name format.
+func ValidateName(id string) error {
+	if !nameRegExp.MatchString(id) {
+		return apiutil.ErrInvalidNameFormat
+	}
+	// Names containing double underscores or double dashes are invalid due to similarity concerns.
+	if strings.Contains(id, "__") || strings.Contains(id, "--") {
+		return apiutil.ErrInvalidNameFormat
```
dborovcanin: We need to wrap this to make it more precise so the user knows what's wrong with the name format.

dborovcanin (changes requested):

arvindh123 (dismissed):

api/http/common.go
```diff
@@ -96,6 +100,19 @@ func ValidateUUID(extID string) (err error) {
...
 }
 
+// ValidateName validates name format.
+func ValidateName(id string) error {
+	if !nameRegExp.MatchString(id) {
+		return apiutil.ErrInvalidNameFormat
+	}
+	// Names containing double underscores or double dashes are invalid due to similarity concerns.
+	if strings.Contains(id, "__") || strings.Contains(id, "--") {
+		return errors.Wrap(apiutil.ErrInvalidNameFormat, errors.New("name containing double underscores or double dashes not allowed"))
```
dborovcanin: Do not create a new error each time, extract `errors.New("name containing double underscores or double dashes not allowed")` to an unexported error var.

dborovcanin, line 110: Do not create a new error each time, extract `errors.New("name containing double underscores or double dashes not allowed")` to an unexported error var.

dborovcanin (changes requested):

api/http/common.go
```diff
@@ -87,6 +89,11 @@ const (
 	MaxIDSize    = 36
 )
 
+var (
+	nameRegExp                = regexp.MustCompile(`^[a-z0-9][a-z0-9_-]{34}[a-z0-9]$`)
+	errDoubleDashOrUnderscore = errors.New("name containing double underscores or double dashes not allowed")
```
dborovcanin: Rename to errUnreadableName or something shorter.

dborovcanin (approved):

PR #2709, opened 2025-02-13
SMQ-2671 - Add request ID and use it to correlate traces, logs, and events

### LINKED ISSUE #2671: Add request ID and use it to correlate traces, logs, and events (i.e. use the same ID for all the telemetry)

nyagamunene:
# What type of PR is this?

- It adds request id in the transport layer.
- It logs the request id using the logger.
- It add request id in the events struct.
- It changes the tracing id to the request id.

## What does this do?
It adds request ID and uses it to correlate trace, logs and events.

- Related Issue #2671
- Resolves #2671

No.

No.

### DIFF
--- a/api/http/authn.go
+++ b/api/http/authn.go
@@ -7,12 +7,10 @@ import (
 	"context"
 	"net/http"
 
-	"github.com/absmach/supermq"
 	apiutil "github.com/absmach/supermq/api/http/util"
 	"github.com/absmach/supermq/auth"
 	smqauthn "github.com/absmach/supermq/pkg/authn"
 	"github.com/go-chi/chi/v5"
-	"github.com/go-chi/chi/v5/middleware"
 )
 
 type sessionKeyType string
@@ -45,21 +43,6 @@ func AuthenticateMiddleware(authn smqauthn.Authentication, domainCheck bool) fun
 
 			ctx := context.WithValue(r.Context(), SessionKey, resp)
 
-			next.ServeHTTP(w, r.WithContext(ctx))
-		})
-	}
-}
-
-func RequestIDMiddleware(idp supermq.IDProvider) func(http.Handler) http.Handler {
-	return func(next http.Handler) http.Handler {
-		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			requestID, err := idp.ID()
-			if err != nil {
-				EncodeError(r.Context(), err, w)
-				return
-			}
-
-			ctx := context.WithValue(r.Context(), middleware.RequestIDKey, requestID)
 			next.ServeHTTP(w, r.WithContext(ctx))
 		})
 	}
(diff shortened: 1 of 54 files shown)


api/http/authn.go
```diff
@@ -47,3 +49,20 @@ func AuthenticateMiddleware(authn smqauthn.Authentication, domainCheck bool) fun
 		})
 	}
 }
+
+func RequestIDMiddleware() func(http.Handler) http.Handler {
```
dborovcanin: This should be in utils.

api/http/authn.go
```diff
@@ -47,3 +49,20 @@ func AuthenticateMiddleware(authn smqauthn.Authentication, domainCheck bool) fun
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			idp := uuid.New()
```
dborovcanin: We should not instantiate a new provider in the handler method, this will result in a new provider for each request.

certs/api/transport.go
```diff
@@ -41,6 +41,8 @@ func MakeHandler(svc certs.Service, authn smqauthn.Authentication, logger *slog.
 
 	r.Group(func(r chi.Router) {
 		r.Use(api.AuthenticateMiddleware(authn, true))
+		r.Use(api.RequestIDMiddleware())
```
dborovcanin: Did you consider using Chi native https://github.com/go-chi/chi/blob/master/middleware/request_id.go? Is the key format the main reason we're not using it?

certs/tracing/tracing.go
```diff
@@ -23,9 +31,28 @@ func New(svc certs.Service, tracer trace.Tracer) certs.Service {
 	return &tracingMiddleware{tracer, svc}
 }
 
+func (tm *tracingMiddleware) startSpan(ctx context.Context, name string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
```
dborovcanin: Can we extract this so we don't repeat it all over?

certs/tracing/tracing.go
```diff
@@ -23,9 +31,28 @@ func New(svc certs.Service, tracer trace.Tracer) certs.Service {
...
+	reqID := middleware.GetReqID(ctx)
+	if reqID != "" {
+		cleanID := strings.ReplaceAll(reqID, separator, emptyString)
+		final := fmt.Sprintf("%032s", cleanID)
```
dborovcanin: Use constant for the format, just like for separator and empty string.

dborovcanin (changes requested):

certs/api/transport.go
nyagamunene: > Did you consider using Chi native https://github.com/go-chi/chi/blob/master/middleware/request_id.go? Is the key format the main reason we're not using it?

The issue is the format which looks like this `1d825ef43136/dFDftFPuGC-000001` which will be challenge to convert to a traceID

dborovcanin (approved):

PR #2717, opened 2025-02-18
NOISSUE - Enable UI prefix when running behind reverse proxy

Musilah:
# What type of PR is this?
This is a feature because it enables UI prefix when running behind reverse proxy.

## What does this do?
Enable UI prefix when running behind reverse proxy.
Now a user can reach the ui via : `http://ui:3000/`

N/A

- Related Issue #
- Resolves #

N/A

N/A


docker/.env
```diff
@@ -151,7 +151,7 @@ SMQ_CLIENTS_URL=http://clients:9006
 SMQ_USERS_URL=http://users:9002
 SMQ_INVITATIONS_URL=http://invitations:9020
 SMQ_DOMAINS_URL=http://domains:9003
-SMQ_UI_HOST_URL=http://localhost:9095
```
ianmuchyri, line 161: This is for the old UI. Remove these env variables

ianmuchyri (changes requested):

docker/nginx/nginx-x509.conf
```diff
@@ -138,6 +138,23 @@ http {
             include snippets/ws-upgrade.conf;
             proxy_pass http://ws-adapter:${SMQ_WS_ADAPTER_HTTP_PORT}/;
         }
+
+       # Proxy Pass to the Magistrala UI
```
arvindh123, line 142: Please align this line with other

docker/docker-compose.yml
```diff
@@ -330,6 +330,8 @@ services:
       - http-adapter
       - ws-adapter
       - coap-adapter
+      - ui
+      - ui-backend
```
arvindh123, line 335: There is no ui or ui-backend in this docker-compose

docker/nginx/nginx-key.conf
```diff
@@ -125,6 +125,23 @@ http {
             include snippets/ws-upgrade.conf;
             proxy_pass http://ws-adapter:${SMQ_WS_ADAPTER_HTTP_PORT}/;
         }
+
+        # Proxy Pass to the Magistrala UI
+        location /ui/ {
```
arvindh123, line 130: In Docker Compose file currently doesn't include ui or ui-backend, so Nginx won't start properly until all the endpoints are accessible. To make this PR functional, we need to add ui and ui-backend to the Docker Compose configuration. Without these, this PR won't be effective.

arvindh123 (changes requested):

docker/nginx/nginx-key.conf
dborovcanin: We will need to add them, as well. Similar to what we did for Magistrala. @Musilah Please take care of this.

dborovcanin:
Given the current limitations on our UI implementation, and that this is only affecting local deployments under reverse proxy, we decided not to introduce this change. It will require a lot of changes on the UI for little benefits. I'm closing this pull request.

PR #2744, opened 2025-03-03
SMQ-2740 - Create events streams per service per action type

### LINKED ISSUE #2740: Create stream per action type

Usually, the event consumer will be interested in only one type of action - create, read, update, delete, and not all of them at the same time. 

We should do event routing accordingly, so instead of having a stream per service, we can have a stream per service per action. We can even tweak it to include a domain, but for now, stream per service per action should be sufficient.

Must-have

### Anything else?

_No response_

felixgateru:
# What type of PR is this?
This is a refactor as it changes the current streams for events from per service to per service per action type.

## What does this do?
- This pr updates the `Publish` method of the events interface to include streamID to enable publishing to different streams at the service level.
- This pr updates all relevent services to use the updated publish method.

- Resolves #2740

Yes, tests have been updated.

No

### DIFF
--- a/channels/events/streams.go
+++ b/channels/events/streams.go
@@ -16,7 +16,19 @@ import (
 	"github.com/go-chi/chi/v5/middleware"
 )
 
-const streamID = "supermq.channels"
+const (
+	supermqPrefix      = "supermq."
+	createStream       = supermqPrefix + channelCreate
+	updateStream       = supermqPrefix + channelUpdate
+	changeStatusStream = supermqPrefix + channelChangeStatus
+	removeStream       = supermqPrefix + channelRemove
+	viewStream         = supermqPrefix + channelView
+	listStream         = supermqPrefix + channelList
+	connectStream      = supermqPrefix + channelConnect
+	disconnectStream   = supermqPrefix + channelDisconnect
+	setParentStream    = supermqPrefix + channelSetParent
+	removeParentStream = supermqPrefix + channelRemoveParent
+)
 
 var _ channels.Service = (*eventStore)(nil)
 
@@ -29,7 +41,7 @@ type eventStore struct {
 // NewEventStoreMiddleware returns wrapper around clients service that sends
 // events to event store.
 func NewEventStoreMiddleware(ctx context.Context, svc channels.Service, url string) (channels.Service, error) {
-	publisher, err := store.NewPublisher(ctx, url, streamID)
+	publisher, err := store.NewPublisher(ctx, url)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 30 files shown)


users/events/streams.go
```diff
@@ -14,7 +14,20 @@ import (
...
+const (
+	supermqPrefix      = "supermq."
+	createStream       = supermqPrefix + userCreate
+	updateStream       = supermqPrefix + userUpdate
+	changeStatusStream = supermqPrefix + changeUserStatus
+	viewStream         = supermqPrefix + userView
+	listStream         = supermqPrefix + userList
+	searchStream       = supermqPrefix + userSearch
+	identifyStream     = supermqPrefix + userIdentify
+	tokenStream        = supermqPrefix + "user.token"
```
arvindh123: @felixgateru Why not we are not using same like operation `issue_token` and `refresh_token`?

arvindh123 (changes requested):

users/events/streams.go
felixgateru: This was done to group the related events in a single stream

dborovcanin (changes requested):
Please rebase. Also, we'll need a new PR to remove operation from the event, since now we have it in the stream.

channels/events/streams.go
```diff
@@ -176,7 +188,7 @@ func (es *eventStore) changeStatus(ctx context.Context, session authn.Session, c
 		Session:   session,
 		requestID: middleware.GetReqID(ctx),
 	}
-	if err := es.Publish(ctx, event); err != nil {
+	if err := es.Publish(ctx, changeStatusStream, event); err != nil {
```
arvindh123: Lets have two status, both `enable` and `disable`, for all entities

channels/events/streams.go
```diff
@@ -89,7 +101,7 @@ func (es *eventStore) update(ctx context.Context, operation string, session auth
 		requestID: middleware.GetReqID(ctx),
 	}
 
-	if err := es.Publish(ctx, event); err != nil {
+	if err := es.Publish(ctx, updateStream, event); err != nil {
```
arvindh123: Lets have two stream id for update channel and update channel tag

arvindh123 (changes requested):
Each service function should have its own stream id and its own operation

dborovcanin:
@felixgateru Please rebase.

arvindh123 (approved):

arvindh123 (changes requested):
Please add these commit to this branch 
https://github.com/felixgateru/supermq/commit/dbcf24b865660e39e42ffe6beed322a1814bfd20

dborovcanin (approved):

arvindh123 (approved):

PR #2747, opened 2025-03-04
NOISSUE - Add default members list base query for role management

SammyOina:
# What type of PR is this?

## What does this do?
- **New Features**
  - Introduced a fallback mechanism for role management, ensuring the system reliably aggregates and presents member data and related actions even when predefined conditions aren’t met.


pkg/roles/repo/postgres/roles.go
```diff
@@ -44,6 +44,8 @@ func NewRepository(db postgres.Database, entityType, tableNamePrefix, entityTabl
 		membersListBaseQuery = groupMembersListBaseQuery()
 	case policies.DomainType:
 		membersListBaseQuery = domainMembersListBaseQuery()
+	default:
```
arvindh123: Instead of having `default` can we provide function to set memberListBaseQuery like `SetMemberListBaseQuery`

arvindh123 (changes requested):

arvindh123 (approved):

PR #2753, opened 2025-03-10
SMQ-2751 - Add search to PATs

### LINKED ISSUE #2751: Feature: Add Search for PAT

The user should be able to search PAT by ID,  name and status with the `ListAll` endpoint like the other entities.

This requires the addition of Status Field in the PAT:
-  `Active`, `Revoked`, `Expired` and `All` status.
-  Requires change in implementation of `RetrieveAll` in postgres.

Must-have

### Anything else?

_No response_

nyagamunene:
# What type of PR is this?
This is a feature because it adds the following functionality: search PAT by name and status with the `ListAll` endpoint.

## What does this do?
It add the following:

- Introduces the following status to PATs: `Active`, `Revoked`, `Expired` and `All` status.
- Changes the implementation of `RetrieveAll` in postgres.

- Related Issue #2751 
- Resolves #2751 

Yes

No

### DIFF
--- a/auth/postgres/pat.go
+++ b/auth/postgres/pat.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/absmach/supermq/auth"
-	repoerr "github.com/absmach/supermq/pkg/errors/repository"
 )
 
 type dbPat struct {
@@ -52,11 +51,7 @@ type dbPagemeta struct {
 	Timestamp   time.Time    `db:"timestamp,omitempty"`
 }
 
-func toAuthPat(db dbPat) (auth.PAT, error) {
-	if db.ID == "" {
-		return auth.PAT{}, repoerr.ErrNotFound
-	}
-
+func toAuthPat(db dbPat) auth.PAT {
 	updatedAt := time.Time{}
 	lastUsedAt := time.Time{}
 	revokedAt := time.Time{}
@@ -73,7 +68,7 @@ func toAuthPat(db dbPat) (auth.PAT, error) {
 		revokedAt = db.RevokedAt.Time
 	}
 
-	pat := auth.PAT{
+	return auth.PAT{
 		ID:          db.ID,
 		User:        db.User,
 		Name:        db.Name,
@@ -87,8 +82,6 @@ func toAuthPat(db dbPat) (auth.PAT, error) {
 		RevokedAt:   revokedAt,
 		Status:      db.Status,
 	}
-
-	return pat, nil
 }
 
 func toAuthScope(dsc []dbScope) ([]auth.Scope, error) {
--- a/auth/postgres/pat.go
+++ b/auth/postgres/pat.go
@@ -21,6 +21,7 @@ type dbPat struct {
 	ExpiresAt   time.Time    `db:"expires_at,omitempty"`
 	UpdatedAt   sql.NullTime `db:"updated_at,omitempty"`
 	LastUsedAt  sql.NullTime `db:"last_used_at,omitempty"`
+	Revoked     bool         `db:"revoked,omitempty"`
 	RevokedAt   sql.NullTime `db:"revoked_at,omitempty"`
 	Status      auth.Status  `db:"status,omitempty"`
 }
@@ -48,6 +49,7 @@ type dbPagemeta struct {
 	Description string       `db:"description"`
 	Secret      string       `db:"secret"`
 	Status      auth.Status  `db:"status"`
+	Timestamp   time.Time    `db:"timestamp,omitempty"`
 }
 
 func toAuthPat(db dbPat) (auth.PAT, error) {
@@ -81,6 +83,7 @@ func toAuthPat(db dbPat) (auth.PAT, error) {
 		ExpiresAt:   db.ExpiresAt,
 		UpdatedAt:   updatedAt,
 		LastUsedAt:  lastUsedAt,
+		Revoked:     db.Revoked,
 		RevokedAt:   revokedAt,
 		Status:      db.Status,
 	}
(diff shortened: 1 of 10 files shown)


ianmuchyri (dismissed):

auth/postgres/repo.go
```diff
@@ -63,18 +63,34 @@ func (pr *patRepo) Retrieve(ctx context.Context, userID, patID string) (auth.PAT
...
+		return auth.PATSPage{}, errors.Wrap(repoerr.ErrViewEntity, err)
+	}
+
+	q := fmt.Sprintf(`
+	WITH updated_pats AS (
+		UPDATE pats
+		SET status = %d
+		WHERE user_id = :user_id AND expires_at < :expires_at
+		RETURNING id
+	)
```
arvindh123: I appreciate this query, but it looks like a hack. 
We are doing update while retrieving

arvindh123 (changes requested):

arvindh123:
Why don't we calculate status while while retrieving in SQL

auth/postgres/repo.go
```diff
@@ -63,18 +63,34 @@ func (pr *patRepo) Retrieve(ctx context.Context, userID, patID string) (auth.PAT
...
+	SELECT 
+		p.id, p.user_id, p.name, p.description, p.secret, p.issued_at, p.expires_at,
+		p.updated_at, p.last_used_at, p.revoked_at, p.status
+		FROM pats p
+		WHERE p.user_id = :user_id %s
+		ORDER BY p.issued_at DESC
+		LIMIT :limit OFFSET :offset`, auth.ExpiredStatus, pageQuery)
```
arvindh123: This PR introduces significant changes, but a more efficient approach is available.

Instead of updating the `status` column in the database, we can dynamically calculate the status using the following SQL query:

```sql
SELECT 
    id,
    name,
    user_id,
    description,
    secret,
    issued_at,
    expires_at,
    updated_at,
    revoked,
    revoked_at,
    last_used_at,
    CASE 
        WHEN revoked = TRUE THEN 'Revoked'
        WHEN expires_at IS NOT NULL AND expires_at < :timestamp THEN 'Expired'
        ELSE 'Active'
    END AS status
FROM pats;
```

This approach avoids unnecessary updates and ensures the `status` is always accurate at retrieval time. It is also more efficient than the approach in this PR, which attempts to update the `status` column.

**Requested Changes:**
- Revert the changes that modify the `revoked` column.
- Revert all modifications to the repository related to updating the `status` column.
- Use the above query instead to calculate the status dynamically.

This will simplify the implementation while ensuring correctness and efficiency.

arvindh123 (changes requested):

auth/postgres/repo.go
```diff
@@ -127,32 +142,69 @@ func (pr *patRepo) RetrieveAll(ctx context.Context, userID string, pm auth.PATSP
...
+		query = append(query, "p.name ILIKE '%' || :name || '%'")
+	}
+
+	if pm.ID != "" {
+		query = append(query, "p.id = :id")
+	}
+
+	if pm.Status != auth.AllStatus {
+		switch pm.Status {
+		case auth.RevokedStatus:
```
arvindh123, line 138: @nyagamunene, In where clause , Does p.Status = :status is not working ? 
Did your tried with this approach?

auth/postgres/repo.go
```diff
@@ -127,32 +142,69 @@ func (pr *patRepo) RetrieveAll(ctx context.Context, userID string, pm auth.PATSP
...
+	}
+
+	var emq string
+	if len(query) > 0 {
+		emq = fmt.Sprintf("AND %s", strings.Join(query, " AND "))
+	}
+	return emq, nil
+}
+
+func (pr *patRepo) RetrieveSecretAndRevokeStatus(ctx context.Context, userID, patID string) (string, auth.Status, error) {
```
arvindh123: I was wondering, is there a particular reason why `RetrieveSecretAndRevokeStatus` uses `auth.Status`? Would it be possible for it to return just `TRUE` or `FALSE` as a `Boolean` instead?

arvindh123 (changes requested):

auth/postgres/repo.go
nyagamunene: @arvindh123 In this approach it is trying to fetch the `status` column which does not exist

auth/postgres/repo.go
nyagamunene: I can revert to how it worked previously

auth/service.go
```diff
@@ -542,6 +545,7 @@ func (svc service) ListPATS(ctx context.Context, token string, pm PATSPageMeta)
 	}
 	patsPage, err := svc.pats.RetrieveAll(ctx, key.User, pm)
 	if err != nil {
+		fmt.Printf("err is %+v\n", err)
```
arvindh123: Please remove this debug line

auth/postgres/pat.go
```diff
@@ -82,6 +85,7 @@ func toAuthPat(db dbPat) (auth.PAT, error) {
 		LastUsedAt:  lastUsedAt,
 		Revoked:     db.Revoked,
 		RevokedAt:   revokedAt,
+		Status:      db.Status,
```
arvindh123, line 83: In this function we can remove 
```diff
-	if db.ID == "" {
-		return auth.PAT{}, repoerr.ErrNotFound
-	}
```

And we can have simply likee 
```
func toAuthPat(db dbPat) (auth.PAT)
```

auth/postgres/repo.go
```diff
@@ -108,10 +122,11 @@ func (pr *patRepo) RetrieveAll(ctx context.Context, userID string, pm auth.PATSP
 			UpdatedAt:   updatedAt,
 			Revoked:     pat.Revoked,
 			RevokedAt:   revokedAt,
+			Status:      pat.Status,
```
arvindh123: Why we could not use `toAuthPat` function here ?

arvindh123 (changes requested):

arvindh123 (approved):

wambui-pixel (approved):

dborovcanin (approved):

PR #2762, opened 2025-03-13
SMQ-2742 - Add entity role check for optional members when adding role

### LINKED ISSUE #2742: Bug: When creating an entity role check optional members part of domain

Create a role for a group, channel or client

We should not be able to create a role with optional members who are not part of the domain

When I create a role for a client , group or channel, I can be able to add users who are not part of the domain.
https://jam.dev/c/ece37b41-d48f-457f-a24f-a122b191c15a

### Steps To Reproduce

The steps are shown in the jam video below
https://jam.dev/c/ece37b41-d48f-457f-a24f-a122b191c15a

Latest supermq

_No response_

nyagamunene:
# What type of PR is this?
This is a bug fix because it fixes the following issue: #2742 

## What does this do?
It adds check for optional members when adding role

- Related Issue #2742 
- Resolves #2742 

No

No

### DIFF
--- a/pkg/roles/rolemanager/middleware/authorization.go
+++ b/pkg/roles/rolemanager/middleware/authorization.go
@@ -9,7 +9,6 @@ import (
 	"github.com/absmach/supermq/pkg/authn"
 	smqauthz "github.com/absmach/supermq/pkg/authz"
 	"github.com/absmach/supermq/pkg/errors"
-	svcerr "github.com/absmach/supermq/pkg/errors/service"
 	"github.com/absmach/supermq/pkg/policies"
 	"github.com/absmach/supermq/pkg/roles"
 	"github.com/absmach/supermq/pkg/svcutil"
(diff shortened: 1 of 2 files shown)


pkg/roles/rolemanager/middleware/authorization.go
```diff
@@ -64,6 +64,9 @@ func (ram RoleManagerAuthorizationMiddleware) AddRole(ctx context.Context, sessi
 	}); err != nil {
 		return roles.RoleProvision{}, err
 	}
+	if err := ram.authorizeMembers(ctx, session, optionalMembers); err != nil {
```
arvindh123: Can we rename this to `validateMembers`

In that function we can return something message like 
`<Member ID> is not member of domain` for other entities
and 
`<Member ID> is not found` for domains

arvindh123 (changes requested):

arvindh123 (approved):

dborovcanin (approved):

PR #2772, opened 2025-03-25
SMQ-2761 - Add route field to channels

### LINKED ISSUE #2761: Add topic field to channels
**Goal**: The goal of this ticket is to replace the channel ID with a human-readable topic.

**Approach**: To achieve this, we can introduce a new field called `topic` in the channels. 

- The `topic` field should be unique within the domain.
- The user should be able to update the topic field.
- The user should provide the topic field during creation.
- If the user doesn't provide a topic field during creation, the ID should be used as the default topic.

**Topic Naming Rules**:
- The topic should start with an alphabetic character, followed by numbers if necessary.
- If the topic is a UUID, we can allow the first character to be either a number or an alphabet.
- The topic can contain symbols, but the first character must always be an alphabet.
- The topic symbol should not conflict with MQTT, NATS, RabbitMQ, or other MQ systems.
  Example:
  In NATS ".", "*", and ">" there use by NATS server
  In MQTT '$,' '*,' '+,' and '# these symbols are used by MQTT server
- Therefore, for now, we will only support the symbols `_` and `-`. 
May be in future we can add more symbols.

felixgateru:
# What type of PR is this?
This is a feature as it adds a route field to channels.

## What does this do?
 -  Introduces  route attribute for channels
 -  Updates relevent repository, service and api methods for this addition
 -  Updates documentation and tests

- Resolves #2761 

Yes, tests have been added for my changes.

Yes, api documentation has been updated for the change

### DIFF
--- a/apidocs/openapi/channels.yaml
+++ b/apidocs/openapi/channels.yaml
@@ -553,6 +553,7 @@ components:
           example: enabled
       required:
         - name
+        - route
 
     ParentGroupReqObj:
       type: object
--- a/apidocs/openapi/channels.yaml
+++ b/apidocs/openapi/channels.yaml
@@ -454,7 +454,7 @@ paths:
           description: Database can't process request.
         "500":
           $ref: "#/components/responses/ServiceError"
- 
+
   /{domainID}/channels/{chanID}/connect:
     post:
       operationId: connectClientsToChannel
@@ -534,14 +534,14 @@ components:
           type: string
           example: channelName
           description: Free-form channel name. Channel name is unique on the given hierarchy level.
-        description:
-          type: string
-          example: long channel description
-          description: Channel description, free form text.
         parent_id:
           type: string
           example: bb7edb32-2eac-4aad-aebe-ed96fe073879
           description: Id of parent channel, it must be existing channel.
+        topic:
+          type: string
+          example: channelTopic
+          description: Channel topic.
         metadata:
           type: object
           example: { "location": "example" }
@@ -557,7 +557,7 @@ components:
     ParentGroupReqObj:
       type: object
       properties:
-         parent_group_id:
+        parent_group_id:
           type: string
           format: uuid
           example: bb7edb32-2eac-4aad-aebe-ed96fe073879
@@ -587,10 +587,10 @@ components:
           format: uuid
           example: bb7edb32-2eac-4aad-aebe-ed96fe073879
           description: Channel parent identifier.
-        description:
+        topic:
           type: string
-          example: long channel description
-          description: Channel description, free form text.
+          example: channelTopic
+          description: Channel topic.
         metadata:
           type: object
           example: { "role": "general" }
@@ -655,10 +655,10 @@ components:
           type: string
           example: channelName
           description: Free-form channel name. Channel name is unique on the given hierarchy level.
-        description:
+        topic:
           type: string
-          example: long description but not too long
-          description: Channel description, free form text.
+          example: channelTopic
+          description: Channel topic.
         metadata:
           type: object
           example: { "role": "general" }
@@ -667,7 +667,7 @@ components:
         - name
         - metadata
         - description
-    
+
     ChannelUpdateTags:
       type: object
       properties:
@@ -817,7 +817,7 @@ components:
         format: uuid
       required: false
       example: bb7edb32-2eac-4aad-aebe-ed96fe073879
-      
+
     Metadata:
       name: metadata
       description: Metadata filter. Filtering is performed matching the parameter with metadata on top level. Parameter is json.
@@ -867,7 +867,7 @@ components:
         application/json:
           schema:
             $ref: "#/components/schemas/ChannelReqObj"
-    
+
     ChannelsCreateReq:
       description: JSON-formatted document describing the new channels to be registered
       required: true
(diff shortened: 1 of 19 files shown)


dborovcanin (changes requested):
@felixgateru Please rebase.

apidocs/openapi/channels.yaml
```diff
@@ -534,14 +534,14 @@ components:
...
           description: Free-form channel name. Channel name is unique on the given hierarchy level.
-        description:
-          type: string
-          example: long channel description
-          description: Channel description, free form text.
         parent_id:
           type: string
           example: bb7edb32-2eac-4aad-aebe-ed96fe073879
           description: Id of parent channel, it must be existing channel.
+        route:
```
arvindh123, line 541: Should the route be added to the required ?

arvindh123 (changes requested):

channels/postgres/init.go
```diff
@@ -62,6 +62,17 @@ func Migration() (*migrate.MemoryMigrationSource, error) {
...
+			{
+				Id: "channels_03",
+				Up: []string{
+					`ALTER TABLE channels ADD COLUMN route VARCHAR(36) NOT NULL;`,
+					`ALTER TABLE channels ADD CONSTRAINT unique_domain_route UNIQUE (domain_id, route);`,
+				},
+				Down: []string{
+					`ALTER TABLE channels DROP CONSTRAINT unique_domain_route;`,
+					`ALTER TABLE channels DROP COLUMN route;`,
+				},
```
arvindh123: I have applied these migration in existing channels 
I get error like this in channels 
```log 
2025-05-21 14:25:56 {"time":"2025-05-21T08:55:56.963692336Z","level":"ERROR","msg":"failed to apply migrations : ERROR: column \"route\" of relation \"channels\" contains null values (SQLSTATE 23502) handling channels_03"}
```
---
In order make this work with existing deployments 
We can simply copy channel id to route 

```sql
-- Step 1: Add the 'route' column, allowing NULLs temporarily
ALTER TABLE channels ADD COLUMN route VARCHAR(36);

-- Step 2: Populate the 'route' column for existing rows with values from 'id'
-- Assuming 'id' is a UUID or a type that can be cast to VARCHAR(36)
-- If 'id' is a different type (e.g., INT), you might need a different casting or conversion
UPDATE channels SET route = id::VARCHAR(36);

-- Step 3: Add the NOT NULL constraint to the 'route' column
-- This will succeed now because all existing rows have a value for 'route'
ALTER TABLE channels ALTER COLUMN route SET NOT NULL;

-- Step 4: Add the unique constraint
ALTER TABLE channels ADD CONSTRAINT unique_domain_route UNIQUE (domain_id, route);
```

arvindh123 (changes requested):

arvindh123 (dismissed):

channels/postgres/init.go
```diff
@@ -62,6 +62,19 @@ func Migration() (*migrate.MemoryMigrationSource, error) {
 					`ALTER TABLE channels ADD CONSTRAINT channels_domain_id_name_key UNIQUE (domain_id, name)`,
 				},
 			},
+			{
+				Id: "channels_03",
+				Up: []string{
+					`ALTER TABLE channels ADD COLUMN route VARCHAR(36);`,
+					`UPDATE channels SET route = id WHERE route IS NULL;`,
+					`ALTER TABLE channels ALTER COLUMN route SET NOT NULL;`,
+					`ALTER TABLE channels ADD CONSTRAINT unique_domain_route UNIQUE (domain_id, route);`,
```
dborovcanin: Why do we make route required?

apidocs/openapi/channels.yaml
dborovcanin: Yes, as of the current design. However, I don't think we should enforce it.

channels/postgres/init.go
dborovcanin: If nothing, routes **must not be in UUID format** so we can easily and quickly make a difference between ID and route; something like username and email.

dborovcanin (changes requested):

channels/api/http/requests.go
```diff
@@ -25,6 +25,9 @@ func (req createChannelReq) validate() error {
 			return apiutil.ErrMissingChannelID
 		}
 	}
+	if req.Channel.Route != "" {
+		return api.ValidateRoute(req.Channel.Route)
+	}
```
arvindh123, line 35:
```suggestion
	if req.Channel.Route != "" {
		if err := api.ValidateRoute(req.Channel.Route); err != nil {
			return err
		}
	}
```

arvindh123 (commented):
LGTM

channels/api/http/endpoint_test.go
```diff
@@ -140,6 +152,18 @@ func TestCreateChannelEndpoint(t *testing.T) {
 			status:      http.StatusBadRequest,
 			err:         apiutil.ErrNameSize,
 		},
+		{
+			desc:     "create channel with invalid route format",
+			token:    validToken,
+			domainID: validID,
+			req: channels.Channel{
+				Name:  valid,
+				Route: "__invalid",
```
dborovcanin, line 161: Also, add a check for UUID format. We must not allow UUIDs in routes to prevent confusion and make our lives easier.

dborovcanin: Add test here as well:
```go
		{
			desc:     "create channel with UUID route",
			token:    validToken,
			domainID: validID,
			req: channels.Channel{
				Name:  valid,
				Route: testsutil.GenerateUUID(t),
			},
			contentType: contentType,
			status:      http.StatusBadRequest,
			err:         apiutil.ErrInvalidRouteFormat,
		},
```
Validation testing is fine, but it only works if **decoding request works properly**, so we to test it here as well.

dborovcanin (approved):

PR #2773, opened 2025-03-26
NOISSUE - Add send_invitation permission

WashingtonKK:
# What type of PR is this?

- Related Issue #
- Resolves #

### DIFF
--- a/docker/spicedb/schema.zed
+++ b/docker/spicedb/schema.zed
@@ -360,7 +360,7 @@ definition domain {
 	permission group_manage_role_permission = group_manage_role + team->group_manage_role + organization->admin
 	permission group_add_role_users_permission = group_add_role_users + team->group_add_role_users + organization->admin
 	permission group_remove_role_users_permission = group_remove_role_users + team->group_remove_role_users + organization->admin
-	permission group_view_role_users_permission = group_view_role_users + team->group_view_role_users + organization->
+	permission group_view_role_users_permission = group_view_role_users + team->group_view_role_users + organization->admin
 	
 	permission send_invitation = admin
 }
--- a/docker/spicedb/schema.zed
+++ b/docker/spicedb/schema.zed
@@ -360,8 +360,9 @@ definition domain {
 	permission group_manage_role_permission = group_manage_role + team->group_manage_role + organization->admin
 	permission group_add_role_users_permission = group_add_role_users + team->group_add_role_users + organization->admin
 	permission group_remove_role_users_permission = group_remove_role_users + team->group_remove_role_users + organization->admin
-	permission group_view_role_users_permission = group_view_role_users + team->group_view_role_users + organization->admin
-
+	permission group_view_role_users_permission = group_view_role_users + team->group_view_role_users + organization->
+	
+	permission send_invitation = admin
 }
 
 // Add this relation and permission in future while adding organization
--- a/domains/middleware/authorization.go
+++ b/domains/middleware/authorization.go
@@ -22,10 +22,7 @@ import (
 var _ domains.Service = (*authorizationMiddleware)(nil)
 
 // ErrMemberExist indicates that the user is already a member of the domain.
-var (
-	ErrMemberExist    = errors.New("user is already a member of the domain")
-	ErrPermissionFail = errors.New("billing permission failed")
-)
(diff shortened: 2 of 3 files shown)


domains/middleware/authorization.go
```diff
@@ -262,3 +266,19 @@ func (am *authorizationMiddleware) extAuthorize(ctx context.Context, subj, perm,
...
+		SubjectType: policies.UserType,
+		SubjectKind: policies.UsersKind,
+		Subject:     subj,
+		Permission:  perm,
+		ObjectType:  policies.DomainType,
+		Object:      domainID,
+	}
+	if err := am.authz.Authorize(ctx, req); err != nil {
+		return err
+	}
```
SammyOina: we do not need a new function, just edit the current permission check

SammyOina (changes requested):

domains/middleware/authorization.go
```diff
@@ -22,7 +22,10 @@ import (
 var _ domains.Service = (*authorizationMiddleware)(nil)
 
 // ErrMemberExist indicates that the user is already a member of the domain.
-var ErrMemberExist = errors.New("user is already a member of the domain")
+var (
+	ErrMemberExist    = errors.New("user is already a member of the domain")
+	ErrPermissionFail = errors.New("billing permission failed")
```
SammyOina: there can be many auth callouts so error will be passed from auth service

domains/middleware/authorization.go
```diff
@@ -151,6 +154,10 @@ func (am *authorizationMiddleware) SendInvitation(ctx context.Context, session a
 		return err
 	}
 
+	if err := am.extAuthorize(ctx, auth.EncodeDomainUserID(invitation.DomainID, session.UserID), policies.SendInvitationPermission, policies.DomainType, invitation.DomainID); err != nil {
+		return errors.Wrap(svcerr.ErrAuthorization, ErrPermissionFail)
```
SammyOina: wrap err and authorization error

SammyOina (changes requested):

SammyOina (approved):

docker/spicedb/schema.zed
```diff
@@ -361,7 +361,8 @@ definition domain {
 	permission group_add_role_users_permission = group_add_role_users + team->group_add_role_users + organization->admin
 	permission group_remove_role_users_permission = group_remove_role_users + team->group_remove_role_users + organization->admin
 	permission group_view_role_users_permission = group_view_role_users + team->group_view_role_users + organization->admin
-
+	
+	permission send_invitation = admin
```
arvindh123, line 365: Why don't we reuse the permission `admin` or `add_role_users_permission`?, 
https://github.com/absmach/supermq/blob/d842182d160a442e943ed97a07902f54512275da/docker/spicedb/schema.zed#L323
Because , invitation is similar to adding user to a role.

arvindh123 (changes requested):

docker/spicedb/schema.zed
SammyOina: adding user to role is not exactly the same. since invitations can be rejected.

The idea here is with billing if we are to limit the number of users in a workspace. we need to limit sending invitations and accepting invitations at the same time. 

some psudocode:

1. Send user inviation
2. Billing callback checks if domain-members limit is exceeded. if not invitation sending is authorized

1. when user is accepting invitation ("adding role user")
2. billing will also check domain members limit. Allow the user to be added to domain or not

arvindh123: Got it, 
Then Can we have separate relation like `relation send_invitations: role#member`? 
and separate permission like `permission send_invitation_permission = send_invitations + organization->admin`

What do you think about this approach  ?

SammyOina: Not sure if a new relation is needed, since it will add the number of options the user has on actions, but I don't mind if you'd like to add it on mg. Billing only cares about the permission in the auth request

arvindh123: @SammyOina 
This can be avoided , if we move callout out of auth service 
I have created issue here for this 
https://github.com/absmach/supermq/issues/2804

WashingtonKK:
Fixed here: https://github.com/absmach/supermq/pull/2823

Issue #2702, opened 2025-02-11
Add routes caching


dborovcanin:
We will allow **messaging API** to use names alongside IDs for simplified topics. We need a cache to map the name to the ID.

dborovcanin:
The main challenge here is the invalidation of cache. If any entity changes its name, we need to invalidate the cache.

dborovcanin:
This implies that both Domain and Channel have **topic** field - it's what we now call domain alias, and a new field we need to add to a channel. @ianmuchyri We will need an explanation at UI, similar to what we need for aliases.

arvindh123:
# Present flow 

![Image](https://github.com/user-attachments/assets/49cf38e3-eead-47cf-aa23-52a608b8058b)

# Problem

In MQTT subscribe, we could not subscribe if topic uses `route` example : `<d_route>/c/<ch_route>/m/<subtopic>`


**Reason:** 
- MQTT client directly subscribe to MQTT server with route topic `<d_route>/c/<ch_route>/m/<subtopic>`
  
- In Broker we publish all the messages with IDs 
  
- Forwarder subscribes to all the topic from broker and forward to format `<d_id>/c/<ch_id>/m/<subtopic>`

- But the MQTT client subscribed to topic `<d_route>/c/<ch_route>/m/<subtopic>`, so it will not get data 

![Image](https://github.com/user-attachments/assets/83bb1ce2-5af5-4b65-9a34-f127c737952a)


# Solution:
If MQTT clients do subcribe with route topic `<d_route>/c/<ch_route>/m/<subtopic>`, By using interceptors, we can modify the packets and forward to MQTT broker to subscribe to topic with ids `<d_id>/c/<ch_id>/m/<subtopic>`

This requires MQTT adapter to access to the domain service and channels service to retrieve the `id and route` 


![Image](https://github.com/user-attachments/assets/3fe7e9a2-fb74-4c5e-98a3-591733bd962b)

**Additional advantage:** This approach remove the requirement of global cache

dborovcanin:
I'm afraid this implicit rewrite of topics is not a good idea since:
1. it can cause hidden bugs and between-protocol communication routing problems
2. We still need a cache, unless we want to do the same procedure for each Publish/HTTP post message

arvindh123:
> I'm afraid this implicit rewrite of topics is not a good idea since:
> 
> 1. it can cause hidden bugs and between-protocol communication routing problems
[...]

PR #2775, opened 2025-03-27
SMQ-2761 - Support domain and channel routes in message topic

### LINKED ISSUE #2761: Add topic field to channels

felixgateru:
# What type of PR is this?
This is a feature as it adds support for domain and channel routes in the message topic.

## What does this do?
This pr:
- Introduces support for domain and channel route in place of respective ids in channel topic
- Updates adapters to resolve channel and domain routes

- Related Issue #2702 

Yes, tests have been updated for the changes

Yes, in code documentation has been updated.

### Notes
To be merged after:
1. https://github.com/absmach/supermq/pull/2765
2. https://github.com/absmach/supermq/pull/2772


dborovcanin (changes requested):
Let's use `{domain,channel}_prefix` instead of `{domain,channel}_id` and `{domain,channel}_route` in doc comments and explain it can be either ID or route in the documentation.

arvindh123 (changes requested):
This PR remove the changes done in PR https://github.com/absmach/supermq/pull/2889
Could you please align with this PR https://github.com/absmach/supermq/pull/2889 and main ?

http/handler.go
```diff
@@ -119,10 +126,18 @@ func (h *handler) Publish(ctx context.Context, topic *string, payload *[]byte) e
...
 		return errors.Wrap(errMalformedTopic, err)
 	}
+	domainID, err := h.resolveDomain(ctx, domain)
+	if err != nil {
+		return errors.Wrap(errFailedResolveDomain, err)
+	}
+	channelID, err := h.resolveChannel(ctx, channel, domainID)
+	if err != nil {
+		return errors.Wrap(errFailedResolveChannel, err)
+	}
```
arvindh123, line 140: Can we try to do this in Pre Handler Interceptor? 
And pass as ID to Handler ?

arvindh123, line 140: I think HTTP have no interceptor, So this is not possible here.

coap/api/transport.go
```diff
@@ -32,13 +36,17 @@ const (
...
+	errFailedResolveDomain  = errors.New("failed to resolve domain route")
+	errFailedResolveChannel = errors.New("failed to resolve channel route")
 )
 
 var (
-	logger  *slog.Logger
-	service coap.Service
+	logger   *slog.Logger
+	service  coap.Service
+	channels grpcChannelsV1.ChannelsServiceClient
```
arvindh123, line 48: We need to revisit this approach later 
Including the `coap.Service`

http/handler.go
arvindh123: We should implement an HTTP/WebSocket interceptor in `mGate`, similar to how we handled MQTT. The related logic can go into `http/main.go`.

dborovcanin:
@felixgateru Please rebase this one. We can handle this in mGate later, it's OK to keep it in the HTTP adapter for the time being.

arvindh123 (approved):

dborovcanin:
This should be part of the `authentication` procedure. I.e., fetching domains via gRPC, then fetching Channels, again using gRPC, then calling AuthN, and then AthZ is an extremely inefficient way to handle routes.

dborovcanin:
Merging this one since it does not affect the latency in the ID-based approach, but for routes, we need to come up with a better and more efficient solution. Caching is one step, but that's not enough.

PR #2784, opened 2025-04-01
NOISSUE - Update Mockery version and config

felixgateru:
# What type of PR is this?
This is a dependency update as it updates the version of Mockery to `v3.0.0` and updates the config file to generate mocks from the config file rather than go generate.

## What does this do?
This pr:
- Updates the version of mockery to `v3.0.0`
- Updates mocks generation from using `go generate` to using the config file as expressed in the version change documentation.

Noissue

No

No


Makefile
```diff
@@ -21,7 +21,7 @@ DOCKER_PROJECT ?= $(shell echo $(subst $(space),,$(USER_REPO)) | sed -E 's/[^a-z
 DOCKER_COMPOSE_COMMANDS_SUPPORTED := up down config restart
 DEFAULT_DOCKER_COMPOSE_COMMAND  := up
 GRPC_MTLS_CERT_FILES_EXISTS = 0
-MOCKERY_VERSION=v2.43.2
+MOCKERY_VERSION=v2.53.3
```
dborovcanin: Let's switch to [v3](https://github.com/absmach/magistrala/blob/main/Makefile#L22). It's beta, but it's coming soon.

dborovcanin (changes requested):

.github/workflows/check-generated-files.yaml
```diff
@@ -115,12 +115,13 @@ jobs:
       - name: Check Mocks are up to Date
         if: steps.changes.outputs.mocks == 'true'
         run: |
-          MOCKERY_VERSION=v2.43.2
-          go install github.com/vektra/mockery/v2@$MOCKERY_VERSION
+          MOCKERY_VERSION=v3.0.0-beta.6
+          go install github.com/vektra/mockery/v3@$MOCKERY_VERSION
```
dborovcanin: Makefile handles Mockery installation, so this line is probably not needed.

.github/workflows/check-generated-files.yaml
```diff
@@ -115,12 +115,13 @@ jobs:
...
 
           mv ./auth/mocks/token_client.go ./auth/mocks/token_client.go.tmp
           mv ./auth/mocks/authz.go ./auth/mocks/authz.go.tmp
-          mv ./auth/mocks/keys.go ./auth/mocks/keys.go.tmp
+          mv ./auth/mocks/key_repository.go ./auth/mocks/key_repository.go.tmp
```
dborovcanin: This whole logic can be simplified by using loops, something like:
```bash
for f in $(find . -type f -path '*/mocks/*'); do
 mv $f $f.tmp
...
done
```
and the same for check.

tools/config/.mockery.yaml
```diff
@@ -1,10 +1,15 @@
...
-outpkg: "mocks"
-boilerplate-file: "./tools/config/boilerplate.txt"
+pkgname: mocks
+mockname: "{{.InterfaceName}}"
+filename: "{{snakecase .InterfaceName}}.go"
+dir: "{{.InterfaceDirRelative}}/mocks"
+template-data:
+  boilerplate-file: ./tools/config/boilerplate.txt
+  with-expecter: true
+force-file-write: true
```
dborovcanin, line 11: I like this, I missed it in the docs. Fixed Magistrala to use it. :)

dborovcanin (changes requested):

dborovcanin (approved):

PR #2788, opened 2025-04-02
NOISSUE - Fix callback

nyagamunene:
# What type of PR is this?

- Related Issue #
- Resolves #

### DIFF
--- a/auth/callback.go
+++ b/auth/callback.go
@@ -81,12 +81,12 @@ func (c *callback) Authorize(ctx context.Context, pr policies.Policy) error {
 	// We use a single URL at a time and others as fallbacks
 	// the first positive result returned by a callback in the chain is considered to be final
 	for i := range c.urls {
-		if err = c.makeRequest(ctx, c.urls[i], payload); err == nil {
-			return nil
+		if err = c.makeRequest(ctx, c.urls[i], payload); err != nil {
+			return err
 		}
 	}
 
-	return err
+	return nil
 }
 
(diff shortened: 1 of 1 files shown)


auth/callback.go
```diff
@@ -81,12 +81,12 @@ func (c *callback) Authorize(ctx context.Context, pr policies.Policy) error {
...
 	for i := range c.urls {
-		if err = c.makeRequest(ctx, c.urls[i], payload); err == nil {
-			return nil
+		if err = c.makeRequest(ctx, c.urls[i], payload); err != nil {
+			return err
 		}
 	}
 
-	return err
+	return nil
```
SammyOina: this changes the logic described in the above comment

nyagamunene: If we will be using multiple URLs this logic will have to be changed.

auth/callback.go
```diff
@@ -81,12 +81,12 @@ func (c *callback) Authorize(ctx context.Context, pr policies.Policy) error {
 	// We use a single URL at a time and others as fallbacks
 	// the first positive result returned by a callback in the chain is considered to be final
 	for i := range c.urls {
-		if err = c.makeRequest(ctx, c.urls[i], payload); err == nil {
-			return nil
+		if err = c.makeRequest(ctx, c.urls[i], payload); err != nil {
```
dborovcanin: Let's iterate like `for _, url := range c.urls ...`

dborovcanin (approved):

PR #2805, opened 2025-04-09
SMQ-2785 - Disable deleting a confirmed or rejected invitation by non invitee user

### LINKED ISSUE #2785: Bug: Invitation can be removed by sender after being rejected/accepted

I'm removing the invitation.

I expect that the inviter is not able to remove the invitation if it's been handled (accepted or rejected) by the invited party.

I can delete it, which removes the trace on how the user has been added to the domain.

### Steps To Reproduce

_No response_

I'm using the latest SuperMQ Docker images.

_No response_

felixgateru:
# What type of PR is this?
This is a bug fix as it fixes the following issue: #2785

## What does this do?
This pr:
- adds checks for invitation state before deleting
- updates tests

- Resolves #2785 

Yes, I have updated tests

No

### DIFF
--- a/domains/service.go
+++ b/domains/service.go
@@ -300,11 +300,12 @@ func (svc *service) DeleteInvitation(ctx context.Context, session authn.Session,
 		return errors.Wrap(svcerr.ErrRemoveEntity, err)
 	}
 
-	if inv.InvitedBy == session.UserID {
-		if err := svc.repo.DeleteInvitation(ctx, inviteeUserID, domainID); err != nil {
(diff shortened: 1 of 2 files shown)


arvindh123 (approved):

arvindh123 (changes requested):
I can able to delete the user after accepting invitation

arvindh123 (approved):
LGTM , 
This requires changes in UI 

UI Doesn't have provision to resent invitation for rejected invitations. 

In UI we should disable delete button for invitations accepted 

For rejected invitation we should have resend button instead of delete button

PR #2825, opened 2025-04-17
NOISSUE - Update mGate version in http and ws adapters

arvindh123:
# What type of PR is this?

- Related Issue #
- Resolves #

### Notes
Should be merged after https://github.com/absmach/mgate/pull/87


dborovcanin:
@arvindh123 mGate PR has been merged. Please update the PR.

ws/api/endpoints.go
```diff
@@ -16,25 +19,54 @@ import (
...
+
+		sessionID, err := generateSessionID()
+		if err != nil {
+			logger.Warn(fmt.Sprintf("Failed to generate session id: %s", err.Error()))
+			http.Error(w, "", http.StatusInternalServerError)
+			return
+		}
+
+		responseHeader := http.Header{}
+		responseHeader.Set("X-SMQ-WS-Session-ID", sessionID)
```
dborovcanin: Why do we use this custom header?

dborovcanin (changes requested):

ws/api/endpoints.go
arvindh123: To identify the WS client.

If two ws client uses same thing id (client id), then if any one disconnect , another will not get message
So to avoid it,  we need unique id for each ws connection.

dborovcanin (approved):

Issue #2831, opened 2025-04-22
Bug: Makefile fails with '@unset' – not recognized across all environments


nintran52:
### What were you trying to achieve?

I wanted to unset a Makefile variable (GRPC_TLS or GRPC_MTLS) when the other is enabled, to ensure mutual exclusivity before running Docker Compose via the make run target.

### What are the expected results?

- When GRPC_MTLS=true, the @unset GRPC_TLS command should remove the variable.

- The make run command should complete without errors, with only one of GRPC_TLS or GRPC_MTLS being active.

### What are the received results?

When running make run, I get the following error:

`make run`

```
make: unset: No such file or directory
make: *** [check_mtls] Error 1
```

This error is due to @unset being used, which is:

- Not available in Windows environments (e.g., CMD or PowerShell)

- Executed in a subshell in Unix-based systems (Linux/macOS), making it ineffective across lines

### Steps To Reproduce

Run:

`make run`

Error occurs:

```
make: unset: No such file or directory
make: *** [check_mtls] Error 1
```

### In what environment did you encounter the issue?

- OS: macOS Ventura / Ubuntu 22.04 / Windows 10

- Make: GNU Make 4.3+

- Shell: Bash 5.x / CMD / PowerShell / Git Bash

### Additional information you deem important

- `unset` is a shell builtin (not an external command), and does not persist in Makefile since each line is executed in a separate shell.

- On Windows, unset is not recognized in CMD or PowerShell.

- Recommended fix: avoid @unset; use $(eval GRPC_TLS=) instead:

  ```$(eval GRPC_TLS=)```

- Or enforce logic using conditional $(error ...) to catch mutual conflict.

dborovcanin:
This has been resolved by the linked PR.

PR #2834, opened 2025-04-22
SMQ-2831 - Fix Makefile '@unset' incompatibility on macOS

nintran52:
# What type of PR is this?
This is a bug fix because it resolves issue #2831 where using `@unset` in Makefile caused errors or had no effect depending on the platform.

## What does this do?
- Replaces `@unset` with `$(eval)` in Makefile rules (`check_tls` and `check_mtls`)
- Ensures variables like `GRPC_TLS` and `GRPC_MTLS` are properly unset within Makefile's scope
- Makes the logic portable and predictable across Linux, macOS, and Windows environments

- Related Issue #2831
- Resolves #

No, because this change affects Makefile logic used at build/config time, which is not covered by unit tests. Manual testing confirms cross-platform compatibility.

No documentation changes were needed as this is an internal bug fix for script logic.

### Notes
This fix is important for developers running `make run` or related commands in environments like macOS or CI/CD systems. It eliminates a class of non-portable shell commands inside the Makefile.


nintran52:
@dborovcanin could you take a look

dborovcanin:
Hello, @nintran52. Thanks for the contribution. I could not test it with MacOS, but I'll trust you with it. :) However, I tested on Linux, and it looks like this PR breaks the Makefile for Linux:
```bash
GRPC_TLS=true make run args=-d
Makefile:229: *** missing separator.  Stop.
```
Can you try something like this, instead?
```make
check_tls:
ifeq ($(GRPC_TLS),true)
	@echo "gRPC TLS is enabled"
	$(eval GRPC_MTLS :=)
else
	$(eval GRPC_TLS :=)
endif

check_mtls:
ifeq ($(GRPC_MTLS),true)
	@echo "gRPC MTLS is enabled"
	$(eval GRPC_TLS :=)
else
	$(eval GRPC_MTLS :=)
endif
```
This works on Linux, but I can't test it on MacOS.

nintran52:
@dborovcanin I updated

dborovcanin (approved):

dborovcanin (changes requested):
It looks like you are using spaces, please replace them with tabs since Make only accepts tabs.

nintran52:
I replaced spaces with tabs for the file @dborovcanin

dborovcanin (approved):

Issue #2838, opened 2025-04-27
Bug: Running make run_addons vault removes supermq-base-net and causes Docker error


nintran52:
### What were you trying to achieve?

I was trying to launch the Vault addon service by running:
`make run_addons vault`

after having already launched the main SuperMQ services via:
`make run`

### What are the expected results?

I expected Vault to start normally and join the existing supermq-base-net Docker network without affecting or removing the network.

### What are the received results?

While trying to run make run_addons vault, I received this error:
```
 ✔ Network nintran52_supermq_supermq-base-net  Removed                                  
Error response from daemon: error while removing network: network nintran52_supermq_supermq-base-net id...
```

This error shows that make run_addons vault attempted to remove the network supermq-base-net, which still had active endpoints attached from running services.

### Steps To Reproduce

Start main SuperMQ services:
`make run`

Start the Vault addon:
`make run_addons vault`

Observe that Docker tries to remove the supermq-base-net network and encounters an error.

### In what environment did you encounter the issue?

OS: macOS 14.4.1 (Sonoma)

Docker Desktop version: 4.40.0

Docker Engine version: 28.0.4

Docker Compose version: 2.34.0

### Additional information you deem important

- The supermq-base-net network is created and used by both main services (make run) and addons (make run_addons vault).

- Running make run_addons vault independently tries to recreate the network even though it already exists and is in use.

- Modify docker/addons/vault/docker-compose.yaml to mark the network supermq-base-net as external: true.

- Example change:
```
networks:
  supermq-base-net:
    external: true
```

- This way, Vault will join the existing network instead of trying to recreate or remove it.

dborovcanin:
This bug has been fixed by the mentioned pull requests.

PR #2839, opened 2025-04-27
SMQ-2838 - Mark supermq-base-net as external in addons docker-compose

nintran52:
# What type of PR is this?
This is a bug fix because it fixes the following issue: active endpoints error when running make run_addons vault after make run.

This PR modifies the docker/addons/.../docker-compose.yaml file to mark the shared Docker network supermq-base-net as external: true.

By doing this, the addons will reuse the existing network instead of trying to recreate or remove it, thus avoiding Docker errors when the network has active endpoints.

- Related Issue #2838 
- Resolves #

No, I have not included tests because this change only modifies the Docker Compose configuration for services orchestration, which is usually tested manually during deployment.

No, I have not updated the documentation because the change is in the deployment layer only (docker-compose network configuration).

### DIFF
--- a/docker/addons/certs/docker-compose.yaml
+++ b/docker/addons/certs/docker-compose.yaml
@@ -8,6 +8,8 @@
 
 networks:
   supermq-base-net:
+    external: true
+    name: supermq-base-net
 
 volumes:
   supermq-certs-db-volume:
@@ -118,7 +120,7 @@ services:
       AM_JAEGER_URL: ${SMQ_JAEGER_URL}
       AM_JAEGER_TRACE_RATIO: ${SMQ_JAEGER_TRACE_RATIO}
     volumes:
-     - ./config.yaml:/config/config.yaml
+      - ./config.yaml:/config/config.yml
     ports:
       - 9010:9010
       - 7012:7012
--- a/docker/addons/certs/docker-compose.yaml
+++ b/docker/addons/certs/docker-compose.yaml
@@ -8,8 +8,7 @@
 
 networks:
   supermq-base-net:
-    external: true
-    name: supermq-base-net
+
 
 volumes:
   supermq-certs-db-volume:
--- a/docker/addons/journal/docker-compose.yaml
+++ b/docker/addons/journal/docker-compose.yaml
@@ -10,8 +10,7 @@
 
 networks:
   supermq-base-net:
-    external: true
-    name: supermq-base-net
+
 
 volumes:
   supermq-journal-volume:
--- a/docker/addons/journal/docker-compose.yaml
+++ b/docker/addons/journal/docker-compose.yaml
@@ -10,9 +10,12 @@
 
 networks:
   supermq-base-net:
+    external: true
+    name: supermq-base-net
 
 volumes:
   supermq-journal-volume:
+
 
 services:
   journal-db:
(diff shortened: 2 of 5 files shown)


nintran52:
@dborovcanin help me review if you have time

docker/addons/journal/docker-compose.yaml
```diff
@@ -10,10 +10,13 @@
 
 networks:
   supermq-base-net:
+    external: true
+    name: supermq-base-net
 
 volumes:
   supermq-journal-volume:
 
+
```
dborovcanin, line 18: Remove unnecessary empty lines (we already have one, there's no need for two).

docker/addons/certs/docker-compose.yaml
```diff
@@ -8,6 +8,8 @@
 
 networks:
   supermq-base-net:
+    external: true
+    name: supermq-base-net
```
dborovcanin: Is this needed? We are usually running using `make` commands, but regardless, the compose networking is written with a [single project on mind](https://docs.docker.com/compose/how-tos/networking/), so setting up the compose project name should solve this problem.

dborovcanin (changes requested):

docker/addons/journal/docker-compose.yaml
nintran52: It is formatted in docker-compose format (because above has `supermq-certs-db-volume:` no value) so it generates 2 line breaks

docker/addons/certs/docker-compose.yaml
nintran52: Currently, since we are running `docker compose` with the `-p project_name` flag (for example, `-p nintran52_supermq` via Makefile),  
the `supermq-base-net` network is **managed by Docker Compose** under the project scope.  

As a result, even if we only run `docker compose up` (not down), Docker Compose **tries to recreate** the network to match the definitions in the current compose file,  
because without `external: true`, it assumes that it owns and controls the network.

If the network already exists and is shared by other active containers (such as `auth`, `clients`, etc.),  
this leads to the error:

```
error while removing network: has active endpoints
```

**Solutions to fix this cleanly:**

**Option 1** (Simpler for development):

- In the main `docker/docker-compose.yaml`, define the network like this:
  ```yaml
  networks:
    supermq-base-net:
      driver: bridge
      name: supermq-base-net
  ```
- This forces the network to have a fixed global name (`supermq-base-net`), without project name prefixing.
- Result: No more conflict across main services and addons.

**Option 2** (More robust and production-friendly):

- In every addon `docker-compose.yaml` (e.g., `docker/addons/vault/docker-compose.yaml`), define the network as:
  ```yaml
  networks:
    supermq-base-net:
      external: true
  ```
- This marks the network as an external resource.
- Docker Compose will **only connect** containers to the network, **without creating or removing it**.

Setting `external: true` in each addon compose file is the cleaner and more robust solution for the long term,  but setting `name: supermq-base-net` in the main compose file is sufficient and simpler for local development.

nintran52:
@dborovcanin wdyt

docker/addons/certs/docker-compose.yaml
dborovcanin: Thanks for the detailed explanation. Since compose is the development-first setup, let's go with option 1), which you have already implemented.

dborovcanin (approved):

PR #2857, opened 2025-05-07
SMQ-2841 - List clients and channels with no parent group

### LINKED ISSUE #2841: Feature: List  clients and channels with no parent group

Yes, Could not able to list  clients and channels which have not parent group 

Adding feature to list clients and channels which have no parent group 

Must-have

### Anything else?

_No response_

nyagamunene:
# What type of PR is this?
It adds `no_parent_id` parameter to `pagemetadata` when listing for both clients and channels.

## What does this do?
- It adds `no_parent_id` parameter to `pagemetadata` when listing for both clients and channels.
- It update the `pageQuery` method in the database query.

- Related Issue #2841 
- Resolves #2841 

Yes

No

### DIFF
--- a/channels/channels.go
+++ b/channels/channels.go
@@ -57,6 +57,7 @@ type Page struct {
 	Tag            string   `json:"tag,omitempty"`
 	Status         Status   `json:"status,omitempty"`
 	Group          string   `json:"group,omitempty"`
+	NoParentGroup  bool     `json:"no_parent_group,omitempty"`
 	Client         string   `json:"client,omitempty"`
 	ConnectionType string   `json:"connection_type,omitempty"`
 	RoleName       string   `json:"role_name,omitempty"`
--- a/channels/channels.go
+++ b/channels/channels.go
@@ -56,8 +56,7 @@ type Page struct {
 	Domain         string   `json:"domain,omitempty"`
 	Tag            string   `json:"tag,omitempty"`
 	Status         Status   `json:"status,omitempty"`
-	Group          string   `json:"group,omitempty"`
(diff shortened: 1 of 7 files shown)


channels/channels.go
```diff
@@ -57,6 +57,7 @@ type Page struct {
 	Tag            string   `json:"tag,omitempty"`
 	Status         Status   `json:"status,omitempty"`
 	Group          string   `json:"group,omitempty"`
+	NoParentGroup  bool     `json:"no_parent_group,omitempty"`
```
dborovcanin: Let's reuse `Group` instead. We need to change it to the pointer, where nil indicates no interest in this query param, an empty value indicates no parent, and a non-empty value indicates a desired parent. We can also use NullableString here, but let's not do it in this PR.

dborovcanin (changes requested):

channels/postgres/channels.go
```diff
@@ -1250,9 +1250,16 @@ func PageQuery(pm channels.Page) (string, error) {
 	if pm.Domain != "" {
 		query = append(query, "c.domain_id = :domain_id")
 	}
-	if pm.Group != "" {
+
+	switch {
```
dborovcanin: You can simplify:
```go
switch {
    case pm.Group == nil:
        break
    case *pm.Group == "":
            query = ....
    default:
        query = ....
   }
}
```
but I prefer:
```go
if pm.Group != nil {
    switch *pm.Group {
    case "":
        query = ...
    default:
        query = ....
    }
}
```

dborovcanin (changes requested):

dborovcanin (approved):

Issue #2842, opened 2025-04-30
Bug:  Journal service could not consume new events


arvindh123:
### What were you trying to achieve?

View the Audit log of a entity

### What are the expected results?

No Audit logs

### What are the received results?

Audit logs

### Steps To Reproduce

Run the SuperMQ + Magistrala 
Then sent lot of messages 

### In what environment did you encounter the issue?

Docker

### Additional information you deem important

Journal service failed to handle message and the failed messages are not acknoweldged by NAST Handler 
So again these messages are queue backed to NATS consumer - `events__-jorunal`

```
{"time":"2025-04-29T14:17:23.054395618Z","level":"WARN","msg":"Save journal failed","duration":"3.453307ms","request_id":"","journal":{"occurred_at":"2025-04-28T14:17:43.625060352Z","operation":"messaging.client_publish"},"error":{"error":"","message":"entity not found"}}
{"time":"2025-04-29T14:17:23.054482056Z","level":"WARN","msg":"Failed to handle SuperMQ message: failed to handle nats event: entity not found"}
{"time":"2025-04-29T14:17:23.058153203Z","level":"WARN","msg":"Save journal failed","duration":"3.549231ms","request_id":"","journal":{"occurred_at":"2025-04-28T14:17:43.625101824Z","operation":"messaging.client_publish"},"error":{"error":"","message":"entity not found"}}
{"time":"2025-04-29T14:17:23.058894671Z","level":"WARN","msg":"Failed to handle SuperMQ message: failed to handle nats event: entity not found"}
```

Journal service could not able to process the events , 
One of the example events:
```
{"occurred_at":1745938409291330344,"operation":"group.role.list_available_actions"}
```

dborovcanin:
This is an urgent one. Please take it as the highest priority.

PR #2858, opened 2025-05-08
SMQ-2842 - Update journal service to handle roles events

felixgateru:
# What type of PR is this?
This is a bug fix as it addresses the following issue:#2842

## What does this do?
This pr:
- Updates the journal service to be able to handle roles related events
- Fixes a bug in updating client telemetry for mqtt subscriptions
- Removes listing of messaging events as clients and channels logs

- Related Issue #2842 

No, manually tested

No

### DIFF
--- a/journal/events/consumer.go
+++ b/journal/events/consumer.go
@@ -11,7 +11,6 @@ import (
 	"github.com/absmach/supermq/journal"
 	"github.com/absmach/supermq/pkg/events"
 	"github.com/absmach/supermq/pkg/events/store"
-	"github.com/absmach/supermq/pkg/messaging"
 )
 
 var ErrMissingOccurredAt = errors.New("missing occurred_at")
@@ -19,11 +18,9 @@ var ErrMissingOccurredAt = errors.New("missing occurred_at")
 // Start method starts consuming messages received from Event store.
 func Start(ctx context.Context, consumer string, sub events.Subscriber, service journal.Service) error {
 	subCfg := events.SubscriberConfig{
-		Consumer:       consumer,
-		Stream:         store.StreamAllEvents,
-		Handler:        Handle(service),
-		Ordered:        true,
-		DeliveryPolicy: messaging.DeliverAllPolicy,
+		Consumer: consumer,
+		Stream:   store.StreamAllEvents,
+		Handler:  Handle(service),
 	}
 
 	return sub.Subscribe(ctx, subCfg)
--- a/journal/events/consumer.go
+++ b/journal/events/consumer.go
@@ -11,6 +11,7 @@ import (
 	"github.com/absmach/supermq/journal"
 	"github.com/absmach/supermq/pkg/events"
 	"github.com/absmach/supermq/pkg/events/store"
+	"github.com/absmach/supermq/pkg/messaging"
 )
 
 var ErrMissingOccurredAt = errors.New("missing occurred_at")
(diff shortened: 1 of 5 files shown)


dborovcanin (dismissed):

dborovcanin:

journal/journal.go
```diff
@@ -67,11 +67,11 @@ func (e EntityType) Query() string {
 	case UserEntity:
 		return "((operation LIKE 'user.%' AND attributes->>'id' = :entity_id) OR (attributes->>'user_id' = :entity_id))"
 	case GroupEntity:
-		return "((operation LIKE 'group.%' AND attributes->>'id' = :entity_id) OR (attributes->>'group_id' = :entity_id))"
+		return "((operation LIKE 'group.%' AND (attributes->>'id' = :entity_id OR attributes->>'entity_id' = :entity_id)) OR (attributes->>'group_id' = :entity_id))"
```
arvindh123: Do we need `group_id` ?

arvindh123 (changes requested):

journal/events/consumer.go
```diff
@@ -11,16 +11,19 @@ import (
...
 
 var ErrMissingOccurredAt = errors.New("missing occurred_at")
 
 // Start method starts consuming messages received from Event store.
 func Start(ctx context.Context, consumer string, sub events.Subscriber, service journal.Service) error {
 	subCfg := events.SubscriberConfig{
-		Consumer: consumer,
-		Stream:   store.StreamAllEvents,
-		Handler:  Handle(service),
+		Consumer:       consumer,
```
arvindh123: @felixgateru  
For Example, If the service replication increase, the newly started service will consume all the messages from first ? 
Could you please confirm this case with NATS Jetstream consumers

arvindh123 (changes requested):

journal/events/consumer.go
felixgateru: According to https://docs.nats.io/nats-concepts/jetstream/streams the behaviour is based on the use of a durable name.
If the consumer durable name is the same among replicated consumers then they will share messages and each will not have the same messages but if the consumer durable name is unique the deliver all policy will apply for each consumer and they will each have the exact same messages which will include those from the beginning of the stream.

dborovcanin: Our events subscriber is using [LimitsPolicy](https://github.com/absmach/supermq/blob/17a884f7e99f53fa7ce3bf7705ab1279d8b7c445/pkg/events/nats/subscriber.go#L29), shouldn't we use `WorkQueuePolicy` instead?

felixgateru: I don't think the policy would be suitable for us as we have multiple overlapping consumers i.e
image

arvindh123: @dborovcanin  This LimitsPolicy for creating a stream, For Consumer we use this configuration https://github.com/absmach/supermq/blob/757ad31f921d7a05df8f1f4255cb41a52ce5f233/pkg/events/nats/subscriber.go#L67-L76 and this will call the NATS subscriber at here https://github.com/absmach/supermq/blob/602025b48bc008d8ec89be55e8ed228f1cebc4cb/pkg/messaging/nats/pubsub.go#L75-L114

arvindh123: What happens if the consumer is deleted ? 
Does the messages will be start from the first ?

arvindh123: If the `Not Acked messages` increase in consumer, Then we will delete consumer and create new consumer. 
This workaround is for to avoid `Not Acked Messages` again getting back to queue.  

So if we have All Messages policy , I think again all the messages from stream will get into consumer. Again the error message will get into queue.  


Another problem , If we delete the consumer and again the all the messages start flow into journal, 

Journal service might throw error, because these messages are already added in journal database. 

So we should handle this case in Journal service too.

arvindh123: We can add this case only if journal service can able to handle the duplicate messages. 
@felixgateru  At present if journal could not able to handle duplicate message, then please revert these changes and we will have this in new PR along with journal service feature to handle duplicate message.

arvindh123 (approved):

dborovcanin (approved):

PR #2864, opened 2025-05-13
SMQ-2842 - Fix client telemetry handling in journal service

### LINKED ISSUE #2842: Bug:  Journal service could not consume new events

felixgateru:
# What type of PR is this?
This is a bug fix because it fixes the following issue: #2842 

## What does this do?
This pr:
- Improves handling of client telemetry in repository
- Fixes bugs in client removal in clients telemetry
- Fixes issue with publishing of `messaging.client_publish` events

- Related Issue #2842
- Resolves #2842

No, tested manually

No

### Notes
none

### DIFF
--- a/cmd/mqtt/main.go
+++ b/cmd/mqtt/main.go
@@ -148,13 +148,6 @@ func main() {
 	}
 	defer mpub.Close()
 
-	mpub, err = msgevents.NewPublisherMiddleware(ctx, mpub, cfg.ESURL)
-	if err != nil {
-		logger.Error(fmt.Sprintf("failed to create event store middleware: %s", err))
-		exitCode = 1
-		return
-	}
-
 	fwd := mqtt.NewForwarder(brokers.SubjectAllChannels, logger)
 	fwd = mqtttracing.New(serverConfig, tracer, fwd, brokers.SubjectAllChannels)
 	if err := fwd.Forward(ctx, svcName, bsub, mpub); err != nil {
--- a/docker/addons/journal/docker-compose.yaml
+++ b/docker/addons/journal/docker-compose.yaml
@@ -16,6 +16,7 @@ networks:
 volumes:
   supermq-journal-volume:
 
+
 services:
   journal-db:
     image: postgres:16.2-alpine
--- a/docker/addons/journal/docker-compose.yaml
+++ b/docker/addons/journal/docker-compose.yaml
@@ -31,8 +31,6 @@ services:
       - supermq-base-net
     volumes:
       - supermq-journal-volume:/var/lib/postgresql/data
-    ports:
-      - 6012:5432
 
   journal:
     image: supermq/journal:${SMQ_RELEASE_TAG}
--- a/docker/addons/journal/docker-compose.yaml
+++ b/docker/addons/journal/docker-compose.yaml
@@ -16,7 +16,6 @@ networks:
 volumes:
   supermq-journal-volume:
 
-
 services:
   journal-db:
     image: postgres:16.2-alpine
@@ -32,6 +31,8 @@ services:
       - supermq-base-net
     volumes:
       - supermq-journal-volume:/var/lib/postgresql/data
+    ports:
+      - 6012:5432
 
   journal:
     image: supermq/journal:${SMQ_RELEASE_TAG}
--- a/journal/events/consumer.go
+++ b/journal/events/consumer.go
@@ -41,6 +41,7 @@ func Handle(service journal.Service) handleFunc {
 
 		operation, ok := data["operation"].(string)
 		if !ok {
+			// Error is logged instead of being returned to avoid redelivering of the event.
 			slog.Error(errMsg, "error", errMissingOperation)
 			return nil
 		}
--- a/journal/events/consumer.go
+++ b/journal/events/consumer.go
@@ -14,7 +14,12 @@ import (
 	"github.com/absmach/supermq/pkg/events/store"
 )
 
-var ErrMissingOccurredAt = errors.New("missing occurred_at")
+var (
+	ErrMissingOccurredAt = errors.New("missing occurred_at")
+	errMissingOperation  = errors.New("missing operation")
+	errMissingAttributes = errors.New("missing attributes")
+	errMsg               = "failed to save journal"
+)
 
 // Start method starts consuming messages received from Event store.
 func Start(ctx context.Context, consumer string, sub events.Subscriber, service journal.Service) error {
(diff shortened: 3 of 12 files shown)


journal/postgres/telemetry.go
```diff
@@ -38,14 +38,11 @@ func (repo *repository) DeleteClientTelemetry(ctx context.Context, clientID, dom
...
-
-	result, err := repo.db.NamedExecContext(ctx, q, dbct)
+	_, err := repo.db.NamedExecContext(ctx, q, dbct)
 	if err != nil {
 		return postgres.HandleError(repoerr.ErrRemoveEntity, err)
 	}
-	if rows, _ := result.RowsAffected(); rows == 0 {
-		return repoerr.ErrNotFound
-	}
+
```
arvindh123, line 49: Could we handle error `repoerr.ErrNotFound` in service layer? 
May be this information might be use full for logging in service layer.

journal/postgres/telemetry.go
```diff
@@ -144,15 +141,11 @@ func (repo *repository) IncrementInboundMessages(ctx context.Context, clientID s
 		return errors.Wrap(repoerr.ErrUpdateEntity, err)
 	}
 
-	result, err := repo.db.NamedExecContext(ctx, q, dbct)
+	_, err = repo.db.NamedExecContext(ctx, q, dbct)
 	if err != nil {
 		return postgres.HandleError(repoerr.ErrUpdateEntity, err)
 	}
```
arvindh123: If inbound message is not found, Instead of skipping them why don't we create a new entry for these messages?

journal/service.go
```diff
@@ -108,7 +108,7 @@ func (svc *service) handleTelemetry(ctx context.Context, journal Journal) error
 }
 
 func (svc *service) addClientTelemetry(ctx context.Context, journal Journal) error {
-	ce, err := toClientEvent(journal)
+	ce, err := toClientEvent(journal, true)
```
arvindh123, line 116: What do you think about handing postgres error `entity already exists` here in `addClientTelementry` at service layer ?

cmd/mqtt/main.go
```diff
@@ -148,13 +148,6 @@ func main() {
 	}
 	defer mpub.Close()
 
-	mpub, err = msgevents.NewPublisherMiddleware(ctx, mpub, cfg.ESURL)
```
arvindh123, line 151: I agree with removal of this events middleware, Forward (consumer) no need to publish their events to NATS.

arvindh123 (changes requested):

journal/service.go
```diff
@@ -228,11 +227,13 @@ func toClientEvent(journal Journal) (clientEvent, error) {
...
-		if err != nil {
-			return clientEvent{}, fmt.Errorf("invalid created_at format")
+	if isCreate {
+		createdAtStr := journal.Attributes["created_at"].(string)
+		if createdAtStr != "" {
+			createdAt, err = time.Parse(time.RFC3339, createdAtStr)
+			if err != nil {
+				return clientEvent{}, fmt.Errorf("invalid created_at format")
+			}
 		}
```
arvindh123, line 249: @felixgateru @dborovcanin 

What do you think about this approach
If we return error then it will again queue back. So we can log them or store in separate place 
```suggestion
	if isCreate {
		createdAtIf, ok := journal.Attributes["created_at"]
		if ok {
			createdAtStr, ok := createdAtIf.(string)
			if ok && createdAtStr != "" {
				createdAt, err = time.Parse(time.RFC3339, createdAtStr)
				if err != nil {
					log.Println("log error : invalid created_at format")
				}
			}
		}
```

journal/postgres/telemetry.go
felixgateru: Initially I was of the idea, I could add it as a feature after we had added domainID to the internal message topic so that client telemetry can also be created on client_subscribe if it did not exist.
I can add for client_publish now and for the client_subscribe later.

journal/service.go
felixgateru: Do we just log the error?

arvindh123: If we are returning error, then the message will again queued back , this will go into a loop

journal/service.go
dborovcanin: I don't mind having a trash to store unsuccessfully processed events.

arvindh123: Yeah I agree, Can we do it in next PR ?

dborovcanin (changes requested):
Instead of logging errors directly in the service and returning nil, we should wrap the error with contextual information and propagate it. This allows us to decouple logging concerns from business logic and delegate error handling to the caller.

arvindh123 (dismissed):

docker/addons/journal/docker-compose.yaml
```diff
@@ -16,7 +16,6 @@ networks:
 volumes:
   supermq-journal-volume:
 
-
```
dborovcanin: Keep this one. One empty line is for the value (empty), and the other isthe  separator for better readability.

journal/events/consumer.go
```diff
@@ -35,22 +41,26 @@ func Handle(service journal.Service) handleFunc {
 
 		operation, ok := data["operation"].(string)
 		if !ok {
-			return errors.New("missing operation")
+			slog.Error(errMsg, "error", errMissingOperation)
```
dborovcanin, line 45: Add a comment with an explanation for logging and replacing errors with nil.

journal/service.go
dborovcanin: Yes, this should be done in a separate PR. @felixgateru Please open a ticket for this.

dborovcanin (changes requested):

journal/service.go
felixgateru: Done :https://github.com/absmach/supermq/issues/2880

arvindh123 (approved):

dborovcanin (approved):

PR #2870, opened 2025-05-14
SMQ-2869 - Switch to UTC time on backends

### LINKED ISSUE #2869: Switch to UTC time on backends
Always opt for UTC in all the `time.Now()` and similar cases. Also, let's convert the time received to endpoints to UTC as soon as we can.

dborovcanin:
# What type of PR is this?

## What does this do?
This sets all the backend times to UTC.

- Resolves #2869

All tests pass, no new tests needed.
No.

### Notes
N/A

### DIFF
--- a/domains/service.go
+++ b/domains/service.go
@@ -119,7 +119,7 @@ func (svc service) RetrieveDomain(ctx context.Context, session authn.Session, id
 }
 
 func (svc service) UpdateDomain(ctx context.Context, session authn.Session, id string, d DomainReq) (Domain, error) {
-	updatedAt := time.Now()
+	updatedAt := time.Now().UTC()
 	d.UpdatedAt = &updatedAt
 	d.UpdatedBy = &session.UserID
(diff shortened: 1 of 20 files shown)


dborovcanin:
Please note that `time.Now().UnixNano()` and similar methods are not affected by the zone, so we are not wasting time converting them to the `UTC`. We're mostly concerned that everything that ends up in the DB is UTC for simplicity and portability.

domains/service.go
```diff
@@ -61,7 +61,7 @@ func (svc service) CreateDomain(ctx context.Context, session authn.Session, d Do
 		return Domain{}, []roles.RoleProvision{}, svcerr.ErrInvalidStatus
 	}
 
-	d.CreatedAt = time.Now()
```
felixgateru, line 64: Instances in update, enable, disable and freeze not updated

felixgateru (changes requested):

felixgateru (approved):

PR #2871, opened 2025-05-14
NOISSUE - Add ability to handle both tls cert file and raw cert data

smithjilks:
# What type of PR is this?
This is an enhancement because it adds ability to load tls file from both file and raw data input

Related ultravioletrs/prism#816

- Related Issue #
- Resolves #

### DIFF
--- a/pkg/server/http/http.go
+++ b/pkg/server/http/http.go
@@ -45,9 +45,13 @@ func (s *httpServer) Start() error {
 			return err
 		}
 
-		s.server.TLSConfig = &tls.Config{
-			Certificates: []tls.Certificate{certs},
+		if s.server.TLSConfig == nil {
+			s.server.TLSConfig = &tls.Config{}
 		}
+
+		tlsConf := s.server.TLSConfig.Clone()
+		tlsConf.Certificates = append(tlsConf.Certificates, certs)
+		s.server.TLSConfig = tlsConf
 		s.Protocol = httpsProtocol
 
 		s.Logger.Info(fmt.Sprintf("%s service %s server listening at %s with TLS", s.Name, s.Protocol, s.Address))
--- a/pkg/server/http/http.go
+++ b/pkg/server/http/http.go
@@ -5,6 +5,7 @@ package http
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"log/slog"
 	"net/http"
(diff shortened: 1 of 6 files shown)


SammyOina (dismissed):

SammyOina (dismissed):

pkg/server/http/http.go
```diff
@@ -39,10 +42,34 @@ func (s *httpServer) Start() error {
...
+		}
+
+		keyFile, err := readFileOrData(s.Config.KeyFile)
+		if err != nil {
+			s.cleanupTempFiles()
+			return fmt.Errorf("failed to process key file: %w", err)
+		}
+		if keyFile != s.Config.KeyFile {
+			s.tempFiles = append(s.tempFiles, keyFile)
+		}
```
dborovcanin: There is no need for all these complications of temp files. Package `tls` has two functions `tls.X509KeyPair()` and `tls.LoadX509KeyPair()`, one loads from bytes, the other from file (using the first one under the hood).

dborovcanin (changes requested):

pkg/server/http/http.go
```diff
@@ -39,10 +42,34 @@ func (s *httpServer) Start() error {
...
+		if err != nil {
+			s.cleanupTempFiles()
+			return fmt.Errorf("failed to load TLS certificate: %w", err)
+		}
+
 		s.Protocol = httpsProtocol
 		s.Logger.Info(fmt.Sprintf("%s service %s server listening at %s with TLS cert %s and key %s", s.Name, s.Protocol, s.Address, s.Config.CertFile, s.Config.KeyFile))
 		go func() {
-			errCh <- s.server.ListenAndServeTLS(s.Config.CertFile, s.Config.KeyFile)
+			errCh <- s.server.ListenAndServeTLS(certFile, keyFile)
```
smithjilks: @dborovcanin 
We don't have control over this method, which expects file paths. That's why I went with an approach for creating temp files.

dborovcanin: You can try with something like this:
```go
case s.Config.CertFile != "" || s.Config.KeyFile != "":
		certs, err := server.LoadX509KeyPair(s.Config.CertFile, s.Config.KeyFile)
		if err != nil {
			return err
		}

		s.Protocol = httpsProtocol
		tlsConf := s.server.TLSConfig.Clone()
		tlsConf.Certificates = append(tlsConf.Certificates, certs)
		s.server.TLSConfig = tlsConf
		s.Logger.Info(fmt.Sprintf("%s service %s server listening at %s with TLS cert %s and key %s", s.Name, s.Protocol, s.Address, s.Config.CertFile, s.Config.KeyFile))
		go func() {
			errCh <- s.server.ListenAndServeTLS("", "")
		}()
```

dborovcanin (changes requested):

pkg/server/http/http.go
```diff
@@ -44,11 +45,12 @@ func (s *httpServer) Start() error {
 			return err
 		}
 
+		s.server.TLSConfig = &tls.Config{
+			Certificates: []tls.Certificate{certs},
+		}
```
dborovcanin: Why not clone?

dborovcanin (approved):

PR #2877, opened 2025-05-16
NOISSUE - Add Nullable type for optional values handling

dborovcanin:
# What type of PR is this?
This is a refactor, but also an optimization.

## What does this do?
This PR introduces a new nullable package that provides a generic `nullable.Value[T]` type, allowing explicit representation of optional values across the codebase. Unlike Go's *T, `nullable.Value[T]` makes it possible to distinguish between:

- Unset values (e.g., field omitted)
- Zero values (e.g., 0, "", false)
- Explicit null values (usually not used)

This pattern improves data integrity, parsing robustness, and overall code clarity, especially when working with JSON, form input, and APIs.

Why?
- Clearer optional semantics than pointers
- Easier handling of queries with missing fields
- Safe defaulting without nil dereference risks
- Performance and GC-friendly for primitives
 

There are not such issues.

No.
However, included micro-benchmarks show that `nullable.Value[T]` performs competitively with pointer-based approaches, with the added benefit of avoiding heap allocations for primitive types.

N/A

### Notes
N/A


dborovcanin:
You can run benchmarks from the correct package like `go test -bench=String .`.

internal/nullable/doc.go
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Abstract Machines
+// SPDX-License-Identifier: Apache-2.0
+
+// Package nullable contains nullabele types used to handle
```
felixgateru:
```suggestion
// Package nullable contains nullable types used to handle
```

internal/nullable/doc.go
```diff
@@ -0,0 +1,7 @@
...
+// scenarios where default values can't be used to inidicate empty,
```
felixgateru:
```suggestion
// scenarios where default values can't be used to indicate empty,
```

internal/nullable/value.go
```diff
@@ -0,0 +1,48 @@
...
+package nullable
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+)
+
+// Value type is used to represent difference betweeen an
+// intentionally omitted value and default type falue.
```
felixgateru:
```suggestion
// intentionally omitted value and default type value.
```

felixgateru (changes requested):

felixgateru (dismissed):

Issue #2885, opened 2025-05-20
Bug: Environment variable not applied & Role missing allow_any_name=true in Vault PKI setup


nintran52:
### What were you trying to achieve?

I was trying to use `vault_set_pki.sh` to automatically configure Vault PKI and generate a server certificate (for example: `supermq.com`) by setting `SMQ_NGINX_SERVER_NAME=supermq.com` in the `.env` file.

### What are the expected results?

- The script should load `.env` early and recognise `SMQ_NGINX_SERVER_NAME` before using it.

- The generated certificate should use the custom common name (`supermq.com`).

- The Vault role should allow issuing a certificate for the requested CN without error.

### What are the received results?

1. The script defaults to `localhost` as CN because `readDotEnv` is called after the check for `SMQ_NGINX_SERVER_NAME`, causing environment variables from `.env` to be ignored.

2. After fixing the first issue, Vault throws:
`common name supermq.com not allowed by this role`
due to missing `allow_any_name=true` in the Vault role config.

### Steps To Reproduce

1. Set `SMQ_NGINX_SERVER_NAME=supermq.com` in `docker/.env`.

2. Run `vault_set_pki.sh` without modifying it.

3. Observe that the CN in the generated certificate is `localhost`, not `supermq.com`.

4. Move `readDotEnv` to the top, rerun.

5. Vault API returns 400 error: `common name supermq.com not allowed by this role`.

6. Fix it by adding `allow_any_name=true` in `vaultSetupServerCertsRole`.

### In what environment did you encounter the issue?

OS: macOS 15.4.1 (Sonoma)

Docker Desktop version: 4.41.2

Docker Engine version: 28.1.1

Docker Compose version: 2.36.0

### Additional information you deem important

![Image](https://github.com/user-attachments/assets/9c3ae8fa-7d4c-4282-bfc1-195ae25f3bc3)
![Image](https://github.com/user-attachments/assets/bc53f12d-7b3a-41a5-857e-5bb3de3bcab8)
 
- Solution:
  - Move readDotEnv before usage of env variables.
  - Add allow_any_name=true to Vault server cert role:
`vault write ... allow_any_name=true \`

![Image](https://github.com/user-attachments/assets/f129ece8-c86d-4c06-a3bc-4e45b50aa6be)

dborovcanin:
This has been resolved by the linked PR.

PR #2886, opened 2025-05-20
SMQ-2885 - fix(pki): load env file early and allow any CN in server cert role

nintran52:
# What type of PR is this?
This is a **bug fix** because it addresses two issues in the `vault_set_pki.sh` script:
1. Environment variable not loaded in time
2. Vault server role not allowing custom CNs

## What does this do?
This PR updates the `vault_set_pki.sh` script with the following changes:
- Calls `readDotEnv` **before** accessing `SMQ_NGINX_SERVER_NAME`, ensuring that values from the `.env` file are correctly loaded.
- Adds the `allow_any_name=true` flag to the Vault server certificate role to support CNs like `supermq.com`.

- Related Issue #2885 
- Resolves #

No. This is a shell script change and Vault config script, so changes were tested manually via local Docker environment with Vault.

No documentation update is needed since this is an internal bug fix to an existing setup script.


nintran52:
Hi @dborovcanin, please help me review this PR. I discovered the issue a few weeks ago but only just found time to fix it.

docker/addons/vault/scripts/vault_set_pki.sh
```diff
@@ -188,6 +190,7 @@ vaultSetupServerCertsRole() {
         echo "Setup Server certificate role"
         vault write -namespace=${SMQ_VAULT_NAMESPACE} -address=${SMQ_VAULT_ADDR} ${SMQ_VAULT_PKI_INT_PATH}/roles/${SMQ_VAULT_PKI_INT_SERVER_CERTS_ROLE_NAME} \
             allow_subdomains=true \
+			allow_any_name=true \
```
dborovcanin: Can you, please, check the alignment in this file and sign off your commits? Also, take care that Vault support will be limited since we'll be switching to our own simple solution (https://github.com/absmach/certs/) or fork of Vault (https://openbao.org/).

dborovcanin (changes requested):

docker/addons/vault/scripts/vault_set_pki.sh
nintran52: do we have a repo for UI and Devops, I saw it before, but don't see it anymore

docker/.env
```diff
@@ -11,6 +11,7 @@ SMQ_NGINX_HTTP_PORT=80
 SMQ_NGINX_SSL_PORT=443
 SMQ_NGINX_MQTT_PORT=1883
 SMQ_NGINX_MQTTS_PORT=8883
+SMQ_NGINX_SERVER_NAME=
```
arvindh123, line 14: If an environment variable is empty by default, it can be removed. It can then be added at any point , when it is required.

arvindh123 (changes requested):

docker/.env
nintran52: Thanks for your input!

While it's true that empty variables can be omitted, we’ve followed the convention of including all configurable environment variables upfront — even if they’re empty — to improve visibility and make it easier for users to override them without diving into the codebase.

In this case, I’d prefer to keep SMQ_NGINX_SERVER_NAME= consistent with the rest of the .env structure (e.g., SMQ_AUTH_HTTP_SERVER_KEY=, SMQ_POSTGRES_SSL_CERT=, etc.).

Let me know if you'd still prefer removing it.

arvindh123 (commented):
LGTM

docker/addons/vault/scripts/vault_set_pki.sh
nintran52: @dborovcanin may I know that and help me to merge this PR

nintran52:
@dborovcanin could you recheck the sign off

PR #2889, opened 2025-05-22
SMQ-2888 - Change internal message topic format

### LINKED ISSUE #2888: Feature: Update the internal message format to `m.<domain_id>.c.<channel_id>.<sub_topic>`

Yes, There is no way to NATS subscribe message for particular domain 

Updating the internal message format to `m.<domain_id>.c.<channel_id>.<sub_topic>` will help support domain-level message subscriptions.

This format should also be applied to **Alarms** and **Writers** messages:

* **Alarms**: `alarms.<domain_id>.c.<channel_id>.<sub_topic>`
* **Writers**: `writers.<domain_id>.c.<channel_id>.<sub_topic>`

Must-have

### Anything else?

_No response_

arvindh123:
# What type of PR is this?

- Related Issue #2888 
- Resolves #2888


dborovcanin (changes requested):
Please fix conflicts.

dborovcanin:
@arvindh123 Please fix the CI.

felixgateru (dismissed):
LGTM

pkg/messaging/msgtopic.go
```diff
@@ -0,0 +1,151 @@
+// Copyright (c) Abstract Machines
+// SPDX-License-Identifier: Apache-2.0
+
+package messaging
```
dborovcanin: Rename the file to `topics.go`

pkg/messaging/msgtopic.go
```diff
@@ -0,0 +1,151 @@
...
+
+func EncodeToInternalSubjectSuffix(domainID string, channelID string, subtopic string) string {
+	subject := fmt.Sprintf("%s.%s.%s", domainID, ChannelTopicPrefix, channelID)
+	if subtopic != "" {
+		subject = fmt.Sprintf("%s.%s", subject, subtopic)
+	}
+	return subject
+}
+
+func (m *Message) EncodeToInternalSubject() string {
```
dborovcanin: Let's not attach any new methods to the generated code - use `EncodeToInternalTopic` directly.

dborovcanin: Let's not use `internal` notion - `EncodeTopic`.

pkg/messaging/msgtopic.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+var (
+	ErrMalformedTopic    = errors.New("malformed topic")
+	ErrMalformedSubtopic = errors.New("malformed subtopic")
+	// Regex to group topic in format m.<domain_id>.c.<channel_id>.<sub_topic> `^\/?m\/([\w\-]+)\/c\/([\w\-]+)(\/[^?]*)?(\?.*)?$`.
+	msgTopicRegExp = regexp.MustCompile(`^\/?` + MsgTopicPrefix + `\/([\w\-]+)\/` + ChannelTopicPrefix + `\/([\w\-]+)(\/[^?]*)?(\?.*)?$`)
+)
+
+func ParsePublishTopic(topic string) (string, string, string, error) {
```
dborovcanin: Let's use named return values for clarity `func ParsePublishTopic(topic string) (domainID, chanID, subtopic string, err error)`.

pkg/messaging/msgtopic.go
```diff
@@ -0,0 +1,151 @@
...
+	subtopic = strings.ReplaceAll(subtopic, "/", ".")
+
+	elems := strings.Split(subtopic, ".")
+	filteredElems := []string{}
+	for _, elem := range elems {
+		if elem == "" {
+			continue
+		}
+
+		if strings.Contains(elem, " ") || strings.Contains(elem, "#") || strings.Contains(elem, "+") || strings.Contains(elem, "*") || strings.Contains(elem, ">") {
```
dborovcanin: No need to do this in the loop; we can do these checks right before the `replaceAll`:
```go
	for _, r := range allowed {
		if strings.Contains(subtopic, r) {
			return "", ErrMalformedSubtopic
		}
	}
```
where `var allowed = []string{" ", "#", "+", "*", ">"}`.

dborovcanin (changes requested):

pkg/messaging/topics.go
```diff
@@ -87,21 +85,55 @@ func ParseSubscribeTopic(topic string) (domainID string, chanID string, subtopic
...
 		}
+		filteredElems = append(filteredElems, elem)
 	}
 
+	subtopic = strings.Join(filteredElems, ".")
+
+	return subtopic, nil
+}
+
+func formatSubTopic(subtopic string) (string, error) {
```
dborovcanin: Let's use `formatSubtopic` i.e. a single word `subtopic` rather than `subTopic`.

pkg/messaging/topics.go
```diff
@@ -26,9 +26,9 @@ var (
 	ErrMalformedTopic    = errors.New("malformed topic")
 	ErrMalformedSubtopic = errors.New("malformed subtopic")
 	// Regex to group topic in format m.<domain_id>.c.<channel_id>.<sub_topic> `^\/?m\/([\w\-]+)\/c\/([\w\-]+)(\/[^?]*)?(\?.*)?$`.
-	msgTopicRegExp             = regexp.MustCompile(`^\/?` + MsgTopicPrefix + `\/([\w\-]+)\/` + ChannelTopicPrefix + `\/([\w\-]+)(\/[^?]*)?(\?.*)?$`)
-	pubSubTopicNotAllowedChars = []string{" ", "#", "+", "*", ">"}
-	subSubTopicNotAllowedChars = []string{" ", "#", "+"}
+	msgTopicRegExp              = regexp.MustCompile(`^\/?` + MsgTopicPrefix + `\/([\w\-]+)\/` + ChannelTopicPrefix + `\/([\w\-]+)(\/[^?]*)?(\?.*)?$`)
+	subTopicNotAllowedWildCards = []rune{'*', '>'}
```
dborovcanin: `subtopicInvalidChars`

pkg/messaging/topics.go
```diff
@@ -26,9 +26,9 @@ var (
 	ErrMalformedTopic    = errors.New("malformed topic")
 	ErrMalformedSubtopic = errors.New("malformed subtopic")
 	// Regex to group topic in format m.<domain_id>.c.<channel_id>.<sub_topic> `^\/?m\/([\w\-]+)\/c\/([\w\-]+)(\/[^?]*)?(\?.*)?$`.
-	msgTopicRegExp             = regexp.MustCompile(`^\/?` + MsgTopicPrefix + `\/([\w\-]+)\/` + ChannelTopicPrefix + `\/([\w\-]+)(\/[^?]*)?(\?.*)?$`)
-	pubSubTopicNotAllowedChars = []string{" ", "#", "+", "*", ">"}
-	subSubTopicNotAllowedChars = []string{" ", "#", "+"}
+	msgTopicRegExp              = regexp.MustCompile(`^\/?` + MsgTopicPrefix + `\/([\w\-]+)\/` + ChannelTopicPrefix + `\/([\w\-]+)(\/[^?]*)?(\?.*)?$`)
```
dborovcanin: `topicRegexp`

pkg/messaging/topics.go
dborovcanin: Can we use `subtopicInvalidWildcards` or even just `invalidWildcards`?

pkg/messaging/topics.go
```diff
@@ -87,21 +85,55 @@ func ParseSubscribeTopic(topic string) (domainID string, chanID string, subtopic
...
 
-	for _, ssnaChars := range subSubTopicNotAllowedChars {
-		if strings.Contains(subtopic, ssnaChars) {
-			return "", ErrMalformedSubtopic
+	subtopic, err = formatSubTopic(subtopic)
+	if err != nil {
+		return "", errors.Wrap(ErrMalformedSubtopic, err)
+	}
+
+	elems := strings.Split(subtopic, ".")
```
dborovcanin, line 101: Why don't we perform checks for invalidChars before the split? Also, consider using `strings.ContainsAny`.

dborovcanin (changes requested):

pkg/messaging/topics.go
arvindh123: They are not invalid wild cards ,
They are valid wild cards but they are not allowed for publish topic

dborovcanin: Then, let's use `invalidPublishWildcards`.

dborovcanin: I understand `notAllowed` or `forbidden` is more precise, but in this context - it's the same and `invalid` is a little more Go idiomatic.

pkg/messaging/topics.go
```diff
@@ -0,0 +1,156 @@
...
+	return EncodeTopicSuffix(m.GetDomain(), m.GetChannel(), m.GetSubtopic())
+}
+
+func (m *Message) EncodeToMQTTTopic() string {
+	topic := fmt.Sprintf("%s/%s/%s/%s", MsgTopicPrefix, m.GetDomain(), ChannelTopicPrefix, m.GetChannel())
+	if m.GetSubtopic() != "" {
+		topic = topic + "/" + strings.ReplaceAll(m.GetSubtopic(), ".", "/")
+	}
+	return topic
+}
```
dborovcanin: Can we not attach methods to the generated code? I am afraid it is using reflection under the hood, and this may mess up with Protobuf. Probably will not, but there is no need to attach it to the message anyway.

dborovcanin (changes requested):

pkg/messaging/topics.go
```diff
@@ -0,0 +1,156 @@
...
+	domainGroup   = 1 // domain group is first in msg topic regexp
+	channelGroup  = 2 // channel group is second in msg topic regexp
+	subtopicGroup = 3 // subtopic group is third in msg topic regexp
+)
+
+var (
+	ErrMalformedTopic    = errors.New("malformed topic")
+	ErrMalformedSubtopic = errors.New("malformed subtopic")
+	// Regex to group topic in format m.<domain_id>.c.<channel_id>.<sub_topic> `^\/?m\/([\w\-]+)\/c\/([\w\-]+)(\/[^?]*)?(\?.*)?$`.
+	topicRegExp          = regexp.MustCompile(`^\/?` + MsgTopicPrefix + `\/([\w\-]+)\/` + ChannelTopicPrefix + `\/([\w\-]+)(\/[^?]*)?(\?.*)?$`)
```
dborovcanin, line 29: Regex is slow, as you can see from the attached benchmarks. Since it is not replaceable in our case, let's not use it. We need to write our parser, and we need to be careful, so let's do it in a separate PR.

pkg/messaging/topics.go
arvindh123: The Goal is to "Generate topic from Message" so only added function `EncodeTopicSuffix` to `Message`.

pkg/messaging/topics.go
```diff
@@ -0,0 +1,156 @@
...
+	subject := fmt.Sprintf("%s.%s.%s", domainID, ChannelTopicPrefix, channelID)
+	if subtopic != "" {
+		subject = fmt.Sprintf("%s.%s", subject, subtopic)
+	}
+	return subject
+}
+
+func EncodeMessageTopicSuffix(m *Message) string {
+	return EncodeTopicSuffix(m.GetDomain(), m.GetChannel(), m.GetSubtopic())
+}
```
dborovcanin: Why not `EncodeTopic` (i.e. suffix is not needed, and it's `messaging.EncodeTopic(message)` so it's clean and concise)? Also, `encodeTopic` instead of `EncodeTopicSuffix`, since it's not used anywhere but in tests (which should also probably be replaced by the exported one).

arvindh123: `EncodeTopic` is used by subscribers , where `m.` will be included
`EncodeTopicSuffix` is used by publishers where `m.` is omitted

dborovcanin (approved):

PR #2921, opened 2025-06-04
SMQ-2884 - Create `domains` and `roles` entity operation

### LINKED ISSUE #2884: Create Individual entity operation instead of using svcutil.Operation
svcutil.Operation is used in all the service to identify the list of operations in the service 

This might not be fool proof, So we should have Operation per entity like clients.Operation, channels.Operation, groups.Operation and domains.Operation

nyagamunene:
# What type of PR is this?

- Related Issue #2884 
- Resolves #2884


dborovcanin:
@nyagamunene Please rebase.

arvindh123:
The entity operations and permission are separated 
Operations are defined in entity and permission are defined in config file 

This was implement in PR https://github.com/absmach/supermq/pull/3281
So this approach is no more needed
Closing this PR on behalf of https://github.com/absmach/supermq/pull/3281

PR #2928, opened 2025-06-06
NOISSUE - Add ACK type for message broker handler

dborovcanin:
# What type of PR is this?
This is a feature improvement.

## What does this do?
This adds ACK handling to the Subscriber.

There is no such issue.

No.

No.

### Notes
N/A


nyagamunene (approved):
LGTM

pkg/messaging/nats/pubsub.go
```diff
@@ -130,25 +132,45 @@ func (ps *pubsub) Unsubscribe(ctx context.Context, id, topic string) error {
...
 		}
 
-		if err := h.Handle(&msg); err != nil {
-			ps.logger.Warn(fmt.Sprintf("Failed to handle SuperMQ message: %s", err))
-			if ackErr {
-				if err := m.Ack(); err != nil {
-					ps.logger.Warn(fmt.Sprintf("Failed to ack message: %s", err))
-				}
-			}
+		err := h.Handle(&msg)
```
arvindh123, line 143: @dborovcanin  
Here If I kept AckErr as `Term` and If i get postgres error like memory full, then the message will not redeliveried 

Instead what do you think about the below ? 

image

or 

```go
Handle(msg *Message) (error, AckType)
```

dborovcanin: This starts to look more like a problem of the consumers than the interface itself. However, it also indicates our interface is not generic enough. I prefer this `Handle(msg *Message) (error, AckType)`. However, is this going to solve the problem, since the source is in `consumers/messages.go` ?

PR #2930, opened 2025-06-07
NOISSUE - Use ack types

dborovcanin:
# What type of PR is this?
This is an update for MQTT messaging QoS configuration.

There is no such issue. Depends on #2928.

Tests are fixed accordingly.

No.

### Notes
N/A


mqtt/forwarder.go
```diff
@@ -48,7 +49,7 @@ func handle(ctx context.Context, pub messaging.Publisher, logger *slog.Logger) h
 		}
 
 		topic := messaging.EncodeMessageMQTTTopic(msg)
-
+		ctx = mqtt.WithQoS(ctx, 1)
```
arvindh123, line 52: @dborovcanin 
The Forwarder subscribes to all topics from the message broker and republishes them to the MQTT broker. However, it doesn't check whether there is an active MQTT client subscribed to those topics. This can lead to unnecessary load on the MQTT broker.
To avoid this, we should have a separate forwarder instance for each MQTT subscriber.

dborovcanin:
This will not be addressed for the time being; we already have basic ack and nack handling.

PR #2942, opened 2025-06-12
SMQ-2924 - Allow bulk deleting of invitations from repository

### LINKED ISSUE #2924: Add `DeleteUsersInvitation` function to invitations repo to bulk delete invitations
We can have repo call where we can delete All the Invitation at once something like 

```go 
svc.repo.DeleteUsersInvitation(ctx,entityID, memberIDs...)
```

This requires new function invitations repo like ```DeleteUsersInvitation(ctx,entityID, memberIDs...)```

_Originally posted by @arvindh123 in https://github.com/absmach/supermq/pull/2835#discussion_r2128893065_

felixgateru:
# What type of PR is this?
This is a feature because it allows deleting multiple invitations from the repo

## What does this do?
This pr:
- creates the `DeleteUserInvitation` method to delete multiple invitations from the repo
- adds tests for the method

- Related Issue #2924
- Resolves #2924

Yes, tests have been included for the changes

Yes, in code documentation has been included

### DIFF
--- a/domains/domains.go
+++ b/domains/domains.go
@@ -263,8 +263,8 @@ type Repository interface {
 	// UpdateRejection updates an invitation by setting the rejection time.
 	UpdateRejection(ctx context.Context, invitation Invitation) (err error)
 
-	// DeleteUserInvitation deletes users invitation associated with a domain.
-	DeleteUserInvitation(ctx context.Context, domainID string, userID ...string) (err error)
+	// DeleteUsersInvitation deletes users invitation associated with a domain.
+	DeleteUsersInvitation(ctx context.Context, domainID string, userID ...string) (err error)
 
 	roles.Repository
 }
--- a/domains/domains.go
+++ b/domains/domains.go
@@ -263,8 +263,8 @@ type Repository interface {
 	// UpdateRejection updates an invitation by setting the rejection time.
 	UpdateRejection(ctx context.Context, invitation Invitation) (err error)
 
-	// DeleteUsersInvitation deletes users invitation associated with a domain.
-	DeleteUsersInvitation(ctx context.Context, domainID string, userID ...string) (err error)
+	// DeleteUsersInvitations deletes invitation to a provided domain for users with provided user IDs.
+	DeleteUsersInvitations(ctx context.Context, domainID string, userID ...string) (err error)
 
 	roles.Repository
 }
--- a/domains/domains.go
+++ b/domains/domains.go
@@ -263,8 +263,8 @@ type Repository interface {
 	// UpdateRejection updates an invitation by setting the rejection time.
 	UpdateRejection(ctx context.Context, invitation Invitation) (err error)
 
-	// Delete deletes an invitation.
-	DeleteInvitation(ctx context.Context, userID, domainID string) (err error)
+	// DeleteUserInvitation deletes users invitation associated with a domain.
+	DeleteUserInvitation(ctx context.Context, domainID string, userID ...string) (err error)
 
 	roles.Repository
 }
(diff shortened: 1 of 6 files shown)


domains/domains.go
```diff
@@ -267,8 +267,8 @@ type Repository interface {
 	// UpdateRejection updates an invitation by setting the rejection time.
 	UpdateRejection(ctx context.Context, invitation Invitation) (err error)
 
-	// Delete deletes an invitation.
-	DeleteInvitation(ctx context.Context, userID, domainID string) (err error)
+	// DeleteUserInvitation deletes users invitation associated with a domain.
+	DeleteUserInvitation(ctx context.Context, domainID string, userID ...string) (err error)
```
arvindh123:
```suggestion
	DeleteUsersInvitation(ctx context.Context, domainID string, userID ...string) (err error)
```

arvindh123 (dismissed):

dborovcanin:
@arvindh123 Please re-review.

dborovcanin:
@arvindh123 Please review.

domains/domains.go
```diff
@@ -263,8 +263,8 @@ type Repository interface {
 	// UpdateRejection updates an invitation by setting the rejection time.
 	UpdateRejection(ctx context.Context, invitation Invitation) (err error)
 
-	// Delete deletes an invitation.
-	DeleteInvitation(ctx context.Context, userID, domainID string) (err error)
+	// DeleteUsersInvitation deletes users invitation associated with a domain.
+	DeleteUsersInvitation(ctx context.Context, domainID string, userID ...string) (err error)
```
dborovcanin: Use plural, `DeleteUsersInvitations`. Also, update the comment to something like: `DeleteUsersInvitations deletes invitation to a provided domain for users with provided user IDs.`

dborovcanin (changes requested):
Please update SDK, and CLI as well.

dborovcanin (approved):

domains/service.go
```diff
@@ -334,11 +334,10 @@ func (svc *service) DeleteInvitation(ctx context.Context, session authn.Session,
...
-	for _, member := range members {
-		if err := svc.repo.DeleteInvitation(ctx, member, entityID); err != nil && err != repoerr.ErrNotFound {
-			return err
-		}
+	if err := svc.repo.DeleteUsersInvitations(ctx, entityID, members...); err != nil && err != repoerr.ErrNotFound {
+		return err
 	}
+
 	return svc.ProvisionManageService.RemoveEntityMembers(ctx, session, entityID, members)
 }
```
arvindh123, line 342: A improvement concern , I will create ticket for this 

Currently, this function performs two independent actions:

1. Removes user invitations from the repo (`DeleteUsersInvitations`).
2. Propagates the removal to the `ProvisionManageService`.

The issue is that if step 2 `ProvisionManageService` fails , the first step `DeleteUsersInvitations` has already been committed, and we cannot roll it back. This creates an inconsistent state between the repo and the external service.

To handle this properly, we likely need **transaction support across multiple operations**:

- **Option 1: Intermediate aggregation layer**  
  Implement an aggregation layer that coordinates multiple repos/services with a **2-phase commit (2PC)**-like pattern.  
  - Repo operation starts a transaction instead of committing immediately.  
  - If the second operation succeeds, we commit the repo transaction.  
  - If the second operation fails, we roll back the repo transaction.

- **Option 2: Return transaction control to the service layer**  
  Expose repo transactions to the service.  
  - Service starts a transaction in the repo.  
  - Executes the repo changes.  
  - Calls the external service.  
  - If the external call fails, the service cancels the repo transaction.  
  - If successful, the service commits the repo transaction.

This way, the system ensures consistency even if one part of the process fails.

arvindh123 (approved):

PR #2965, opened 2025-06-25
SMQ-2719 - Update list domain invitations endpoint

### LINKED ISSUE #2719: Bug: Invitations could not be filter by domain_id by invitee

As an invitee user, tried retrieve the received invitation by domain_id 

Response with applied filter `domain_id`

Authorization error

### Steps To Reproduce

1.) Create new domain 
2.) Send invitation to another user 
3.) Login as another user to which you have sent invitation
4.) List the invitations with `domain_id` filter
```bash
curl --location 'http://localhost/invitations?domain_id=<domain_id>' \
--header 'Authorization: Bearer <TOKEN>'
```

Docker

At present we have single domain service function `ListInvitation` , So we could not allow invitee to filter by domain_id, 
We need to have different domain service function and repo function `ListInviteeInvitations` ,  which should list only the invitee user received invitations. Then we can able to apply filter in repo.

felixgateru:
# What type of PR is this?
This is a refactor as it enables dedicated invitee invitation listing by adding `ListInviteeInvitations`

## What does this do?
This pr:
- Adds ability for users to list domain invitations specifically addressed to them using `ListDomainInvitations`
- Updates the authorization check for this

- Related Issue #2719
- Resolves #2719 

Yes, tests have been included

Yes, in code documentation is included


domains/middleware/authorization.go
```diff
@@ -253,6 +253,21 @@ func (am *authorizationMiddleware) ListInvitations(ctx context.Context, session
 	return am.svc.ListInvitations(ctx, session, page)
 }
 
+func (am *authorizationMiddleware) ListInviteeInvitations(ctx context.Context, session authn.Session, page domains.InvitationPageMeta) (invs domains.InvitationPage, err error) {
```
arvindh123: I think we can have two function , 
`ListInvitations` and `ListDomainInvitations`

- `ListInvitations` list invitee invitations 
- `ListDomainInvitations` list all the invitations sent on domain 

No need for `CheckAdmin` because SuperAdmin inherently get the permission

arvindh123 (changes requested):

arvindh123 (changes requested):
I could not see like this in `domainID: chi.URLParam(r, "domainID"),` in decoder requests of  SendDomainInvitations, ListDomainInvitation

felixgateru:
> I could not see like this in `domainID: chi.URLParam(r, "domainID"),` in decoder requests of SendDomainInvitations, ListDomainInvitation

They are obtained from the session since it already fetches them from the url

domains/api/http/endpoint.go
```diff
@@ -236,7 +236,7 @@ func listDomainInvitationsEndpoint(svc domains.Service) endpoint.Endpoint {
 		}
 		req.InvitationPageMeta.DomainID = session.DomainID
```
arvindh123: Please move this logic to service layer

arvindh123 (changes requested):

arvindh123 (changes requested):
https://github.com/absmach/supermq/blob/f3c11ef4eaeacdf1115ac9b82158d3f253240c00/domains/api/http/endpoint.go#L261
This might not be needed , Please verify this one

domains/events/events.go
```diff
@@ -378,6 +380,37 @@ func (lie listInvitationsEvent) Encode() (map[string]interface{}, error) {
...
+		"token_type":      lie.session.Type.String(),
+		"super_admin":     lie.session.SuperAdmin,
+	}
+
+	if lie.InvitedBy != "" {
+		val["invited_by"] = lie.InvitedBy
+	}
+	if lie.DomainID != "" {
+		val["domain_id"] = lie.DomainID
+	}
```
arvindh123: Domain ID is mandatory, It could not be optional 

in `listInvitationsEvent` super admin is not required , instead we can have  requested user id

arvindh123 (changes requested):

arvindh123 (dismissed):

domains/events/events.go
```diff
@@ -351,10 +353,43 @@ type listInvitationsEvent struct {
...
+	if lie.RoleID != "" {
+		val["role_id"] = lie.RoleID
+	}
+	if lie.State.String() != "" {
+		val["state"] = lie.State.String()
+	}
```
dborovcanin: This will never be true. Please check `State`.

dborovcanin (changes requested):

PR #2999, opened 2025-07-03
NOISSUE - Update nullable handling

dborovcanin:
# What type of PR is this?
This is a refactor.

## What does this do?
This changes the Nullable parser naming and adds a factory method to return a valid nullable with value.

There is no such issue.

Yes.

N/A

### Notes
N/A

### DIFF
--- a/internal/nullable/value.go
+++ b/internal/nullable/value.go
@@ -12,13 +12,13 @@ import (
 // Value type is used to represent difference betweeen an
 // intentionally omitted value and default type value.
 type Value[T any] struct {
-	Set   bool
+	Valid bool
 	Value T
 }
 
-func New[T nullable](v T, set bool) Value[T] {
+func New[T nullable](v T) Value[T] {
 	return Value[T]{
-		Set:   set,
+		Valid: true,
 		Value: v,
 	}
 }
@@ -30,7 +30,7 @@ type Parser[T nullable] func(string) (Value[T], error)
 
 // MarshalJSON encodes the value if set, otherwise returns `null`.
 func (n Value[T]) MarshalJSON() ([]byte, error) {
-	if !n.Set {
+	if !n.Valid {
 		return []byte("null"), nil
 	}
 	return json.Marshal(n.Value)
@@ -39,7 +39,7 @@ func (n Value[T]) MarshalJSON() ([]byte, error) {
 // UnmarshalJSON decodes JSON and sets the value and Set flag.
 func (n *Value[T]) UnmarshalJSON(data []byte) error {
 	if bytes.Equal(data, []byte("null")) {
-		n.Set = false
+		n.Valid = false
 		var empty T
 		n.Value = empty
 		return nil
@@ -50,6 +50,6 @@ func (n *Value[T]) UnmarshalJSON(data []byte) error {
 		return fmt.Errorf("nullable: failed to unmarshal: %w", err)
 	}
 	n.Value = val
-	n.Set = true
+	n.Valid = true
 	return nil
 }
--- a/internal/nullable/value.go
+++ b/internal/nullable/value.go
@@ -16,7 +16,7 @@ type Value[T any] struct {
 	Value T
 }
 
-func New[T any](v T, set bool) Value[T] {
+func New[T Nullable](v T, set bool) Value[T] {
 	return Value[T]{
 		Set:   set,
 		Value: v,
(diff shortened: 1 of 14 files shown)


internal/nullable/value.go
```diff
@@ -16,10 +16,17 @@ type Value[T any] struct {
...
 }
 
-// FromString[T any] represents a parser function. It is used to avoid
+func New[T any](v T, set bool) Value[T] {
+	return Value[T]{
+		Set:   set,
+		Value: v,
+	}
+}
+
```
arvindh123: What do you think about below approach ? 

```suggestion
type Nullable interface {
	~string | ~int | ~uint
}

// Value type is used to represent difference betweeen an
// intentionally omitted value and default type value.
type Value[T Nullable] struct {
	Set   bool
	Value T
}

func New[T Nullable](v T) Value[T] {
	return NewNull[T].Set(v)
}

func NewNull[T Nullable]() Value[T] {
	return Value[T]{}
}

// Set sets the value and marks it as set.
func (v *Value[T]) Set(val T) {
	v.Set = true
	v.Value = val
}

```

arvindh123 (changes requested):

internal/nullable/value.go
arvindh123: If `T` is `any` we could not determine `v == nil`
```go
func New[T](v T) Value[T] {
   if v == nil  { // x not possible 
   ...
}
```

If we have constraint like below, then it makes sure that pointer or interface can't be passed so `New[T Nullable]` function can set without checking for `nil`
```go
type Nullable interface {
	~string | ~int | ~uint
}
```

```
func New[T Nullable](v T) Value[T] {
	return NewNull[T].Set(v)
}
```

dborovcanin: I have added a union, but let's keep the factory method signature as is - it's simple enough and works for all cases. Setting the value is a nice idea, but we'd have to make it non-exported and use getters and setter, which is not Go idiomatic, and also does not bring anything to the table unless we make both `set` and `value` non-exported. Then, we can enforce that Set always makes `set` true, but I'm not sure that's even desired behavior. New(value, bool) should be enough and it's also the most flexible.

internal/nullable/value.go
```diff
@@ -16,10 +16,17 @@ type Value[T any] struct {
 	Value T
 }
 
-// FromString[T any] represents a parser function. It is used to avoid
+func New[T nullable](v T, set bool) Value[T] {
```
arvindh123: If we have `nullable`, then `value` could not be null   
example:
`var a int = 0`
Then the following is not true
`New(a,false)`
a is not null, I see `Set` is used for to determine any value is set or value is `nil`,  
by default `int` goes to `0` which is not `nil`

So new function can be like `New(a)`

If some one need to have int with null, then we can provide `NewNull[int]()`
`a := NewNull[int]()`
or 
`var a Value[int]` which defaults to null

internal/nullable/value.go
```diff
@@ -16,10 +16,17 @@ type Value[T any] struct {
...
+		Value: v,
+	}
+}
+
+// Parser[T any] represents a parser function. It is used to avoid
 // a single parser for all nullables for improved readability and performance.
-// FromString should always return Nullable with Set=true, error otherwise.
-type FromString[T any] func(string) (Value[T], error)
+// Parser should always return Nullable with Set=true, error otherwise.
+type Parser[T any] func(string) (Value[T], error)
```
arvindh123: T can be `any` or it should be `nulllable` ?

internal/nullable/value.go
dborovcanin: It's OK to restrict null value, that's why we have the `Set` field. If Set is false, we don't care about value. If it's true, we take the value, even if default.

internal/nullable/value.go
dborovcanin: Should be `nullable`, fixed.

pkg/sdk/groups_test.go
```diff
@@ -681,7 +681,7 @@ func TestUpdateGroup(t *testing.T) {
 
 	group.ID = generateUUID(t)
 
-	updatedDesc := nullable.Value[string]{Set: true, Value: updatedDescription}
+	updatedDesc := nullable.New(updatedDescription, true)
```
arvindh123: Please update this function

arvindh123 (changes requested):

PR #3000, opened 2025-07-03
SMQ-2977 - Add cache for message topic caching

### LINKED ISSUE #2977: Feature: Add Cache Topic Parser Cache for Handlers

No

The `ParsePublishTopic` function is invoked whenever a topic is provided. It parses the topic to extract `domainID`, `channelID`, `subtopic`, and returns an `error` if malformed. 
This is a CPU-consuming operation, especially under high load with repeated topics.

To optimize this, we propose introducing an in-memory cache (LRU or ARC) where the key is the full topic string, and the value is the parsed result, including errors for invalid topics.

This allows us to skip parsing entirely for repeated topics — valid or invalid — significantly reducing CPU overhead.

The cache is lightweight and scoped locally at the handler level, designed solely to accelerate topic parsing. It does not affect resolvers or any other logic.

**Topics using Route-based addressing must be resolved every time and cannot be cached**. 
**Topics with UUID-based addressing are stable and can be safely cached.**
**Since the resolver runs before the handler and converts Route-based topics into UUID-based topics before passing them to handlers, this cache can be effectively applied within the handler to avoid redundant parsing.**

Known cache types 

| Cache Type     | Use When                                   | Libraries                                                                          |
| -------------- | ------------------------------------------ | ---------------------------------------------------------------------------------- |
| **LRU Cache**  | Recent topics are likely to repeat         | [`hashicorp/golang-lru`](https://github.com/hashicorp/golang-lru) (very popular)   |
| **LFU Cache**  | Frequently used topics, not just recent    | [`dgraph-io/ristretto`](https://github.com/dgraph-io/ristretto) (High performance) |
| **ARC Cache**  | Balances recency & frequency               | \[`hashicorp/golang-lru` ARC mode]                                                 |
| **Simple map** | Small, bounded keyspace (e.g., 10K topics) | Native `map[string]ParsedTopic`                                                    |
| **Ristretto**  | Ultra-high performance, memory-bounded     | \[`dgraph-io/ristretto`] (✅ production grade)                                      |

Must-have

### Anything else?

_No response_

felixgateru:
# What type of PR is this?
This is a feature because it adds a cache in the adapters to cache message topics

## What does this do?
This pr:
- introduces configurable topic parsing with caching for all protocol adapters (HTTP, MQTT, CoAP, WS), improving performance and flexibility.
- adds new environment variables for cache configuration, allowing fine-tuning of cache size and behavior for each adapter
- replaces topic resolver logic with a unified, cache-enabled topic parser throughout all adapters and handlers.

- Resolves #2977

Yes, tests have been updated and added for new features

Yes, documentation has been added and existing dos updated


pkg/messaging/parser.go
```diff
@@ -0,0 +1,124 @@
...
+	if resolve {
+		domainID, channelID, isRoute, err = p.resolver.Resolve(ctx, domainID, channelID)
+		if err != nil {
+			return "", "", "", err
+		}
+	}
+	if !isRoute {
+		p.saveToCache(topic, domainID, channelID, subtopic, nil)
+	}
+
```
dborovcanin: Why don't we store it if it's a route? One of the main reasons we want caching is so we don't have to resolve routes all the time, since that's a very costly operation.

pkg/messaging/parser.go
```diff
@@ -0,0 +1,124 @@
...
+		if err != nil {
+			return "", "", "", err
+		}
+	}
+
+	return domainID, channelID, subtopic, nil
+}
+
+func (p *parser) saveToCache(topic string, domainID, channelID, subtopic string, err error) {
+	p.cache.Set(topic, &parsedTopic{
```
dborovcanin: Consider hashing the topic key. Fixed key length can benefit the underlying cache implementation (but not necessarily) and will certainly help memory usage.

dborovcanin (changes requested):
Since `pkg/messaging` is getting too messy, let's move topic-related stuff under `pkg/messaging/topics`.

pkg/messaging/parser.go
felixgateru: Ristretto point out that they already hash the key internally:
https://hypermode.com/blog/introducing-ristretto-high-perf-go-cache/

And I did notice a change in bechmarks for the hashed key:
image
Without adding a hash

image
With `hash/fnv` hash

dborovcanin:
@felixgateru Please resolve conflicts.

dborovcanin (approved):

PR #3015, opened 2025-07-14
SMQ-2873 - Switch to `TIMESTAMPZ` in Postgres

### LINKED ISSUE #2873: Switch to TIMESTAMPZ in Postgres
Related #2869 - We should use TIMESTAMPZ in PG instead of TIMESTAMP.

nyagamunene:
# What type of PR is this?
This is a bug fix because it fixes the following issue: #2873 

## What does this do?
It updates postgres from `TIMESTAMP` to `TIMESTAMPZ`

- Related Issue #2873 
- Resolves #2873

Yes


dborovcanin (changes requested):
Since this is a breaking change, let's add a migration for it. We will eventually reset migrations for v1.0, but since we are approaching the 1.0 release, let's make a migration to help seamless update.

dborovcanin (approved):

Issue #2801, opened 2025-04-09
Add health check endpoints to protocol adapters


dborovcanin:
We need to expose a health topic for different protocols so we check if messaging is working without polluting the actual messag topics.

arvindh123:
@dborovcanin  Can we have topic with prefix `h`  for health check ?

dborovcanin:
Yes, we can go with `hc` maybe. It's on the root level, like `m` for messages.

nyagamunene:
@arvindh123 From my understanding of this issue, we need health check endpoints to verify whether the adapter services are running correctly. However, I’d like some clarification: since `http`, `coap`, and `ws` already have health adapters, could I be misunderstanding something?

dborovcanin:
No, we need this for end-devices to check if connectivity works. I.e. we need an endpoint in each adapter that will not publish to NATS, but just receive the message and return the equivalent of HTTP OK to the device.

arvindh123:
All the PR has been merged, Closing this ticket.

PR #3024, opened 2025-07-21
SMQ-2801 - Add health check endpoint for MQTT adapter 

felixgateru:
# What type of PR is this?
This is a feature because it adds a health check endpoint for the MQTT adapter

## What does this do?
This pr adds a health check endpoint for the MQTT adapter

- Related Issue #2801

No

No

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -113,7 +113,7 @@ func (h *handler) Publish(ctx context.Context, topic *string, payload *[]byte) e
 		return errors.Wrap(errFailedPublish, errClientNotInitialized)
 	}
 
-	domainID, channelID, subtopic, err := h.parser.ParsePublishTopic(ctx, *topic, true)
+	domainID, channelID, subtopic, _, err := h.parser.ParsePublishTopic(ctx, *topic, true)
 	if err != nil {
 		return errors.Wrap(errMalformedTopic, err)
 	}
--- a/mqtt/handler.go
+++ b/mqtt/handler.go
@@ -113,12 +113,12 @@ func (h *handler) AuthPublish(ctx context.Context, topic *string, payload *[]byt
 		return ErrClientNotInitialized
 	}
 
-	domainID, chanID, _, err := h.parser.ParsePublishTopic(ctx, *topic, false)
+	domainID, chanID, _, topicType, err := h.parser.ParsePublishTopic(ctx, *topic, false)
 	if err != nil {
 		return err
 	}
 
-	return h.authAccess(ctx, string(s.Username), domainID, chanID, connections.Publish)
+	return h.authAccess(ctx, string(s.Username), domainID, chanID, connections.Publish, topicType)
 }
 
 // AuthSubscribe is called on device subscribe,
@@ -133,12 +133,12 @@ func (h *handler) AuthSubscribe(ctx context.Context, topics *[]string) error {
 	}
 
 	for _, topic := range *topics {
-		domainID, chanID, _, err := h.parser.ParseSubscribeTopic(ctx, topic, false)
+		domainID, chanID, _, topicType, err := h.parser.ParseSubscribeTopic(ctx, topic, false)
 		if err != nil {
 			return err
 		}
 
-		if err := h.authAccess(ctx, string(s.Username), domainID, chanID, connections.Subscribe); err != nil {
+		if err := h.authAccess(ctx, string(s.Username), domainID, chanID, connections.Subscribe, topicType); err != nil {
 			return err
 		}
 	}
@@ -164,7 +164,7 @@ func (h *handler) Publish(ctx context.Context, topic *string, payload *[]byte) e
 	}
 	h.logger.Info(fmt.Sprintf(LogInfoPublished, s.ID, *topic))
 
-	domainID, chanID, subTopic, err := h.parser.ParsePublishTopic(ctx, *topic, false)
+	domainID, chanID, subTopic, topicType, err := h.parser.ParsePublishTopic(ctx, *topic, false)
 	if err != nil {
 		return errors.Wrap(ErrFailedPublish, err)
 	}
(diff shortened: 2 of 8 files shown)


dborovcanin:
@felixgateru What's the status of this one?

dborovcanin:
The goal of MQTT HC is not that our service send a heartbeat to the broker, but to enable device to verify connectivity by sending a simple publish to the unauthorized, reserved topic used only for that purpose. However, to make sense out of it, we actually need a "special" channel and this message needs to be able to have authenticated (verify that relevant client exist) the client.

mqtt/handler.go
```diff
@@ -219,21 +221,26 @@ func (h *handler) Disconnect(ctx context.Context) error {
...
-	if err != nil {
-		return err
-	}
-	if !res.GetAuthorized() {
-		return svcerr.ErrAuthorization
-	}
+func (h *handler) authAccess(ctx context.Context, clientID, domainID, chanID string, msgType connections.ConnType, topicType messaging.TopicType) error {
+	switch topicType {
+	case messaging.HealthType:
+		return nil
```
dborovcanin, line 227: In this case, we need to _authenticate_ the client. I.e. we can't just ignore **any** and channelID client.

ws/handler.go
```diff
@@ -84,7 +84,7 @@ func (h *handler) AuthPublish(ctx context.Context, topic *string, payload *[]byt
 		return errClientNotInitialized
 	}
 
-	domainID, channelID, _, err := h.parser.ParsePublishTopic(ctx, *topic, true)
+	domainID, channelID, _, _, err := h.parser.ParsePublishTopic(ctx, *topic, true)
```
dborovcanin, line 87: Since this is already intrusive change, let's add the same HC endpoint to other protocols as well.

http/handler.go
```diff
@@ -113,7 +113,7 @@ func (h *handler) Publish(ctx context.Context, topic *string, payload *[]byte) e
 		return errors.Wrap(errFailedPublish, errClientNotInitialized)
 	}
 
-	domainID, channelID, subtopic, err := h.parser.ParsePublishTopic(ctx, *topic, true)
+	domainID, channelID, subtopic, _, err := h.parser.ParsePublishTopic(ctx, *topic, true)
```
dborovcanin, line 116: Same here, since we are changing message parser, we can as well add HC for HTTP as well.

dborovcanin (changes requested):

mqtt/handler.go
felixgateru: in the case of MQTT the client is already authenticated here:
https://github.com/absmach/supermq/blob/1d935f05198e8f22f4ca764a6726670222d37141/mqtt/handler.go#L90-L100

before getting to AuthPublish and AuthSubscribe for channel authorization
Do we add channelID to the health check topic and check that the client is authorized to publish or subscribe to the channel as well?

dborovcanin: No, let's not do it for the time being; mostly because we'd need to enforce channels which would make things complicated.

pkg/messaging/topics.go
```diff
@@ -53,15 +61,16 @@ type parsedTopic struct {
 	domainID  string
 	channelID string
 	subtopic  string
+	topicType TopicType
```
dborovcanin: No need for this. We don't have to cache anything but `MessageType`; which makes it unnecessary here.

pkg/messaging/topics.go
```diff
@@ -347,33 +360,61 @@ func EncodeMessageMQTTTopic(m *Message) string {
 	return topic
 }
 
-func EncodeAdapterTopic(domain, channel, subtopic string) string {
-	topic := fmt.Sprintf("%s/%s/%s/%s", string(MsgTopicPrefix), domain, string(ChannelTopicPrefix), channel)
-	if subtopic != "" {
-		topic = topic + "/" + subtopic
+func EncodeAdapterTopic(domain, channel, subtopic string, topicType TopicType) string {
```
dborovcanin: There is no need for this to be exported.

dborovcanin (changes requested):

pkg/messaging/topics.go
```diff
@@ -165,14 +173,17 @@ func NewTopicResolver(channelsClient grpcChannelsV1.ChannelsServiceClient, domai
...
+	if domain == "" {
 		return "", "", false, ErrEmptyRouteID
 	}
 
 	domainID, isdomainRoute, err := r.resolveDomain(ctx, domain)
 	if err != nil {
 		return "", "", false, errors.Wrap(ErrFailedResolveDomain, err)
 	}
+	if channel == "" {
+		return domainID, "", isdomainRoute, nil
```
dborovcanin: Not related to this PR; please fix naming to `isDomainRoute`.

pkg/messaging/topics.go
```diff
@@ -347,33 +358,61 @@ func EncodeMessageMQTTTopic(m *Message) string {
...
 	if n > 0 && topic[0] == '/' {
 		start = 1
 	}
+	if n <= start {
+		return "", "", "", InvalidType, ErrMalformedTopic
+	}
+
+	// Healthcheck: "hc/<domain_id>"
+	// Check first because it's shortest and avoids extra work.
+	if n >= start+3 && topic[start] == HealthTopicPrefix[0] && topic[start+1] == HealthTopicPrefix[1] && topic[start+2] == '/' {
```
dborovcanin: This looks a little overegineered. Try:
```go
	if n > start+3 && topic[start:start+2] == HealthTopicPrefix { // notice > 3, not >= 3 so you don't have to check if there is a domain - there is)
```

dborovcanin (changes requested):

dborovcanin (changes requested):
Add publish tests to `hc` endpoint.

dborovcanin (approved):

PR #3030, opened 2025-07-29
SMQ-2866 - MQTT auth fails with identical secrets across different domains

### LINKED ISSUE #2866: Bug: MQTT auth fails with identical secrets across different domains

Was using MQTT to send message.

Clients should send message via MQTT

`{"time":"2025-05-12T04:44:17.98368156Z","level":"WARN","msg":"failed to proxy from MQTT broker to client with id 19eef5c7-d2ef-456b-84b7-336f596a78d3-pub with error: invalid user id"}`

### Steps To Reproduce

1. Create `client1` in `domain1`  with `secret1`
2. Create `channel1` in `domain1` and connect `client1` and `channel1`
3. Create `client2` in `domain2`  with `secret1`
4. Create `channel2` in `domain2` and connect `client2` and `channel2`
5. Send a message via `MQTT` with either client

Docker supermq

_No response_

dborovcanin:
# What type of PR is this?
This is a bugfix.

## What does this do?
This pull request enables messaging for multiple clients with the same key in different domains.

- Resolves #2866

Yes.

N/A

### Notes
N/A


pkg/authn/authn.go
```diff
@@ -48,3 +51,57 @@ type Session struct {
...
+	}
+	return "Unknown"
+}
+
+// ErrNotEncoded acts similarly to EOF - it does indicate there is no suffix in
+// the token, but that does not have to be treated as the error in some cases.
+// If token is not base64-encoded, the token is returned as a key alongside with the error.
+var ErrNotEncoded = errors.New("token is not encoded with domain ID suffix")
+
+func AuthUnpack(token string) (AuthPrefix, string, string, error) {
```
arvindh123, line 84: **Why don't we handle this at the protocol layer, like in HTTP or CoAP headers?**
HTTP supports basic authentication, so we can perform this at the HTTP layer and send the ID and key to the Clients service for verification.
We can do the same with CoAP, since it functions similarly to an HTTP server.

dborovcanin: Backward compatibility is the only reason for the time being. We can switch to basic Auth, but then we will break the API - `Client` prefix will make no sense anymore.

nyagamunene (dismissed):
LGTM

PR #3038, opened 2025-08-01
NOISSUE - List domain invitations on sdk

WashingtonKK:
# What type of PR is this?

- Related Issue #
- Resolves #

### DIFF
--- a/pkg/sdk/invitations.go
+++ b/pkg/sdk/invitations.go
@@ -6,6 +6,7 @@ package sdk
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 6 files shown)


pkg/sdk/invitations.go
```diff
@@ -68,6 +68,26 @@ func (sdk mgSDK) Invitation(ctx context.Context, userID, domainID, token string)
 	return invitation, nil
 }
 
+func (sdk mgSDK) DomainInvitations(ctx context.Context, pm PageMetadata, token, domainID string) (invitations InvitationPage, err error) {
+	url := domainsEndpoint + "/" + domainID + "/" + invitationsEndpoint
```
dborovcanin: Use `fmt.Printf` instead. Do the same for all the other cases across SDK package.

dborovcanin (changes requested):

pkg/sdk/invitations.go
SammyOina: sprintf

dborovcanin (approved):

PR #3049, opened 2025-08-06
SMQ-2799 - Add support for basic auth for HTTP and WS adapters

### LINKED ISSUE #2799: Feature: Clients authentication : Consider for moving to basic authentication with `client_id` and `client_key` for `Websocket` and `HTTP` adapters
Support Authentication with Both `client_id` and `client_key` for Enhanced Security

Currently, we use only the `client_key` for authenticating clients during messaging. This approach requires `client_key` to be globally unique, which introduces the risk of brute-force discovery.

### Proposed Enhancement:

Introduce authentication using **both `client_id` and `client_key`**. This provides several benefits:

* Eliminates the need for a globally unique `client_key`.
* Reduces the risk of brute-force attacks, since both values must match.
* Simplifies lookup: the system can directly validate that the provided `client_id` and `client_key` pair exists without needing to look up the `client_id` from the `client_key`.

### Implementation Suggestions:

* **HTTP Protocol:** Use Basic Authentication or a Base64-encoded string of `client_id:client_key` in the HTTP headers.
* **WebSocket Protocol:** Support credentials via query parameters:

  ```
  ws://localhost/ws?auth=<client_id>|<client_key>
  ```

  or a Base64-encoded form:

  ```
  ws://localhost/ws?auth=<base64(client_id:client_key)>
  ```

felixgateru:
# What type of PR is this?

- Related Issue #2799
- Resolves #2799


dborovcanin:
@felixgateru Is this ready?

ws/tracing/tracing.go
```diff
@@ -31,11 +31,11 @@ func New(tracer trace.Tracer, svc ws.Service) ws.Service {
 }
 
 // Subscribe traces the "Subscribe" operation of the wrapped ws.Service.
-func (tm *tracingMiddleware) Subscribe(ctx context.Context, sessionID, clientKey, domainID, chanID, subtopic string, client *ws.Client) error {
+func (tm *tracingMiddleware) Subscribe(ctx context.Context, sessionID, clientUsername, clientPassword, domainID, chanID, subtopic string, client *ws.Client) error {
```
dborovcanin: No need for `client` prefix; use `username` and `password`.

ws/adapter.go
```diff
@@ -135,3 +148,15 @@ func (svc *adapterService) authorize(ctx context.Context, authKey, domainID, cha
 
 	return clientID, nil
 }
+
+func (svc *adapterService) clientAuthenticate(ctx context.Context, token string) (string, error) {
+	authnRes, err := svc.clients.Authenticate(ctx, &grpcClientsV1.AuthnReq{Token: token})
+	if err != nil {
+		return "", svcerr.ErrAuthentication
+	}
```
dborovcanin: Let's wrap the error with `ErrAuthentication` instead.

ws/handler.go
```diff
@@ -226,3 +241,31 @@ func (h *handler) authAccess(ctx context.Context, token, domainID, chanID string
 
 	return clientID, nil
 }
+
+func (h *handler) clientAuthenticate(ctx context.Context, token string) (string, error) {
+	authnRes, err := h.clients.Authenticate(ctx, &grpcClientsV1.AuthnReq{Token: token})
+	if err != nil {
+		return "", svcerr.ErrAuthentication
+	}
```
dborovcanin: Same, wrap instead.

dborovcanin (changes requested):

http/handler.go
```diff
@@ -120,19 +122,33 @@ func (h *handler) Publish(ctx context.Context, topic *string, payload *[]byte) e
...
+		clientID, err = h.clientAuthenticate(ctx, smqauthn.AuthPack(smqauthn.DomainAuth, domainID, secret))
 		if err != nil {
-			h.logger.Warn(fmt.Sprintf(logInfoFailedAuthNClient, secret, *topic, err))
-			return mgate.NewHTTPProxyError(http.StatusUnauthorized, svcerr.ErrAuthentication)
-		}
-		if !authnRes.Authenticated {
-			h.logger.Warn(fmt.Sprintf(logInfoFailedAuthNClient, secret, *topic, svcerr.ErrAuthentication))
-			return mgate.NewHTTPProxyError(http.StatusUnauthorized, svcerr.ErrAuthentication)
+			h.logger.Warn(fmt.Sprintf(logInfoFailedAuthNClient, s.Username, *topic, err))
+			return mgate.NewHTTPProxyError(http.StatusUnauthorized, err)
```
dborovcanin: A lot of repetition between cases. Use switch to determine credentials and auth type only, and then do authentication and error handling.

dborovcanin (changes requested):
Remove unused constants in http/handler.go:
```go
const (
	protocol                = "http"
	clientIDCtxKey   ctxKey = "client_id"
	clientTypeCtxKey ctxKey = "client_type"
)
```
to
```go
const protocl = "http"
```
Also, Handler should be wrapping logging rather than having it coupled with the logic, but that can be addressed in a separate PR.

http/handler.go
```diff
@@ -146,6 +143,14 @@ func (h *handler) Publish(ctx context.Context, topic *string, payload *[]byte) e
...
 	}
 
+	if clientType == policies.ClientType {
+		clientID, err = h.clientAuthenticate(ctx, authReq)
+		if err != nil {
+			h.logger.Warn(fmt.Sprintf(logInfoFailedAuthNClient, s.Username, *topic, err))
+			return mgate.NewHTTPProxyError(http.StatusUnauthorized, err)
+		}
+	}
+
```
dborovcanin: Consider simplifying flow to something like:
```go
// Publish - after client successfully published.
func (h *handler) Publish(ctx context.Context, topic *string, payload *[]byte) error {
	if topic == nil {
		return errMissingTopicPub
	}
	topic = &strings.Split(*topic, "?")[0]
	s, ok := session.FromContext(ctx)
	if !ok {
		return errors.Wrap(errFailedPublish, errClientNotInitialized)
	}

	domainID, channelID, subtopic, err := h.parser.ParsePublishTopic(ctx, *topic, true)
	if err != nil {
		return errors.Wrap(errMalformedTopic, err)
	}

	var id, clientType string
	var token string
	pass := string(s.Password)
	switch {
	case s.Username != "" && string(s.Password) != "":
		token = smqauthn.AuthPack(smqauthn.BasicAuth, s.Username, string(s.Password))
		clientType = policies.ClientType
	case strings.HasPrefix(pass, apiutil.BasicAuthPrefix):
		username, password, err := decodeAuth(strings.TrimPrefix(string(s.Password), apiutil.BasicAuthPrefix))
		if err != nil {
			h.logger.Warn(fmt.Sprintf(logInfoFailedAuthNClient, s.Username, *topic, err))
			return mgate.NewHTTPProxyError(http.StatusUnauthorized, err)
		}

		token = smqauthn.AuthPack(smqauthn.BasicAuth, username, password)
		clientType = policies.ClientType
	case strings.HasPrefix(pass, "Client"):
		secret := strings.TrimPrefix(string(s.Password), apiutil.ClientPrefix)
		token = smqauthn.AuthPack(smqauthn.DomainAuth, domainID, secret)
		clientType = policies.ClientType
	case strings.HasPrefix(pass, apiutil.BearerPrefix):
		token = strings.TrimPrefix(string(s.Password), apiutil.BearerPrefix)
		clientType = policies.UserType
	default:
		return mgate.NewHTTPProxyError(http.StatusUnauthorized, svcerr.ErrAuthentication)
	}

	id, err = h.authenticate(ctx, clientType, token)
	if err != nil {
		h.logger.Warn(fmt.Sprintf(logInfoFailedAuthNClient, id, *topic, err))
		return mgate.NewHTTPProxyError(http.StatusUnauthorized, err)
	}
```
where `h.authenticate` is:
```go

func (h *handler) authenticate(ctx context.Context, authType, token string) (string, error) {
	switch authType {
	case policies.UserType:
		authnSession, err := h.authn.Authenticate(ctx, token)
		if err != nil {
			return "", err
		}
		return authnSession.UserID, nil
	case policies.ClientType:
		authnRes, err := h.clients.Authenticate(ctx, &grpcClientsV1.AuthnReq{Token: token})
		if err != nil {
			return "", svcerr.ErrAuthentication
		}
		if !authnRes.Authenticated {
			return "", svcerr.ErrAuthentication
		}

		return authnRes.GetId(), nil
	default:
		return "", errors.New("invalid client type")
	}
}
```

dborovcanin (changes requested):

docker/addons/certs/openbao-entrypoint.sh
```diff
@@ -270,7 +270,10 @@ if [ ! -f /opt/openbao/data/configured ]; then
     key_usage=\"DigitalSignature,KeyEncipherment,KeyAgreement\" \
     ext_key_usage=\"ServerAuth,ClientAuth,OCSPSigning\" \
     use_csr_common_name=true \
-    use_csr_sans=false \
+    use_csr_sans=true \
+    copy_extensions=true \
+    allowed_extensions=\"*\" \
+    basic_constraints_valid_for_non_ca=true \
```
dborovcanin, line 276: What does this have to do with Basic Auth?

http/handler.go
```diff
@@ -118,34 +115,36 @@ func (h *handler) Publish(ctx context.Context, topic *string, payload *[]byte) e
...
+		token = strings.TrimPrefix(pass, apiutil.BearerPrefix)
 		clientType = policies.UserType
-		clientID = authnSession.UserID
 	default:
 		return mgate.NewHTTPProxyError(http.StatusUnauthorized, svcerr.ErrAuthentication)
 	}
 
+	clientID, err := h.authenticate(ctx, clientType, token)
+	if err != nil {
+		h.logger.Warn(fmt.Sprintf(logInfoFailedAuthNClient, clientID, *topic, err))
```
dborovcanin: It's not necessarily the client, it can also be a user. Solve that by returning and handling a proper error from `authenticate`

dborovcanin (changes requested):

docker/addons/certs/openbao-entrypoint.sh
felixgateru: The CI was failing on Fetch Certs, so i run make fetch_certs

http/handler.go
```diff
@@ -31,9 +31,8 @@ const protocol = "http"
 
 // Log message formats.
 const (
-	logInfoPublished         = "published with client_type %s client_id %s to the topic %s"
-	logInfoFailedAuthNToken  = "failed to authenticate token for topic %s with error %s"
-	logInfoFailedAuthNClient = "failed to authenticate client key %s for topic %s with error %s"
+	logInfoPublished   = "published with client_type %s client_id %s to the topic %s"
+	logInfoFailedAuthN = "failed to authenticate client_type %s for topic %s with error %s"
```
dborovcanin: This is not info, but a warning. It's probably safer and more accurate to use `failedAuthnFmt` and `publishedInfoFmt`.

dborovcanin (changes requested):

dborovcanin (approved):

dborovcanin (dismissed):

dborovcanin:
@felixgateru Please fix CI.

docker/addons/certs/.env
```diff
@@ -33,6 +33,8 @@ AM_CERTS_GRPC_CLIENT_TLS=
 AM_CERTS_GRPC_CA_CERTS=
 AM_CERTS_INSTANCE_ID=
 AM_CERTS_RELEASE_TAG=latest
+# WARNING: This is a development/testing secret only.
+# NEVER use this weak secret in production! Generate a strong random secret for production deployments.
```
dborovcanin, line 37: This is not necessary (also I suspect AI generated).

http/handler.go
```diff
@@ -118,37 +114,39 @@ func (h *handler) Publish(ctx context.Context, topic *string, payload *[]byte) e
...
+		username, password, err := decodeAuth(strings.TrimPrefix(pass, apiutil.BasicAuthPrefix))
 		if err != nil {
-			h.logger.Warn(fmt.Sprintf(logInfoFailedAuthNToken, *topic, err))
-			return mgate.NewHTTPProxyError(http.StatusUnauthorized, svcerr.ErrAuthentication)
+			h.logger.Warn(fmt.Sprintf(failedAuthnFmt, policies.ClientType, *topic, err))
+			return mgate.NewHTTPProxyError(http.StatusUnauthorized, err)
 		}
+		token = smqauthn.AuthPack(smqauthn.BasicAuth, username, password)
+		clientType = policies.ClientType
+	case strings.HasPrefix(pass, "Client"):
```
dborovcanin: Use `apiutil.ClientPrefix` here.

dborovcanin (changes requested):

docker/addons/certs/.env
felixgateru: This is from certs, it is introduced in this pr because of running `make fetch_certs` to pass the ci

dborovcanin (approved):

PR #3052, opened 2025-08-07
NOISSUE - Allow sorting

Musilah:
# What type of PR is this?
Updates the entities to use order and direction

## What does this do?
Allows entities to correctly use sorting when listing

N/A

- Related Issue https://github.com/absmach/magistrala-ui/issues/795

N/A

N/A


channels/api/http/requests.go
```diff
@@ -93,6 +93,10 @@ func (req listChannelsReq) validate() error {
 		return apiutil.ErrNameSize
 	}
 
+	if req.Order != "" && req.Order != api.NameOrder && req.Order != api.CreatedAtOrder && req.Order != api.UpdatedAtOrder {
```
ianmuchyri: can we also check `direction`. Do the same for the rest.

ianmuchyri (changes requested):

ianmuchyri (dismissed):

channels/api/http/requests.go
```diff
@@ -93,6 +93,14 @@ func (req listChannelsReq) validate() error {
 		return apiutil.ErrNameSize
 	}
 
+	if req.Order != "" && (req.Order != api.NameOrder && req.Order != api.CreatedAtOrder && req.Order != api.UpdatedAtOrder) {
+		return apiutil.ErrInvalidOrder
+	}
```
arvindh123: Please use switch case for readable clean

```go
	switch req.Order {
	case  "", api.NameOrder ,  api.CreatedAtOrder ,  api.UpdatedAtOrder:
		// Do nothing
	default:
		return apiutil.ErrInvalidOrder
	}
```

clients/api/http/requests.go
```diff
@@ -85,6 +85,14 @@ func (req listClientsReq) validate() error {
 		return apiutil.ErrNameSize
 	}
 
+	if req.Order != "" && (req.Order != api.NameOrder && req.Order != api.CreatedAtOrder && req.Order != api.UpdatedAtOrder) {
+		return apiutil.ErrInvalidOrder
+	}
+
+	if req.Dir != "" && (req.Dir != api.AscDir && req.Dir != api.DescDir) {
+		return apiutil.ErrInvalidDirection
+	}
```
arvindh123: Is there need for multiple column sorts ? 
```sql
SELECT name, age
FROM people
ORDER BY name DESC, age ASC;
```

arvindh123 (changes requested):

users/api/requests.go
```diff
@@ -99,7 +99,16 @@ func (req listUsersReq) validate() error {
 	if req.limit > maxLimitSize || req.limit < 1 {
 		return apiutil.ErrLimitSize
 	}
-	if req.dir != "" && (req.dir != api.AscDir && req.dir != api.DescDir) {
+
+	switch req.order {
+	case "", api.NameOrder, api.CreatedAtOrder, api.UpdatedAtOrder:
```
arvindh123: What happens if "first_name", "last_name", "username", "email" are passed in query

arvindh123 (changes requested):
Does Clients, Channels and Domain postgres function have implemented ordering ?

users/api/requests.go
Musilah: Yes they do. I just had to add that check to the api layer.

Musilah: Yes I will update that and send the fix. We should be using those fields instead.

arvindh123 (approved):

arvindh123 (changes requested):
Please fix CI

dborovcanin (approved):

PR #3079, opened 2025-08-25
NOISSUE - Replace interface{} with any

dborovcanin:
# What type of PR is this?
This is a refactor.

## What does this do?
It replaces `interface{}` with a more modern `any`.

There is no such issue.

N/A

N/A

### Notes
Any is equivalent to `interface{}` in all ways.


felixgateru (approved):

pkg/domains/events/consumer/decode.go
```diff
@@ -36,7 +36,7 @@ var (
 	errUpdatedAt     = errors.New("failed to parse 'updated_at' time")
 )
 
-func ToDomains(data map[string]interface{}) (domains.Domain, error) {
+func ToDomains(data map[string]any) (domains.Domain, error) {
```
arvindh123, line 39: Mock file is auto generated, So we could not edit this files

arvindh123 (changes requested):

pkg/domains/events/consumer/decode.go
dborovcanin: This is not a mock file; i.e. `make mocks` will not result in any changes.

pkg/callout/mocks/callout.go
```diff
@@ -41,15 +41,15 @@ func (_m *Callout) EXPECT() *Callout_Expecter {
 }
 
 // Callout provides a mock function for the type Callout
-func (_mock *Callout) Callout(ctx context.Context, perm string, pl map[string]interface{}) error {
+func (_mock *Callout) Callout(ctx context.Context, perm string, pl map[string]any) error {
```
arvindh123, line 44: Sorry, I intended to mention this change.

arvindh123 (changes requested):

pkg/callout/mocks/callout.go
dborovcanin: I did not manually change this; `make mocks` does it since the `Callout` method interface changed.

arvindh123 (approved):

arvindh123 (approved):

PR #3039, opened 2025-08-01
NOISSUE - Handle callout error

arvindh123:
# What type of PR is this?

## Which issue(s) does this PR fix/relate to?

- Related Issue #
- Resolves #

PR #3101, opened 2025-08-29
SMQ-3093 - User email verification

arvindh123:
# What type of PR is this?

- Resolves #3039


pkg/authn/middleware.go
```diff
@@ -0,0 +1,160 @@
+package authn
```
felixgateru, line 4: Add license header

users/postgres/users.go
```diff
@@ -565,6 +607,15 @@ func ToUser(dbu DBUser) (users.User, error) {
 	if dbu.UpdatedAt.Valid {
 		updatedAt = dbu.UpdatedAt.Time.UTC()
 	}
+	var verifiedAt time.Time
+	if dbu.UpdatedAt.Valid {
```
felixgateru:
```suggestion
	if dbu.VerifiedAt.Valid {
```

users/postgres/init.go
```diff
@@ -97,6 +97,23 @@ func Migration() *migrate.MemoryMigrationSource {
...
+				Up: []string{
+					`ALTER TABLE users
+						ADD COLUMN verified BOOLEAN DEFAULT FALSE,
+						ADD COLUMN verified_at TIMESTAMP,
+						ADD COLUMN verification_token TEXT,
+						ADD COLUMN verification_token_expires_at TIMESTAMP;`,
+				},
+				Down: []string{
+					`ALTER TABLE users
+						DROP COLUMN verified BOOLEAN DEFAULT FALSE,
```
felixgateru: I think down migrations should not have a default

users/api/users.go
```diff
@@ -28,13 +28,14 @@ import (
...
 // usersHandler returns a HTTP handler for API endpoints.
-func usersHandler(svc users.Service, authn smqauthn.Authentication, tokenClient grpcTokenV1.TokenServiceClient, selfRegister bool, r *chi.Mux, logger *slog.Logger, pr *regexp.Regexp, idp supermq.IDProvider, providers ...oauth2.Provider) *chi.Mux {
+func usersHandler(svc users.Service, authn smqauthn.Authn, tokenClient grpcTokenV1.TokenServiceClient, selfRegister bool, r *chi.Mux, logger *slog.Logger, pr *regexp.Regexp, idp supermq.IDProvider, providers ...oauth2.Provider) *chi.Mux {
 	passRegex = pr
 
 	opts := []kithttp.ServerOption{
 		kithttp.ServerErrorEncoder(apiutil.LoggingErrorEncoder(logger, api.EncodeError)),
 	}
-
+	//All endpoints in users service doesn't required Domain check
```
felixgateru:
```suggestion
	// All endpoints in users service don't required Domain check
```

users/users.go
```diff
@@ -143,6 +164,7 @@ type Page struct {
 	FirstName  string   `json:"first_name,omitempty"`
 	LastName   string   `json:"last_name,omitempty"`
 	Email      string   `json:"email,omitempty"`
+	Verified   bool     `json:"Verified,omitempty"`
```
felixgateru:
```suggestion
	Verified   bool     `json:"verified,omitempty"`
```

users/users.go
```diff
@@ -29,6 +29,19 @@ type User struct {
...
+	// VerificationToken is sent to the user's email and should not be exposed via API.
+	VerificationToken          string    `json:"-"`
+	VerificationTokenExpiresAt time.Time `json:"-"`
+}
+
+func (u *User) IsVerificationTokenExpired() bool {
+	if u.VerificationTokenExpiresAt.IsZero() {
+		return true
+	}
+	return time.Now().After(u.VerificationTokenExpiresAt)
```
felixgateru: Should we use .UTC() here?

felixgateru (changes requested):

users/api/requests.go
```diff
@@ -39,15 +38,16 @@ func (req createUserReq) validate() error {
 		return err
 	}
 	// Username must not be a valid email format due to username/email login.
-	if _, err := mail.ParseAddress(req.User.Credentials.Username); err == nil {
+	if err := api.ValidateEmail(req.User.Email); err == nil {
 		return apiutil.ErrInvalidUsername
 	}
+
```
felixgateru:
```suggestion
	if err := api.ValidateEmail(req.User.Credentials.Username); err == nil {
		return apiutil.ErrInvalidUsername
	}

```

this logic is to ensure that a user does not have an email address as a username

users/api/requests.go
```diff
@@ -67,6 +67,24 @@ func (req createUserReq) validate() error {
 	return req.User.Validate()
 }
 
+type sendVerificationReq struct{}
+
+func (req sendVerificationReq) validate() error {
+	return nil
+}
```
felixgateru: why have an empty validate function?

users/api/users.go
```diff
@@ -196,6 +212,30 @@ func usersHandler(svc users.Service, authn smqauthn.Authentication, tokenClient
...
 
+func decodeSendVerification(_ context.Context, r *http.Request) (any, error) {
+	if !strings.Contains(r.Header.Get("Content-Type"), api.ContentType) {
+		return nil, errors.Wrap(apiutil.ErrValidation, apiutil.ErrUnsupportedContentType)
+	}
+
+	req := sendVerificationReq{}
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return nil, errors.Wrap(apiutil.ErrValidation, errors.Wrap(err, errors.ErrMalformedEntity))
+	}
```
felixgateru: Why decode into an empty request struct?

users/api/users.go
```diff
@@ -196,6 +212,30 @@ func usersHandler(svc users.Service, authn smqauthn.Authentication, tokenClient
 	return r
 }
 
+func decodeSendVerification(_ context.Context, r *http.Request) (any, error) {
+	if !strings.Contains(r.Header.Get("Content-Type"), api.ContentType) {
+		return nil, errors.Wrap(apiutil.ErrValidation, apiutil.ErrUnsupportedContentType)
+	}
```
felixgateru: Is this required as the request is empty?

felixgateru (changes requested):

dborovcanin (changes requested):
CI is failing.

users/README.md
```diff
@@ -12,48 +12,51 @@ For in-depth explanation of the aforementioned scenarios, as well as thorough un
...
+| SMQ_USERS_DB_SSL_ROOT_CERT        | Path to the PEM encoded root certificate file                           | ""                                |
+| SMQ_EMAIL_HOST                    | Mail server host                                                        | localhost                         |
+| SMQ_EMAIL_PORT                    | Mail server port                                                        | 25                                |
+| SMQ_EMAIL_USERNAME                | Mail server username                                                    | ""                                |
+| SMQ_EMAIL_PASSWORD                | Mail server password                                                    | ""                                |
+| SMQ_EMAIL_FROM_ADDRESS            | Email "from" address                                                    | ""                                |
+| SMQ_EMAIL_FROM_NAME               | Email "from" name                                                       | ""                                |
+| SMQ_PASSWORD_RESET_URL_PREFIX     | Password reset URL prefix                                               | http://localhost/password/reset>  |
```
felixgateru:
```suggestion
| SMQ_PASSWORD_RESET_URL_PREFIX     | Password reset URL prefix                                               | <http://localhost/password/reset>  |
```

users/README.md
```diff
@@ -12,48 +12,51 @@ For in-depth explanation of the aforementioned scenarios, as well as thorough un
...
+| SMQ_PASSWORD_RESET_EMAIL_TEMPLATE | Password reset email template                                           | reset-password-email.tmpl         |
+| SMQ_VERIFICATION_URL_PREFIX       | Verification URL prefix                                                 | http://localhost/verify-email>    |
```
felixgateru:
```suggestion
| SMQ_VERIFICATION_URL_PREFIX       | Verification URL prefix                                                 | <http://localhost/verify-email>    |
```

users/postgres/init.go
```diff
@@ -97,6 +97,23 @@ func Migration() *migrate.MemoryMigrationSource {
 					`ALTER TABLE users ALTER COLUMN updated_at TYPE TIMESTAMP;`,
 				},
 			},
+			{
+				Id: "clients_07",
+				Up: []string{
+					`ALTER TABLE users
+						ADD COLUMN verified BOOLEAN DEFAULT FALSE,
+						ADD COLUMN verified_at TIMESTAMP,
```
felixgateru:
```suggestion
						ADD COLUMN verified_at TIMESTAMPZ,
```

users/postgres/init.go
```diff
@@ -97,6 +97,23 @@ func Migration() *migrate.MemoryMigrationSource {
...
+						ADD COLUMN verification_token TEXT,
+						ADD COLUMN verification_token_expires_at TIMESTAMP;`,
```
felixgateru:
```suggestion
						ADD COLUMN verification_token_expires_at TIMESTAMPZ;`,
```

auth/jwt/tokenizer.go
```diff
@@ -36,6 +38,7 @@ const (
 	tokenType              = "type"
 	userField              = "user"
 	RoleField              = "role"
+	VerifiedFiled          = "verified"
```
felixgateru:
```suggestion
	VerifiedField          = "verified"
```

users/service.go
```diff
@@ -19,10 +22,18 @@ import (
 	"github.com/absmach/supermq/pkg/policies"
 )
 
+const (
+	verificationTokenExpiryDuration = 24 * time.Hour
+	verificationTokenSeparator      = "."
```
dborovcanin: Remove if unused.

users/service.go
```diff
@@ -19,10 +22,18 @@ import (
 	"github.com/absmach/supermq/pkg/policies"
 )
 
+const (
```
dborovcanin: Once you remove unused const, make this inline and rename `const verificationTokenDuration`.

users/service.go
```diff
@@ -92,6 +103,82 @@ func (svc service) Register(ctx context.Context, session authn.Session, u User,
 	return user, nil
 }
 
+func (svc service) SendVerification(ctx context.Context, session authn.Session) error {
+	dbUser, err := svc.users.RetrieveVerificationToken(ctx, session.UserID)
+	if err != nil {
+		if err == repoerr.ErrNotFound {
+			return nil
+		}
+		return err
```
dborovcanin: No need for nested `if`:
```go
if err == repoerr.ErrNotFound {
    return nil
}
if err != nil {
    return err
}
```
Why do we return nil in case of notFound?

cmd/users/main.go
```diff
@@ -121,12 +125,21 @@ func main() {
...
+	if err := env.Parse(&resetPasswordEmailConfig); err != nil {
+		logger.Error(fmt.Sprintf("failed to load reset password email configuration : %s", err.Error()))
 		exitCode = 1
 		return
 	}
+	resetPasswordEmailConfig.Template = cfg.PasswordResetEmailTemplate
+
+	verificationEmailConfig := email.Config{}
+	if err := env.Parse(&verificationEmailConfig); err != nil {
+		logger.Error(fmt.Sprintf("failed to load reset password email configuration : %s", err.Error()))
```
felixgateru:
```suggestion
		logger.Error(fmt.Sprintf("failed to load verification email configuration : %s", err.Error()))
```

pkg/authn/middleware.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/absmach/supermq/pkg/errors"
+	"github.com/go-chi/chi/v5"
+)
+
+type sessionKeyType string
+
+const (
+	allowUnverifiedUserEnv = "SMQ_ALLOW_UNVERIFIED_USER"
+	jsonContentType        = "application/json"
+	textContentType        = "text/plain"
```
felixgateru: I think this is unused

users/service.go
```diff
@@ -110,7 +197,7 @@ func (svc service) IssueToken(ctx context.Context, identity, secret string) (*gr
 		return &grpcTokenV1.Token{}, errors.Wrap(svcerr.ErrLogin, err)
 	}
 
-	token, err := svc.token.Issue(ctx, &grpcTokenV1.IssueReq{UserId: dbUser.ID, UserRole: uint32(dbUser.Role + 1), Type: uint32(smqauth.AccessKey)})
+	token, err := svc.token.Issue(ctx, &grpcTokenV1.IssueReq{UserId: dbUser.ID, UserRole: uint32(dbUser.Role + 1), Type: uint32(smqauth.AccessKey), Verified: dbUser.Verified})
```
dborovcanin: Not related to this PR, but `Role+1` is too cryptic.

pkg/errors/service/types.go
```diff
@@ -93,4 +93,13 @@ var (
 
 	// ErrSuperAdminAction indicates that the user is not a super admin.
 	ErrSuperAdminAction = errors.New("not authorized to perform admin action")
+
+	// ErrUserAlreadyVerified indicates user is already verified.
+	ErrUserAlreadyVerified = errors.New("user already verified")
+
+	// ErrInvalidVerificationToken indicates user is already verified.
```
felixgateru:
```suggestion
	// ErrInvalidVerificationToken indicates verification token is invalid.
```

users/service.go
```diff
@@ -574,3 +666,43 @@ func (svc service) updateUserPolicy(ctx context.Context, userID string, role Rol
...
+		return tokenPayload{}, errors.Wrap(errFailedToDecodeToken, err)
+	}
+
+	var payload tokenPayload
+	if err := json.Unmarshal(decodedPayload, &payload); err != nil {
+		return tokenPayload{}, errInvalidTokenFormat
+	}
+
+	return payload, nil
+}
```
dborovcanin: Let's move this to a separate file.

users/users.go
```diff
@@ -29,6 +29,19 @@ type User struct {
...
+
+	// VerificationToken is sent to the user's email and should not be exposed via API.
+	VerificationToken          string    `json:"-"`
+	VerificationTokenExpiresAt time.Time `json:"-"`
+}
+
+func (u *User) IsVerificationTokenExpired() bool {
+	if u.VerificationTokenExpiresAt.IsZero() {
+		return true
+	}
```
dborovcanin: Should we check if verificationToken is empty first?

felixgateru (changes requested):

users/users.go
```diff
@@ -81,6 +92,12 @@ type Repository interface {
 	// RetrieveByUsername retrieves user by its unique credentials.
 	RetrieveByUsername(ctx context.Context, username string) (User, error)
 
+	// RetrieveVerificationToken retrieves verification token of given user id.
+	RetrieveVerificationToken(ctx context.Context, id string) (User, error)
+
+	// UpdateToUserVerificationDetails update given user id verification details.
+	UpdateToUserVerificationDetails(ctx context.Context, user User) (User, error)
+
```
dborovcanin: Update instead of UpdateTo.

users/users.go
```diff
@@ -90,6 +107,10 @@ type Repository interface {
 	// UpdateSecret updates secret for user with given email.
 	UpdateSecret(ctx context.Context, user User) (User, error)
 
+	UpdateEmail(ctx context.Context, user User) (User, error)
+
+	UpdateRole(ctx context.Context, user User) (User, error)
```
dborovcanin, line 96: Add godoc comments.

users/users.go
```diff
@@ -152,6 +174,10 @@ type Service interface {
 	// non-nil error value is returned.
 	Register(ctx context.Context, session authn.Session, user User, selfRegister bool) (User, error)
 
+	SendVerification(ctx context.Context, session authn.Session) error
+
+	VerifyEmail(ctx context.Context, verificationToken string) (User, error)
```
dborovcanin: Same, add godoc.

users/service_test.go
```diff
@@ -776,13 +776,13 @@ func TestUpdateRole(t *testing.T) {
 		repoCall := cRepo.On("CheckSuperAdmin", context.Background(), mock.Anything).Return(tc.checkSuperAdminErr)
 		policyCall := policies.On("AddPolicy", context.Background(), mock.Anything).Return(tc.addPolicyErr)
 		policyCall1 := policies.On("DeletePolicyFilter", context.Background(), mock.Anything).Return(tc.deletePolicyErr)
-		repoCall1 := cRepo.On("Update", context.Background(), mock.Anything, mock.Anything).Return(tc.updateRoleResponse, tc.updateRoleErr)
+		repoCall1 := cRepo.On("UpdateRole", context.Background(), mock.Anything).Return(tc.updateRoleResponse, tc.updateRoleErr)
```
dborovcanin, line 780: We are missing tests for the added methods. Either do it now or open a ticket for the follow-up PR, since this one is already large.

dborovcanin (changes requested):

users/postgres/init.go
```diff
@@ -97,6 +97,23 @@ func Migration() *migrate.MemoryMigrationSource {
 					`ALTER TABLE users ALTER COLUMN updated_at TYPE TIMESTAMP;`,
 				},
 			},
+			{
+				Id: "clients_07",
+				Up: []string{
+					`ALTER TABLE users
+						ADD COLUMN verified BOOLEAN DEFAULT FALSE,
```
dborovcanin: Do we need this redundancy? Why don't we just use verified_at as nullable?

users/api/users.go
```diff
@@ -47,23 +49,45 @@ func usersHandler(svc users.Service, authn smqauthn.Authentication, tokenClient
...
+				api.EncodeResponse,
+				opts...,
+			), "refresh_token").ServeHTTP)
+			r.Patch("/{id}/email", otelhttp.NewHandler(kithttp.NewServer(
+				updateEmailEndpoint(svc),
+				decodeUpdateUserEmail,
+				api.EncodeResponse,
+				opts...,
+			), "update_user_email").ServeHTTP)
+		})
```
dborovcanin, line 87: Why is this allowed for unverified user?

dborovcanin (changes requested):

users/api/users.go
arvindh123: 1. Change Email Endpoint:
If a user enters the wrong email address during signup, they must have the ability to correct it. Allowing unverified users to change their email ensures they can complete the verification process successfully.

2. Refresh Token Endpoint:    
The refresh token enables users to request new access tokens regardless of their verification status. Both access and refresh tokens include a verified field. Before verification, this field is false. Once the user completes verification and refreshes their token, the newly issued tokens will correctly reflect verified = true

users/postgres/init.go
arvindh123: Yes. We use a `verified` (boolean) field in JWTs and for other condition checks. Relying only on `verified_at` would make the logic less clear.

For example, if we only had `verified_at`, we’d need to write something like:

```go
verified: !verified_at.IsZero()
```

By keeping a `verified` field, the code is cleaner and easier to understand.

felixgateru (commented):
image

users/users.go
arvindh123: I will think about this, What we should return if token is empty ?

users/api/users.go
dborovcanin: > 1. Change Email Endpoint:
>    If a user enters the wrong email address during signup, they must have the ability to correct it. Allowing unverified users to change their email ensures they can complete the verification process successfully.

No need -> they can restart the procedure process with the correct email.
> 2. Refresh Token Endpoint:
>    The refresh token enables users to request new access tokens regardless of their verification status. Both access and refresh tokens include a verified field. Before verification, this field is false. Once the user completes verification and refreshes their token, the newly issued tokens will correctly reflect verified = true

Why? Isn't it simpler to disable issuing tokens to who don't have `VerifiedAt` field?

arvindh123: > No need -> they can restart the procedure process with the correct email.
- A username cannot be reused until the associated account is deleted.
- This ensures usernames can be properly recycled.
- It helps prevent the buildup of unwanted or unverified user accounts.

arvindh123: We can provide some basic operation without verification of email 
This provides flexibility.

During this basic operation time, If user want to reset password, then they need to verify the email.

dborovcanin:
CI is failing.

users/postgres/init.go
```diff
@@ -97,6 +97,21 @@ func Migration() *migrate.MemoryMigrationSource {
...
+				Up: []string{
+					`ALTER TABLE users
+						ADD COLUMN verified_at TIMESTAMPTZ DEFAULT NULL,
+						ADD COLUMN verification_token TEXT,
+						ADD COLUMN verification_token_expires_at TIMESTAMPTZ;`,
+				},
+				Down: []string{
+					`ALTER TABLE users
+						DROP COLUMN verified_at,
+						DROP COLUMN verification_token,
```
dborovcanin: Is the token needed at all? Can we put user ID in JWT we send for verification?

users/postgres/init.go
```diff
@@ -97,6 +97,21 @@ func Migration() *migrate.MemoryMigrationSource {
...
+						DROP COLUMN verification_token_expires_at;`,
```
dborovcanin: Is this needed if we are going to use JWT that already contains the expiration?

dborovcanin (changes requested):

felixgateru (dismissed):

users/postgres/init.go
```diff
@@ -97,6 +97,27 @@ func Migration() *migrate.MemoryMigrationSource {
...
+			{
+				Id: "clients_07",
+				Up: []string{
+					`ALTER TABLE users ADD COLUMN verified_at TIMESTAMPTZ DEFAULT NULL;`,
+					`CREATE TABLE users_verifications (
+						user_id VARCHAR(36) NOT NULL,
+						email VARCHAR(254) NOT NULL,
+						otp VARCHAR(255),
+						created_at TIMESTAMPTZ,
+						expiry_at TIMESTAMPTZ,
```
dborovcanin: Let's use `expires_at` since we used `created_at`.

users/service.go
```diff
@@ -92,6 +93,85 @@ func (svc service) Register(ctx context.Context, session authn.Session, u User,
...
+	return nil
+}
+
+func (svc service) VerifyEmail(ctx context.Context, ruvs string) (User, error) {
+	var ruv UserVerification
+	if err := ruv.Decode(ruvs); err != nil {
+		return User{}, errors.Wrap(svcerr.ErrMalformedEntity, err)
+	}
+
+	oguv, err := svc.users.RetrieveUserVerification(ctx, ruv.UserID, ruv.Email)
```
dborovcanin: Oguv?

users/service.go
```diff
@@ -255,14 +335,23 @@ func (svc service) UpdateEmail(ctx context.Context, session authn.Session, userI
...
+	usr := User{
+		ID:         userID,
+		Email:      email,
+		UpdatedAt:  time.Now().UTC(),
+		UpdatedBy:  session.UserID,
+		VerifiedAt: time.Time{},
 	}
-	user, err := svc.users.Update(ctx, userID, usr)
+
+	user, err := svc.users.UpdateEmail(ctx, usr)
```
dborovcanin, line 354: Should we send a verification email?

users/postgres/users_verfications.go
```diff
@@ -0,0 +1,131 @@
+// Copyright (c) Abstract Machines
```
dborovcanin: Let's rename to `verifications`, no need for `users` prefix.

dborovcanin (changes requested):

users/service.go
arvindh123: `oguv` - Original (OG) User Verification

users/service.go
arvindh123: No we can send email on demand when user requests
We can have verification button in UI  next to email, After successfull change of email
Or
For better user experience, This can be done in UI, `After successful change of email , UI can send send verification request to backend`

dborovcanin (changes requested):
CI is failing.

smithjilks (changes requested):
@arvindh123 
I get no token/invalid token when I call sdk verify email method with the token gotten from link query params.

users/service.go
dborovcanin: Both `ruvs` and `oguv` are not descriptive names. Please update. Rename `ruvs` to `user`; and `user` to `updatedUser` or just `updated`. Rename `ruvs` to `token`, and `ruv := UserVerification` to `uv := userVerification`. Especially since this does not have to be _registration_ (`r`) related.

felixgateru (dismissed):

smithjilks (changes requested):
Image
You can update the template so that it says welcome to 'host' instead of 'host/endpoint'

smithjilks (dismissed):

groups/api/http/transport.go
```diff
@@ -19,14 +19,14 @@ import (
 )
 
 // MakeHandler returns a HTTP handler for Groups API endpoints.
-func MakeHandler(svc groups.Service, authn authn.Authentication, mux *chi.Mux, logger *slog.Logger, instanceID string, idp supermq.IDProvider) *chi.Mux {
+func MakeHandler(svc groups.Service, authnMW authn.AuthNMiddleware, mux *chi.Mux, logger *slog.Logger, instanceID string, idp supermq.IDProvider) *chi.Mux {
```
dborovcanin: Let's just use `authn` for this variable.

users/service.go
```diff
@@ -92,6 +93,85 @@ func (svc service) Register(ctx context.Context, session authn.Session, u User,
...
+	}
+
+	uvs, err := uv.Encode()
+	if err != nil {
+		return errors.Wrap(svcerr.ErrCreateEntity, err)
+	}
+
+	if err := svc.email.SendVerification([]string{dbUser.Email}, dbUser.Credentials.Username, uvs); err != nil {
+		return errors.Wrap(svcerr.ErrCreateEntity, err)
+	}
```
dborovcanin, line 128: Does it make more sense to send an email and then save in the DB (i.e. it's the same for us, but it's more likely that connection to email provider that's not in our control will fail, than connection to the DB in our control).

dborovcanin (changes requested):

users/service.go
arvindh123: @dborovcanin 

We’re currently following this flow for email OTP:

1. **Generate OTP**
2. **Save OTP in DB**
3. **Send Email**

**Reasoning:**

* By saving OTP **before** sending the email, we make sure the OTP is always persisted and retrievable.
* If sending the email fails (provider down, user didn’t receive it, etc.), the user can re-trigger verification. In that case, we simply:

  * Retrieve the already stored OTP from DB
  * Check if it’s still valid (expiry)
  * Resend the email

This way, even if the email provider (which is outside our control) fails, the OTP isn’t lost, and the flow remains consistent.

If we did it the other way (send email first, then save to DB), we’d risk a situation where the user receives an OTP but it never got stored in DB due to a failure — which would break verification.

So saving first is safer because DB is within our control, and it guarantees consistency in the flow.

users/service.go
```diff
@@ -129,18 +129,18 @@ func (svc service) SendVerification(ctx context.Context, session authn.Session)
 	return nil
 }
 
-func (svc service) VerifyEmail(ctx context.Context, ruvs string) (User, error) {
-	var ruv UserVerification
-	if err := ruv.Decode(ruvs); err != nil {
+func (svc service) VerifyEmail(ctx context.Context, token string) (User, error) {
+	var uv UserVerification
```
dborovcanin: Let's use `received` and `stored` instead, so we have `stored.Match(received)`.

dborovcanin (approved):

PR #3146, opened 2025-09-17
SMQ-3028 - Unify SMQ certs with Absmach Certs

### LINKED ISSUE #3028: Unify SMQ certs with Absmach Certs

nyagamunene:
# What type of PR is this?

- Related Issue #3028 
- Resolves #3028


docker/addons/certs/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+RUN apk update \
+    && apk add make upx\
+    && make $SVC \
+    && upx build/$SVC \
+    && mv build/$SVC /exe
+
+FROM scratch
+
+# Required for billing service
```
felixgateru, line 25:
```suggestion
# Required for certs service
```

docker/addons/certs/Dockerfile.dev
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Abstract Machines
+# SPDX-License-Identifier: Apache-2.0
+
+FROM scratch
+ARG SVC
+COPY --from=alpine:latest /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY ./build/$SVC /exe
+ENTRYPOINT ["/exe"]
```
felixgateru, line 8: Add empty line

docker/addons/certs/docker-compose.yaml
```diff
@@ -1,134 +1,115 @@
...
   certs:
-    image: supermq/certs:${SMQ_RELEASE_TAG}
-    container_name: supermq-certs
+    image: ghcr.io/absmach/certs:${AM_CERTS_RELEASE_TAG}
+    container_name: certs
     depends_on:
       - openbao
       - certs-db
-    restart: on-failure
+    restart: always
```
felixgateru, line 21: Why restart always here?

docker/certs-docker-compose-override.yml
felixgateru, line 1: remove if unused

pkg/sdk/clients_test.go
```diff
@@ -30,6 +30,8 @@ import (
 	"github.com/stretchr/testify/mock"
 )
 
+var clientID = "fe6b4e92-cc98-425e-b0aa-000000000001"
```
felixgateru, line 33: is this used?

felixgateru (changes requested):

scripts/certs.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
arvindh123, line 1: Why do we need to use the same docker compose?

arvindh123 (changes requested):

pkg/sdk/clients_test.go
nyagamunene: It is used in multiple places in the code

scripts/certs.sh
dborovcanin: I agree, this is probably not needed. We can just put a simple compose file that uses absmach/certs to `docker/addons`.

nyagamunene: What about when we update certs and we need the changes reflected on supermq will it not be alot of work making the changes on superMQ side?

dborovcanin: I don't think so. SMQ is only using Compose from Certs. However, now that we're working on integrations service, I'm thinking that Certs service will become even more important and I'm afraid we can't avoid coupling between Certs and core service. Anyways, let's keep this as is, and we'll revisit the approach during this sprint as we progress with integrations.

dborovcanin (approved):

PR #3155, opened 2025-09-25
SMQ-3108 - Add support for public and private metadata for users and clients

### LINKED ISSUE #3108: Add public metadata to Users and Clients
Add public metadata to Users and Clients so it can be shared with other users safely.

We should have both public and private metadata.

1. **Public metadata** can contains non-sensitive user information that can be safely exposed to other users. Examples include a profile photo, language preferences, or theme settings.
2. **Private metadata** should sensitive or personally identifiable information that must be protected and only accessible internally. Examples include a user’s address, phone number, etc.

Separating public and private metadata is important because:

- prevents sensitive user information from being unintentionally exposed and reduces the risk of data leaks.

-  allows safe sharing of public data while keeping private data restricted, which simplifies user access control and strengthens user trust.

 _Originally posted by @dborovcanin in [#963](https://github.com/absmach/magistrala-ui/issues/963#issuecomment-3242789194)_

felixgateru:
# What type of PR is this?
This is a feature as it adds support for private and public metadata for Users and Clients

## What does this do?
This pr:
- adds a database column for public metadata and subsequent changes to Users and Clients structs

- Related Issue #3108
- Resolves #3108

Yes, tests have been updated.

Yes, api docs have been updated for the changes


dborovcanin (changes requested):
To keep it backward compatible and not break API, let's consider our Metadata private, and just add public one. I.e. we have `metadata` and `publicMetadata` fields. Private ones should not be visible even to superadmin, probably, but for now - let's just add public meta.

dborovcanin:
@felixgateru Please resolve conflicts.

users/events/events.go
```diff
@@ -89,8 +89,8 @@ func (uce createUserEvent) Encode() (map[string]any, error) {
 	if len(uce.Tags) > 0 {
 		val["tags"] = uce.Tags
 	}
-	if uce.Metadata != nil {
-		val["metadata"] = uce.Metadata
+	if uce.PublicMetadata != nil {
+		val["public_metadata"] = uce.PublicMetadata
```
arvindh123, line 96: Why we are not emitting metadata ?

arvindh123 (changes requested):

pkg/clients/events/consumer/decode.go
```diff
@@ -90,9 +90,9 @@ func ToClient(data map[string]any) (clients.Client, error) {
 		c.Tags = tags
 	}
 
-	meta, ok := data["metadata"].(map[string]any)
+	meta, ok := data["public_metadata"].(map[string]any)
 	if ok {
-		c.Metadata = meta
+		c.PublicMetadata = meta
 	}
```
arvindh123: Missing `metadata` in decoding

arvindh123 (changes requested):

arvindh123 (dismissed):

cli/clients_test.go
```diff
@@ -305,6 +305,26 @@ func TestUpdateClientCmd(t *testing.T) {
 		client        smqsdk.Client
 		logType       outputLog
 	}{
+		{
+			desc: "update client name and public metadata successfully",
```
dborovcanin, line 313: Add a test case for updating public metadata only (also add unhappy flow for it).

users/postgres/users.go
```diff
@@ -178,7 +178,7 @@ func (repo *userRepo) RetrieveAll(ctx context.Context, pm users.Page) (users.Use
 func (repo *userRepo) UpdateUsername(ctx context.Context, user users.User) (users.User, error) {
 	q := `UPDATE users SET username = :username, updated_at = :updated_at, updated_by = :updated_by
         WHERE id = :id AND status = :status
-		RETURNING id, tags, metadata, status, created_at, updated_at, updated_by, first_name, last_name, username, email, role, verified_at`
+		RETURNING id, tags, public_metadata, status, created_at, updated_at, updated_by, first_name, last_name, username, email, role, verified_at`
```
dborovcanin: Why are not we returning metadata also?

users/service_test.go
```diff
@@ -546,6 +551,17 @@ func TestUpdateUser(t *testing.T) {
 			token:            validToken,
 			err:              nil,
 		},
+		{
+			desc:   "update public metadata successfully as normal user",
```
dborovcanin, line 556: Add unhappy flow also (invalid format).

dborovcanin (changes requested):

users/postgres/users.go
felixgateru: I initially treated metadata as private to the user and client and only returned on save or on retrieval. I have updated to include on update but it is not returned on listing entities.

PR #3162, opened 2025-10-02
NOISSUE - Auto verify oauth user during OAuth Callback

arvindh123:
# What type of PR is this?
This PR does the following 
- Add verified status to newly signing user via OAuth
- If an already unverified user logins via OAuth provider, then automatically make user as verified

- Related Issue #
- Resolves #


dborovcanin:
@arvindh123 Please fix tests.

Issue #1744, opened 2023-03-14
Improve error handling


AryanGodara:
1. Is there an open issue addressing this request? If it does, please add a "+1" reaction to the
   existing issue, otherwise proceed to step 2.
#1723 Continuation of this issue
2. Describe the feature you are requesting and the possible use case(s) for it.
The error handling for Magistrala has become very complex, and there are a lot of repeated errors, so it needs to be made simpler and more readable.
Related issues: #2059, #1964, #2260, #1683.

3. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).
should have

arvindh123:
## There are two approach 
- Using Existing Error [SMQ-1744 - Error handling with TypedError created on top existing Error](https://github.com/absmach/supermq/pull/3170)
- New error [SMQ-1744 - Add new error and new error types #3169](https://github.com/absmach/supermq/pull/3169)

arvindh123:
This can be closed on behalf of PR https://github.com/absmach/supermq/pull/3170

PR #3170, opened 2025-10-06
SMQ-1744 - Error handling with TypedError created on top existing Error 

arvindh123:
# What type of PR is this?

- Related Issue #1744 
- Resolves #


users/postgres/init.go
```diff
@@ -118,6 +118,15 @@ func Migration() *migrate.MemoryMigrationSource {
...
 				},
 			},
+			{
+				Id: "clients_08",
+				Up: []string{
+					`ALTER TABLE users RENAME CONSTRAINT clients_identity_key TO clients_email_key;`,
+				},
+				Down: []string{
+					`ALTER TABLE users RENAME CONSTRAINT clients_email_key TO clients_identity_key;`,
+				},
```
dborovcanin, line 128: Does not belong to this PR.

pkg/errors/errors.go
```diff
@@ -75,30 +67,23 @@ func Contains(e1, e2 error) bool {
...
 		return wrapper
 	}
-	if w, ok := wrapper.(Error); ok {
-		return &customError{
-			msg: w.Msg(),
-			err: cast(err),
-		}
+
+	if ne, ok := err.(TypedError); ok {
+		return ne.Wrap(wrapper)
```
arvindh123:
```suggestion
func Wrap(wrapper, err error) error {
	if wrapper == nil || err == nil {
		return wrapper
	}
	if ne, ok := err.(TypedError); ok {
		return ne.Wrap(wrapper)
	}

	if w, ok := wrapper.(Error); ok {
		return &customError{
			msg: w.Msg(),
			err: cast(err),
		}
	}
	return &customError{
		msg: wrapper.Error(),
		err: cast(err),
	}
}
```

pkg/errors/errortypes.go
```diff
@@ -0,0 +1,200 @@
...
+var _ TypedError = (*RequestError)(nil)
+
+type RequestError struct {
+	customError
+}
+
+func wrap(wrapper, err error) Error {
+	if wrapper == nil || err == nil {
+		return &customError{
+			msg: wrapper.Error(),
```
arvindh123: Slop code

users/postgres/errors.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"github.com/absmach/supermq/pkg/errors"
+)
+
+var _ errors.Mapper = (*duplicateErrors)(nil)
+
+type duplicateErrors struct{}
+
+func (d duplicateErrors) GetError(key string) (error, bool) {
```
arvindh123: rename the `key` to `constraint` , please write comment documentation here

users/service.go
```diff
@@ -97,72 +120,72 @@ func (svc service) Register(ctx context.Context, session authn.Session, u User,
 	}()
 	user, err := svc.users.Save(ctx, u)
 	if err != nil {
-		return User{}, errors.Wrap(svcerr.ErrCreateEntity, err)
+		return User{}, svcerr.ErrFailedToSaveEntityDB.Embed(err)
```
arvindh123: The original error orginate in repo layer , that was intentend to  expose in publish will be hidden by the service layers
Because service layer users embed, so the original public error which are orignated in repo layer will not  exposed


The  `err`  can have the following
- `errors.NewRequestError("email id already registered")`
- `errors.NewRequestError("username not available")`
These are intendented to expose to public , in API response

pkg/errors/service/types.go
```diff
@@ -8,37 +8,62 @@ import "github.com/absmach/supermq/pkg/errors"
...
-	// ErrLogin indicates wrong login credentials.
-	ErrLogin = errors.New("invalid credentials")
-
-	// ErrMalformedEntity indicates a malformed entity specification.
-	ErrMalformedEntity = errors.New("malformed entity specification")
+	ErrDomainAuthorization = errors.NewAuthZError("failed to perform authorization over the domain")
 
-	// ErrNotFound indicates a non-existent entity request.
-	ErrNotFound = errors.New("entity not found")
+	// ErrUnauthorizedPAT indicates failure occurred while authorizing PAT.
```
arvindh123, line 22: I think now, we can move service and repo errors to  common errors , Since we have typed errors now

api/http/util/errors.go
```diff
@@ -19,259 +19,262 @@ var (
...
 	// ErrInvalidVerification indicates invalid email verification.
-	ErrInvalidVerification = errors.New("invalid verification")
+	ErrInvalidVerification = errors.NewRequestError("invalid verification")
 
 	// ErrEmailNotVerified indicates invalid email not verified.
-	ErrEmailNotVerified = errors.New("email not verified")
+	ErrEmailNotVerified = errors.NewRequestError("email not verified")
+
+	// ErrMalformedRequest indicates malformed request body.
+	ErrMalformedRequestBody = errors.NewRequestError("malformed request body")
```
arvindh123: we can return something like `Request body is not a valid JSON, Expecting a valid JSON` , Only for all the JSON parsed failed errors

groups/postgres/errors.go
```diff
@@ -0,0 +1,24 @@
...
+
+var _ errors.Mapper = (*duplicateErrors)(nil)
+
+type duplicateErrors struct{}
+
+// GetError maps constraint names to known errors.
+func (d duplicateErrors) GetError(constraint string) (error, bool) {
+	switch constraint {
+	case "groups_pkey":
+		return errors.NewRequestError("group already exists"), true
```
arvindh123: The Group ID or the Group Name is already exists ?
Please add more details to this error,

groups/postgres/groups.go
arvindh123: Please return a typed error here, This should be returned to user

arvindh123 (commented):
Now that we have a mechanism to return the correct error in API responses, this PR may still not cover all API scenarios.

I suggest we incrementally fix this by returning the correct error on the fly whenever we identify cases where the proper error is not being propagated back to the API response.

api/http/util/errors.go
```diff
@@ -19,259 +19,262 @@ var (
...
 	// ErrInvalidVerification indicates invalid email verification.
-	ErrInvalidVerification = errors.New("invalid verification")
+	ErrInvalidVerification = errors.NewRequestError("invalid verification")
 
 	// ErrEmailNotVerified indicates invalid email not verified.
-	ErrEmailNotVerified = errors.New("email not verified")
+	ErrEmailNotVerified = errors.NewRequestError("email not verified")
+
+	// ErrMalformedRequest indicates malformed request body.
+	ErrMalformedRequestBody = errors.NewRequestError("Request body is not a valid JSON, Expecting a valid JSON")
```
arvindh123: please make them in smaller case ,just want to align with other errors

arvindh123 (commented):
LGTM

pkg/errors/error_types.go
```diff
@@ -0,0 +1,206 @@
...
+
+func NewInternalErrorWithErr(err error) NestError {
+	return &InternalError{
+		customError: customError{
+			msg: "internal server error",
+			err: cast(err),
+		},
+	}
+}
+
```
dborovcanin, line 145: Since we have attached `Embed` to `customError` - it masks here that InternalError **does not implement Embed**, so calling it on this type will return **customError** instead.

dborovcanin: Example:
```go
  err := errors.NewInternalErrorWithErr(errors.New("test"))
  embedded := err.(errors.NestError).Embed(errors.New("context"))
  // Type is *errors.customError, not *errors.InternalError
 ```

File: pkg/errors/errortypes.go [hunk 11]
```diff
@@ -0,0 +1,206 @@
...
+			err: errors.New(message),
+		},
+	}
+}
+
+func NewServiceErrorWithErr(message string, err error) NestError {
+	return &ServiceError{
+		customError: customError{
+			msg: message,
+			err: cast(err),
```
dborovcanin: Somewhere we use `cast(err)`, and somewhere `err` directly (see L48, L79). This is borderline dangerous, please stay consistent.

pkg/errors/service/types.go
```diff
@@ -8,101 +8,116 @@ import "github.com/absmach/supermq/pkg/errors"
...
-
-	// ErrUserAlreadyVerified indicates user is already verified.
-	ErrUserAlreadyVerified = errors.New("user already verified")
-
-	// ErrInvalidUserVerification indicates user verification is invalid.
-	ErrInvalidUserVerification = errors.New("invalid verification")
-
 	// ErrUserVerificationExpired indicates user verification is expired.
-	ErrUserVerificationExpired = errors.New("verification expired, please generate new verification")
+	ErrUserVerificationExpired = errors.New("verification expired, please generate New verification")
```
dborovcanin: Fix typo: `New` -> `new`.

pkg/errors/errors_test.go
```diff
@@ -276,66 +276,6 @@ func TestWrap(t *testing.T) {
 	}
 }
 
-func TestUnwrap(t *testing.T) {
```
dborovcanin, line 279: Why is this test removed?

api/http/common.go
```diff
@@ -184,146 +183,39 @@ func EncodeError(_ context.Context, err error, w http.ResponseWriter) {
...
-		errors.Contains(err, svcerr.ErrRemoveEntity),
-		errors.Contains(err, svcerr.ErrEnableClient),
-		errors.Contains(err, svcerr.ErrEnableUser),
-		errors.Contains(err, svcerr.ErrDisableUser):
-		err = unwrap(err)
-		w.WriteHeader(http.StatusUnprocessableEntity)
-
-	case errors.Contains(err, svcerr.ErrNotFound):
-		err = unwrap(err)
-		w.WriteHeader(http.StatusNotFound)
```
dborovcanin, line 295: Did we lose not found case completely? That does not sound right. We **must** keep backward compatibility when it comes to status codes.

pkg/errors/service/types.go
```diff
@@ -8,101 +8,116 @@ import "github.com/absmach/supermq/pkg/errors"
...
+	// ErrInvalidUserVerification indicates user verification is invalid.
+	ErrInvalidUserVerification = errors.NewServiceError("invalid verification")
+
+	// ErrIssueProviderID indicates failure to issue unique ID from ID provider.
+	ErrIssueProviderID = errors.NewServiceError("failed to issue unique ID from id provider")
+
+	// ErrHashPassword indicates failure to hash password.
+	ErrHashPassword = errors.NewServiceError("failed to hash password")
+
+	// ErrStatusAlreadyAssigned indicated that the client or group has already been assigned the status.
```
dborovcanin: Indicated? :)

pkg/errors/error_types.go
```diff
@@ -0,0 +1,206 @@
+// Copyright (c) Abstract Machines
```
dborovcanin, line 1: Use `error_types.go`; do the same for the other files with multiple words ( such as `errorhandler.go` and so on).

dborovcanin (changes requested):

pkg/errors/errors_test.go
felixgateru: I have removed the unwrap function

dborovcanin (approved):

PR #3178, opened 2025-10-08
NOISSUE - Auto generation username for oauth user during singup 

arvindh123:
# What type of PR is this?

- Related Issue #
- Resolves #

### DIFF
--- a/users/service.go
+++ b/users/service.go
@@ -8,6 +8,7 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"fmt"
+	mrand "math/rand"
 	"net/mail"
 	"regexp"
 	"strings"
@@ -23,12 +24,17 @@ import (
 	repoerr "github.com/absmach/supermq/pkg/errors/repository"
 	svcerr "github.com/absmach/supermq/pkg/errors/service"
 	"github.com/absmach/supermq/pkg/policies"
+	"github.com/gofrs/uuid/v5"
 )
+
+const defaultUsernamePrefix = "user"
 
 var (
 	errIssueToken       = errors.New("failed to issue token")
 	errRecoveryToken    = errors.New("failed to generate password recovery token")
 	errLoginDisableUser = errors.New("failed to login in disabled user")
+
+	usernameRegExp = regexp.MustCompile(`^[a-z0-9][a-z0-9_-]{34}[a-z0-9]$`)
 )
 
 type service struct {
@@ -689,12 +695,11 @@ func generateUsername(email string) string {
 func getEmailPrefix(email string) string {
 	parts := strings.Split(email, "@")
 	if len(parts) == 0 {
-		return "user"
+		return defaultUsernamePrefix
 	}
 
 	prefix := parts[0]
-	reg := regexp.MustCompile("[^a-zA-Z0-9]+")
-	cleaned := reg.ReplaceAllString(prefix, "")
+	cleaned := usernameRegExp.ReplaceAllString(prefix, "")
 	cleaned = strings.ToLower(cleaned)
 
 	if len(cleaned) > 15 {
@@ -703,7 +708,7 @@ func getEmailPrefix(email string) string {
 
 	cleaned = removeSpecialChars(cleaned)
 	if cleaned == "" {
-		cleaned = "user"
+		cleaned = defaultUsernamePrefix
 	}
 
 	return cleaned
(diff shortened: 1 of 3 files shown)


users/service.go
```diff
@@ -673,3 +679,51 @@ func (svc service) updateUserPolicy(ctx context.Context, userID string, role Rol
...
 	}
 }
+
+func generateUsername(email string) string {
+	uniqueSuffix := generateRandomID()
+	emailPrefix := getEmailPrefix(email)
+	return fmt.Sprintf("%s_%s", emailPrefix, uniqueSuffix)
+}
+
+func getEmailPrefix(email string) string {
```
dborovcanin: Rename to `extractEmailPrefix`. Also, extract the email in a more robust way, since valid mails can contain special characters such as `+` or `"`. You can even use `mail.ParseAddress`. In case you stick to regex, move it to service field and pre-compile during service instantiation; there is no need to create a new regex for each call.

dborovcanin (changes requested):

SammyOina (dismissed):

users/service.go
```diff
@@ -555,8 +566,12 @@ func (svc service) OAuthCallback(ctx context.Context, user User) (User, error) {
 	u, err := svc.users.RetrieveByEmail(ctx, user.Email)
 
 	if errors.Contains(err, repoerr.ErrNotFound) {
+		user.Credentials.Username = generateUsername(user.Email)
 		u, err = svc.Register(ctx, authn.Session{}, user, true)
 		if err != nil {
+			if errors.Contains(err, errors.ErrUsernameNotAvailable) {
+				return User{}, errors.ErrTryAgain
+			}
```
dborovcanin, line 573: This will result in losing the proper error in logs. TryAgain should be returned by the API layer, but service layer should not mask the actual error or we will be blind when such an error happens.

users/service.go
```diff
@@ -673,3 +688,57 @@ func (svc service) updateUserPolicy(ctx context.Context, userID string, role Rol
...
+	id, err := uuid.NewV4()
+	if err == nil {
+		randomBytes = append(randomBytes, id.Bytes()...)
+	}
+
+	mrand.Shuffle(len(randomBytes), func(i, j int) {
+		randomBytes[i], randomBytes[j] = randomBytes[j], randomBytes[i]
+	})
+	return hex.EncodeToString(randomBytes)[:10]
+}
```
dborovcanin, line 728: Why don't we either use hash of email or UUID instead of random bytes?

users/service.go
```diff
@@ -673,3 +688,57 @@ func (svc service) updateUserPolicy(ctx context.Context, userID string, role Rol
...
+
+// removeSpecialChars removes non-ASCII and special characters.
+func removeSpecialChars(s string) string {
+	return strings.Map(func(r rune) rune {
+		if unicode.IsLetter(r) || unicode.IsNumber(r) {
+			return r
+		}
+		return -1
+	}, s)
+}
```
dborovcanin: Can we stick to ASCII and do something like this:
```go
func removeSpecialChars(s string) string {
	buf := make([]byte, 0, len(s))
	for i := 0; i < len(s); i++ {
		c := s[i]
		if (c >= 'A' && c <= 'Z') ||
			(c >= 'a' && c <= 'z') ||
			(c >= '0' && c <= '9') {
			buf = append(buf, c)
		}
	}
	return string(buf)
}
```

dborovcanin (changes requested):

users/service.go
arvindh123: Anyone who knows this  logic can able to create Username and make a conflict while actual user signup via oauth 
 So only adding random bytes 
 May be instead of UUID we can have (hashed email + random bytes) and pick random 10 from this byte array

users/service.go
```diff
@@ -733,12 +727,51 @@ func generateRandomID() string {
...
+
+	for i := 0; i < len(s); i++ {
+		c := s[i]
+
+		// Keep alphanumeric, hyphen, underscore
+		if (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') || c == '-' || c == '_' {
+			// Skip if current char is hyphen/underscore and same as last char
+			if (c == '-' || c == '_') && c == lastChar {
+				continue // Skip consecutive hyphens or underscores
+			}
```
dborovcanin, line 756: Can we extract from nested if? Also, remove `|| c == '-' || c == '_'` from the other if in that case.

dborovcanin (approved):

PR #3191, opened 2025-10-16
NOISSUE - Use structured requests in callouts

dborovcanin:
# What type of PR is this?

## What does this do?
This updates callout requests to use structure instead of any for at least some fields. Those fields are supposed to be common for majority of the requests:
- operation
- object type
- subject id
- subject type

There is no such issue.

Yes.

N/A
### Notes
Please note that there is no `object type` in the common fields. That's because issuing a call, we usually want to **do** (operation) and need to auth{N,Z} (subject id and type), on something (object type) but sometimes we don't have the exact object (it can be create operation, or listing operation, say). When needed, it can be added as a payload.

### DIFF
--- a/pkg/callout/callout.go
+++ b/pkg/callout/callout.go
@@ -28,7 +28,7 @@ type Request struct {
 	Payload     map[string]any `json:"payload,omitempty"`
 }
 
-// Callout send request to an external service.
+// Callout send a request to an external service.
 type Callout interface {
 	Callout(ctx context.Context, req Request) error
 }
@@ -144,7 +144,9 @@ func (c *callout) makeRequest(ctx context.Context, urlStr string, req Request) e
 			return jsonErr
 		}
 		r, err = http.NewRequestWithContext(ctx, c.method, urlStr, bytes.NewReader(data))
-		r.Header.Set("Content-Type", "application/json")
+		if err == nil {
+			r.Header.Set("Content-Type", "application/json")
+		}
 	}
 
 	if err != nil {
--- a/pkg/callout/callout.go
+++ b/pkg/callout/callout.go
@@ -80,14 +80,12 @@ func (c *callout) Callout(ctx context.Context, req Request) error {
 		return nil
 	}
 
-	// Check if the operation is in the allowed list
-	// Otherwise, only call webhook if the operation is in the map
 	if _, exists := c.allowedOperation[req.Operation]; !exists {
 		return nil
 	}
 
-	// We iterate through all URLs in sequence
-	// if any request fails, we return the error immediately
+	// Make requests sequentially as they appear in the URL
+	// slice and fail fast as soon as any request fails.
 	for _, url := range c.urls {
 		if err := c.makeRequest(ctx, url, req); err != nil {
 			return err
@@ -135,7 +133,7 @@ func (c *callout) makeRequest(ctx context.Context, urlStr string, req Request) e
 		query.Set("operation", req.Operation)
 		query.Set("subject_id", req.SubjectID)
 		query.Set("subject_type", req.SubjectType)
-		query.Set("entity_type", req.ObjectType)
+		query.Set("object_type", req.ObjectType)
 		for key, value := range req.Payload {
 			query.Set(key, fmt.Sprintf("%v", value))
 		}
--- a/pkg/callout/callout.go
+++ b/pkg/callout/callout.go
@@ -24,7 +24,7 @@ type Request struct {
 	Operation   string         `json:"operation,omitempty"`
 	SubjectID   string         `json:"subject_id,omitempty"`
 	SubjectType string         `json:"subject_type,omitempty"`
-	EntityType  string         `json:"entity_type,omitempty"`
+	ObjectType  string         `json:"object_type,omitempty"`
 	Payload     map[string]any `json:"payload,omitempty"`
 }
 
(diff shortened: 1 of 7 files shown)


pkg/callout/callout.go
```diff
@@ -21,6 +20,19 @@ import (
...
 
+type Request struct {
+	Operation   string         `json:"operation,omitempty"`
+	SubjectID   string         `json:"subject_id,omitempty"`
+	SubjectType string         `json:"subject_type,omitempty"`
+	ObjectType  string         `json:"object_type,omitempty"`
+	Payload     map[string]any `json:"payload,omitempty"`
+}
+
+// Callout send request to an external service.
```
felixgateru:
```suggestion
// Callout sends a request to an external service.
```

pkg/callout/callout.go
```diff
@@ -103,48 +123,35 @@ func newCalloutClient(ctls bool, certPath, keyPath, caPath string, timeout time.
...
-
-		data, jsonErr := json.Marshal(calloutReq)
+		data, jsonErr := json.Marshal(req)
 		if jsonErr != nil {
 			return jsonErr
 		}
-		req, err = http.NewRequestWithContext(ctx, c.method, urlStr, bytes.NewReader(data))
-		req.Header.Set("Content-Type", "application/json")
+		r, err = http.NewRequestWithContext(ctx, c.method, urlStr, bytes.NewReader(data))
+		r.Header.Set("Content-Type", "application/json")
```
felixgateru: Is this not unsafe? If the err is not nil then r may be nil and will lead to a panic

pkg/callout/callout_test.go
```diff
@@ -339,43 +342,25 @@ func TestCallout_Operations(t *testing.T) {
...
-			desc:       "empty operations list calls always",
-			operations: []string{},
-			payload: map[string]any{
-				"entity_type": entityType,
-				"sender":      userID,
-				"domain":      domainID,
-				"time":        time.Now().UTC(),
-				"operation":   operation,
-			},
+			desc:         "empty operations list calls always",
```
felixgateru, line 363:
```suggestion
			desc:         "empty operations list never calls",
```

pkg/roles/rolemanager/middleware/authorization.go
```diff
@@ -500,18 +499,19 @@ func (ram RoleManagerAuthorizationMiddleware) validateMembers(ctx context.Contex
...
-	pl := map[string]any{
-		"entity_type":  ram.entityType,
-		"subject_type": policies.UserType,
-		"subject_id":   session.UserID,
-		"domain":       session.DomainID,
-		"time":         time.Now().UTC(),
-	}
+func (ram RoleManagerAuthorizationMiddleware) callOut(ctx context.Context, session authn.Session, op string, pld map[string]any) error {
+	pld["time"] = time.Now().UTC()
+	pld["domain"] = session.DomainID
```
felixgateru: if these two are always used why not define them in the struct?

felixgateru (changes requested):

arvindh123:
I think there might be a small misunderstanding around the naming here.   

In our case, this **callout** isn’t related to **AuthN/AuthZ** or subject–object actions — it’s just a request to an external service (like **billing**, **notification**, **reporting**, **restriction check**, etc.)    

**AuthN/AuthZ** is already done before callout.  

The fields like `entity_id`, `entity_type`, `subject_id`, and `subject_type` are just context info we pass along with the callout, not part of any authorization model.  

So changing `entity_type` to `object_type` might be a bit misleading here, since there’s no “object” in the AuthZ sense — it’s really just the **entity the callout is about**.   

I’d prefer to keep `entity_type` since it matches our use case better and avoids confusion with **AuthZ** terminology.   

Also, right now the **callout logic** sits inside the **AuthZ middleware**, but it actually runs **after AuthZ is completed**. It would be cleaner to move this into a **separate middleware** so that responsibilities are properly separated.   

One more note — in the **billing service**, the field name was `ObjectType`. I had actually planned to rename it to `EntityType` when I was working on one of the billing PRs and intended to include that change in my last commit, but I forgot to do it.  
I think it would be better to make that update now.

pkg/callout/callout.go
```diff
@@ -21,6 +20,27 @@ import (
...
 var errFailedToRead = errors.New("failed to read callout response body")
 
+// Can be used in the implementation of
+// callout service with structured payload.
+type BaseRequest struct {
+	Operation  string `json:"operation,omitempty"`
+	EntityType string `json:"entity_type,omitempty"`
+	EntityID   string `json:"entity_id,omitempty"`
+	CallerID   string `json:"caller_id,omitempty"`
+	CallerType string `json:"caller_type,omitempty"`
```
arvindh123: Caller can interpreted as the service calling the callout (e.g., domain service calling callout , so here domain service is caller). 

What do you think about rename CallerID and CallerType to something more descriptive like RequesterID / RequesterType ?

pkg/callout/callout.go
```diff
@@ -103,48 +131,37 @@ func newCalloutClient(ctls bool, certPath, keyPath, caPath string, timeout time.
...
 	var err error
 
 	switch c.method {
 	case http.MethodGet:
 		query := url.Values{}
-		for key, value := range params {
+		query.Set("operation", req.Operation)
+		query.Set("subject_id", req.CallerID)
+		query.Set("subject_type", req.CallerType)
+		query.Set("object_type", req.EntityType)
```
felixgateru: entityID and domainID are not added to the query

felixgateru (changes requested):

pkg/callout/callout.go
dborovcanin: Modified to use flattened JSON marshalling and all the available fields (respecting omit tag). If the callout service does not care about some fields - they can ignore.

pkg/callout/callout.go
dborovcanin: I don't think this would help, since `requester` can also be interpreted as a service. The same is true for `client` (imagine, we have client of type user). Only `user` is not ambiguous, but this can be SMQ Client ID, not only user, to it would be misleading. That's why we have  `CallerType`. I.e. type is at the moment only user for our calls, so I think we are good with `caller` notion.

felixgateru (approved):

PR #3210, opened 2025-10-29
NOISSUE - Start both Coap dtls and without dtls servers

felixgateru:
# What type of PR is this?

- Related Issue #
- Resolves #


dborovcanin:
This  has been moved to mGate/mProxy.

PR #3216, opened 2025-10-31
NOISSUE - Send email on invitation

### LINKED ISSUE #3213: Send an email notification when user receives a notification

Currently, invitation are separate from Users service and users don't get notified via email once someone send them an invitation to join the domain.

I'd like that user gets notified via email there is an invitation waiting for them.

Must-have

### Anything else?

_No response_

WashingtonKK:
# What type of PR is this?

- Related Issue #
- Resolves #


internal/proto/users/v1/users.proto
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Abstract Machines
```
dborovcanin: Let's rename to file `emails.proto`.

internal/proto/users/v1/users.proto
```diff
@@ -0,0 +1,27 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+
+package users.v1;
+option go_package = "github.com/absmach/supermq/api/grpc/users/v1";
+
+// UsersService is a service that provides user-related functionalities
+// for SuperMQ services.
+service UsersService {
```
dborovcanin: Same, let's use `EmailService`.

internal/proto/users/v1/users.proto
```diff
@@ -0,0 +1,27 @@
...
+  rpc SendEmailWithUserId(SendEmailWithUserIdReq) returns (SendEmailWithUserIdRes) {}
```
dborovcanin: `SendEmail`

internal/proto/users/v1/users.proto
```diff
@@ -0,0 +1,27 @@
...
+}
+
+message SendEmailWithUserIdReq {
```
dborovcanin: `EmailReq`

internal/proto/users/v1/users.proto
```diff
@@ -0,0 +1,27 @@
...
+  repeated string users = 1;    // Email recipients
+  string from = 2;              // Sender email address (optional)
+  string subject = 3;           // Email subject
+  string header = 4;            // Email header text
+  string user = 5;              // User name
+  string content = 6;           // Email content/URL
+  string footer = 7;            // Email footer text
```
dborovcanin: Let's make it something like:
```
enum ContactType {
  ID = 0;
  Email = 1;
}

message EmailReq {
  string from = 1;              // Sender
  ContactType fromType = 2; // indicate sender id or email
  repeated string to = 3;       // recipients
  ContactType toType = 4; //  indicate recipient id or email
  string subject = 5;           // Email subject
  string content = 6;           // Email content/URL  
  optional string template = 10; // the whole template to be parsed by emailer
  optional string template_file = 11; // template file name to be parsed by emailer
  map<string, string> options = 12; // map that can be used in template
}
```
We will see what to put instead of this map as we progress with the server implementation. This can be used in emailer something like:
```go
if req.Template != "" {
	t, err := template.New("body").Parse(req.Template)
	if err != nil {
		return err
	}
	buff := new(bytes.Buffer)
	if err := t.Execute(buff, req.Options); err != nil {
		return err
	}
}
```

dborovcanin: In the future, we can be able to send body already prepared (template parsed), but also we should be able to send something like `template_file` name, so that emailer takes care of it.

internal/proto/users/v1/users.proto
```diff
@@ -0,0 +1,27 @@
...
+}
+
+message SendEmailRes {
+  bool sent = 1;                // Whether the email was sent successfully
```
dborovcanin: Let's use error here. So we have very go-ish interface where we inspect if response has an error.

dborovcanin (changes requested):

domains/service.go
```diff
@@ -205,6 +208,17 @@ func (svc *service) SendInvitation(ctx context.Context, session authn.Session, i
 	if err := svc.repo.SaveInvitation(ctx, invitation); err != nil {
 		return errors.Wrap(svcerr.ErrCreateEntity, err)
 	}
+
+	if _, err := svc.usersClient.SendEmail(ctx, &grpcEmailsV1.EmailReq{
+		Tos:      []string{invitation.InviteeUserID},
+		ToType:   grpcEmailsV1.ContactType_CONTACT_TYPE_ID,
+		Subject:  "Invitation to join Domain",
```
SammyOina: needs customization, probably just from template since prism has workspaces not domains

internal/proto/emails/v1/emails.proto
```diff
@@ -0,0 +1,35 @@
...
+// EmailService is a service that provides email-related functionalities
+// for SuperMQ services.
+service EmailService {
+  rpc SendEmail(EmailReq) returns (SendEmailRes) {}
+}
+
+enum ContactType {
+  CONTACT_TYPE_UNSPECIFIED = 0;
+  CONTACT_TYPE_ID = 1;
+  CONTACT_TYPE_EMAIL = 2;
```
SammyOina, line 18: more go idiomatic constant names

SammyOina: also use camel case

SammyOina (changes requested):

internal/proto/emails/v1/emails.proto
WashingtonKK: protolint requires enums to be in caps.

SammyOina (dismissed):

users/api/grpc/server.go
```diff
@@ -0,0 +1,76 @@
...
+	}, nil
+}
+
+func encodeSendEmailResponse(_ context.Context, grpcRes any) (any, error) {
+	res := grpcRes.(sendEmailRes)
+	errMsg := ""
+	if !res.sent {
+		errMsg = "failed to send email"
+	}
+	return &grpcUsersV1.SendEmailRes{Error: errMsg}, nil
```
dborovcanin, line 69: Rather than this, please use the actual error from the service.

dborovcanin (changes requested):

users/service.go
```diff
@@ -814,3 +814,21 @@ func changed(updated *string, old string) bool {
 
 	return *updated != old
 }
+
+func (svc service) SendEmailWithUserId(ctx context.Context, userIds []string, from, subject, header, user, content, footer string) error {
```
dborovcanin: Rename to `sendEmail` and encapsulate logic (fetch ID vs send directly via address), as well as params using DTO similarly how we do it in gPRC API.

dborovcanin: Also, add tests for this method.

dborovcanin (changes requested):

users/emailer.go
```diff
@@ -10,4 +10,10 @@ type Emailer interface {
 
 	// SendVerification sends an email to the user with a verification token.
 	SendVerification(To []string, user, verificationToken string) error
+
+	// Send sends an email with custom parameters.
+	Send(To []string, from, subject, header, user, content, footer string) error
+
+	// SendCustom sends an email with custom parameters using a custom email agent.
+	SendCustom(To []string, from, subject, header, user, content, footer string) error
```
dborovcanin: What's the difference between these two methods? This sounds more like 2 implementations of the emailer, rather than two methods in a single interface.

users/service.go
```diff
@@ -814,3 +815,39 @@ func changed(updated *string, old string) bool {
...
+		default:
+			return errors.Wrap(svcerr.ErrMalformedEntity, errors.New("invalid contact type for recipients"))
+		}
+	}
+
+	// Convert sender based on contact type
+	var senderName string
+	switch fromType {
+	case grpcEmailsV1.ContactType_CONTACT_TYPE_ID:
+		inviter, err := svc.users.RetrieveByID(ctx, from)
```
dborovcanin: There should only be one `SendEmail` method (the same for the emailer) that encapsulates fetching ID if sender type is ID rather than email and so on. Also, encapsulate params to a DTO that's similar to the `protobuf` email request.

dborovcanin (changes requested):

dborovcanin:
@WashingtonKK Please fix the CI.

dborovcanin:
Closing in favor of https://github.com/absmach/supermq/pull/3254.

PR #3221, opened 2025-11-03
SMQ-3264 - Make NATS as MQTT broker 

### LINKED ISSUE #3264: Use NATS as MQTT broker

arvindh123:
# What type of PR is this?

- Related Issue #3264
- Resolves #3264


dborovcanin:
This pull request is going to be obsolete as soon as we switch to https://github.com/absmach/fluxmq as a message broker, so it does not make sense.

Issue #1672, opened 2022-11-22
Verification of JWT Token by service itself (possible with present symmetric encryption) & Consideration of asymmetric encryption for a JWT token


arvindh123:
**FEATURE REQUEST**

1. Is there an open issue addressing this request? If it does, please add a "+1" reaction to the
   existing issue, otherwise proceed to step 2.

2. Describe the feature you are requesting, as well as the possible use case(s) for it.

### Objective  
Verification of JWT token by service, without depending on auth service. 
Auth service will issue only token, Service by them self will verify the token 

### Present scenario 
Every request received by service need to verify token, So the service call the auth service to validate the token, 
So, Auth service will have more network traffic and computation all the time. 

### Proposal 
If service can be able to verify the JWT token signature by itself, then no need to send request to auth for verification. 

At present auth service use HMAC encryption  
To verify the JWT token by service itself, the HMAC secret key need to share with all the service.  
Distribution of HMAC secret key to service is not good way I think, since with this HMAC symmetric key we can able to verify the certificate and we can able to create JWT Token with any other tools or codes , 

Anyone who knows the HMAC symmetrical secret key can tamper the system.

## Consideration of asymmetric encryption for a JWT token (For the case service verifies JWT token by itself)
If we use asymmetric encryption key RSA256,  then we need to share only PUBLIC key with all service. 

### Implementation rough idea: 
Auth service will generate private and public key 
Private key is used by auth service to generate JWT with signature 
Private key is not shared or not distributed, It will remains in auth service

Public Key is shared to service as JSON Web Key Set - JWKS Key through HTTP endpoint 
example: http://mainflux-auth:9999/.well-known/jwks.json 
 
And also we should have two token for this case , Access token with 10 minutes or less than 10 minutes expiry time and Refresh token with 12 hours or more than it .

Access Token is to access the resource 
Refresh Token to get access token 

### Additional security by rotation of asymmetric key
In additional, in auth service, we can rotate the programmatically public and private key on daily basis, and keep the public and private key in memory so no one (even us) could tamper them,
Or we can keep it them as file whenever it is generated.

3. Indicate the importance of this feature to you (must-have, should-have, nice-to-have).
should-have

arvindh123:
Request for Comments 
@mainflux/contributors 
@mainflux/maintainers

drasko:
[...]

PR #3228, opened 2025-11-07
SMQ-1672 - Add asymmetric key authentication

felixgateru:
# What type of PR is this?
This is a refactor as it replaces symmetric key auth centered authentication with asymmetric per service based authentication.

## What does this do?
New Features:
- Introduce KeyManager implementation providing SignJWT, ParseJWT, Rotate, and PublicJWKS methods
- Add HTTP endpoint at /.well-known/jwks.json to serve JSON Web Key Sets
- Add pkg/authn/jwks module for local token validation using fetched JWKS

Enhancements:
- Refactor JWT Tokenizer and Auth service to integrate with KeyManager instead of static secret
- Extend Auth service, middleware, and CLI clients to support JWKS-based authentication
- Update service interface and tests to include RetrieveJWKS and mock KeyManager/Tokenizer interactions

- Related Issue #1672

Yes, tests have been updated

Yes, in code documentation is included

### Notes
```mermaid
sequenceDiagram
    participant Client
    participant JWKSAuthn as "JWKS AuthN Module"
    participant AuthService
    participant JWKS_EP as "JWKS HTTP Endpoint"
    Client->>JWKSAuthn: Send JWT for authentication
    JWKSAuthn->>JWKS_EP: GET /.well-known/jwks.json
    JWKS_EP->>AuthService: RetrieveJWKS()
    AuthService->>JWKS_EP: JWKS response
    JWKS_EP->>JWKSAuthn: JWKS keys
    JWKSAuthn->>Client: Validate JWT and respond
```


dborovcanin:
@arvindh123 Please review.

auth/jwt/tokenizer.go
```diff
@@ -34,26 +34,22 @@ var (
...
-	secret []byte
+	keyManager auth.KeyManager
 }
 
 var _ auth.Tokenizer = (*tokenizer)(nil)
 
-// NewRepository instantiates an implementation of Token repository.
-func New(secret []byte) auth.Tokenizer {
+// New instantiates an implementation of Tokenizer service.
+func New(keyManager auth.KeyManager) auth.Tokenizer {
```
arvindh123, line 50: Lets have two tokenizer, One is symmentric and another u asymmetric and also we can provide option for asymmetric and symmentric alogirthms used

auth/middleware/logging.go
```diff
@@ -100,6 +101,16 @@ func (lm *loggingMiddleware) Identify(ctx context.Context, token string) (id aut
 	return lm.svc.Identify(ctx, token)
 }
 
+func (lm *loggingMiddleware) RetrieveJWKS() (jwks []jwk.Key) {
```
arvindh123: instead of jwk.Key can use our own structure? 
I'm not sure, please give our comments, 
I dont want to depend on external library for struct

arvindh123 (changes requested):
We need something like symmentric and asymmetric tokenizer with option to choose the alogrithm via env 

In Asymmentirc tokenizer, we need the key rotation option we the duration can be configured via env

auth/api/http/keys/responses.go
```diff
@@ -69,3 +70,23 @@ func (res revokeKeyRes) Headers() map[string]string {
...
+	Keys []jwk.Key `json:"keys"`
+}
+
+func (res retrieveJWKSRes) Code() int {
+	return http.StatusOK
+}
+
+func (res retrieveJWKSRes) Headers() map[string]string {
+	headers := map[string]string{
+		"Cache-Control": "public, max-age=900, stale-while-revalidate=60",
```
arvindh123: Is it possible to make this option provide via env

arvindh123 (changes requested):

dborovcanin (approved):

dborovcanin (changes requested):
Please fix the CI.

dborovcanin:
Also, let's use shared cache and make services share the **lock** if needed, to make this work in a scalable way.

nyagamunene:
PATs authentication is broken. 
image

pkg/authn/jwks/authn.go
```diff
@@ -0,0 +1,175 @@
+// Copyright (c) Abstract Machines
```
arvindh123, line 1: we can rename like asymmentric auth or something else 
`authasym` , `authsym` both method depend on auth service 

- In Symmentric we need to pass token
- In Asymmentric we are getting the JWKS from Auth Service for time interval

docker/.env
```diff
@@ -98,13 +98,18 @@ SMQ_AUTH_DB_SSL_MODE=disable
 SMQ_AUTH_DB_SSL_CERT=
 SMQ_AUTH_DB_SSL_KEY=
 SMQ_AUTH_DB_SSL_ROOT_CERT=
-SMQ_AUTH_SECRET_KEY=HyE2D4RUt9nnKG6v8zKEqAp6g6ka8hhZsqUpzgKvnwpXrNVQSH
-SMQ_AUTH_ACCESS_TOKEN_DURATION="1h"
-SMQ_AUTH_REFRESH_TOKEN_DURATION="24h"
+SMQ_AUTH_KEYS_ROTATION_INTERVAL="24h"
```
arvindh123: Is there any way to turn off the  Auth Keys rotation ?

auth/keymanager/symmetric/key_manager.go
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Abstract Machines
```
arvindh123, line 1: Can we provide feature to rotate the symmetrical key too? 

Not in this PR, May be in future PR

If yes, Can we add in  the next PR "rotation of both symmetrical and asymmetrical keys" ?

arvindh123 (changes requested):

auth/keymanager/symmetric/key_manager.go
dborovcanin: Let's not address this in this PR - that would be a new feature. Opening a new ticket for that.

dborovcanin:
@felixgateru Please fix CI.

auth/keymanager/asymmetric/key_manager.go
```diff
@@ -0,0 +1,251 @@
...
+				return
+			}
+		}()
+	}
+
+	go func() {
+		if err := km.cleanupHandler(ctx); err != nil {
+			return
+		}
+	}()
```
dborovcanin: These must not fail silently. It's a critical error, and here we have 0 visibility.

auth/keymanager/asymmetric/key_manager.go
```diff
@@ -0,0 +1,251 @@
...
+	km.activeKey.mu.RLock()
+	toRetire := km.activeKey.keyID
+	km.activeKey.mu.RUnlock()
+
+	err = km.repo.Retire(ctx, toRetire, time.Now().UTC())
+	if err != nil {
+		return err
+	}
+	km.setActiveKey(newID, newPair.privateKey)
+
```
dborovcanin: This lock is not correct - the other routine can access and change it before retire is called, resulting in retire failure. We need this whole section to be under the lock.

dborovcanin: In fact - this whole logic is problematic. What happens when saving succeeds and retiring fails and what if the opposite happens? This should at least be done in a single transaction.

cmd/auth/main.go
```diff
@@ -144,7 +149,40 @@ func main() {
...
+	case isSymmetric:
+		keyManager, err = symmetric.NewKeyManager(keyManagerCfg, []byte(cfg.SecretKey))
+		if err != nil {
+			logger.Error(fmt.Sprintf("failed to create symmetric key manager: %s", err.Error()))
+			exitCode = 1
+			return
+		}
+	default:
+		keyRepo := apostgres.NewPublicKeyRepo(db)
+		keyManager, err = asymmetric.NewKeyManager(ctx, keyManagerCfg, uuid.New(), keyRepo)
```
dborovcanin: This will result in starting a new key manager for each instance of the Auth service - which implies multiple cleanup and rotate routines competing for active key. This is unnecessary, suboptimal, and dangerous.

docker/.env
```diff
@@ -98,13 +98,18 @@ SMQ_AUTH_DB_SSL_MODE=disable
...
+SMQ_AUTH_KEYS_ACCESS_TOKEN_DURATION="1h"
+SMQ_AUTH_KEYS_REFRESH_TOKEN_DURATION="24h"
+SMQ_AUTH_KEYS_ALGORITHM="RS512"
+SMQ_AUTH_KEYS_SIZE=2048
```
dborovcanin: The common size for RS512 is 4096.

auth/postgres/public_keys.go
```diff
@@ -0,0 +1,168 @@
...
+	underlyingKey := j.JWK.Key()
+	if underlyingKey == nil {
+		return nil, nil
+	}
+
+	return json.Marshal(underlyingKey)
+}
+
+func (pkr *publicKeyRepo) Save(ctx context.Context, key auth.PublicKey) error {
+	q := `
```
dborovcanin: Saved keys should be encrypted.

auth/keymanager/asymmetric/key_manager.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/lestrrat-go/jwx/v2/jwa"
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/lestrrat-go/jwx/v2/jws"
+	"github.com/lestrrat-go/jwx/v2/jwt"
+)
+
+var (
+	cleanupInterval = 10 * time.Minute // How often to clean up expired keys
+	skewDuration    = 15*time.Second + cleanupInterval
+)
```
dborovcanin, line 27: Why not `const`? (AI slop?)

auth/keymanager/asymmetric/key_manager.go
```diff
@@ -0,0 +1,251 @@
...
+func (km *manager) SignJWT(token jwt.Token) ([]byte, error) {
+	km.activeKey.mu.RLock()
+	privateKey := km.activeKey.privateKey
+	km.activeKey.mu.RUnlock()
+
+	return jwt.Sign(token, jwt.WithKey(jwa.RS256, privateKey))
+}
+
+func (km *manager) ParseJWT(ctx context.Context, token string) (jwt.Token, error) {
+	keys, err := km.PublicJWKS(ctx)
```
dborovcanin: So, parsing JWT **always calls DB methods**? This is not acceptable.

dborovcanin (changes requested):
We will need to redesign the whole approach.

arvindh123 (dismissed):

auth/postgres/init.go
```diff
@@ -125,6 +125,23 @@ func Migration() *migrate.MemoryMigrationSource {
...
+						retired_at  		TIMESTAMPTZ,
+						status      		SMALLINT NOT NULL DEFAULT 0 CHECK (status >= 0)
+					);`,
+					`CREATE INDEX IF NOT EXISTS idx_public_keys_status ON public_keys(status);`,
+					`CREATE INDEX IF NOT EXISTS idx_public_keys_retired_at ON public_keys(retired_at);`,
+				},
+				Down: []string{
+					`DROP TABLE IF EXISTS public_keys;`,
+				},
+			},
```
dborovcanin: This should be removed.

dborovcanin (changes requested):

docker/.env
```diff
@@ -98,13 +98,17 @@ SMQ_AUTH_DB_SSL_MODE=disable
 SMQ_AUTH_DB_SSL_CERT=
 SMQ_AUTH_DB_SSL_KEY=
 SMQ_AUTH_DB_SSL_ROOT_CERT=
-SMQ_AUTH_SECRET_KEY=HyE2D4RUt9nnKG6v8zKEqAp6g6ka8hhZsqUpzgKvnwpXrNVQSH
 SMQ_AUTH_ACCESS_TOKEN_DURATION="1h"
 SMQ_AUTH_REFRESH_TOKEN_DURATION="24h"
+SMQ_AUTH_KEYS_ALGORITHM="EdDSA"
+SMQ_AUTH_KEYS_PRIVATE_KEY_PATH="./ssl/keys/private.key"
```
dborovcanin, line 104: Add an example of the key to the relevant path so compose works out of the box.

felixgateru: this is added

PR #3241, opened 2025-11-19
SMQ-1672 - Revoke refresh token

felixgateru:
# What type of PR is this?
This is a feature because it adds token revocation functionality to the auth service

## What does this do?
This pr enables refresh tokens to be revoked and stored.

- Related Issue #1672 

Yes, tests have been included

Yes, in code documentation has been included

### Notes
To be merged after: #3228


users/service.go
```diff
@@ -229,6 +229,22 @@ func (svc service) RefreshToken(ctx context.Context, session authn.Session, refr
...
 }
 
+func (svc service) RevokeRefreshToken(ctx context.Context, session authn.Session, refreshToken string) error {
+	dbUser, err := svc.users.RetrieveByID(ctx, session.UserID)
+	if err != nil {
+		return errors.Wrap(svcerr.ErrAuthentication, err)
+	}
+	if dbUser.Status == DisabledStatus {
+		return errors.Wrap(svcerr.ErrAuthentication, errLoginDisableUser)
+	}
```
arvindh123, line 245: If admin want to revoke the token of disabled user, then it is not possible, I think we can remove this logic

auth/cache/tokens.go
```diff
@@ -0,0 +1,62 @@
...
+		keyDuration: duration,
+	}
+}
+
+func (tc *tokensCache) Save(ctx context.Context, value string) error {
+	if err := tc.client.SAdd(ctx, defKey, value).Err(); err != nil {
+		return errors.Wrap(repoerr.ErrCreateEntity, err)
+	}
+
+	return nil
```
arvindh123: It would be nice if we had a TTL here because the refresh JWT already has an expiry. Once the refresh token expires, it becomes invalid, and there is no point in storing its JWT ID anymore, so the record should be automatically removed after that time

TTL should be like `expiry time - current time`

auth/postgres/token.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return repo.eh.HandleError(repoerr.ErrCreateEntity, err)
+	}
+	if rows == 0 {
+		return repoerr.ErrCreateEntity
+	}
+
+	return nil
+}
+
```
arvindh123: It would be nice if we had a TTL here because the refresh JWT already has an expiry. Once the refresh token expires, it becomes invalid, and there is no point in storing its JWT ID anymore, so the record should be automatically removed after that time

TTL should be like `expiry time - current time`

arvindh123 (changes requested):

auth/tokenizer/asymmetric/tokenizer.go
```diff
@@ -105,60 +107,56 @@ func (km *tokenizer) Issue(key auth.Key) (string, error) {
...
 	}
 
+	if key.Type == auth.RefreshKey && key.ID != "" && key.Subject != "" {
+		ttl := time.Until(key.ExpiresAt)
+		if ttl > 0 {
+			if err := tok.cache.SaveActive(ctx, key.Subject, key.ID, ttl); err != nil {
+				return "", err
+			}
+		}
+	}
```
arvindh123: @felixgateru 
I have question here, Just want to know reason 
Why we are keeping the active token cache with Token , not in service layer ? 
No need to change, If needed we can do it different PR

arvindh123: Because tokenizer helps to create and parse the token.
So may be, we can have separate layer for token management irrespective of asymmetric or symmetric token

arvindh123 (changes requested):

arvindh123 (commented):
The approach is good, but the Active token is embedded inside the Tokenizer, 
It will be nice if we have separate layer for Active token , 
So that it can plugged or unplugged based on need

arvindh123 (dismissed):

dborovcanin:
@felixgateru Please rebase.

nyagamunene:
@felixgateru I have tested and it works as expected. Can we update users apidocs in this PR or a different one?

nyagamunene (dismissed):

dborovcanin (commented):
Please rebase.

internal/proto/token/v1/token.proto
```diff
@@ -31,3 +39,20 @@ message Token {
...
+}
+
+message ListUserRefreshTokensRes {
+  repeated RefreshToken refresh_tokens = 1;
+}
+
+message RefreshToken {
+  string id = 1;
+  string description = 2;
+}
```
dborovcanin: Add a newline.

users/api/responses.go
```diff
@@ -80,6 +81,36 @@ func (res tokenRes) Empty() bool {
...
+
+func (res revokeRes) Headers() map[string]string {
+	return map[string]string{}
+}
+
+func (res revokeRes) Empty() bool {
+	return true
+}
+
+type listRefreshTokensRes struct {
```
dborovcanin, line 99: Missing
```go
_ supermq.Response = (*listRefreshTokensRes)(nil)
```

users/events/events.go
```diff
@@ -492,6 +492,17 @@ func (rte refreshTokenEvent) Encode() (map[string]any, error) {
 	}, nil
 }
 
+type revokeRefreshTokenEvent struct {
+	requestID string
```
dborovcanin, line 497: Let's also add token ID.

auth/middleware/logging.go
```diff
@@ -110,6 +110,41 @@ func (lm *loggingMiddleware) RetrieveJWKS() (jwks []auth.PublicKeyInfo) {
...
+			slog.String("duration", time.Since(begin).String()),
+			slog.String("user_id", userID),
+			slog.Int("tokens_count", len(tokens)),
+		}
+		if err != nil {
+			args = append(args, slog.String("error", err.Error()))
+			lm.logger.Warn("List user refresh tokens failed", args...)
+			return
+		}
+		lm.logger.Info("List user sessions completed successfully", args...)
```
dborovcanin: Fix the message.

auth/api/grpc/token/server.go
```diff
@@ -16,8 +16,10 @@ var _ grpcTokenV1.TokenServiceServer = (*tokenGrpcServer)(nil)
...
 	grpcTokenV1.UnimplementedTokenServiceServer
-	issue   kitgrpc.Handler
-	refresh kitgrpc.Handler
+	issue                 kitgrpc.Handler
+	refresh               kitgrpc.Handler
+	revoke                kitgrpc.Handler
+	listUserRefreshTokens kitgrpc.Handler
 }
 
 // NewAuthServer returns new AuthnServiceServer instance.
```
dborovcanin: Fix docs string.

auth/api/grpc/token/client.go
dborovcanin: Fix docs string.

users/users.go
```diff
@@ -264,13 +264,19 @@ type Service interface {
...
 
 	// RefreshToken refreshes expired access tokens.
 	// After an access token expires, the refresh token is used to get
 	// a new pair of access and refresh tokens.
 	RefreshToken(ctx context.Context, session authn.Session, refreshToken string) (*grpcTokenV1.Token, error)
 
+	// RevokeRefreshToken revokes a refresh token by its ID.
+	RevokeRefreshToken(ctx context.Context, session authn.Session, tokenID string) error
+
+	// ListActiveSessions lists all active refresh token sessions for the authenticated user.
```
dborovcanin: Fix docs string.

users/service.go
```diff
@@ -229,6 +235,39 @@ func (svc service) RefreshToken(ctx context.Context, session authn.Session, refr
...
+	_, err = svc.token.Revoke(ctx, &grpcTokenV1.RevokeReq{TokenId: tokenID})
+	if err != nil {
+		return errors.Wrap(svcerr.ErrAuthorization, err)
```
dborovcanin: Does this mask the real error? For example, not found?

dborovcanin: Should not found result in any error? Probably yes, but the status code should reflect it. Please check status code as well, and relevant API docs.

auth/cache/tokens.go
```diff
@@ -0,0 +1,165 @@
...
+type tokenData struct {
+	UserID      string `json:"user_id"`
+	Description string `json:"description,omitempty"`
+}
+
+var _ auth.UserActiveTokensCache = (*tokensCache)(nil)
+
+type tokensCache struct {
+	client      *redis.Client
+	keyDuration time.Duration
```
dborovcanin, line 26: This field is not used.

dborovcanin: Rather than passing TTL to the function, let `tokenCache`handle it.

dborovcanin: Don't allow it to be 0.

auth/cache/tokens.go
```diff
@@ -0,0 +1,165 @@
...
+
+	dataJSON, err := json.Marshal(data)
+	if err != nil {
+		return err
+	}
+
+	pipe := tc.client.TxPipeline()
+
+	pipe.Set(ctx, tc.tokenKey(tokenID), dataJSON, ttl)
+	pipe.SAdd(ctx, tc.userTokensKey(userID), tokenID)
```
dborovcanin: Do we need a TTL here?

auth/cache/tokens.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	var data tokenData
+	if err := json.Unmarshal([]byte(dataJSON), &data); err != nil {
+		pipe := tc.client.TxPipeline()
+		pipe.Del(ctx, tokenKey)
+		pipe.SRem(ctx, tc.userTokensKey(dataJSON), tokenID)
+		_, err = pipe.Exec(ctx)
+		return err
+	}
```
dborovcanin: This whole section has a smell. Also, cleanup on listing looks weird. Finally, 
```go
if err != nil {
		continue
	}
```
on L107-L109 don't look reassuring. Please revisit this whole approach.

users/api/endpoint_test.go
```diff
@@ -2384,7 +2384,9 @@ func TestIssueToken(t *testing.T) {
 	defer us.Close()
 
 	validUsername := "valid"
+	validDescription := "test token"
 	dataFormat := `{"username": "%s", "password": "%s"}`
+	dataFormatWithDesc := `{"username": "%s", "password": "%s", "description": "test token"}`
```
dborovcanin: Fix `"description:" %s`.

dborovcanin (changes requested):

auth/cache/tokens.go
```diff
@@ -0,0 +1,171 @@
...
+
+	"github.com/absmach/supermq/auth"
+	"github.com/absmach/supermq/pkg/errors"
+	svcerr "github.com/absmach/supermq/pkg/errors/service"
+	"github.com/redis/go-redis/v9"
+)
+
+const (
+	refreshPrefix = "refresh_tokens:"
+)
```
dborovcanin, line 20: Inline `const refreshPrefix = ...`.

arvindh123 (dismissed):

dborovcanin (changes requested):

auth/cache/tokens.go
```diff
@@ -0,0 +1,169 @@
...
+	return nil
+}
+
+func (tc *tokensCache) tokenKey(tokenID string) string {
+	return fmt.Sprintf("%s:token:%s", refreshPrefix, tokenID)
+}
+
+func (tc *tokensCache) userTokensKey(userID string) string {
+	return fmt.Sprintf("%s:user_tokens:%s", refreshPrefix, userID)
+}
```
dborovcanin: No need to make these tokenCache methods - make it package-level functions.

auth/cache/tokens.go
```diff
@@ -0,0 +1,169 @@
...
+	if err = json.Unmarshal([]byte(dataJSON), &data); err != nil {
+		return err
+	}
+
+	pipe.SRem(ctx, tc.userTokensKey(data.UserID), tokenID)
+	return nil
+}
+
+func (tc *tokensCache) tokenKey(tokenID string) string {
+	return fmt.Sprintf("%s:token:%s", refreshPrefix, tokenID)
```
dborovcanin: Use simpler string concatenation - since we have only `%s` - it's more efficient and there is no need to use reflection. This is against our internal guidelines, but let's change rules in service of progress. :)

auth/cache/tokens.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	dataJSON, err := json.Marshal(data)
+	if err != nil {
+		return err
+	}
+
+	pipe := tc.client.TxPipeline()
+
+	pipe.Set(ctx, tc.tokenKey(tokenID), dataJSON, tc.keyDuration)
```
dborovcanin: Does it make sense to check token expiration here? I.e. always set smaller value between token.Expiry and cache.Expiry?

auth/cache/tokens.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if len(stale) > 0 {
+		cleanup := tc.client.Pipeline()
+		for _, tokenID := range stale {
+			cleanup.SRem(ctx, key, tokenID)
+		}
+		if _, err := cleanup.Exec(ctx); err != nil {
+			return nil, err
+		}
+	}
```
dborovcanin: This will happen **only if** List is called with the correct token IDs. I.e. we potentially can leave with a lot of sets with empty values, but keys that are never used. This is practically a memory leak. We need either a different approach or a background cleanup process.

dborovcanin (changes requested):

auth/cache/tokens.go
```diff
@@ -0,0 +1,174 @@
...
+
+	pipe := tc.client.TxPipeline()
+	pipe.Del(ctx, tokenKeyStr)
+	defer func() {
+		_, execErr := pipe.Exec(ctx)
+		if err == nil {
+			err = execErr
+		}
+	}()
```
dborovcanin: Is this safe? Pipe is executed in defer, but commands are added **after** the error; i.e. ZRem will not execute in case unmarshalling fails. This all looks a bit clunky. Dropping defer would make it much cleaner.

auth/cache/tokens.go
```diff
@@ -0,0 +1,174 @@
...
+			ID:          tokenID,
+			Description: data.Description,
+		})
+	}
+
+	return valid, nil
+}
+
+// RemoveActive removes an active refresh token ID for a user.
+func (tc *tokensCache) RemoveActive(ctx context.Context, tokenID string) (err error) {
```
dborovcanin: Why don' we also pass User ID in this interface? It seems appropriate and it would simplify the the implementation a lot - no JSON or payload inspection. I am probably missing something here. We can deal with super admins revoking tokens on behalf of the users in a separate PR (or never; I am not sure this is even desired).

dborovcanin (changes requested):
OK, let's do this the old-fashioned way, try to do this refactoring **without AI tools** - they started hurting in this PR, especially in tokens cache file.
Zset seems like a good idea and appropriate choice.
Interface and implementation seem a bit off.

auth/cache/tokens.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if len(tokenIDs) == 0 {
+		return nil, nil
+	}
+
+	getPipe := tc.client.Pipeline()
+	getCmds := make(map[string]*redis.StringCmd, len(tokenIDs))
+	for _, tokenID := range tokenIDs {
+		getCmds[tokenID] = getPipe.Get(ctx, tokenKey(tokenID))
+	}
```
dborovcanin: Let's not use map to avoid non-deterministic order of the loop.

dborovcanin (changes requested):

dborovcanin (approved):

PR #3252, opened 2025-11-24
SMQ-3253 - Update CLI usage

### LINKED ISSUE #3253: Bug: Update CLI Usage

The currently the usage is not clear after changing the ordering of arguments. For example:

`usage: users <user_id_or_all> <operation> [args...]`

I expect the user to get the commands that are also available:

`usage: users <user_id|all|create|token|refreshtoken|profile|resetpasswordrequest|resetpassword|password|search> [operation] [args...]`

`usage: users <user_id_or_all> <operation> [args...]`

### Steps To Reproduce

Run the CLI without all the commands.

Docker

_No response_

nyagamunene:
# What type of PR is this?

- Related Issue #3253
- Resolves #3253

### DIFF
--- a/cli/users.go
+++ b/cli/users.go
@@ -16,10 +16,10 @@ import (
 
 const (
 	token                = "token"
-	refreshtoken         = "refreshtoken"
+	refreshToken         = "refreshtoken"
 	profile              = "profile"
-	resetpasswordrequest = "resetpasswordrequest"
-	resetpassword        = "resetpassword"
+	resetPasswordRequest = "resetpasswordrequest"
+	resetPassword        = "resetpassword"
 	password             = "password"
 	search               = "search"
 	username             = "username"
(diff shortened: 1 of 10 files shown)


cli/users.go
```diff
@@ -83,25 +94,58 @@ Examples:
 				handleVerify(cmd, args[1])
 				return
 			case token:
-				handleUserToken(cmd, args[1], args[2:])
+				if len(args) < 2 {
+					handleUserToken(cmd, "", []string{})
+				} else {
+					handleUserToken(cmd, args[1], args[2:])
+				}
```
dborovcanin: Previous approach was cleaner. Why do we need these checks, what's the reason for an empty first param?

dborovcanin (changes requested):

dborovcanin (changes requested):
Since there are no cases where we need <2 args, can we avoid these checks by extracting them before `switch` and after zero check, like:
```go
			if len(args) < 2 {
				logUsageCmd(*cmd, args[0])
				return
			}
```

Also, please use this PR to fix camelcase - `refreshToken`, `ResetPasswordReq`, `ResetPassword`...

dborovcanin (approved):

PR #3254, opened 2025-11-25
SMQ-3234 - Add notifications service

WashingtonKK:
# What type of PR is this?

- Related Issue #
- Resolves #


notifications/emailer/emailer.go
```diff
@@ -0,0 +1,54 @@
...
+package emailer
+
+import (
+	"fmt"
+
+	"github.com/absmach/supermq/internal/email"
+	"github.com/absmach/supermq/notifications"
+)
+
+var _ notifications.Emailer = (*emailer)(nil)
```
SammyOina: you should have a notifier interface and implmentations  for different notifiers (SMS, email, slack ...)

notifications/emailer/emailer.go
```diff
@@ -0,0 +1,54 @@
...
+		return nil, err
+	}
+
+	return &emailer{
+		invitationSentAgent:     invitationSentAgent,
+		invitationAcceptedAgent: invitationAcceptedAgent,
+	}, nil
+}
+
+func (e *emailer) SendInvitationSentEmail(to []string, inviteeName, domainName, inviterName string) error {
```
SammyOina: One send method is enough and message strcture can be defined  with events especially subject and headers need to be dynamic to avoid grouping messages from prism and smq in yahoo

notifications/emailer/emailer.go
```diff
@@ -0,0 +1,54 @@
...
+	subject := fmt.Sprintf("You've been invited to join %s", domainName)
+	header := "Domain Invitation"
+	content := fmt.Sprintf("%s has invited you to join the domain %s. Please log in to accept or reject this invitation.", inviterName, domainName)
+	footer := "SuperMQ Team"
```
SammyOina: needs to be dynamic

SammyOina (changes requested):

dborovcanin (changes requested):
@WashingtonKK Check our old notifier approach:
https://github.com/absmach/supermq-contrib/tree/main/consumers/notifiers.

Notification service is **consumer**, i.e. we subscribe to events. What we need is:
1. actual Notification sender - think something like an interface like:
```go
type Notifier interface {
    Notify(Notification) error
}
```
So, notification struct is 👍🏽.
2. Notification service contains a collection of notifiers, think of it like email, Slack, SMS, any other notifier. Notification service is used to subscribe to provided events, convert them to Notification, and route Notification to the registered notifiers.

Since notifications are not decoupled from other services (nobody should be importing them), you can import users or any other SMQ package directly.

users/emailer/emailer.go
```diff
@@ -48,3 +62,15 @@ func (e *emailer) SendVerification(to []string, user, verificationToken string)
...
+	subject := fmt.Sprintf("You've been invited to join %s", domainName)
+	content := fmt.Sprintf("%s has invited you to join %s as %s.", inviterName, domainName, roleName)
+	return e.invitationAgent.Send(to, "", subject, "", inviteeName, content, "SuperMQ Team")
+}
+
+func (e *emailer) SendInvitationAccepted(to []string, inviterName, inviteeName, domainName, roleName string) error {
+	subject := fmt.Sprintf("%s accepted your invitation to %s", inviteeName, domainName)
+	content := fmt.Sprintf("%s has accepted your invitation to join %s as %s.", inviteeName, domainName, roleName)
+	return e.invitationAcceptedAgent.Send(to, "", subject, "", inviterName, content, "SuperMQ Team")
+}
```
SammyOina: we just need one send/ notify method

users/events/consumer.go
```diff
@@ -0,0 +1,233 @@
...
+	operation, ok := data["operation"].(string)
+	if !ok {
+		return nil
+	}
+
+	switch operation {
+	case invitationSend:
+		return handleInvitationSent(ctx, data, h.emailer, h.userRepo, h.logger)
+	case invitationAccept:
+		return handleInvitationAccepted(ctx, data, h.emailer, h.userRepo, h.logger)
```
SammyOina: there should be more events here

SammyOina: there should be more events here

SammyOina: there should be more events here

users/emailer.go
SammyOina, line 1: emailer should implement notifier interface rather than having its own interface

users/notifier.go
```diff
@@ -0,0 +1,38 @@
...
+	// NotificationEmailVerification is sent when a user needs to verify their email.
+	NotificationEmailVerification NotificationType = "email_verification"
+	// NotificationInvitationSent is sent when a user is invited to a domain.
+	NotificationInvitationSent NotificationType = "invitation_sent"
+	// NotificationInvitationAccepted is sent when a user accepts an invitation.
+	NotificationInvitationAccepted NotificationType = "invitation_accepted"
+)
+
+// NotificationData contains the data needed to send a notification.
+type NotificationData struct {
```
dborovcanin: Rename to Notification.

users/emailer.go
dborovcanin: This has not been addressed.

users/service.go
```diff
@@ -41,17 +41,17 @@ type service struct {
 	idProvider supermq.IDProvider
 	policies   policies.Service
 	hasher     Hasher
-	email      Emailer
+	notifier   Notifier
 }
 
 // NewService returns a new Users service implementation.
-func NewService(token grpcTokenV1.TokenServiceClient, urepo Repository, policyService policies.Service, emailer Emailer, hasher Hasher, idp supermq.IDProvider) Service {
+func NewService(token grpcTokenV1.TokenServiceClient, urepo Repository, policyService policies.Service, notifier Notifier, hasher Hasher, idp supermq.IDProvider) Service {
```
dborovcanin: Ideally, there would be a slice of notifiers for different types of notifications. It's OK to keep it a single Notifier since we are talking about emailer only.

users/emailer/notifier.go
```diff
@@ -0,0 +1,72 @@
...
+	case users.NotificationInvitationSent:
+		return n.notifyInvitationSent(data)
+	case users.NotificationInvitationAccepted:
+		return n.notifyInvitationAccepted(data)
+	default:
+		return fmt.Errorf("unknown notification type: %s", data.Type)
+	}
+}
+
+func (n *emailNotifier) notifyPasswordReset(data users.NotificationData) error {
```
dborovcanin: Always propagate context from the caller (Notify).

users/events/consumer.go
```diff
@@ -0,0 +1,255 @@
...
+
+	return sub.Subscribe(ctx, subCfg)
+}
+
+// Handle handles invitation events.
+func (h *eventHandler) Handle(ctx context.Context, event events.Event) error {
+	data, err := event.Encode()
+	if err != nil {
+		h.logger.Error("failed to encode event", slog.Any("error", err))
+		return nil
```
dborovcanin: Why don't we return an error?

users/emailer/notifier.go
```diff
@@ -0,0 +1,72 @@
...
+	token := data.Metadata["verification_token"]
+
+	return n.emailer.SendVerification(data.Recipients, user, token)
+}
+
+func (n *emailNotifier) notifyInvitationSent(data users.NotificationData) error {
+	inviteeName := data.Metadata["invitee_name"]
+	inviterName := data.Metadata["inviter_name"]
+	domainName := data.Metadata["domain_name"]
+	roleName := data.Metadata["role_name"]
```
dborovcanin: We are missing a lot of validation here. Also, DRY, decode once, and validate depending on the kind of the email you want to send. `notifyInvitationSent` and `notifyInvitationAccepted` are the same. What about rejected?

users/events/consumer.go
```diff
@@ -0,0 +1,255 @@
...
+		logger.Error("failed to retrieve inviter user",
+			slog.String("user_id", invitedBy),
+			slog.Any("error", err),
+		)
+		return nil
+	}
+
+	// Normalize names for display
+	inviteeName := invitee.FirstName + " " + invitee.LastName
+	if inviteeName == " " || inviteeName == "" {
```
dborovcanin: `inviteeName == ""` will never be the case because of the previous line.

users/notifier.go
```diff
@@ -0,0 +1,38 @@
...
+	// Type is the type of notification to send.
+	Type NotificationType
+
+	// Recipients is the list of recipients for the notification.
+	Recipients []string
+
+	// Metadata contains additional data specific to the notification type.
+	Metadata map[string]string
```
dborovcanin: While very generic, this way makes handling notification more cryptic and complex then it needs to be. I.e. I'd prefer even `Notify(ctx context.Context, notification any) error` over extracting it from the map and expecting keys to be there - we'd have at least type safety once we cast `any` and instead of `NotificationData.Type` use `switch notification.(type)`. Check two other (unlikely) options:
1. Making Notification an interface (notifications holds data, so it's probably a dead-end)
2. Using generics (probably will end up mostly the same code as `Notify(ctx context.Context, notification any) error`)

dborovcanin (changes requested):

users/emailer/emailer.go
```diff
@@ -4,23 +4,39 @@
...
-	verificationURL string
-	resetAgent      *email.Agent
-	verifyAgent     *email.Agent
+	resetURL                string
+	verificationURL         string
+	resetAgent              *email.Agent
+	verifyAgent             *email.Agent
+	invitationAgent         *email.Agent
+	invitationAcceptedAgent *email.Agent
+	invitationRejectedAgent *email.Agent
```
SammyOina: we do not need multiple email agents

users/emailer/emailer.go
```diff
@@ -31,20 +47,142 @@ func New(resetURL, verificationURL string, resetConfig, verifyConfig *email.Conf
...
+	}
+
+	url := fmt.Sprintf("%s?token=%s", e.resetURL, notif.Token)
+	return e.resetAgent.Send(notif.To, "", "Password Reset Request", "", notif.User, url, "")
 }
 
-func (e *emailer) SendVerification(to []string, user, verificationToken string) error {
-	url := fmt.Sprintf("%s?token=%s", e.verificationURL, verificationToken)
-	return e.verifyAgent.Send(to, "", "Email Verification", "", user, url, "")
+func (e *emailer) sendEmailVerification(_ context.Context, notif *users.EmailVerificationNotification) error {
```
SammyOina: don't pass context if not used

SammyOina: same in other places

users/events/consumer.go
```diff
@@ -0,0 +1,343 @@
...
+	logger.Info("invitation accepted notification sent",
+		slog.String("to", inviter.Email),
+		slog.String("domain", domainName),
+	)
+
+	return nil
+}
+
+func handleInvitationRejected(ctx context.Context, data map[string]any, notifier users.Notifier, userRepo users.Repository, logger *slog.Logger) error {
+	inviteeUserID, _ := data["invitee_user_id"].(string)
```
SammyOina: will panic if assetion is not valid, do a check

SammyOina (changes requested):

users/emailer/emailer.go
WashingtonKK: https://github.com/absmach/supermq/issues/3257

users/events/consumer.go
```diff
@@ -0,0 +1,402 @@
...
+	case invitationAccept:
+		return handleInvitationAccepted(ctx, data, h.notifier, h.userRepo, h.logger)
+	case invitationReject:
+		return handleInvitationRejected(ctx, data, h.notifier, h.userRepo, h.logger)
+	default:
+		return nil
+	}
+}
+
+func handleInvitationSent(ctx context.Context, data map[string]any, notifier users.Notifier, userRepo users.Repository, logger *slog.Logger) error {
```
SammyOina: this functions seems to only ever return nil and never any error

users/events/consumer.go
```diff
@@ -0,0 +1,402 @@
...
+
+	logger.Info("invitation notification sent",
+		slog.String("to", invitee.Email),
+		slog.String("domain", domainName),
+	)
+
+	return nil
+}
+
+func handleInvitationAccepted(ctx context.Context, data map[string]any, notifier users.Notifier, userRepo users.Repository, logger *slog.Logger) error {
```
SammyOina: same here

users/events/consumer.go
```diff
@@ -0,0 +1,402 @@
...
+
+	logger.Info("invitation accepted notification sent",
+		slog.String("to", inviter.Email),
+		slog.String("domain", domainName),
+	)
+
+	return nil
+}
+
+func handleInvitationRejected(ctx context.Context, data map[string]any, notifier users.Notifier, userRepo users.Repository, logger *slog.Logger) error {
```
SammyOina: same here, you type assertion and if they fail you just continue with the invalid data type, meaning there is likely to be a runtime panic

internal/email/email.go
```diff
@@ -106,5 +107,7 @@ func (a *Agent) Send(to []string, from, subject, header, user, content, footer s
 		return errors.Wrap(errSendMail, err)
 	}
 
+	fmt.Println("Email Sent")
```
SammyOina: clean up the pr

SammyOina (changes requested):

users/events/consumer.go
```diff
@@ -0,0 +1,391 @@
...
+func (h *eventHandler) Handle(ctx context.Context, event events.Event) error {
+	data, err := event.Encode()
+	if err != nil {
+		h.logger.Error("failed to encode event", slog.Any("error", err))
+		return err
+	}
+
+	operation, ok := data["operation"].(string)
+	if !ok {
+		return nil
```
SammyOina: should not be nil if type assertion failed

users/events/consumer.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+
+	// Retrieve invitee user
+	invitee, err := userRepo.RetrieveByID(ctx, inviteeUserID)
+	if err != nil {
+		logger.Error("failed to retrieve invitee user",
+			slog.String("user_id", inviteeUserID),
+			slog.Any("error", err),
+		)
+		return nil
```
SammyOina: why return nil here

users/events/consumer.go
```diff
@@ -0,0 +1,391 @@
...
```
SammyOina: a number of lines where we have an error but return nil

users/events/consumer.go
```diff
@@ -0,0 +1,391 @@
...
+
+	if err := notifier.Notify(ctx, notification); err != nil {
+		logger.Error("failed to send invitation rejected notification",
+			slog.String("to", inviter.Email),
+			slog.Any("error", err),
+		)
+		return nil
+	}
+
+	logger.Info("invitation rejected notification sent",
```
SammyOina: I see a lot of logging, we can do this in middleware

SammyOina (changes requested):

SammyOina (dismissed):

users/notifier.go
```diff
@@ -0,0 +1,48 @@
...
+	InviterName string
+	DomainName  string
+	RoleName    string
+}
+
+// Notifier is an interface for sending notifications through various channels.
+type Notifier interface {
+	// Notify sends a notification based on the notification type.
+	Notify(ctx context.Context, notification any) error
+}
```
arvindh123: This can be emailer, because it sends email not notifications

notifications/events/consumer.go
```diff
@@ -0,0 +1,112 @@
...
+
+	return notifications.Notification{
+		InviterID:  invitedBy,
+		InviteeID:  inviteeUserID,
+		DomainID:   domainID,
+		DomainName: domainName,
+		RoleID:     roleID,
+		RoleName:   roleName,
+	}, nil
+}
```
arvindh123, line 111: Can we move this parser to common package `pkg/domains/events/consumer/decode.go`

SammyOina (changes requested):
Image

subject on verification will still result in grouping on yahoo

notifications/notifier.go
```diff
@@ -0,0 +1,37 @@
...
+// Notification contains the data needed to send a notification.
+type Notification struct {
+	Type       NotificationType
+	InviterID  string
+	InviteeID  string
+	DomainID   string
+	DomainName string
+	RoleID     string
+	RoleName   string
+}
```
arvindh123, line 31: Is this specific to invitations ?

notifications/events/consumer.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/absmach/supermq/notifications"
+	"github.com/absmach/supermq/pkg/events"
+)
+
+const (
+	// Stream names.
+	sendInvitationStream   = "events.supermq.invitation.send"
+	acceptInvitationStream = "events.supermq.invitation.accept"
+	rejectInvitationStream = "events.supermq.invitation.reject"
```
arvindh123, line 19: @dborovcanin
Just saying my thoughts 

Having stream topic as constant make notification service works only for invitations 
We need to make this configurable via file 

The file should have event topic,event payload parser, notification condition check , then emailer or slacker or webhooker

notifications/events/consumer.go
dborovcanin: For the time being, let's keep it here. If we introduce more kinds of notification, it will make sense to move it to a shared decoder.

PR #3255, opened 2025-11-25
SMQ-3222 - Add OAuth support to CLI

### LINKED ISSUE #3222: Add OAuth support to CLI

Get token for User who signup via google oauth

Getting token

No way to get token

### Steps To Reproduce

This is the command to get token `supermq-cli users token <username> <password>` where <username> is google oauth email id 
Since there is no password for oauth user, i could not put any password here 

latest cli

_No response_

dborovcanin:
# What type of PR is this?
This is a feature.

## What does this do?
It adds an OAuth support to SMQ CLI.

Resolves #3222.

Yes.

Yes.

### Notes
N/A


cli/oauth.go
```diff
@@ -0,0 +1,547 @@
...
+)
+
+const (
+	callbackPath    = "/callback"
+	localServerPort = "9090"
+	callbackTimeout = 5 * time.Minute
+	shutdownTimeout = 5 * time.Second
+)
+
+const successHTML = `<!DOCTYPE html>
```
nyagamunene: Since we have the preview.html files can we remove this constants and use `embed` library

dborovcanin: Removed preview files and move strings to a separate go file. We are trying to use approach as simple as possible, and also standalone since this is something that will be running on Users end - i.e. we start a small server used for redirection. We are not even using it as a template, as you can see, but just serve it as a string with HTML content type.

nyagamunene (dismissed):
LGTM

arvindh123:
@dborovcanin  
I think CLI should open the SuperMQ Login page in browser and then user should choose the oauth 
Where SuperMQ UI or backend should capable of redirecting to CLI server 

We can take reference flow from  Digital ocean doctl cli tool and Github CLI tool

dborovcanin:
@arvindh123 What we can do is use Device Auth Flow instead https://datatracker.ietf.org/doc/html/rfc8628. I have a draft prepared, but haven't sent it yet.

dborovcanin:
@arvindh123 The reason is simple - there is no reason to assume device using CLI has broweser installed. For example, you could and are likely to use CLI when SSH-ed to the cloud component.

dborovcanin:
This is obsolete, no point in updating. It requires a new PR. Closing.

PR #3295, opened 2025-12-18
SMQ-2627 - Align PATs with new architecture

### LINKED ISSUE #2627: Feature:  PAT align with new architecture

No

Present PAT is designed for old authorization model, So PAT need to be revisited and need to align with new architecture 

Must-have

### Anything else?

_No response_

nyagamunene:
# What type of PR is this?

- Related Issue #2627 
- Resolves #2627


pkg/authn/authsvc/authn.go
```diff
@@ -46,9 +43,11 @@ func (a authentication) Authenticate(ctx context.Context, token string) (authn.S
 		return authn.Session{}, errors.Wrap(errors.ErrAuthentication, err)
 	}
 
-	if strings.HasPrefix(token, patPrefix) {
+	tokenType := authn.TokenType(res.GetTokenType())
```
dborovcanin: This creates possible misconfiguration issue. I.e. what if token prefix here and tokenType returned from the service are not the same?  While we are not able to validate token - we are able to parse it locally (even if only for prefix) and we can do this verification to prevent mismatch.

channels/middleware/authorization.go
```diff
@@ -350,30 +343,33 @@ func (am *authorizationMiddleware) authorize(ctx context.Context, session authn.
...
 		case channels.OpDisconnectClient:
-			req.Operation = auth.DeleteOp
-		case domains.OpCreateDomainChannels,
-			domains.OpListDomainChannels:
-			if op == domains.OpCreateDomainChannels {
-				req.Operation = auth.CreateOp
-			} else {
-				req.Operation = auth.ListOp
-			}
+			req.Operation = auth.ChannelDisconnectClientOp
```
dborovcanin: Let's stay consistent in naming, compare this to [ClientConnectToChannelOp](https://github.com/absmach/supermq/pull/3295/changes#diff-b78fc26b2e6dbc94bc9d8a3dba809777dbd4891a71147cc00536e5cc4ea53adbR300).

dborovcanin (changes requested):

auth/pat.go
```diff
@@ -18,86 +18,194 @@ const AnyIDs = "*"
...
-	CreateOp Operation = iota
-	ReadOp
-	ListOp
-	UpdateOp
-	DeleteOp
-	ShareOp
-	UnshareOp
-	PublishOp
-	SubscribeOp
+	ClientCreateOp Operation = iota + 100
```
arvindh123: Is it possible to reuse the existing service Operations ?

arvindh123 (changes requested):

auth/pat.go
nyagamunene: we will have cyclic imports

auth/pat.go
```diff
@@ -214,25 +310,145 @@ func (et *EntityType) UnmarshalText(data []byte) (err error) {
 	return err
 }
 
+var ValidOperationsForEntity = map[EntityType][]Operation{
```
felixgateru: I noticed here that we have not included domains as an entity, does this mean that pats can not be used for actions on domains?

nyagamunene: From past discussion with @arvindh123, PATs are being used for entities like channels, clients, groups, dashboards and messages. It is not used to create users and domains. @dborovcanin can you confirm this?

dborovcanin: This is fine; we can limit PATS to this at the moment. I.e. PATS are used for users to manage entities within the domain.

auth/pat.go
```diff
@@ -288,6 +504,10 @@ func (s *Scope) Validate() error {
 		}
 	}
 
+	if !IsValidOperationForEntity(s.EntityType, s.Operation) {
+		return errors.Wrap(apiutil.ErrInvalidQueryParams, errors.New("operation not valid for entity type"))
```
felixgateru: You can define this error, instead of calling new each time

felixgateru (changes requested):

felixgateru (dismissed):

internal/proto/auth/v1/auth.proto
```diff
@@ -40,12 +40,13 @@ message PolicyReq {
...
-  string optional_domain_id = 4;  // Optional domain id (PAT)
-  uint32 operation = 5;           // Operation (PAT)
-  string entity_id = 6;           // EntityID (PAT)
+  uint32 token_type = 1;          // Token type
+  string user_id = 2;             // User id
+  string pat_id = 3;              // Pat id
+  uint32 entity_type = 4;         // Entity type
+  string optional_domain_id = 5;  // Optional domain id
+  uint32 operation = 6;           // Operation
+  string entity_id = 7;           // EntityID
```
dborovcanin: Remove slop comments.

internal/proto/auth/v1/auth.proto
```diff
@@ -40,12 +40,13 @@ message PolicyReq {
...
-  string pat_id = 2;              // Pat id
-  uint32 entity_type = 3;         // Entity type (PAT)
-  string optional_domain_id = 4;  // Optional domain id (PAT)
-  uint32 operation = 5;           // Operation (PAT)
-  string entity_id = 6;           // EntityID (PAT)
+  uint32 token_type = 1;          // Token type
+  string user_id = 2;             // User id
+  string pat_id = 3;              // Pat id
+  uint32 entity_type = 4;         // Entity type
+  string optional_domain_id = 5;  // Optional domain id
```
dborovcanin: Why not `optional string domain_id = 5`?

dborovcanin (changes requested):

internal/proto/auth/v1/auth.proto
nyagamunene: Since users was remove I think it should be optional

pkg/roles/rolemanager/middleware/authorization.go
```diff
@@ -291,14 +292,65 @@ func (ram RoleManagerAuthorizationMiddleware) RoleRemoveMembers(ctx context.Cont
...
+		case roles.OpRoleListMembers:
+			pr.Operation = auth.RoleListMembersOp
+		case roles.OpRoleCheckMembersExists:
+			pr.Operation = auth.RoleCheckMembersExistsOp
+		case roles.OpRoleRemoveMembers:
+			pr.Operation = auth.RoleRemoveMembersOp
+		case roles.OpRoleRemoveAllMembers:
+			pr.Operation = auth.RoleRemoveAllMembersOp
+		}
+	}
```
arvindh123, line 320: I suspect this code need , These should be passed from permissions.yaml file

internal/proto/auth/v1/auth.proto
```diff
@@ -19,39 +19,40 @@ message AuthNReq {
...
+  string domain = 2;
+  string subject_type = 3;
+  string subject_kind = 4;
+  string subject_relation = 5;
+  string subject = 6;
+  string relation = 7;
+  string permission = 8;
+  string object = 9;
+  string object_type = 10;
 }
```
arvindh123: Is it possible to fit the PATreq within here? Because we have  token_type here

arvindh123 (changes requested):

auth/postgres/init.go
```diff
@@ -93,11 +93,11 @@ func Migration() *migrate.MemoryMigrationSource {
 					`CREATE TABLE IF NOT EXISTS pat_scopes (
 						id              	VARCHAR(36) PRIMARY KEY,
 						pat_id          	VARCHAR(36) REFERENCES pats(id) ON DELETE CASCADE,
-						optional_domain_id	VARCHAR(36),
```
felixgateru, line 96: Why modify an existing migration instead of adding a new one to rename the column

auth/pat.go
```diff
@@ -456,15 +675,15 @@ type PATSRepository interface {
 
 	RemoveScope(ctx context.Context, userID string, scopesIDs ...string) error
 
-	CheckScope(ctx context.Context, userID, patID string, entityType EntityType, optionalDomainID string, operation Operation, entityID string) error
+	CheckScope(ctx context.Context, userID, patID string, entityType EntityType, domainID string, operation Operation, entityID string) error
```
felixgateru: for the sake of consistency you can order the parameters uniformly across the two interfaces.
in this one domainid is after entitype and in the one in Cache it is after.
If you do make the change ensure to change everywhere the function is used

auth/scope_test.go
```diff
@@ -14,126 +14,126 @@ import (
...
-			entityID:         "any-entity",
-			expected:         true,
+			entityType: auth.GroupsType,
+			domainID:   "domain1",
+			operation:  auth.GroupCreateOp,
+			entityID:   "any-entity",
+			expected:   true,
 		},
 		{
 			desc: "Authorized without domain ID",
```
felixgateru, line 57: isnt this test description misleading since there is a domain in the scope

auth/service.go
```diff
@@ -211,8 +211,11 @@ func (svc service) RetrieveJWKS() []PublicKeyInfo {
 }
 
 func (svc service) Authorize(ctx context.Context, pr policies.Policy) error {
+	// For PAT authorization, first check if the PAT has the required scope,
+	// then continue with normal authorization to ensure the user still has permissions.
```
felixgateru: the comment suggest that we continue with authorization after the PAT check but that is not the case since we have the early return on line 221. You can update the comment

felixgateru (changes requested):

felixgateru (dismissed):

internal/proto/auth/v1/auth.proto
```diff
@@ -19,40 +19,33 @@ message AuthNReq {
...
+  string domain = 1;
+  string subject_type = 2;
+  string subject_kind = 3;
+  string subject_relation = 4;
+  string subject = 5;
+  string relation = 6;
+  string permission = 7;
+  string object = 8;
+  string object_type = 9;
+  string user_id = 10;
```
arvindh123: Is it possible to use the `subject_type` and `subject` instead of this `user id` ?

internal/proto/auth/v1/auth.proto
```diff
@@ -19,40 +19,33 @@ message AuthNReq {
...
+  string pat_id = 11;
+  uint32 entity_type = 12;
```
arvindh123: Is it possible to use the `object_type` instead of `entity_type` ?

internal/proto/auth/v1/auth.proto
```diff
@@ -19,40 +19,33 @@ message AuthNReq {
...
+  uint32 operation = 13;
+  string entity_id = 14;
```
arvindh123: Is it possible to use the `object` instead of `entity_id` ?

arvindh123 (changes requested):
I have idea now, no need to implement in this PR.

We can have  the permission map in auth service only 
And in policy request we can send only the operations 

Since PAT also sends the operation , we can make the users authz to send the operation instead of permission and 
In auth service we can convert operations to Permissions

What do you think?

nyagamunene:
> I have idea now, no need to implement in this PR.
> 
> We can have the permission map in auth service only And in policy request we can send only the operations
> 
> Since PAT also sends the operation , we can make the users authz to send the operation instead of permission and In auth service we can convert operations to Permissions
> 
> What do you think?

This is a good idea. It would simplify the authorization process and have no need of having the permissions file for all the entities.

pkg/authn/authsvc/authn.go
```diff
@@ -44,9 +48,20 @@ func (a authentication) Authenticate(ctx context.Context, token string) (authn.S
...
-		return authn.Session{Type: authn.PersonalAccessToken, PatID: res.GetId(), UserID: res.GetUserId(), Role: authn.Role(res.GetUserRole())}, nil
+	isPAT := strings.HasPrefix(token, patPrefix)
+	tokenType := authn.TokenType(res.GetTokenType())
+
+	if isPAT && tokenType != authn.PersonalAccessToken {
+		return authn.Session{}, errTokenTypeMismatch
+	}
+	if !isPAT && tokenType == authn.PersonalAccessToken {
+		return authn.Session{}, errTokenTypeMismatch
 	}
```
arvindh123: This logic can be we merged into single case

arvindh123: Or something like this 
```go
if isPAT != (tokenType == authn.PersonalAccessToken) {
	return authn.Session{}, errTokenTypeMismatch
}
```

arvindh123: Since this is library, i think we don't want to do this check here 
We can do this check in Auth Service itself and believe response that auth service has returned

pkg/authz/authsvc/authz.go
```diff
@@ -83,19 +61,39 @@ func (a authorization) Authorize(ctx context.Context, pr authz.PolicyReq) error
...
 			},
+		}
+		patRes, err := a.authSvcClient.Authorize(ctx, &patReq)
+		if err != nil {
+			return errors.Wrap(errors.ErrAuthorization, err)
+		}
+		if !patRes.GetAuthorized() {
+			return errors.ErrAuthorization
+		}
+	}
```
arvindh123: @nyagamunene I have question, Does this TokenType is needed? 
How this TokenType is helpfull for us

pkg/permissions/config.go
```diff
@@ -41,44 +41,73 @@ func (pc *PermissionConfig) GetEntityPermissions(entityType string) (map[string]
...
-) (EntitiesPermission, EntitiesOperationDetails[Operation], error) {
-	entitiesPermission := make(EntitiesPermission)
-	entitiesOperationDetails := make(EntitiesOperationDetails[Operation])
-
-	for _, entityType := range entityTypes {
-		ops, _, err := config.GetEntityPermissions(entityType)
-		if err != nil {
-			return nil, nil, fmt.Errorf("failed to get permissions for %s: %w", entityType, err)
-		}
+func (pc *PermissionConfig) GetAuthOperations(entityType string) (map[string]string, map[string]string, error) {
```
arvindh123: Is it possible to reuse the existing `GetPermission(et string, op K)` , instead of GetAuthOperations

Or 

Could please explain the reason for choosing having the new function `GetAuthOperations`

arvindh123 (changes requested):

pkg/permissions/config.go
nyagamunene: PATs requires us to extract `auth_operation` from the permission yaml file.

internal/proto/auth/v1/auth.proto
```diff
@@ -19,40 +19,28 @@ message AuthNReq {
...
+  uint32 operation = 11;
 }
 
 message AuthZReq {
-  oneof auth_type {
-    PolicyReq policy = 1;         // Policy-based authorization
-    PATReq pat = 2;               // PAT authorization
-  }
+  PolicyReq policy = 1;
 }
```
arvindh123: I think we can convert  `PolicyReq` to `AuthZReq`, Since `AuthZReq` have only one field

cmd/domains/main.go
```diff
@@ -317,20 +317,39 @@ func newDomainService(ctx context.Context, domainsRepo domainsSvc.Repository, ca
...
 	}
 
+	authOps := make(map[string]auth.Operation)
+	for opName, opInfo := range domainRoleOps {
+		if opInfo.AuthOperation != "" {
+			if authOp, err := auth.ParseOperation(opInfo.AuthOperation); err == nil {
+				authOps[opName] = authOp
+			}
+		}
+	}
```
arvindh123: `authOps` is needed now, Since we have remove `GetAuthOperations`?

arvindh123 (changes requested):

cmd/domains/main.go
nyagamunene: Yes. Its the operations used by the PATs

auth/pat.go
```diff
@@ -214,39 +312,159 @@ func (et *EntityType) UnmarshalText(data []byte) (err error) {
 	return err
 }
 
+var ValidOperationsForEntity = map[EntityType][]Operation{
+	ClientsType: {
+		ClientCreateOp,
+		ClientListOp,
+		ClientViewOp,
+		ClientUpdateOp,
+		ClientUpdateTagsOp,
```
arvindh123: Is it possible to reuse the `OperationDetails` for each entities

pkg/permissions/operations.go
```diff
@@ -35,6 +35,7 @@ type OperationName[K OperationKey] map[K]string
 type OperationDetails struct {
 	Name               string
 	PermissionRequired bool
+	AuthOperation      string
```
arvindh123: Does this `AuthOperation` is still required here ?

pkg/permissions/config.go
```diff
@@ -15,8 +15,12 @@ type PermissionConfig struct {
...
 type EntityPermissions struct {
-	Operations      []map[string]string `yaml:"operations"`
-	RolesOperations []map[string]string `yaml:"roles_operations"`
+	Operations      []map[string]interface{} `yaml:"operations"`
+	RolesOperations []map[string]interface{} `yaml:"roles_operations"`
+}
+
+type OperationInfo struct {
+	Permission Permission
 }
```
arvindh123: Now this struct is needed ?

arvindh123 (changes requested):

pkg/policies/service.go
```diff
@@ -52,16 +52,16 @@ type Policy struct {
...
-	Operation uint32 `json:"operation,omitempty"`
+	Operation string `json:"operation,omitempty"`
+	// UserID contains the user ID for PAT authorization.
+	UserID string `json:"user_id,omitempty"`
+	// EntityType contains the entity type for PAT authorization.
+	EntityType string `json:"entity_type,omitempty"`
+	// DomainID contains the domain ID for PAT authorization.
+	DomainID string `json:"domain_id,omitempty"`
 	// EntityID contains the entity ID for PAT authorization.
 	EntityID string `json:"entity_id,omitempty"`
```
arvindh123, line 64: This should be same as `PolicyReq` in `auth.proto`

arvindh123 (changes requested):

pkg/permissions/config.go
```diff
@@ -15,8 +15,8 @@ type PermissionConfig struct {
...
 
 type EntityPermissions struct {
-	Operations      []map[string]string `yaml:"operations"`
-	RolesOperations []map[string]string `yaml:"roles_operations"`
+	Operations      []map[string]interface{} `yaml:"operations"`
+	RolesOperations []map[string]interface{} `yaml:"roles_operations"`
```
arvindh123: I think this can be converted back to string, Because we strictly want string and don't want to have other values.

pkg/permissions/config.go
```diff
@@ -41,44 +41,30 @@ func (pc *PermissionConfig) GetEntityPermissions(entityType string) (map[string]
...
-
-		entitiesPermission[entityType] = ops
-
-		detailsFunc, ok := operationDetailsFuncs[entityType]
-		if !ok {
-			return nil, nil, fmt.Errorf("operation details function not found for entity type %s", entityType)
-		}
-
-		entitiesOperationDetails[entityType] = detailsFunc()
+func extractPermission(value interface{}) string {
```
arvindh123: If we are strictly having string, then i think this function is no more needed

arvindh123 (changes requested):

groups/middleware/authorization.go
```diff
@@ -382,46 +383,21 @@ func (am *authorizationMiddleware) checkSuperAdmin(ctx context.Context, session
...
-			pr.Operation = auth.UpdateOp
-		case groups.OpDeleteGroup,
-			groups.OpRemoveParentGroup,
-			groups.OpRemoveChildrenGroups,
-			groups.OpRemoveAllChildrenGroups:
-			pr.Operation = auth.DeleteOp
-		case domains.OpCreateDomainGroups:
-			pr.Operation = auth.CreateOp
+		pr.EntityType = auth.GroupsScopeStr
+		pr.Operation = auth.OperationStringForEntity(auth.GroupsType, permissions.Operation(op))
```
arvindh123: Is it possible to use  `am.entitiesOps.OperationName()` ?

arvindh123 (changes requested):

pkg/authz/authsvc/authz.go
```diff
@@ -83,20 +60,34 @@ func (a authorization) Authorize(ctx context.Context, pr authz.PolicyReq) error
...
+		Domain:          pr.Domain,
+		SubjectType:     pr.SubjectType,
+		SubjectKind:     pr.SubjectKind,
+		SubjectRelation: pr.SubjectRelation,
+		Subject:         pr.Subject,
+		Relation:        pr.Relation,
+		Permission:      pr.Permission,
+		Object:          pr.Object,
+		ObjectType:      pr.ObjectType,
 	}
```
arvindh123, line 78: I think we need to send both PAT and AuthZ request in gRPC. 

if the PAT exists , then `Authorize` function in `Auth` service will check `PAT` and do `AuhtZ` , 
else do only `AuthZ`. 

Please give your thought on this

nyagamunene: This would work but we will have to have the PAT fields then we can send both `PAT` and `AuthZ` of them together

arvindh123 (commented):
We have two policy object `Policy` at `pkg/policies/service.go` and `PolicyReq` in `pkg/authz/authz.go`

Does both are needed?

nyagamunene:
> We have two policy object `Policy` at `pkg/policies/service.go` and `PolicyReq` in `pkg/authz/authz.go`
> 
> Does both are needed?

I think we can remove one of them

felixgateru:
> We have two policy object `Policy` at `pkg/policies/service.go` and `PolicyReq` in `pkg/authz/authz.go`
> 
> Does both are needed?

We do not need both. With the cahnges made to the pr it would be now easier to use one object instead of the nested object that was defined here:
image

arvindh123:
> `pkg/policies/service.go`

`Policy` in `pkg/policies/service.go`  is abstraction of Policy engines. 
So we can plugin any other policy engine like Keto or OpenFGA

auth/pat.go
```diff
@@ -11,114 +11,176 @@ import (
...
-	shareOpStr     = "share"
-	UnshareOpStr   = "unshare"
-	PublishOpStr   = "publish"
-	SubscribeOpStr = "subscribe"
+	MessagePublishOp Operation = iota + 500
+	MessageSubscribeOp
 )
 
-func (op Operation) String() string {
+func OperationString(op Operation) string {
```
arvindh123: This `OperationString` function is used no where 
Does it needed ? 

And also `lookupServiceOperationName` and `lookupRoleOperationName` ?

arvindh123 (changes requested):

auth/pat.go
```diff
@@ -12,135 +12,72 @@ import (
...
+	OpSubscribe = "subscribe"
 
-func (op *Operation) UnmarshalJSON(data []byte) error {
-	str := strings.Trim(string(data), "\"")
-	val, err := ParseOperation(str)
-	*op = val
-	return err
-}
+	OpMessagePublish   = "message_publish"
+	OpMessageSubscribe = "message_subscribe"
```
arvindh123, line 44: Still these const are needed? Since we are using the Operation name of entities

arvindh123 (changes requested):

auth/pat.go
nyagamunene: > Still these const are needed? Since we are using the Operation name of entities

These are used in unmarshalling the JSON API requests

arvindh123 (approved):

dborovcanin (approved):

PR #3301, opened 2025-12-24
NOISSUE - Improve JWKS

dborovcanin:
# What type of PR is this?
This is a fix and improvement on the current JWKS design.

## What does this do?
This updates JWKS handling and abstractions to make it simpler to work with and safer in case of key rotation.

There is no such issue.

Yes.

N/A

### Notes
N/A


docker/docker-compose.yaml
```diff
@@ -121,7 +121,7 @@ services:
       SMQ_AUTH_ACCESS_TOKEN_DURATION: ${SMQ_AUTH_ACCESS_TOKEN_DURATION}
       SMQ_AUTH_REFRESH_TOKEN_DURATION: ${SMQ_AUTH_REFRESH_TOKEN_DURATION}
       SMQ_AUTH_KEYS_ALGORITHM: ${SMQ_AUTH_KEYS_ALGORITHM}
-      SMQ_AUTH_KEYS_PRIVATE_KEY_PATH: ${SMQ_AUTH_KEYS_PRIVATE_KEY_PATH:+/keys/private.key}
+      SMQ_AUTH_KEYS_PRIVATE_KEY_PATH: ${SMQ_AUTH_KEYS_PRIVATE_KEY_PATH:+/keys/}
```
arvindh123: This can renamed to directory olike `SMQ_AUTH_PRIVATE_KEYS_DIRECTORY_PATH`

auth/key_manager.go
```diff
@@ -4,47 +4,104 @@
...
-// KeyManager represents a manager for JWT keys.
+// PublicKeyInfo represents a public key for external distribution via JWKS.
+// This follows RFC 7517 (JSON Web Key) specification.
+type PublicKeyInfo struct {
+	KeyID     string `json:"kid"`
+	KeyType   string `json:"kty"`
+	Algorithm string `json:"alg"`
+	Use       string `json:"use,omitempty"`
+
+	Key any `json:"-"`
```
arvindh123: Why this should be `any` ? Is it possible to have type here?

arvindh123 (changes requested):

auth/key_manager.go
dborovcanin: It is. I have changed to any just in the last commit, in fact: https://github.com/absmach/supermq/pull/3301/changes/3686dacfb1a84c300d42d6efbee65374c5d295a3

The reason I did that is that different algorithms have different relevant fields, but maybe this is an overegineering since we will likely not support more than one algorithm and one key at a time anyways. I can revert this change if you insist. Actually, let me do that regardless. This is an unnecessary abstraction.

cmd/auth/main.go
```diff
@@ -258,6 +265,30 @@ func initSchema(ctx context.Context, client *authzed.ClientWithExperimental, sch
 	return nil
 }
 
+// validateKeyConfig validates key configuration based on the algorithm type.
```
arvindh123: Comments can be removed, we dont have like this comments in main

cmd/auth/main.go
arvindh123, line 153: May we can have enum here for Symmetric and Asymmetric keys

cmd/auth/main.go
```diff
@@ -258,6 +265,30 @@ func initSchema(ctx context.Context, client *authzed.ClientWithExperimental, sch
...
+func validateKeyConfig(isSymmetric bool, cfg config) error {
+	if isSymmetric {
+		// Validate symmetric key configuration
+		if len(cfg.SecretKey) < 32 {
+			return fmt.Errorf("symmetric key must be at least 32 bytes, got %d bytes", len(cfg.SecretKey))
+		}
+		if cfg.SecretKey == "secret" {
```
arvindh123, line 269: Either we should check for lot other insecure keys or we can simply remove this case , giving the responsibility to platform admin

cmd/auth/main.go
```diff
@@ -69,7 +69,7 @@ type config struct {
 	AccessDuration                time.Duration `env:"SMQ_AUTH_ACCESS_TOKEN_DURATION"             envDefault:"1h"`
 	RefreshDuration               time.Duration `env:"SMQ_AUTH_REFRESH_TOKEN_DURATION"            envDefault:"24h"`
 	KeyAlgorithm                  string        `env:"SMQ_AUTH_KEYS_ALGORITHM"                    envDefault:"EdDSA"`
-	PrivateKeyPath                string        `env:"SMQ_AUTH_KEYS_PRIVATE_KEY_PATH"             envDefault:"./ssl/keys/private.pem"`
+	PrivateKeyPath                string        `env:"SMQ_AUTH_KEYS_PRIVATE_KEY_DIR"              envDefault:"./keys/"`
```
arvindh123: May we be can change the env name to `PrivateKeysDirPath`

auth/api/http/keys/responses.go
```diff
@@ -76,20 +75,18 @@ func (res revokeKeyRes) Empty() bool {
...
+	CacheStaleWhileRevalidate int                  `json:"-"`
 }
 
 func (res retrieveJWKSRes) MarshalJSON() ([]byte, error) {
-	set := jwk.NewSet()
-	for _, k := range res.Keys {
-		if err := set.AddKey(k.Key()); err != nil {
-			return nil, err
-		}
+	// Convert PublicKeyInfo to JWK Set format
```
arvindh123: Can we remove these comments

auth/key_manager.go
```diff
@@ -5,46 +5,53 @@ package auth
...
+	KeyType   string `json:"kty"`
+	Algorithm string `json:"alg"`
+	Use       string `json:"use,omitempty"`
 
-// NewJWK creates a new JWK from a jwk.Key.
-func NewJWK(key jwk.Key) JWK {
-	return JWK{key: key}
-}
+	// EdDSA (Ed25519) fields
+	Curve string `json:"crv,omitempty"`
```
arvindh123, line 26: Why don't we have separate object for this? 
Why this should be in struct that going to be returned in Public ? 
Is it safe to expose these information in public ? 
Is this needed for Public JWKS

arvindh123 (changes requested):

auth/jwt/tokenizer.go
```diff
@@ -14,27 +14,34 @@ import (
 )
 
 var (
-	// errInvalidIssuer is returned when the issuer is not supermq.auth.
-	errInvalidIssuer = errors.New("invalid token issuer value")
+	// ErrInvalidIssuer is returned when the issuer is not supermq.auth.
+	ErrInvalidIssuer = errors.New("invalid token issuer value")
+
 	// errInvalidType is returned when there is no type field.
```
arvindh123: I think we can remove comment for un exported errors

auth/tokenizer/asymmetric/rotation_test.go
```diff
@@ -0,0 +1,329 @@
+// Copyright (c) Abstract Machines
```
arvindh123, line 1: Do we need this test file ? 
The  file name is rotation_test.go 
So are we rotating the keys ?

auth/tokenizer/symmetric/tokenizer.go
```diff
@@ -30,18 +31,54 @@ func NewKeyManager(algorithm string, secret []byte) (auth.KeyManager, error) {
 	}, nil
 }
 
-func (km *manager) SignJWT(token jwt.Token) ([]byte, error) {
-	return jwt.Sign(token, jwt.WithKey(km.algorithm, km.secret))
+func (km *manager) Sign(key auth.Key) (string, error) {
+	builder := jwt.NewBuilder()
```
arvindh123: The builder is common for both Asymmetric and Symmetric keys,
So this logic can be moved back to `tokenizer.go`

arvindh123 (changes requested):

auth/tokenizer/symmetric/tokenizer.go
```diff
@@ -30,18 +31,54 @@ func NewKeyManager(algorithm string, secret []byte) (auth.KeyManager, error) {
...
 		jwt.WithKey(km.algorithm, km.secret),
 	)
+	if err != nil {
+		return auth.Key{}, err
+	}
+
+	// Validate issuer
+	if tkn.Issuer() != smqjwt.IssuerName {
+		return auth.Key{}, smqjwt.ErrInvalidIssuer
+	}
```
arvindh123, line 77: We didn't had this logic before, So our previous code are vulnerable ?

auth/key_manager.go
dborovcanin: We don't have them as separate object to avoid custom JSON marshaling - they should be embedded in the response according to standard. It is safe to expose, `crv` essentially identifies algorithm (like Ed255519 is algorithm in  Edwards-curve Digital Signature Algorithm (EdDSA) scheme), and X is practically the "value" of the public key. You can check for examples: https://datatracker.ietf.org/doc/html/rfc7517#page-25 and https://datatracker.ietf.org/doc/html/rfc7517#section-3.

auth/keymanager/asymmetric/key_manager.go
```diff
@@ -24,64 +27,178 @@ var (
...
+	activeKeyID string
+	// Field keys is populated during initialization and never modified afterward,
+	// making it safe for concurrent reads from Sign(), Verify(), and PublicKeys().
+	keys map[string]*keyPair
 }
 
 var _ auth.KeyManager = (*manager)(nil)
 
-// NewKeyManager creates a new asymmetric key manager that loads the private key from a file.
 func NewKeyManager(privateKeyPath string, idProvider supermq.IDProvider) (auth.KeyManager, error) {
```
arvindh123: `privateKeyPath` can renamed to `privateKeyDir`
If we are going to support multiple private keys

auth/tokenizer/asymmetric/rotation_test.go
dborovcanin: Yes, let's keep it. This tests the way manager works if we have more than one key, and one of them is retiring/retired/expired.

auth/tokenizer/symmetric/tokenizer.go
dborovcanin: It is accessing internal fields of key manager, so it would make it too difficult. What we can do is consider removing Tokenizer completely; i.e. merging Tokenizer and Manager.

docker/keys/keys.json
```diff
@@ -0,0 +1,19 @@
+{
+  "active_key_id": "active",
```
arvindh123: Instead of this file, we can  get private and public key pair from platform admin

arvindh123: So there will be no need to keys.json file and we also no need to derive the public key from Private key 
We can check the public key and private key are matching

arvindh123 (changes requested):

auth/tokenizer/symmetric/tokenizer.go
dborovcanin: Well, we should check it. It's not a huge risk, but it's a nice addition.

pkg/authn/jwks/authn.go
arvindh123, line 4: we can rename this package to `asym` or some other better name instead of `jwks`

pkg/authn/jwks/authn.go
```diff
@@ -105,45 +129,63 @@ func (a authentication) Authenticate(ctx context.Context, token string) (authn.S
...
-		cached := jwksCache.jwks
-		jwksCache.RUnlock()
-		return cached, nil
+func isSignatureError(err error) bool {
+	return !errors.Contains(err, errJWTExpiryKey) &&
+		!errors.Contains(err, errInvalidIssuer) &&
+		!errors.Contains(err, smqauth.ErrExpiry)
+}
+
+func (a authentication) fetchJWKS(ctx context.Context, forceRefresh bool) (jwk.Set, error) {
```
arvindh123, line 138: I guess some library will be having this function 
But if we going to be like `NATS team`, then we can have this functions

arvindh123 (changes requested):

auth/tokenizer/asymmetric/tokenizer.go
```diff
@@ -0,0 +1,252 @@
...
+		builder.Subject(key.Subject)
+	}
+	if key.ID != "" {
+		builder.JwtID(key.ID)
+	}
+
+	tkn, err := builder.Build()
+	if err != nil {
+		return "", err
+	}
```
arvindh123: Can we abstract this function to common , because this is used between asymmetric and symmetric ?   
   
We can build a abstract object, that can be used between asymmetric and symmetric ?

auth/tokenizer/symmetric/tokenizer.go
```diff
@@ -0,0 +1,101 @@
...
+		return "", err
+	}
+
+	return string(signedBytes), nil
+}
+
+func (km *tokenizer) Parse(ctx context.Context, tokenString string) (auth.Key, error) {
+	if len(tokenString) >= 3 && tokenString[:3] == patPrefix {
+		return auth.Key{Type: auth.PersonalAccessToken}, nil
+	}
```
arvindh123, line 61: Is this needed ? 
I think this check happens in before layer

auth/tokenizer/asymmetric/tokenizer.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+		logger.Info("loaded retiring key for rotation grace period", slog.String("key_id", retiringKID))
+	}
+
+	return mgr, nil
+}
+
+func keyIDFromPath(path string) string {
+	base := filepath.Base(path)
+	ext := filepath.Ext(base)
```
arvindh123: IF the both file have same name, then this will not work 
May we need to add the file hash here or some other thing

docker/.env
```diff
@@ -101,7 +101,8 @@ SMQ_AUTH_DB_SSL_ROOT_CERT=
 SMQ_AUTH_ACCESS_TOKEN_DURATION="1h"
 SMQ_AUTH_REFRESH_TOKEN_DURATION="24h"
 SMQ_AUTH_KEYS_ALGORITHM="EdDSA"
-SMQ_AUTH_KEYS_PRIVATE_KEY_PATH="./ssl/keys/private.key"
+SMQ_AUTH_KEYS_ACTIVE_KEY_PATH="./keys/active.key"
+SMQ_AUTH_KEYS_RETIRING_KEY_PATH="./keys/retiring.key"
```
arvindh123, line 105: Can we provide option for retiring key auto delete duration ? 

Since we are already recommending  below:

- **Recommended:** 168 hours (7 days)
- **Minimum:** 24 hours
- **Maximum:** 720 hours (30 days)

arvindh123 (changes requested):

auth/api/http/keys/responses.go
```diff
@@ -76,20 +75,17 @@ func (res revokeKeyRes) Empty() bool {
 }
 
 type retrieveJWKSRes struct {
-	Keys                      []auth.JWK `json:"-"`
-	CacheMaxAge               int        `json:"-"`
-	CacheStaleWhileRevalidate int        `json:"-"`
+	Keys                      []auth.PublicKeyInfo `json:"-"`
+	CacheMaxAge               int                  `json:"-"`
+	CacheStaleWhileRevalidate int                  `json:"-"`
```
arvindh123, line 80: Is it possible to config cache related settings via env?

arvindh123 (changes requested):

cmd/auth/main.go
```diff
@@ -159,17 +159,23 @@ func main() {
...
 	idProvider := uuid.New()
 
-	var keyManager auth.KeyManager
+	if err := validateKeyConfig(isSymmetric, cfg); err != nil {
+		logger.Error(fmt.Sprintf("invalid key configuration: %s", err.Error()))
+		exitCode = 1
+		return
+	}
+
+	var keyManager auth.Tokenizer
```
arvindh123: Can we make this variable name better, or rename the struct `auth.Tokenizer`.

arvindh123 (changes requested):

cmd/auth/main.go
```diff
@@ -258,7 +264,34 @@ func initSchema(ctx context.Context, client *authzed.ClientWithExperimental, sch
...
+		if cfg.SecretKey == "secret" {
+			return fmt.Errorf("default secret key is insecure - please set SMQ_AUTH_SECRET_KEY environment variable")
+		}
+		return nil
+	}
+
+	// Validate active key path
+	_, err := os.Stat(cfg.ActiveKeyPath)
+	if err != nil {
+		if os.IsNotExist(err) {
```
arvindh123, line 278: Is this needed  `os.IsNotExist`?

cmd/auth/main.go
```diff
@@ -258,7 +264,34 @@ func initSchema(ctx context.Context, client *authzed.ClientWithExperimental, sch
...
+			return fmt.Errorf("active key file not found: %s - please set SMQ_AUTH_KEYS_ACTIVE_KEY_PATH", cfg.ActiveKeyPath)
+		}
+		return fmt.Errorf("failed to access active key file: %w", err)
+	}
+
+	// Retiring key is optional - only validate if path is provided
+	if cfg.RetiringKeyPath != "" {
+		if _, err := os.Stat(cfg.RetiringKeyPath); err != nil {
```
arvindh123, line 286: Why don't here not checking `os.IsNotExist` ?

arvindh123 (changes requested):

cmd/auth/main.go
dborovcanin: Yes, active key must always exist.

cmd/auth/main.go
dborovcanin: This should not return an error, fixed.

auth/api/http/keys/responses.go
dborovcanin: These are configured using env vars, and returned as headers, so we ignore them in JSON body of response.

auth/tokenizer/symmetric/tokenizer.go
arvindh123: The issuer and the verifier are the same service here. 
There is chance to tamper or mint unless the HMAC is shared 

In Asymmetric case :
The issuer is used by the client to validate JWT is issued by the client expected issuer.
JWT may be issued by have multiple issuers. 
Example 
Mobile related service create JWT with `iss: mobile` and request authority to sign the JWT 
Web related service create JWT with `iss: web` and request authority  to sign the JWT

The asymmetric key is used to sign the JWT.
On the receiving side, the client verifies that the issuer matches one of the expected issuers before accepting the token.

cmd/auth/main.go
arvindh123: I think , like how we do for active keys , here also we need to check for `os.IsNotExist`

arvindh123 (dismissed):

docker/.env
dborovcanin: Let's not do it now. It hurts nobody and would require extra complexity.

auth/tokenizer/symmetric/tokenizer.go
dborovcanin: It is used in Identify method of the service. It is an ugly and implicit way to do it, though. Opening an issue for this.

PR #3312, opened 2026-01-15
NOISSUE - Fix publisher id on message published by user

### LINKED ISSUE #3311: Bug: Publisher on message sent by non domain admin has domain id in publisher

View channel messages

The publisher for a message sent via http by a non admin user should have the userID only as the publisher

The publisher is the userID concatenated with the domainID

### Steps To Reproduce

1. Send message via HTTP with a non admin user
2. Use readers to view messages

Local docker compose deployment

Image

felixgateru:
# What type of PR is this?
This is a bug fix because it fixes the following issue: #3311

## What does this do?
This pr fixes the issue with the publisher having the domain id

- Resolves #3311

No

No

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -230,7 +230,7 @@ func (h *handler) authAccess(ctx context.Context, username, password, domainID, 
 		clientType = policies.ClientType
 	}
 
-	id, err := h.authenticate(ctx, clientType, token, domainID)
+	id, subject, err := h.authenticate(ctx, clientType, token, domainID)
 	if err != nil {
 		return "", mgate.NewHTTPProxyError(http.StatusUnauthorized, errors.Wrap(svcerr.ErrAuthentication, err))
 	}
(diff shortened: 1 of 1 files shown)


http/handler.go
```diff
@@ -258,29 +258,29 @@ func (h *handler) authAccess(ctx context.Context, username, password, domainID,
 	return id, nil
 }
 
-func (h *handler) authenticate(ctx context.Context, authType, token, domainID string) (string, error) {
+func (h *handler) authenticate(ctx context.Context, authType, token, domainID string) (string, string, error) {
```
dborovcanin, line 261: This looks a bit obscure now. Let's keep a logic clear - this method returns an ID or error rather than 3 values. Construct subject in the caller method if needed.
```go
id, err := h.authenticate
...
if clientType == policies.UserType {
id = policies.EncodeDomainUserID(domainID, authnSession.UserID)
}
....
```

dborovcanin (changes requested):

http/handler.go
felixgateru: the subject for authorization depends on whether the client of UserType is an admin or not an admin, that is why i proposed the 3 values. Otherwise we would need to return the role from the session to construct the subject in the caller

dborovcanin: Let's not return role. Can we always return subject from here in that case?

dborovcanin: I.e. does that make any difference for health check (where ID is used)?

felixgateru: it makes a difference in the publisher field for the message payload. If we always return the subject we get something like this:
image

dborovcanin (approved):

PR #3320, opened 2026-01-21
SMQ-3298 -  Add support for symmetric key rotation

### LINKED ISSUE #3298: Add a feature to rotate the symmetrical key
Can we provide feature to rotate the symmetrical key too? 

Not in this PR, May be in future PR

If yes, Can we add in  the next PR "rotation of both symmetrical and asymmetrical keys" ?

_Originally posted by @arvindh123 in https://github.com/absmach/supermq/pull/3228#discussion_r2639723473_

felixgateru:
# What type of PR is this?
This is a feature because it adds support for symmetric key rotation

## What does this do?
This pr allows zero downtime symmetric key rotation

- Resolves #3298

Yes

Yes


dborovcanin:
This is not needed for symmetric key - it's offloaded to the infrastructure. We do support 2 keys for 0-downtime rotation, but it's up to the deployment to decide how and when to do the actual rotation.

PR #3323, opened 2026-01-26
SMQ-2601 - Remove user from domain

### LINKED ISSUE #2601: Remove user from domain

felixgateru:
# What type of PR is this?
This is a feature because it enables a user to be removed from a domain

## What does this do?
This pr:
- implements repo methods to delete all user roles 

- Resolves #2601

No

No


pkg/roles/provisionmanage.go
```diff
@@ -226,6 +227,35 @@ func (r ProvisionManageService) AddNewEntitiesRoles(ctx context.Context, domainI
...
+		}
+
+		if err := r.repo.RemoveMemberFromDomain(ctx, domainID, memberID); err != nil {
+			return err
+		}
+		return nil
+	default:
+		return errInvalidOperation
+	}
+}
```
arvindh123, line 253: This should be generic, so that it should work with `MG`

arvindh123 (changes requested):

pkg/roles/roles.go
```diff
@@ -159,6 +159,8 @@ type RoleManager interface {
 
 	RemoveEntityMembers(ctx context.Context, session authn.Session, entityID string, members []string) (err error)
 
+	RemoveMemberFromDomain(ctx context.Context, domainID, memberID string) error
```
arvindh123, line 162: We could not have this `RemoveMemberFromDomain` because this have logic related to particular entity `domain`. 
We should handle this via Entity Event Consumers ( for example in clients, channels, groups service , where it consumes domain delete event)

arvindh123 (changes requested):

dborovcanin:
Atom made this obsolete.

PR #3327, opened 2026-01-28
NOISSUE - Add MG fine grain control

nyagamunene:
# What type of PR is this?

- Related Issue #https://github.com/absmach/magistrala/issues/370
- Resolves #


arvindh123 (changes requested):
This should done in MG and not SuperMQ

Please check the possibilities to fit this in MG

arvindh123:
These update should be in MG not in SMQ

PR #3329, opened 2026-01-30
SMQ-2603 - Remove user from all entities on user deletion

felixgateru:
# What type of PR is this?

- Related Issue #
- Resolves #


SammyOina (commented):
If possible we should go more level on spice db  to delete the graph and all branches related to the user entity so this can work with modified spicedb schemas

dborovcanin:
Atom made this obsolete.

PR #3332, opened 2026-02-02
NOISSUE - Refactor PATs to support MG entities

nyagamunene:
# What type of PR is this?

- Related Issue #
- Resolves #


arvindh123:
These update should be in MG not in SMQ

PR #3393, opened 2026-03-12
NOISSUE - Refactor PAT to use yaml

nyagamunene:
# What type of PR is this?

This is a refactor because it changes the following functionality: 
- It add new PAT yaml file
- It removes the hardcorded entity type and entity operations

- Related Issue #https://github.com/absmach/magistrala/issues/440
- Resolves #


pkg/permissions/entities.go
```diff
@@ -18,6 +18,7 @@ type EntitiesOperations[K OperationKey] interface {
 	GetPermission(et string, op K) (Permission, error)
 	GetPermissionAndRequired(et string, op K) (Permission, bool, error)
 	OperationName(et string, op K) string
+	PATOperationName(et string, op K) string
```
arvindh123, line 21: We have single operation , operations are belongs to service 

Example: 
Client service will have single create operation 

This operation will be call either if the request have Bearer token or PAT Toekn 
There is nothing like PAT Operation Name

arvindh123 (changes requested):

dborovcanin:
We switched to Atom, that will hold Access Tokens. This PR is obsolete.

PR #3416, opened 2026-04-01
SMQ-3415 - Return correct error in auth gRPC response

### LINKED ISSUE #3415: Auth token errors cause redirect loop due to internal server error instead of 401

Logging in to SuperMQ cloud via the auth service. Expected to get a valid authentication token or a clear error message if authentication fails.

If a token is revoked or invalid, or other authentication-related errors occur, the API should return 401 Unauthorized instead of 500 Internal Server Error, thus avoiding browser redirect loops or misleading error states.

The login endpoint returns a 500 Internal Server Error when gRPC error (e.g., "token is revoked") occurs in token-related functions like RefreshToken and IssueToken. The browser displays a redirect loop (ERR_TOO_MANY_REDIRECTS) as seen in the attached screenshot. The gRPC error is wrapped as Internal and bubbles up as an unhandled 500.

### Steps To Reproduce

1. Attempt to log in via SuperMQ's auth service with a revoked/expired token or a token that should be invalid.
2. Observe a 500 Internal Server Error and browser redirect loop.
3. See that the error traces back to token-related gRPC errors in the API backend.

See screenshot for reference:

Auth error screen with redirect loop

Production environment (cloud.magistrala.absmach.eu), browser-based login flow. SuperMQ cloud / users service.

Token-related functions like RefreshToken (https://github.com/absmach/supermq/blob/main/users/service.go#L222-L236) and IssueToken (https://github.com/absmach/supermq/blob/main/users/service.go#L186-L220) currently propagate certain gRPC errors (like "token is revoked") as Internal, causing API to respond with 500 instead of wrapping appropriately for 401 responses.

**Solution:**
- Inspect gRPC error code/message in all token-related functions.
- If the error is authentication-related (e.g. token revoked/invalid), wrap with `ErrAuthentication` from [pkg/errors/service/types.go#L11](https://github.com/absmach/supermq/blob/main/pkg/errors/service/types.go#L11). Example: `ErrAuthentication = errors.NewAuthNError("failed to perform authentication over the entity")`.
- With this wrapper, the API layer will automatically return 401 Unauthorized.
- Apply this pattern to all token-related functions to ensure consistent error reporting and fix the redirect loop.

References:
- [users/service.go - RefreshToken](https://github.com/absmach/supermq/blob/main/users/service.go#L222-L236)
- [users/service.go - IssueToken](https://github.com/absmach/supermq/blob/main/users/service.go#L186-L220)
- [pkg/errors/service/types.go#L11](https://github.com/absmach/supermq/blob/main/pkg/errors/service/types.go#L11)

This will prevent misleading 500 errors, provide accurate feedback to the client, and stop redirect loops in the authentication flow.

arvindh123:
# What type of PR is this?

- Related Issue #3415
- Resolves #3415

### DIFF
--- a/auth/service.go
+++ b/auth/service.go
@@ -331,7 +331,7 @@ func (svc service) invitationKey(ctx context.Context, key Key) (Token, error) {
 func (svc service) refreshKey(ctx context.Context, token string, key Key) (Token, error) {
 	k, err := svc.tokenizer.Parse(ctx, token)
 	if err != nil {
-		return Token{}, errors.Wrap(errRetrieve, err)
+		return Token{}, errors.Wrap(svcerr.ErrAuthentication, err)
 	}
 	if k.Type != RefreshKey {
 		return Token{}, errIssueUser
--- a/auth/tokenizer/symmetric/tokenizer.go
+++ b/auth/tokenizer/symmetric/tokenizer.go
@@ -9,7 +9,6 @@ import (
 	"github.com/absmach/magistrala/auth"
 	smqjwt "github.com/absmach/magistrala/auth/tokenizer/util"
 	"github.com/absmach/magistrala/pkg/errors"
-	svcerr "github.com/absmach/magistrala/pkg/errors/service"
 	"github.com/lestrrat-go/jwx/v2/jwa"
 	"github.com/lestrrat-go/jwx/v2/jwt"
 )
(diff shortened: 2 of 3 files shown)


dborovcanin:
@arvindh123 What's the status here?

dborovcanin (dismissed):

dborovcanin:
@arvindh123 Please remove unused error (address CI comment).

PR #3423, opened 2026-04-02
NOISSUE - Fix listing members for `rule` and `reports` methods

nyagamunene:
# What type of PR is this?

- Related Issue #
- Resolves #

### DIFF
--- a/pkg/roles/repo/postgres/roles.go
+++ b/pkg/roles/repo/postgres/roles.go
@@ -45,6 +45,10 @@ func NewRepository(db postgres.Database, entityType, tableNamePrefix, entityTabl
 		membersListBaseQuery = groupMembersListBaseQuery()
 	case policies.DomainType:
 		membersListBaseQuery = domainMembersListBaseQuery()
+	case policies.RulesType:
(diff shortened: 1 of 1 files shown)


dborovcanin:
@arvindh123 Please review.

pkg/roles/repo/postgres/roles.go
```diff
@@ -45,6 +45,10 @@ func NewRepository(db postgres.Database, entityType, tableNamePrefix, entityTabl
 		membersListBaseQuery = groupMembersListBaseQuery()
 	case policies.DomainType:
 		membersListBaseQuery = domainMembersListBaseQuery()
+	case policies.RulesType:
+		membersListBaseQuery = rulesMembersListBaseQuery()
+	case policies.ReportsType:
+		membersListBaseQuery = reportsMembersListBaseQuery()
```
arvindh123, line 51: This can be optimized generic query with passing entity type and required action filter 
We can do this in separate PR 

We should think, how the external service attach to policy can able to list their entities with action in filter

arvindh123 (approved):

PR #3467, opened 2026-04-09
NOISSUE - Add transient retry to auth

dborovcanin:
# What type of PR is this?
This is an improvement adding transient retry to auth calls.

## What does this do?
This avoids latency in auth calls.

There is no such issue.

Yes.

N/A

### Notes
N/A


dborovcanin:
This is obsolete.

PR #3472, opened 2026-04-13
NOISSUE - Fix `refreshKey` method

nyagamunene:
# What type of PR is this?

This is a bug fix because it fixes refreshed access tokens being issued with the refresh token ID.

## What does this do?

This PR fixes the refresh token flow so that the newly issued access token does not inherit the refresh token ID.

The refresh token ID is now only kept on the refreshed refresh token, where it is needed for active-token tracking and revocation. This prevents refreshed access tokens from being treated like PAT-backed sessions during authorization.

It also moves the refresh-token ID regression coverage into `TestIssue` and adds an edge case for refresh-token issuance failure.

- Related Issue #
- Resolves #

Yes, I updated `TestIssue` to cover the refresh-token flow and verify that the access token is issued without the refresh token ID while the refresh token keeps it.

No, this is a bug fix and does not add or modify a user-facing feature.

### Notes

This does not add PAT entity type support for reports or rules. That will be handled in a separate PR.


JeffMboya (dismissed):
LGTM

auth/service.go
```diff
@@ -343,14 +343,15 @@ func (svc service) refreshKey(ctx context.Context, token string, key Key) (Token
 	if !ok {
 		return Token{}, ErrRevokedToken
 	}
-	key.ID = k.ID
+	refreshID := k.ID
 	key.Type = AccessKey
+	key.ID = ""
```
arvindh123: Why we are sending empty key.ID ? 
If it is empty key.ID, then auth/tokenizer/util/jwt.go BuildToken will not add key id in token 
Does the key id is not requierd for building the refresh token?

JackMangBana (approved):

arvindh123 (approved):

PR #3476, opened 2026-04-15
NOISSUE - Update bootstrap and provision service

nyagamunene:
# What type of PR is this?

- Related Issue #
- Resolves #


bootstrap/service.go
```diff
@@ -229,35 +242,31 @@ func (bs bootstrapService) UpdateConnections(ctx context.Context, session smqaut
 	}
 
 	cfg.Channels = channels
-	var connect, disconnect []string
 
+	nextState := Active
```
JeffMboya: The state toggle here is backwards. `nextState` is initialized to `Active`, then set to `Inactive` when the config is already `Active`. This means:
- Active config: disconnects from all channels and sets state to Inactive
- Inactive config: connects to the new channels and sets state to Active

Calling this endpoint twice on an Active config will silently deactivate a live device. This is a regression from the original behavior, which computed a delta (channels to add vs. remove) without touching state.

The removed `updateList()` function was doing the right thing. Suggest restoring the delta logic and removing the state toggle from `UpdateConnections` entirely. State changes belong in `ChangeState()`.

bootstrap/service.go
```diff
@@ -335,23 +344,12 @@ func (bs bootstrapService) ChangeState(ctx context.Context, session smqauthn.Ses
...
 	case Active:
-		for _, c := range cfg.Channels {
-			if err := bs.sdk.ConnectClients(ctx, c.ID, []string{cfg.ClientID}, []string{"Publish", "Subscribe"}, session.DomainID, token); err != nil {
-				// Ignore conflict errors as they indicate the connection already exists.
-				if errors.Contains(err, svcerr.ErrConflict) {
-					continue
-				}
-				return ErrClients
-			}
+		if err := bs.connectChannels(ctx, session.DomainID, token, cfg.ClientID, bs.toIDList(cfg.Channels)); err != nil {
```
JeffMboya: **P1: ChangeState crashes on configs with no channels**

`bs.toIDList(cfg.Channels)` returns an empty slice when a config has no associated channels. That empty slice is passed directly to `connectChannels`/`disconnectChannels`, which call `sdk.Connect`/`sdk.Disconnect` with `ChannelIDs: []`. The channels service rejects that payload with a 400 Bad Request, so `ChangeState` permanently fails for any config that has zero channels.

Before calling the helpers, add:
```go
channelIDs := bs.toIDList(cfg.Channels)
if len(channelIDs) == 0 {
    return bs.configs.ChangeState(ctx, session.DomainID, id, state)
}
```

The same guard is also missing in `connectChannels`/`disconnectChannels` themselves (lines 396-412).

bootstrap/service.go
```diff
@@ -164,8 +165,21 @@ func (bs bootstrapService) Add(ctx context.Context, session smqauthn.Session, to
...
 	cfg.ClientID = mgClient.ID
 	cfg.DomainID = session.DomainID
-	cfg.State = Inactive
 	cfg.ClientSecret = mgClient.Credentials.Secret
+	cfg.State = Inactive
+
+	// Create actual connections via SDK if there are channels to connect
+	if len(toConnect) > 0 {
+		for _, channelID := range toConnect {
+			if err := bs.sdk.ConnectClients(ctx, channelID, []string{cfg.ClientID}, connTypes, session.DomainID, token); err != nil {
```
JeffMboya: The channel-connection loop in `Add()` has no rollback path. If channels 0..N-1 connect successfully but channel N fails, those already-connected channels are silently orphaned. The config record is never saved (the error returns early), but the live channel connections remain in the Channels service.

This is also the one place that still uses the old `sdk.ConnectClients()` per-channel API while the rest of the service has moved to the new batch `connectChannels` helper. The inconsistency is confusing and the new helper has the same gap (no rollback on partial failure either).

Fix: either use the batch helper here too and add a deferred disconnect-on-error cleanup, or document explicitly that `Add()` may leave dangling connections on failure and the caller must reconcile.

bootstrap/service.go
```diff
@@ -395,6 +393,22 @@ func (bs bootstrapService) DisconnectClientHandler(ctx context.Context, channelI
 	return nil
 }
 
+func (bs bootstrapService) connectChannels(ctx context.Context, domainID, token, clientID string, channelIDs []string) error {
```
JeffMboya: The old per-channel loops explicitly continued on `ErrConflict` (connect) and `ErrNotFound` (disconnect). Those are expected idempotent conditions (already connected, already gone). The new batch helpers propagate every SDK error as a hard failure.

This means:
- Connecting a channel that already exists results in a hard error instead of continuing
- Disconnecting a channel that does not exist results in a hard error instead of continuing

Both are regressions from the old behavior. Add guards before returning:
```go
func (bs bootstrapService) connectChannels(...) error {
    err := bs.sdk.Connect(...)
    if err != nil && !errors.Contains(err, svcerr.ErrConflict) {
        return err
    }
    return nil
}
```
Same pattern for `disconnectChannels` with `repoerr.ErrNotFound`.

bootstrap/postgres/init.go
```diff
@@ -78,31 +74,242 @@ func Migration() *migrate.MemoryMigrationSource {
...
+					`ALTER TABLE IF EXISTS configs ADD COLUMN IF NOT EXISTS client_secret CHAR(36)`,
+					`ALTER TABLE IF EXISTS configs ADD COLUMN IF NOT EXISTS magistrala_client TEXT`,
+					`ALTER TABLE IF EXISTS configs ADD COLUMN IF NOT EXISTS magistrala_secret TEXT`,
+					`ALTER TABLE IF EXISTS configs ADD COLUMN IF NOT EXISTS mainflux_client TEXT`,
+					`ALTER TABLE IF EXISTS configs ADD COLUMN IF NOT EXISTS mainflux_key TEXT`,
+					`UPDATE configs SET client_id = magistrala_client WHERE client_id IS NULL AND magistrala_client IS NOT NULL`,
+					`UPDATE configs SET client_secret = magistrala_secret WHERE client_secret IS NULL AND magistrala_secret IS NOT NULL`,
+					`UPDATE configs SET client_id = mainflux_client WHERE client_id IS NULL AND mainflux_client IS NOT NULL`,
+					`UPDATE configs SET client_secret = mainflux_key WHERE client_secret IS NULL AND mainflux_key IS NOT NULL`,
+					`ALTER TABLE IF EXISTS configs ALTER COLUMN client_id SET NOT NULL`,
```
JeffMboya: `configs_6` runs `ALTER COLUMN client_id SET NOT NULL` (and `client_secret SET NOT NULL`) immediately after an UPDATE that attempts to backfill from `thing_id`. If any row still has a NULL `client_id` after the UPDATE, either because the old column was also NULL or the UPDATE predicate did not cover it, the ALTER fails with a constraint violation and the migration is aborted.

The migration has no rollback. On a real database with legacy NULL rows this will hard-fail on upgrade and leave the schema in a partially-migrated state.

Fix: either run a `DELETE FROM configs WHERE client_id IS NULL` before the ALTER (with an explicit warning in release notes), or make the constraint `DEFAULT ''` with a follow-up migration to tighten it once data is confirmed clean.

bootstrap/postgres/init.go
```diff
@@ -78,31 +74,242 @@ func Migration() *migrate.MemoryMigrationSource {
...
+					`ALTER TABLE IF EXISTS connections ADD COLUMN IF NOT EXISTS domain_id VARCHAR(256)`,
+					`ALTER TABLE IF EXISTS connections ADD COLUMN IF NOT EXISTS client_id VARCHAR(36)`,
+					`ALTER TABLE IF EXISTS connections ADD COLUMN IF NOT EXISTS type SMALLINT`,
+					`ALTER TABLE IF EXISTS connections ADD COLUMN IF NOT EXISTS config_id TEXT`,
+					`ALTER TABLE IF EXISTS connections ADD COLUMN IF NOT EXISTS channel_owner VARCHAR(256)`,
+					`ALTER TABLE IF EXISTS connections ADD COLUMN IF NOT EXISTS config_owner VARCHAR(256)`,
+					`CREATE UNIQUE INDEX IF NOT EXISTS connections_pkey
+						ON connections (channel_id, domain_id, client_id, type)
+						WHERE client_id IS NOT NULL AND type IS NOT NULL`,
+					`INSERT INTO config_channels (channel_id, config_id, domain_id)
```
JeffMboya: `COALESCE(config_owner, channel_owner, domain_id, '')` falls back to an empty string `''` when all source columns are NULL. If `domain_id` has a foreign key constraint referencing the domains table, inserting `''` will fail unless `''` is a valid domain ID, which it almost certainly is not.

This means the migration silently drops any rows that had all-NULL ownership columns instead of migrating them, or hard-fails if the FK is enforced.

The safer approach: log (or INSERT into a quarantine table) rows where all owner columns are NULL, then exclude them from the INSERT with a `WHERE COALESCE(...) != ''` predicate. At minimum, document that orphaned rows will be skipped.

bootstrap/postgres/configs.go
```diff
@@ -568,10 +583,9 @@ func updateConnections(domainID, id string, connections []string, tx *sqlx.Tx) e
 	if len(connections) == 0 {
```
JeffMboya: `updateConnections` returns `nil` immediately when `len(connections) == 0`, but the caller (`UpdateConnections` in the service) may intend to clear all channel connections for a config. With an empty input, the existing `config_channels` rows for that config are never deleted. The config appears connected to channels it should no longer be connected to.

If "empty connections = remove all" is the intended semantics (which the new `UpdateConnections` service logic suggests), this early return is wrong. The function should execute a `DELETE FROM config_channels WHERE config_id = $1 AND domain_id = $2` before returning when the list is empty.

JeffMboya (dismissed):
LGTM

bootstrap/postgres/profiles.go
```diff
@@ -0,0 +1,210 @@
...
+func NewProfileRepository(db postgres.Database, log *slog.Logger) bootstrap.ProfileRepository {
+	return &profileRepository{db: db, log: log}
+}
+
+func (pr profileRepository) Save(ctx context.Context, p bootstrap.Profile) (bootstrap.Profile, error) {
+	q := `INSERT INTO profiles (id, domain_id, name, description, template_format, content_template, defaults, binding_slots, version, created_at, updated_at)
+		  VALUES (:id, :domain_id, :name, :description, :template_format, :content_template, :defaults, :binding_slots, :version, :created_at, :updated_at)`
+
+	now := time.Now().UTC()
+	p.Version = 0
```
rodneyosodo: **Bug: This resets the version to 0, overriding the service layer's explicit `p.Version = 1`.**

The service layer sets `p.Version = 1` before calling `Save()`, but this line overwrites it to `0`. Every newly created profile will have `version: 0` in both the database and API response, contradicting the schema default of `1`.

The first `Update` will bump it to 1 via `version = version + 1`, but the initial state is wrong.

**Fix:** Remove this line so the service-assigned version is preserved:

```go
// Remove: p.Version = 0
```

Or if a default is needed in the repo layer, use `p.Version = 1`.

bootstrap/postgres/configs.go
```diff
@@ -508,264 +336,79 @@ func buildRetrieveQueryParams(domainID string, clientIDs []string, filter bootst
...
 	if dbcfg.Name.Valid {
 		cfg.Name = dbcfg.Name.String
 	}
-
 	if dbcfg.Content.Valid {
 		cfg.Content = dbcfg.Content.String
 	}
-
+	if len(dbcfg.RenderContext) > 0 && string(dbcfg.RenderContext) != jsonNull {
+		_ = json.Unmarshal(dbcfg.RenderContext, &cfg.RenderContext)
```
rodneyosodo: **Bug: The `json.Unmarshal` error is silently discarded.**

A corrupted `render_context` column will produce a config with `nil` `RenderContext` instead of surfacing the data integrity issue. The same operation in `RetrieveAll()` (line 116) properly handles the error and returns an empty page.

**Fix:** Return the error, consistent with the surrounding code:

```go
if err := json.Unmarshal(dbcfg.RenderContext, &cfg.RenderContext); err != nil {
    return Config{}, errors.Wrap(svcerr.ErrViewEntity, err)
}
```

bootstrap/service.go
```diff
@@ -316,55 +289,85 @@ func (bs bootstrapService) Bootstrap(ctx context.Context, externalKey, externalI
...
+	return cfg, nil
+}
+
+func (bs bootstrapService) changeConfigStatus(ctx context.Context, domainID, id string, status Status) (Config, error) {
+	cfg, err := bs.configs.RetrieveByID(ctx, domainID, id)
+	if err != nil {
+		return Config{}, errors.Wrap(svcerr.ErrViewEntity, err)
+	}
+	if cfg.Status == status {
+		return Config{}, svcerr.ErrStatusAlreadyAssigned
```
rodneyosodo: **Behavioral change: This is a breaking change for idempotent enable/disable calls.**

The old `ChangeState` returned `nil` (success) when the state already matched. Now `EnableConfig` on an already-enabled config returns `ErrStatusAlreadyAssigned`.

Clients that idempotently call enable/disable (common in reconciliation loops and retry logic) will now receive errors they didn't before.

If this is intentional, consider documenting it in the PR description. If not, returning the current config without error would preserve backward compatibility:

```go
if cfg.Status == status {
    return cfg, nil
}
```

bootstrap/api/requests.go
```diff
@@ -154,10 +125,153 @@ func (req changeStateReq) validate() error {
...
+
+// --- Profile requests ---
+
+type createProfileReq struct {
+	bootstrap.Profile
+}
+
+func (req createProfileReq) validate() error {
+	if req.Name == "" {
+		return apiutil.ErrMissingID
```
rodneyosodo: **`ErrMissingID` is misleading when the missing field is `Name`.** This also applies to `uploadProfileReq.validate()` at line 150.

**Suggestion:** Use a more descriptive error:

```go
if req.Name == "" {
    return apiutil.ErrMissingName // or a dedicated error
}
```

bootstrap/api/requests.go
```diff
@@ -154,10 +125,153 @@ func (req changeStateReq) validate() error {
...
+	return nil
+}
+
+type updateProfileReq struct {
+	profileID string
+	bootstrap.Profile
+}
+
+func (req updateProfileReq) validate() error {
+	if req.profileID == "" || req.Name == "" {
```
rodneyosodo: **`Name` is required on every update, preventing partial updates.** This means callers cannot update only `content_template` without also resubmitting the current `Name`.

If this is intentional (i.e. all fields must be provided on update), it would be worth documenting. If partial updates should be supported, `Name` should only be required when actually changing the name:

```go
if req.profileID == "" {
    return apiutil.ErrMissingID
}
```

rodneyosodo (commented):
## Review Summary

Major refactor with profile/template-based system. Found 3 bugs and 3 structural issues:

**Bugs:**
1. Profile version is always 0 on creation — repo layer overrides service-assigned version
2. `json.Unmarshal` error for `RenderContext` silently swallowed in `toConfig()`
3. Breaking behavioral change: enable/disable now errors when status already matches

**Structural:**
4. `ErrMissingID` returned for missing profile `Name`
5. `updateProfileReq.validate()` requires `Name`, preventing partial updates
6. Unrelated test changes (`.Maybe()` removals) mixed into this PR

rodneyosodo (commented):
.

cmd/bootstrap/main.go
```diff
@@ -176,20 +173,16 @@ func main() {
...
-	svc, err := newService(ctx, authz, policySvc, db, tracer, logger, cfg, dbConfig)
+	svc, err := newService(ctx, authz, policySvc, database, tracer, logger, cfg)
 	if err != nil {
 		logger.Error(fmt.Sprintf("failed to create %s service: %s", svcName, err))
 		exitCode = 1
 		return
 	}
 
-	if err = subscribeToClientsES(ctx, svc, cfg, logger); err != nil {
-		logger.Error(fmt.Sprintf("failed to subscribe to clients event store: %s", err))
```
rodneyosodo, line 188: The subscribeToClientsES call was removed but no replacement was added. The pkg/bootstrap/events/consumer package (specifically BootstrapEventsSubscribe) exists and handles client.remove events, but it is never called from main.go.

**Impact:** When a client is deleted via the clients service, its associated bootstrap config will **not** be automatically removed. Orphaned bootstrap configs will accumulate over time. This is a regression from the previous behavior.

**Suggestion:** Wire up the event subscription after service creation using the new pkg/bootstrap/events/consumer package. Also, the log message on the next line ("Subscribed to Event Store") is now misleading since nothing is actually subscribed.

provision/service.go
```diff
@@ -362,40 +473,21 @@ func (ps *provisionService) recover(ctx context.Context, e *error, ths *[]smqSDK
 		return
 	}
 
-	if errors.Contains(err, ErrFailedBootstrapValidate) || errors.Contains(err, ErrFailedCertCreation) {
+	if errors.Contains(err, ErrFailedBootstrapValidate) || errors.Contains(err, ErrFailedCertCreation) || errors.Contains(err, ErrFailedBootstrapBinding) {
 		clean(ctx, ps, clients, channels, domainID, token)
```
rodneyosodo, line 477: ErrFailedBootstrap (returned when AddBootstrap fails) is **not handled** by any of the recover branches:

1. ErrFailedClientRetrieval || ErrFailedChannelCreation => cleans clients only
2. ErrFailedBootstrapValidate || ErrFailedCertCreation || ErrFailedBootstrapBinding => cleans clients, channels, bootstraps
3. ErrClientUpdate || ErrGatewayUpdate => cleans clients, channels, certs, bootstraps

**Scenario:** Provisioning creates a client and channels successfully, but AddBootstrap returns an error (e.g., duplicate external_id). The recover defer runs, but none of the errors.Contains checks match ErrFailedBootstrap. Result: orphaned clients and channels remain in the system.

**Suggestion:** Add ErrFailedBootstrap to the first branch since bootstraps have not been created yet when this error occurs:

```go
if errors.Contains(err, ErrFailedClientRetrieval) || errors.Contains(err, ErrFailedChannelCreation) || errors.Contains(err, ErrFailedBootstrap) {
    // clean clients and channels
    ...
}
```

rodneyosodo (commented):
## Review Summary

Two bugs found in this PR:

1. **Bootstrap service no longer subscribes to client remove events** (High) - The subscribeToClientsES call was removed from cmd/bootstrap/main.go but no replacement was added. The pkg/bootstrap/events/consumer package exists but is never wired up. Client deletions will leave orphaned bootstrap configs.

2. **Missing cleanup for ErrFailedBootstrap in provision recover** (Medium) - In provision/service.go, the recover defer handles various error cases but none match ErrFailedBootstrap. If AddBootstrap fails after clients/channels are created, those resources are not cleaned up.

provision/service.go
nyagamunene: `ErrFailedBootstrap` is already handled at `provision/service.go:471`. Moving it to the first branch (clients-only cleanup) would be incorrect — at the point AddBootstrap fails, channels have already been created and also need to be cleaned up. The current second branch calls clean() which removes both clients and channels, which is the right behavior.

rodneyosodo (dismissed):
LGTM

bootstrap/postgres/configs.go
```diff
@@ -381,117 +295,34 @@ func (cr configRepository) ChangeState(ctx context.Context, domainID, id string,
...
 	queries := []string{}
 
 	if len(clientIDs) != 0 {
-		queries = append(queries, fmt.Sprintf("magistrala_client IN ('%s')", strings.Join(clientIDs, "','")))
+		var arr pgtype.TextArray
+		if err := arr.Set(clientIDs); err != nil {
+			return "", nil
+		}
+		params = append(params, arr)
+		queries = append(queries, fmt.Sprintf("id = ANY($%d)", len(params)))
```
dborovcanin: This uses id in clientIDs, but in service, Config ID is no longer linked to client ID. Am I missing something?

dborovcanin (changes requested):
Running test in Bootstrap service fails with:
```
=== RUN   TestEnableConfig/enable_already_active_config
    service_test.go:940: 
                Error Trace:    /home/dusan/go/src/github.com/absmach/magistrala/bootstrap/service_test.go:940
                Error:          Should be true
                Test:           TestEnableConfig/enable_already_active_config
                Messages:       enable already active config: expected status already assigned got %!s(<nil>)
--- FAIL: TestEnableConfig (0.00s)
    --- PASS: TestEnableConfig/enable_non-existing_config (0.00s)
    --- PASS: TestEnableConfig/enable_inactive_config (0.00s)
    --- FAIL: TestEnableConfig/enable_already_active_config (0.00s)
FAIL
```
Please check and check why CI didn't catch this.

dborovcanin (approved):

PR #3515, opened 2026-05-15
NOISSUE - Update bootstrap content format, update profile method and add profile search

nyagamunene:
# What type of PR is this?

This is a feature/refactor because it improves the Bootstrap profile management API with better content format handling, profile listing search, proper DB query patterns, and a corrected update response.

## What does this do?

This PR makes several improvements to the Bootstrap profile management layer:

**1. Rename `template_format` → `content_format`**
The field was renamed across the API, service, Postgres layer, and mocks. A new migration (`configs_16`) renames the column in the `profiles` table via `ALTER TABLE`. The field describes the output format of the rendered `content_template` (go-template, raw, json, yaml, toml), not the format of the uploaded document.

**2. Infer `content_format` from `Content-Type` on profile upload**
When uploading a profile with a JSON, YAML, or TOML body, if `content_format` is not explicitly set in the document, it is now inferred from the `Content-Type` header. Previously the format always defaulted to JSON regardless of the uploaded file type.

**3. Add `config_id` to render-preview endpoint**
The render-preview endpoint now accepts an optional `config_id` in the request body. When provided, the config and its bindings are loaded from the database instead of requiring the caller to supply them inline. Inline `config` and `bindings` fields still work for ad-hoc preview.

**4. Profile listing search by name**
`GET /{domain_id}/clients/bootstrap/profiles` now accepts an optional `name` query parameter for case-insensitive substring filtering (e.g. `?name=gateway`). The Postgres implementation uses named parameters and the `PageQuery`/`applyOrdering` helper pattern consistent with the clients listing implementation.

**5. Fix positional SQL parameters in profiles repository**
All remaining `$1`/`$2` positional parameters in `bootstrap/postgres/profiles.go` have been replaced with named parameters (`:id`, `:domain_id`) using `NamedQueryContext`/`NamedExecContext` with typed DB structs, eliminating a SQL injection risk.

**6. `UpdateProfile` returns the updated profile from DB**
The `Update` SQL query now includes a `RETURNING` clause so the freshly-persisted row (with correct `updated_at` and `version`) is returned to the caller. `ProfileRepository.Update`, `Service.UpdateProfile`, and all middleware/tracing/event-store layers now return `(Profile, error)` instead of just `error`. The update endpoint responds with the DB-fresh profile.

- Related Issue #

Yes. Tests have been added or updated at multiple levels:

- **API endpoint tests** (`bootstrap/api/endpoint_test.go`) — new `TestListProfiles` table-driven test covering name-filter and limit-exceeded cases; new upload test cases for JSON/YAML/TOML `content_format` inference; extended `TestRenderPreview` with `config_id` cases.
- **Service tests** (`bootstrap/service_test.go`) — new `"list profiles filtered by name"` case in `TestListProfiles`; updated `TestUpdateProfile` mock to return `(Profile, error)`.
- **Repository tests** (`bootstrap/postgres/configs_test.go`) — existing `render_context` test updated to match renamed field.

No documentation update required. The changes are API-compatible improvements (additive fields, column renamed via migration, corrected response body).

### Notes

- The `content_format` rename is backward-compatible at the API level; only the Postgres column is renamed via migration.
- The `config_id` field in render-preview is optional — existing callers passing inline `config`/`bindings` are unaffected.
- Profile listing name search uses `ILIKE '%' || :name || '%'` (case-insensitive substring match), consistent with how client listing works.

### DIFF
--- a/bootstrap/postgres/profiles.go
+++ b/bootstrap/postgres/profiles.go
@@ -138,30 +138,33 @@ func applyProfilesOrdering(q string) string {
 	return fmt.Sprintf("%s ORDER BY created_at DESC", q)
 }
 
-func (pr profileRepository) Update(ctx context.Context, p bootstrap.Profile) error {
+func (pr profileRepository) Update(ctx context.Context, p bootstrap.Profile) (bootstrap.Profile, error) {
 	q := `UPDATE profiles SET name = :name, description = :description, content_format = :content_format,
 		  content_template = :content_template, defaults = :defaults, binding_slots = :binding_slots, version = version + 1, updated_at = :updated_at
-		  WHERE id = :id AND domain_id = :domain_id`
+		  WHERE id = :id AND domain_id = :domain_id
+		  RETURNING id, domain_id, name, description, content_format, content_template, defaults, binding_slots, version, created_at, updated_at`
 
 	p.UpdatedAt = time.Now().UTC()
 	dbp, err := toDBProfile(p)
 	if err != nil {
-		return errors.Wrap(repoerr.ErrUpdateEntity, err)
+		return bootstrap.Profile{}, errors.Wrap(repoerr.ErrUpdateEntity, err)
 	}
 
-	res, err := pr.db.NamedExecContext(ctx, q, dbp)
+	rows, err := pr.db.NamedQueryContext(ctx, q, dbp)
 	if err != nil {
-		return errors.Wrap(repoerr.ErrUpdateEntity, err)
+		return bootstrap.Profile{}, errors.Wrap(repoerr.ErrUpdateEntity, err)
 	}
-	cnt, err := res.RowsAffected()
-	if err != nil {
-		return errors.Wrap(repoerr.ErrUpdateEntity, err)
+	defer rows.Close()
+
+	if !rows.Next() {
+		return bootstrap.Profile{}, repoerr.ErrNotFound
 	}
-	if cnt == 0 {
-		return repoerr.ErrNotFound
+	var updated dbProfile
+	if err := rows.StructScan(&updated); err != nil {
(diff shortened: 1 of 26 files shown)


bootstrap/postgres/profiles.go
```diff
@@ -106,35 +113,82 @@ func (pr profileRepository) RetrieveAll(ctx context.Context, domainID string, of
...
+	}
+	if p.ContentTemplate != "" {
+		query = append(query, "content_template = :content_template,")
+	}
+	if p.Defaults != nil {
+		query = append(query, "defaults = :defaults,")
+	}
+	if p.BindingSlots != nil {
+		query = append(query, "binding_slots = :binding_slots,")
+	}
```
dborovcanin, line 156: Should we enable empty values?

bootstrap/profiles.go
```diff
@@ -25,7 +25,7 @@ type Profile struct {
 	DomainID        string         `json:"domain_id,omitempty"`
 	Name            string         `json:"name"`
 	Description     string         `json:"description,omitempty"`
-	TemplateFormat  TemplateFormat `json:"template_format"`
+	ContentFormat   ContentFormat  `json:"content_format"`
```
dborovcanin, line 28: What's the difference - why content vs template format?

nyagamunene: I saw this was confusing for the user. The output is `content` and the `templateformat` was being used to describe the output format so I renamed to `contentformat`

bootstrap/postgres/profiles.go
nyagamunene: `defaults` and `binding_slots` support clearing, sending {} or [] results in a non-nil value in Go, so the update runs. Omitting the field entirely skips it as expected. `content_template` intentionally blocks empty string since a profile with no template would break rendering.

PR #3591, opened 2026-08-20
Reject gateway chaining: a device may not be both is_gateway and declare gateways of its own

ianmuchyri:
## Summary

- Adds `pkg/atom.ValidateGatewayChain`: a device may not simultaneously carry `is_gateway: true` and a non-empty `gateways []` — a gateway may still report its own telemetry directly (A12's composite device is unaffected), but it may not itself be reachable through another gateway. Self-reference is rejected by the same rule.
- Enforced at both existing writers of these attributes: `SetDeviceGateways` (`pkg/atom/devices.go`) and the CLI's device create/update commands (`cli/devices.go`), which bypass `SetDeviceGateways` entirely.
- Exports `AttributeIsGateway` alongside the already-exported `AttributeGateways`, so a future writer (MG-09's HTTP API/SDK, not yet built) can find both halves of the contract.
- Records the decision as **A15** in `edge/architecture.md`'s decision log, including why it doesn't touch A12/ATOM-05 and why it's Magistrala-only (Atom has no notion of "gateway" — `is_gateway`/`gateways` are attribute conventions on a generic `device` entity).

## Why

`magistrala-ui`'s Access-page review flagged that its UI-side gateway-promotion logic already prohibits nested gateway topology, but nothing on this side enforced the same rule — `SetDeviceGateways` had no check against it, and even special-cased a device naming itself as one of its own gateways as supported, working behavior. This closes that gap so the two repos agree, before MG-09 freezes the public API shape.

No data migration is included: existing data was only ever reachable through the CLI (nothing on the UI or API side enforced anything before this), so this only stops new writes from creating the state.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l` clean
- [x] `go test ./pkg/atom/... ./cli/... ./readers/...` — updated existing self-reference tests to assert rejection instead of tolerance, added new tests for the chain-rejection and clearing-an-empty-list cases

### DIFF
--- a/cli/devices.go
+++ b/cli/devices.go
@@ -114,6 +114,12 @@ func handleDeviceCreate(cmd *cobra.Command, args []string) {
 	device.Kind = atomKindDevice
 	device.TenantID = args[1]
 
+	// A new device has no id yet, so self-reference cannot arise here --
+	// only the is_gateway-plus-gateways combination can.
+	if err := validateGatewayChain(cmd, "", device); err != nil {
+		return
+	}
+
 	created, err := atomClient.CreateEntity(cmd.Context(), device)
 	if err != nil {
 		logErrorCmd(*cmd, err)
@@ -181,6 +187,10 @@ func handleDeviceUpdate(cmd *cobra.Command, deviceID string, args []string) {
 		return
 	}
 
+	if err := validateGatewayChain(cmd, deviceID, device); err != nil {
+		return
+	}
+
 	updated, err := atomClient.UpdateEntity(cmd.Context(), deviceID, device)
 	if err != nil {
 		logErrorCmd(*cmd, err)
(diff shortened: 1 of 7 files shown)


cli/devices.go
```diff
@@ -195,6 +205,27 @@ func handleDeviceUpdate(cmd *cobra.Command, deviceID string, args []string) {
...
+// something here calls it too. deviceID is "" for create, where
+// self-reference cannot yet arise.
+//
+// Checked against exactly the payload being written, not a re-read of the
+// device: UpdateEntity replaces the whole attributes value rather than
+// merging (see fakeAtomDevices in pkg/atom's tests), so whatever this
+// payload sets for is_gateway and gateways is exactly what will be
+// persisted -- there is no stale prior state to additionally account for.
+func validateGatewayChain(cmd *cobra.Command, deviceID string, device atom.Entity) error {
+	isGateway, _ := device.Attributes[atom.AttributeIsGateway].(bool)
```
dborovcanin, line 221: This does not look safe - i.e. missing key will panic.

pkg/atom/devices_test.go
```diff
@@ -346,7 +346,7 @@ func TestSetDeviceGatewaysMarksGatewaysInOneBatchedRoundTrip(t *testing.T) {
 	}
 
 	for _, id := range gatewayIDs {
-		isGateway, _ := fa.entities[id].Attributes[atomAttributeIsGateway].(bool)
+		isGateway, _ := fa.entities[id].Attributes[AttributeIsGateway].(bool)
```
dborovcanin, line 349: Same here.

dborovcanin (changes requested):

cli/devices.go
ianmuchyri: This is a comma-ok type assertion (`v, _ := x.(T)`), which never panics regardless of whether the map is nil or the key is missing — a nil-map read yields the zero value (untyped `nil`), and asserting that against `bool` via the two-value form just returns `false, false`. Only the single-value form (`x.(T)` without the second return) panics on a failed assertion, and that's not what's used here.

Verified this empirically too (`go run` against a zero-value `atom.Entity` with a nil `Attributes` map) — no panic. This also mirrors the exact same pattern already in use a few lines up in this file, e.g. `markDeclaredGateways`'s handling, and in `pkg/atom/devices.go`'s `MarkGateways`.

Let me know if you had a different scenario in mind — happy to add a short comment here noting the safety if that'd help future readers.

pkg/atom/devices_test.go
ianmuchyri: Same reasoning as the sibling thread on `cli/devices.go`: `fa.entities` is `map[string]Entity` (value type, not pointer), so `fa.entities[id]` for a missing id yields the zero-value `Entity{}` rather than `nil` — and its `Attributes` field is then a nil map, which a comma-ok assertion reads safely (`false, false`), same as any nil-map read.

This is also unchanged behavior — the rename here is just `atomAttributeIsGateway` → `AttributeIsGateway` (exporting the constant for A15's cross-package invariant), the access pattern itself is identical to what was already here.

PR #3602, opened 2026-08-26
NOISSUE - Align health commands and docs

fbugarski:
## What type of PR is this?

This is a bug fix and documentation update. It aligns the CLI defaults, health commands, and documentation with the current Atom-backed local Compose stack.

## What does this do?

* Increases the default CLI Atom timeout to `90s` when `ATOM_TIMEOUT` is not explicitly set.
* Updates `docker/.env` to use `ATOM_TIMEOUT=90s`.
* Removes stale `certs` support from the CLI health command.
* Updates the default FluxMQ health URL to use the local nginx route.
* Makes successful raw `ok` command output JSON-parseable.
* Updates the CLI and root README documentation to reflect the current behavior.

* Related Issue #
* Resolves #

Yes. Added focused CLI tests covering:

* Atom timeout handling
* Health command behavior
* Raw success output

## Did you document any new/modified features?

Yes. Updated the CLI documentation and the CLI section in the root README.

### Notes

Verified with:

```bash
go test ./cli ./pkg/sdk
git diff --check
make cli
```

### DIFF
--- a/pkg/sdk/health.go
+++ b/pkg/sdk/health.go
@@ -8,7 +8,7 @@ import (
 	"fmt"
 	"net/http"
 
-	smqerrors "github.com/absmach/magistrala/pkg/errors"
+	"github.com/absmach/magistrala/pkg/errors"
 )
 
 // HealthInfo contains version endpoint response.
@@ -29,7 +29,7 @@ type HealthInfo struct {
 	BuildTime string `json:"build_time"`
 }
 
-func (sdk mgSDK) Health(service string) (HealthInfo, smqerrors.SDKError) {
+func (sdk mgSDK) Health(service string) (HealthInfo, errors.SDKError) {
 	var url string
 	switch service {
 	case "certs":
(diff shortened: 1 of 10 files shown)


pkg/sdk/health.go
```diff
@@ -29,28 +29,28 @@ type HealthInfo struct {
...
-func (sdk mgSDK) Health(service string) (HealthInfo, errors.SDKError) {
+func (sdk mgSDK) Health(service string) (HealthInfo, smqerrors.SDKError) {
 	var url string
 	switch service {
-	case "certs":
-		url = fmt.Sprintf("%s/health", sdk.certsURL)
 	case "fluxmq":
 		url = fmt.Sprintf("%s/health", sdk.httpAdapterURL)
+	default:
+		return HealthInfo{}, smqerrors.NewSDKError(fmt.Errorf("unsupported health service %q; supported services: fluxmq", service))
```
arvindh123: This removes the documented `Health("certs")` path from the public SDK, so existing callers configured with `Config.CertsURL` now receive an immediate unsupported-service error even though `certs/api/http` still exposes `/health`. The PR only intends to retire the CLI command; keep the SDK `certs` case and reject it in the CLI, or coordinate removal as a separate SDK API change. This regression blocks merging.

arvindh123 (commented):
Blocking public SDK regression found.

pkg/sdk/health.go
fbugarski: Fixed, thanks. I restored the public SDK Health("certs") path to avoid breaking
existing Config.CertsURL callers, and kept the removal scoped to the CLI by
rejecting unsupported health services before calling the SDK. Force-pushed the
update in 4fbcc6963.

pkg/sdk/health.go
```diff
@@ -8,7 +8,7 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/absmach/magistrala/pkg/errors"
+	smqerrors "github.com/absmach/magistrala/pkg/errors"
```
dborovcanin: Why did you change this? Previous `errors` with no alias was fine, but `smqerrors` alias is not only not needed, but misleading (`mgerrors` is a better choice).

dborovcanin (changes requested):

dborovcanin (approved):

PR #3605, opened 2026-08-27
NOISSUE - Add JSON format timescale-writer addon

felixgateru:
## Summary
- The default `timescale-writer` addon only runs the SenML transformer, so any message published under the `json` subtopic fails SenML validation and is silently dropped instead of persisted.
- Adds a second `timescale-writer` instance (`docker/addons/timescale-writer-json`) configured with the JSON transformer, scoped to `writers/+/c/+/json` so it lands in its own FluxMQ consumer group and doesn't compete with the SenML writer for the same messages. Mirrors the existing `postgres-writer` addon pattern.
- Companion of the equivalent addon added in magistrala-ee (absmach/magistrala-ee#12), which fixed the Rules Engine's Internal DB (JSON) output to actually route messages to this subtopic.


dborovcanin (changes requested):
Let's not make this add-on. Make this a separate instance in the default compose. But first - explore if we can simply do this via multiple transformers and topics.

dborovcanin (approved):

